PR #24, opened 2018-05-27
Improve FAQ instructions for vendoring golangci-lint

y0ssar1an:
```toml
required = [
  "github.com/golangci/golangci-lint/cmd/golangci-lint",
]
```
[The section on vendoring `golangci-lint`](https://github.com/golangci/golangci-lint/tree/61463e305fc23e1246e17b54db2b8daec3220ad4#faq) fails to mention that it needs to be added to the `required` section in `Gopkg.toml`. Without `required`, the vendored `golangci-lint` would be deleted on the next `dep ensure -update`. Also, some of the wording was a bit hard to understand.


golangci:
@y0ssar1an very good, thank you!

Issue #45, opened 2018-05-30
gocyclo linter is not executed when other linters enabled


sdzyba:
GolangCI-Lint version: 1.3.6
Go version: 1.10

I've noticed that when using default configuration (i.e., default set of linters) and enabling `gocyclo` both using command line arguments or `.golangci.yml`, `gocyclo` is actually not running or it can rarely give one output (although I know it should output a ton of warnings).
Looks like something interrupting it's work.
So I've tried to disable all the linters (including enabled by default) but left `gocyclo` enabled and it started to work like a charm.
Then I enabled `govet` and `gocyclo` continued to work correctly.
But when I was trying to enable some other linters from default linter set, `gocyclo` went back to incorrect work (or no work at all, to be precise).

Here's the config:
```
linters:
    enable:
        - gocyclo
    disable:
        - errcheck
        - staticcheck
        - unused
        - gosimple
        - gas
        - structcheck
        - varcheck
        - ineffassign
        - deadcode
        - typecheck

linters-settings:
    gocyclo:
        min-complexity: 5
    golint:
        min-confidence: 0
    govet:
        check-shadowing: true
    maligned:
        suggest-new: true

issues:
  exclude:
    - should have a package comment
    - declaration of "err" shadows declaration
```

golangci:
thank you for the issue,
fixed it in https://github.com/golangci/golangci-lint/pull/46, check, please

sdzyba:
@golangci 
thanks, works perfectly now.

PR #46, opened 2018-05-30
#45: fix no results for gocyclo

golangci:


pkg/enabled_linters_test.go
```diff
@@ -370,3 +371,46 @@ func BenchmarkWithGometalinter(b *testing.B) {
 		compare(b, runGometalinter, runGolangciLint, bc.name, "", lc/1000)
 	}
 }
+
+func TestGetEnabledLintersSet(t *testing.T) {
+	type cs struct {
+		cfg  config.Linters
+		name string   // test case name
```
golangci: `name` is unused

PR #47, opened 2018-05-31
Add Depguard to supported linters

dixonwille:
So I created this linter that is very helpful for administrating a repository. It was created because I help maintain a few go repositories and I would prefer the CI to pickup imports I don't like then have to go through and find them myself.

https://github.com/OpenPeeDeeP/depguard

I was not sure what all needed to change (tests and docs), but I did get it running on my machine and found a few places in the Readme.

Feedback on the linter itself is always welcome as well.


golangci:
@dixonwille thank you very much! I understand how difficult it was without any documentation. I made [wiki page](https://github.com/golangci/golangci-lint/wiki/How-to-add-a-custom-linter) about adding a new linter, feel free to contribute if you found anything incomprehensible.

.golangci.yml
```diff
@@ -15,6 +15,8 @@ linters-settings:
   goconst:
     min-len: 2
     min-occurrences: 2
+  depguard:
+    list-type: blacklist
```
golangci: What does it mean?

README.md
```diff
@@ -196,6 +197,7 @@ golangci-lint linters
 - [goimports](https://godoc.org/golang.org/x/tools/cmd/goimports): Goimports does everything that gofmt does. Additionally it checks unused imports
 - [maligned](https://github.com/mdempsky/maligned): Tool to detect Go structs that would take less memory if their fields were sorted
 - [megacheck](https://github.com/dominikh/go-tools/tree/master/cmd/megacheck): 3 sub-linters in one: unused, gosimple and staticcheck
+- [depguard](https://github.com/OpenPeeDeeP/depguard): Go linter that checks if package imports are in a list of acceptable packages
```
golangci: add it to [thanks](https://github.com/golangci/golangci-lint/blob/master/README.md#thanks) section please

pkg/enabled_linters.go
```diff
@@ -125,6 +126,7 @@ func GetAllSupportedLinterConfigs() []LinterConfig {
 		newLinterConfig(golinters.Maligned{}).WithFullImport().WithPresets(PresetPerformance).WithSpeed(10),
 		newLinterConfig(golinters.Megacheck{GosimpleEnabled: true, UnusedEnabled: true, StaticcheckEnabled: true}).
 			WithSSA().WithPresets(PresetStyle, PresetBugs, PresetUnused).WithSpeed(1),
+		newLinterConfig(golinters.Depguard{}).WithPresets(PresetAdmin).WithSpeed(6),
```
golangci: why not add it to `style` preset?

pkg/golinters/depguard.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	msgSuffix := "is in the blacklist"
+	if dg.ListType == depguardAPI.LTWhitelist {
+		msgSuffix = "is not in the whitelist"
+	}
+	res := make([]result.Issue, 0, len(issues))
+	for _, i := range issues {
+		res = append(res, result.Issue{
+			Pos:        i.Position,
+			Text:       fmt.Sprintf("%s %s", i.PackageName, msgSuffix),
```
golangci: use `formatCode` for `i.PackageName` please: it will be nice-formatted in GitHub comments

pkg/enabled_linters.go
golangci: `WithFullImport()` is needed: without that it won't load `loader.Program` if only this linter is enabled

dixonwille:
I'll address these tonight. At work currently.

.golangci.yml
dixonwille: So depguard is based off a whitelist or a blacklist. If nothing is passed it, it defaults to a whitelist with your current packages in it. Since this project has many imports, and empty blacklist will suffice (since it will allow everything). I could disable the linter if you would prefer?

pkg/enabled_linters.go
dixonwille: I didn't really think it fit the style prefix but I will gladly move it if you think that is more proper. I was thinking it from a repository maintainer administrating what imports are allowed. Let me know what you think?

golangci: I think it's a kind of style guide: what allowed and what not.

.golangci.yml
golangci: Some users run `golangci-lint` like we do: `golangci-lint run --enable-all`, and it shouldn't blacklist anything if not explicitly configured. Remove this configuration please and ensure that we don't need to disable linter explicitly after that. Also, linter should do nothing if it finds that nothing was configured.

pkg/golinters/depguard.go
```diff
@@ -0,0 +1,58 @@
+package golinters
```
golangci: add a test file to `pkg/testdata` please

.golangci.yml
dixonwille: So change the default to an empty blacklist if no configuration is passed?

Not sure I understand the reasoning behind not running a linter if it is enabled but not configured. Should it not have a default configuration (empty blacklist which allows everything)? Which brings up a good point I can short cut the code in that case.

pkg/golinters/depguard.go
dixonwille: Found it. I use Fedora and the golang package that I installed strips out those files...

dixonwille:
Having a few issues with fixing the test for depguard. I'll push it up when I get it all fixed

dixonwille:
So manually testing what I added works. But for some reason when ran against the `testdata`, it seems that the loader.Program does not have an initial set of packages. So depguard falls through as successful...

dixonwille:
I was also seeing something kind of strange. Seems like the list I was passing in was getting duplicated. Will look more into this later as it does not cause an issue for the linter itself. But it is more memory then it is needed

dixonwille:
@golangci So I did get everything working they way you would like! Just to document my issues 

- Some package managers do not install the `$(GOROOT)/test` directory
  - **Solution** was to download the precompiled from golang.org and extract out the `go/test/errchk` file and move it to my `$GOROOT`
- My default values for depguard are very permissive (allows everything) but it also doesn't make sense to have flags for it either as the values will not change often.
  - **Solution** add hidden pflags in the run command so that in the test I could set them to what the test expects
- There was an issue with my linter `*loader.Program.Imported` doesn't include what I would have expected. 
  - **Solution** You will almost always want to use `*loader.Program.InitialPackages()`. I am not sure exactly what the difference is (something around created and imported versus just imported).

golangci:
thank you very much! and thank you for documenting all issues, it will help us.
1. errchck - I will rewrite it to Go to not depend on side perl script.
2. it's awful, I will think about refactoring config and options parsing.
3. yes, all linters must use `loader.Program.InitialPackages()`. It's not obvious and should be refactored too.

golangci:
I've refactored code to simplify its changing and made automatic generation of README.

Issue #30, opened 2018-05-28
allow for skipping file names by patterns


srenatus:
I'm not sure if this is again only something I haven't found in the docs, but -- it would be very nice if `golangci-lint` could be configured to ignore files that have been generated. One way we've used this with GML was to configure an "exclude pattern" of `*.pb.*.go`, like

```
{
  "Exclude": [
     ".*\\.pb\\.go"
  ]
}
```

Any way to achieve that, short of (pre-)filtering the list of files to check?

golangci:
@srenatus thank you for the suggestion!
We will implement it. Now these files analyzed, but issues from them aren't reported (we see first line of these autogenerated files and understand they are autogenerated).

srenatus:
Ah! Thanks for the pointer -- indeed, [the checks](https://github.com/golangci/golangci-lint/blob/fb3929085c21f2d8108e76a9e4b4fdea9f0a91bd/pkg/result/processors/nolint.go#L50-L51) catch our protobuf files (`*.pb.go`), but not stuff generated with a custom code generator:
```
// Code generated by protoc-gen-validate
// source: foo.proto
// DO NOT EDIT!!!
```

In this case, the generator is https://github.com/lyft/protoc-gen-validate.

Likewise, files generated using https://github.com/twitter/go-bindata don't seem to be excluded -- they start with
```
// Code generated by go-bindata.
// sources:
// foo/bar.baz
// DO NOT EDIT!
```

kop:
It would be nice to be able to skip at least dirs/packages for now.
Another use case for this is third-party code that is not part of the vendor (needed in some cases).
I do not want to fix issues there in order to be able to update it later without issues.

jirfag:
try in master, please. Options `--skip-files` and `--skip-dirs`

PR #48, opened 2018-05-31
change isGenerated heuristic to match more generated files

srenatus:
As observed in #30, there's tools out there that don't comply 100% with
the referenced golang convention.

With this change, golangci-lint will skip some more of those generated
files.


golangci:
thank you! good feature

Issue #37, opened 2018-05-29
Additional --out-format (tabwriter)


frou:
`golangci-lint` has the flag:

    --out-format string    Format of output: colored-line-number|line-number|json (default "colored-line-number")

It would be nice if there was an additional format that made use of Go's standard library [tabwriter](https://golang.org/pkg/text/tabwriter/), that would output something like:

```
cmd\modd\main.go:70:15:     errcheck      Error return value of `runner.Reset` is not checked
daemon.go:123:23:           errcheck      Error return value of `d.cmd.Process.Signal` is not checked
modd_test.go:39:17:         errcheck      Error return value of `ioutil.ReadFile` is not checked
notify\notify.go:40:12:     errcheck      Error return value of `cmd.Run` is not checked
notify\notify.go:52:12:     errcheck      Error return value of `cmd.Run` is not checked
proc.go:37:36:              interfacer    fp can be io.Reader
```

rather than:

```
cmd\modd\main.go:70:15: Error return value of `runner.Reset` is not checked (errcheck)
daemon.go:123:23: Error return value of `d.cmd.Process.Signal` is not checked (errcheck)
modd_test.go:39:17: Error return value of `ioutil.ReadFile` is not checked (errcheck)
notify\notify.go:40:12: Error return value of `cmd.Run` is not checked (errcheck)
notify\notify.go:52:12: Error return value of `cmd.Run` is not checked (errcheck)
proc.go:37:36: fp can be io.Reader (interfacer)
```

jirfag:
hi! thank you for the issue,
try `--out-format=tab` please, supported in https://github.com/golangci/golangci-lint/pull/62

frou:
It works well. Thank you, I appreciate it!

PR #62, opened 2018-06-02
#37: add tab output format: --out-format=tab

jirfag:


pkg/printers/tab.go
```diff
@@ -0,0 +1,61 @@
+package printers
```
golangci: should have a package comment, unless it's in another file for this package

pkg/printers/tab.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+func (p Tab) SprintfColored(ca color.Attribute, format string, args ...interface{}) string {
+	c := color.New(ca)
+	return c.Sprintf(format, args...)
+}
+
+func (p *Tab) Print(ctx context.Context, issues <-chan result.Issue) (bool, error) {
+	w := tabwriter.NewWriter(StdOut, 0, 0, 2, ' ', 0)
+	defer w.Flush()
```
golangci: Error return value of `w.Flush` is not checked

PR #63, opened 2018-06-03
skip issues of file "C"

tgulacsi:
Solves

    WARN[0005] running error: can\'t print 99 issues: can\'t read file C for printing issued line: open C: no such file or directory

errors of checking projects using cgo.

You can check it on `gopkg.in/goracle.v2`

### DIFF
--- a/pkg/printers/text.go
+++ b/pkg/printers/text.go
@@ -67,6 +67,9 @@ func (p *Text) Print(ctx context.Context, issues <-chan result.Issue) (bool, err
 
 	issuesN := 0
 	for i := range issues {
+		if i.FilePath() == "C" {
+			continue
+		}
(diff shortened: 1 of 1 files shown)


jirfag:
thank you for the contribution!
But this fix should be done in a next way:
1. checking filename in [nolint processor](https://github.com/golangci/golangci-lint/blob/master/pkg/result/processors/nolint.go): now it's in `text` printer, but there are also `json` and `tab` printers.
2. adding a test for this case similar to [existing tests](https://github.com/golangci/golangci-lint/tree/master/pkg/result/processors/testdata) for `nolint` processor

jirfag:
fixed in https://github.com/golangci/golangci-lint/pull/70

PR #81, opened 2018-06-10
Feature/fix common failures

jirfag:


pkg/lint/load.go
```diff
@@ -138,6 +163,7 @@ func separateNotCompilingPackages(lintCtx *linter.Context) {
 	}
 }
 
+//nolint:gocyclo
```
golangci: comment on exported function LoadContext should be of the form "LoadContext ..."

jirfag: this issue reportin was fixed in current pull request

Issue #93, opened 2018-06-12
Support checkstyle output format 


mxpv:
`gometalinter` has `checkstyle` output format which can be parsed by TeamCity (https://confluence.jetbrains.com/display/TCD10/XML+Report+Processing).
Would be nice to have this feature in `golangci-lint`

PR #95, opened 2018-06-13
Checkstyle support

mxpv:
Adds checkstyle output format #93

### DIFF
--- a/pkg/printers/checkstyle.go
+++ b/pkg/printers/checkstyle.go
@@ -1,0 +1,17 @@
+package printers
+
+import (
+	"context"
+
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+type Checkstyle struct{}
+
+func NewCheckstyle() *Checkstyle {
+	return &Checkstyle{}
+}
+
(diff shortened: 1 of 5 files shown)


pkg/printers/checkstyle.go
```diff
@@ -0,0 +1,78 @@
...
+			Line:     issue.Line(),
+			Message:  issue.Text,
+			Source:   issue.FromLinter,
+			Severity: defaultSeverity,
+		}
+
+		file.Errors = append(file.Errors, newError)
+	}
+
+	out.Files = make([]*checkstyleFile, len(files))
```
jirfag: maybe `make([]*checkstyleFile, 0, len(files))`?

jirfag:
thank you for pull request!

pkg/printers/checkstyle.go
mxpv: This kind of syntax is not supported for maps.
It'll give an error `too many arguments to make(map)`
https://play.golang.org/p/b7b5e8jgEVp

jirfag: But it’s a slice, not map

mxpv: oh, crap. yes, you're right :)

PR #103, opened 2018-06-14
Remove extra backtick in install step

srizzling:
This just removes the extra back in the `go install` step. Helps people (like me) that just copy paste blindly into a Makefile or into the cmdline!


jirfag:
thank you!

Issue #110, opened 2018-06-15
Suppress congrats message


dahankzter:
It would be nice to be able to suppress the congrats message: `Congrats! No issues were found`

Perhaps a flag: `--no-congrats` or something?

jirfag:
Hi,
What’s the use-case? How it can help you?

dahankzter:
The use case is simply to reduce noise. Our integration servers runs a lot of stuff and having to scroll through "positives" to find any errors just adds some overhead. Not much but it adds up from all the tools we use.

dahankzter:
I can implement it if you think it would be a good addition.

mmatczuk:
@jirfag this break Unix Rule of Silence [1], which states that programs should, by default, say nothing.

[1] http://www.linfo.org/rule_of_silence.html
[2] http://www.linfo.org/unix_philosophy.html

ps.

Probably the same is true for json

```
$ golangci-lint run --out-format json ./...
{"Issues":[]} 
```

does not seem very helpful.

jirfag:
@dahankzter it will be great if you will implement it.
@mmatczuk ok, thank you, I agree, except json: no json is not valid json, but `{}` is valid.

mmatczuk:
Thanks @jirfag, congrats on this awesome tool!

dahankzter:
Indeed an awesome tool! Ill get back with a PR asap.

dahankzter:
I submitted some code for this @jirfag

Carpetsmoker:
I agree that the human-readable output should follow the "rule of silcence" whenever feasible @mmatczuk , but I'm not sure about the JSON:

> Probably the same is true for json
>
>     $ golangci-lint run --out-format json ./...
>     {"Issues":[]} 

An empty string can't be parsed by most (all?) JSON parsers, so people implementing this need to add an extra if `output == ""` check (and even if it would be parsable, you'd still need to check for the existence for the `Issues` key).

The "rule of silence" is for *user interfaces*, but the JSON output isn't a user interface but a programming API to integrate the tool with editors etc. I don't think that having it follow the same rules make sense.

onodera-punpun:
I personally think that -s should be the default. The UNIX philosophy to check if a command was successful is to check the status (`$?`) variable and on success not print any output (this is why the `go` commands print nothing unless you pass some verbose flag)

The UNIX philosophy calls this "the rule of silence":

[...]

PR #112, opened 2018-06-15
disable the congrats message

dahankzter:
There is now an extra switch '-s' to disable the congrats message when
there are no issues detected

Fixes: #110

### DIFF
--- a/pkg/commands/root.go
+++ b/pkg/commands/root.go
@@ -83,6 +83,7 @@ func (e *Executor) needVersionOption() bool {
 
 func initRootFlagSet(fs *pflag.FlagSet, cfg *config.Config, needVersionOption bool) {
 	fs.BoolVarP(&cfg.Run.IsVerbose, "verbose", "v", false, wh("verbose output"))
+	fs.BoolVarP(&cfg.Run.Silent, "silent", "s", false, wh("disables congrats outputs"))
 	fs.StringVar(&cfg.Run.CPUProfilePath, "cpu-profile-path", "", wh("Path to CPU profile output file"))
 	fs.StringVar(&cfg.Run.MemProfilePath, "mem-profile-path", "", wh("Path to memory profile output file"))
 	fs.IntVarP(&cfg.Run.Concurrency, "concurrency", "j", getDefaultConcurrency(), wh("Concurrency (default NumCPU)"))
--- a/README.md
+++ b/README.md
@@ -294,6 +294,7 @@ Global Flags:
   -j, --concurrency int           Concurrency (default NumCPU) (default 8)
       --cpu-profile-path string   Path to CPU profile output file
       --mem-profile-path string   Path to memory profile output file
+  -s, --silent                    disables congrats outputs
   -v, --verbose                   verbose output
 
 ```
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -258,10 +258,10 @@ func (e *Executor) runAndPrint(ctx context.Context, args []string) error {
 		p = printers.NewJSON()
 	case config.OutFormatColoredLineNumber, config.OutFormatLineNumber:
 		p = printers.NewText(e.cfg.Output.PrintIssuedLine,
-			format == config.OutFormatColoredLineNumber, e.cfg.Output.PrintLinterName,
+			format == config.OutFormatColoredLineNumber, e.cfg.Output.PrintLinterName, e.cfg.Run.Silent,
 			e.log.Child("text_printer"))
 	case config.OutFormatTab:
-		p = printers.NewTab(e.cfg.Output.PrintLinterName,
+		p = printers.NewTab(e.cfg.Output.PrintLinterName, e.cfg.Run.Silent,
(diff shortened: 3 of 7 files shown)


mmatczuk:
How about making it default behaviour?

dahankzter:
That would change the current behavior and potentially break someone who for example greps for the congrats message.

pkg/commands/root.go
```diff
@@ -84,6 +84,7 @@ func (e *Executor) needVersionOption() bool {
 
 func initRootFlagSet(fs *pflag.FlagSet, cfg *config.Config, needVersionOption bool) {
 	fs.BoolVarP(&cfg.Run.IsVerbose, "verbose", "v", false, wh("verbose output"))
+	fs.BoolVarP(&cfg.Run.DisableCongrats, "no-congrats", "s", false, wh("disable congrats output"))
```
jirfag: 1. maybe `silent` instead of `no-congrats` and `Silent` instead of `DisableCongrats`? It will be more general logic. Also in some places options disables not only congrats message, e.g. in JSON
2. `s/disable/Disable`

pkg/printers/json.go
```diff
@@ -8,22 +8,30 @@ import (
...
 
-func (JSON) Print(ctx context.Context, issues <-chan result.Issue) (bool, error) {
+func (j JSON) Print(ctx context.Context, issues <-chan result.Issue) (bool, error) {
 	allIssues := []result.Issue{}
 	for i := range issues {
 		allIssues = append(allIssues, i)
 	}
 
+	if len(allIssues) == 0 && j.disableCongrats {
+		return true, nil
```
jirfag: I know it's ugly and we need to refactor it. Here need to return `false`: it's a flag that issues were found

pkg/printers/json.go
```diff
@@ -8,22 +8,30 @@ import (
 	"github.com/golangci/golangci-lint/pkg/result"
 )
 
-type JSON struct{}
+type JSON struct {
+	disableCongrats bool
```
jirfag: JSON knows nothing about congrats, `silent` or more specific for JSON like `omitEmpty`

jirfag:
thank you for the pull request!

pkg/commands/root.go
dahankzter: Sure it makes sense.

pkg/printers/json.go
dahankzter: Ah ok I missed the logic a bit here

pkg/printers/json.go
dahankzter: My vote is for silent here as well I think.

dahankzter:
I will make the suggested changes asap.

pkg/printers/json.go
```diff
@@ -9,22 +9,30 @@ import (
...
 
-func (JSON) Print(ctx context.Context, issues <-chan result.Issue) (bool, error) {
+func (j JSON) Print(ctx context.Context, issues <-chan result.Issue) (bool, error) {
 	allIssues := []result.Issue{}
 	for i := range issues {
 		allIssues = append(allIssues, i)
 	}
 
+	if len(allIssues) == 0 && j.silent {
+		return false, nil
```
Carpetsmoker: As I [mentioned here](https://github.com/golangci/golangci-lint/issues/110#issuecomment-397825420), I'm not sure this is actually useful to do for JSON output. It's off by default so not a *huge* deal as such, but the "silent" flag is pretty generically named and could also change other behaviour in the future, and having it output invalid JSON as a kind of side-effect may surprise people.

It may also be confusing in some cases where users provide custom flags to the editor integration or some such, and the editor isn't prepared to deal with invalid JSON, causing confusing errors.

Personally I think always outputting valid JSON is much better, and would simplify the life of people implementing this (i.e. me :smile:).

jirfag: I agree about valid JSON, it would be better to not change JSON output at all

mmatczuk:
@Carpetsmoker I agree that JSON should not be removed, my point was that maybe it could be just `{}` but it's a knob...

@jirfag I'm not sure if adding the silent flag is a move in right direction

1. it does not really solve the Rule of Silence - *by default* say nothing
1. another flag adds complexity to the tool
1. it's not really clear how silent relates to verbose flag
1. I do not see a lot of possibilities to reuse the silent flag for other things

Please reconsider, thanks.

pkg/printers/json.go
dahankzter: Sure for JSON we can just emit {} perhaps or simply let it be as it was. @mmatczuk has a point though to consider.

What about the checkstyle xml? Perhaps it should also produce valid xml?

Tab and text can be silenced then is this the temporary consensus until a major decision is made as to the default behavior?

dahankzter: Ah I just now saw you mentioned checkstyle as well @jirfag

dahankzter:
The Checkstyle and JSON output is back to where it was now.

jirfag:
thank you!

PR #152, opened 2018-07-03
Fixed example for 'exclude-use-default'

beono:
Hi,

There is a mistake in the `README` and `.golangci.example.yml`.

Default value is actually `true`

Proof: https://github.com/golangci/golangci-lint/blob/master/pkg/commands/run.go#L141

```go
fs.BoolVar(&ic.UseDefaultExcludes, "exclude-use-default", true, getDefaultExcludeHelp())
```


jirfag:
hi!
we did so stupid mistake :(
thank you

PR #176, opened 2018-08-01
Improve lll parsing for very long lines

jfrabaute:
lll is using scanner.Scan() to read the file line by line.
scanner.Scan() might fail if the line is longer than bufio.MaxScanTokenSize
In the case where the specified maxLineLen is smaller than bufio.MaxScanTokenSize
we can return this line as a long line instead of returning an error.
The reason for this change is that this case might happen with autogenerated files
The go-bindata tool for instance might generate a file with a very long line.
In this case, as it's a auto generated file, the warning returned by lll will
be ignored.
But if we return a linter error here, and this error happens for an autogenerated
file the error will be discarded (fine), but all the subsequent errors for lll will
be discarded for other files and we'll miss legit error.


jirfag:
perfect, thank you!

PR #179, opened 2018-08-02
Update depguard with Glob matching support

dixonwille:
This did add a vendor github.com/gobwas/glob, but when I ran `dep ensure`, cobra files were added to the vendors as well. Let me know if something doesn't seem right!

Also, awesome job on increasing the supported linters so fast. Can't wait to start using misspell!


dixonwille:
Found out I was on an outdated version of dep. Ran again with the new version.

jirfag:
thank you!

PR #185, opened 2018-08-07
point users to --new-from-rev=HEAD~ for CI setups in --help

rski:
Although the FAQ does eventually mention that
--new-from-rev=origin/master is the right way, --help does not and
using --new in CI is a terrible pitfall.

Make --help point users to -new-from-rev=HEAD~ for CI setups.

Also use HEAD~ instead of origin/master, since HEAD~ is more
universally applicable, e.g. if golangci-lint is ran against a release
branch.


jirfag:
thank you!

PR #225, opened 2018-09-11
Use a multi-staged Dockerfile

schoentoon:
This way we have consistent docker images rather than locally build binaries dumped into an image. Ideally we'd build with 1.11 right away, but this doesn't seem to work correctly just yet sadly

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,13 @@
+FROM golang:1.10 AS builder
+
+COPY . $GOPATH/src/github.com/golangci/golangci-lint
+
+RUN make -C $GOPATH/src/github.com/golangci/golangci-lint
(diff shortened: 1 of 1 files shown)


schoentoon:
It does actually build it, after which it runs all the tests. I just looked into building it with goreleaser inside the docker container but didn't get it to work. First off it would error because my git is in a dirty stage, my main motivation for building it with docker is to always have a completely clean go environment, thus great to use during development. After having turned off this check in the Makefile it however it still wouldn't work as it errored with `missing GITHUB_TOKEN`. Unless there's some option in goreleaser to just build it the binary I don't think actually using goreleaser to build it inside the container is such a great idea.

PR #236, opened 2018-09-28
Add install.sh to the repo and update readme. Fixes #233

### LINKED ISSUE #233: Installation in a restricted CI environment
## Problem 1
  In restricted go ci agents i often find that access to `install.goreleaser.com` is blocked where as `raw.githubusercontent.com` is not.

## Proposed solution
 If install script file is checked in as part of the repo it can be retrived via `raw.githubusercontent.com`.  Currently [dep](https://github.com/golang/dep) project has an [install.sh](https://github.com/golang/dep/blob/master/install.sh) file enabling it to be "curled" from `raw.githubusercontent.com`

Installation from `install.goreleaser.com` can still and always remain the primary mode of installation and `raw.githubusercontent.com` can be a backup

## Problem 2

In readme 
```bash
curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | bash -s -- -b $GOPATH/bin vX.Y.Z
```
In alpine linux boxes bash will not be present

## Proposed solution

Replace the installation script in read me as

```
curl -sfL https://install.goreleaser.com/github.com/golangci/golangci-lint.sh | sh
```

If you are fine with this i will submit a PR :slightly_smiling_face:

kishaningithub:
Fix for #233


jirfag:
thank you, perfect!

PR #248, opened 2018-10-19
readme: fix typo in sample config.

### LINKED ISSUE #216: Config file issues.max-same is actually issues.max-same-issues
Docs say that
```yaml
issues:
  max-same: 0
```

Should disable maximum count of issues with the same text, but actually this does it:
```yaml
issues:
  max-same-issues: 0
```

diegs:
Should be `max-same-issues`, not `max-same`.

Also adds a trailing newline to files, which is good form, and updates
the generator to handle that cleanly.

Fixes #216.


jirfag:
thank you!

PR #252, opened 2018-10-22
Improve error handling when executing of command fails

barthr:
When using golangci-lint I noticed sometimes it panics. This stacktrace looks very ugly in the console. I changed it to use `Fprintf` followed by an `os.Exit(1)` so the output looks a bit nicer.


jirfag:
thank you!

Issue #229, opened 2018-09-19
`--version` flag not works if binary installed via `go get`


derElektrobesen:
```
~/.go/src/github.com/golangci/golangci-lint git:(master)
$ g log -1
commit 0c82b641236764940235a15c3d51a4623b8045a1
Author: Denis Isaev <denis@golangci.com>
Date:   Wed Sep 5 20:49:57 2018 +0300

$ go version
go version go1.11 linux/amd64

$ go env
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/p.berezhnoy/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/p.berezhnoy/.go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build666675245=/tmp/go-build -gno-record-gcc-switches"

$ golangci-lint --version
Error: unknown flag: --version
```

barthr:
That is because https://github.com/golangci/golangci-lint/blob/4570c043a9b56ccf0372cb6ba7a8b645d92b3357/cmd/golangci-lint/main.go#L11-L14 is populated during build by goreleaser. The flag will not be available whenever `date == ""`. Which is the case when using go get.

If you wan't to build from source and wan't this variables to be present (so including the --version flag). You need to run this:

```shell
$ go get -u github.com/golangci/golangci-lint
$ cd $GOPATH/src/github.com/golangci/golangci-lint/cmd/golangci-lint
$ go install -ldflags "-X 'main.version=$(git rev-parse --abbrev-ref HEAD)' -X 'main.commit=$(git rev-parse --short HEAD)' -X 'main.date=$(date)'"

$ golangci-lint --version
```

jirfag:
[...]

PR #256, opened 2018-10-26
Improve readme with extended installation instruction when building from source.

barthr:
Building from source gave some problems with for example `--version` see #229 . This pull request updates the documentation so that building from source includes the needed ldfags.


jirfag:
thank you!

PR #257, opened 2018-10-27
Add without-tests flag for errcheck

osamingo:
We would be able to hand over `-ignoretests` flag to `errcheck` linter.
ref. https://github.com/kisielk/errcheck#the-deprecated-method

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -59,6 +59,10 @@ linters-settings:
     # report about assignment of errors to blank identifier: `num, _ := strconv.Atoi(numStr)`;
     # default is false: such cases aren't reported by default.
     check-blank: false
+
+    # report about checking of _test.go files is disabled.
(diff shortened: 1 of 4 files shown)


jirfag:
hi!
1. sorry, it won't work because this flag is used by errcheck only when loading program, but loading is performed by golangci-lint.
2. why do you need this flag?

osamingo:
@jirfag Thanks for your reply!

> 1. sorry, it won't work because this flag is used by errcheck only when loading program, but loading is performed by golangci-lint.

Oh, sorry.
It was my lack of understanding.

> 2. why do you need this flag?

I use this flag when check by `errcheck`.
e.g. https://godoc.org/github.com/favclip/testerator

jirfag:
it looks like solution with configuration of tests per linter in more convenient and can be reused for other linters

osamingo:
OK, I just understand it. I close this PR.

PR #258, opened 2018-10-28
Tweak log entry for "not compiling" packages.

scode:
Motivation: Make it clear that the message is indicating packages
*which do not compile*, rather than packages which *won't be
compiled*.


jirfag:
thank you!
message was really incomprehensible

PR #279, opened 2018-11-07
fix #199: support gochecknoinits, gochecknoglobals

### LINKED ISSUE #199: Feature request: add gochecknoinits
It would be rad to include this linter:

https://github.com/leighmcculloch/gochecknoinits

We want to avoid the use of `init` in our code, and would love it if `golangci-lint` did it for us.

jirfag:
Add 2 new linters

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -31,3 +31,4 @@ linters:
     - maligned
     - prealloc
     - gosec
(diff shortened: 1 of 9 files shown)


jirfag:
merging manually: failures are because of new linter name in `.golangci.yml`

PR #283, opened 2018-11-08
Update astutil to fix goimports

### LINKED ISSUE #282: goimports removes imports that don't match package names
Hi there!

It seems like that goimports fails to resolve imported packages if their paths don't match their names but the latest binary from the upstream doesn't do that.

Here's the script to reproduce it: 

```bash
export GO111MODULE=on
echo 'module goimports' > go.mod
mkdir sub
echo 'package subpkg

func Foo() {}' > sub/sub.go
echo 'package main

import "goimports/sub"

func main() {
	subpkg.Foo()
}' > main.go
```

Run:

```bash
golangci-lint run --disable-all --enable=goimports
```

Output:

```
main.go:3: File is not `goimports`-ed (goimports)
import "goimports/sub"
```

amenzhinsky:
Fixes #282

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,7 +65,7 @@ require (
 	github.com/spf13/viper v1.0.2
 	github.com/stretchr/testify v1.2.1
 	golang.org/x/crypto v0.0.0-20180505025534-4ec37c66abab // indirect
-	golang.org/x/tools v0.0.0-20180831211245-6c7e314b6563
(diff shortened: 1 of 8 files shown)


amenzhinsky:
It was a wrong shot, I'm seeing that there's activity in the tools repo, recent changes that caused the issue in the first place were  reverted.

PR #309, opened 2018-11-27
Fix typo in gometalinter comparison example

bvwells:
Fix typo in gometalinter comparison example.


jirfag:
thank you! I'm waiting for CLA to merge

bvwells:
No problems @jirfag!

bvwells:
@jirfag, it looks like golangci-lint also supports excludes on the command line (```--exclude=<default golangci-lint excludes>```). I was thinking about adding this to the gometalinter example comparison. What do you think?

jirfag:
these excludes are used by default, even without an explicit argument.

bvwells:
I see what you mean. Thanks for the clarification.

Issue #239, opened 2018-10-03
Bump gosec version


jirfag:

mastertinner:
This should include gosec 1.3: https://github.com/securego/gosec/releases/tag/1.3.0

jirfag:
In fact, we already use updated `gosec`: we don't have only the following commits:
```
62b5195 Report for Golang errors (#284)
9cdfec4 Change test
8048b15 Add more badges in the README file
```

But these commits don't change the behaviour of `gosec`.

PR #310, opened 2018-11-29
Use upstream version of gosec

dhui:
Addresses: https://github.com/golangci/golangci-lint/issues/239

There's a lot of noise in the PR diff since I ran `go mod tidy` and `go mod vendor`. The only changes I made were in `pkg/golinters/gas.go`, the rest of the changes were made by the Go tool chain.

Also, `github.com/stretchr/testify` was automatically updated from `v1.2.1` to `v1.2.2` by the Go tool chain, which shouldn't cause any issues...


dhui:
Ahhh, turns out using the upstream version of `gosec` isn't that simple since the golangci forked version adds the `ProcessProgram()` method to `gosec.Analyzer` which works with loaded programs instead of package path strings.

jirfag:
closing, I will do it myself

PR #311, opened 2018-12-03
Add .pre-commit-hooks.yaml

aneeshusa:
This file will allow using golangci-lint with https://pre-commit.com.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,7 @@
+- id: golangci-lint
+  name: golangci-lint
+  description: Fast linters runner for Go.
+  entry: golangci-lint run
+  types: [go]
+  language: golang
+  pass_filenames: false


aneeshusa:
cc @diegs

.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,7 @@
+- id: golangci-lint
+  name: golangci-lint
+  description: Fast linters runner for Go.
+  entry: golangci-lint run
+  types: [go]
+  language: golang
+  pass_filenames: false
```
diegs: I think should support passing filenames, right? Otherwise this won't have the ability to do somewhat incremental diffs.

Separately, it would be nice to pass the sha and only do incremental diffs on a per-commit basis. I know golangci-lint supports this, but I don't think pre-commit does.

aneeshusa: While that's true, golangci-lint AFAIK expects go package directories, not filenames, and I'm not sure whether a wrapper script here in golangci-lint, or a converter in pre-commit (since other Go tools are likely to use the same convention) is better.

pre-commit can be smart about only checking staged files during an actual pre-commit hook, but I haven't thought about how to integrate that with golangci-lint; fixing the `pass_filenames` thing would be a good first step there.

I'd like to get this in as a basic implementation first, and then refine it to get the improved performance/benefits of incremental linting afterwards since that will require a bit of design work.

diegs:
@aneeshusa LGTM, though I have no access to this repo to grant any real approvals :)

jirfag:
thank you for the pull request!

Issue #301, opened 2018-11-21
Don't load and analyze dirs from --skip-dirs


nicksnyder:
I am trying to run golangci-lint on [sourcegraph/sourcegraph](https://github.com/sourcegraph/sourcegraph) but it is failing due to the presence of some Go code in one of our node_modules dependencies (weird but ¯\\\_(ツ)\_/¯). I tried using `--skip-dirs node_modules`, but it seems to not have any effect.

Reproduce:
```
git clone https://github.com/sourcegraph/sourcegraph
cd sourcegraph
yarn
golangci-lint -v run --skip-dirs node_modules
```

Requested info:
1. Version of golangci-lint: `golangci-lint --version` (or git commit if you don't use binary distribution)
55a18ae18a3c49cb7151da117c0ca0a566f58ce5

2. Config file: `cat .golangci.yml`
None

3. Go environment: `go version && go env`
```
go version go1.11.2 darwin/amd64
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/nick/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/nick/dev/gopath"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/l6/djg_zw1j5lgbwz4h5q92lnwr0000gn/T/go-build800779018=/tmp/go-build -gno-record-gcc-switches -fno-common"
```

4. Verbose output of running: `golangci-lint run -v`

```
$ golangci-lint -v run --skip-dirs node_modules
[...]

PR #313, opened 2018-12-06
Fix the skip-dirs option.

sean-:
Prior to this change the SkipDir runner was not skipping files such as `foo/bar/baz.go` with a `skip-dir` entry of `foo/bar` when the `run` command was invoked with an argument of `foo/...`.  This is both a surprising and problematic behavior change.

The pathology was:

1. `shouldPassIssue()` was receiving an input like `foo/bar/baz.go`
2. `shouldPassIssue()` would call `p.getLongestArgRelativeIssuePath()` which was returning `bar`, not `foo/bar` as expected.
3. The reason for this was because inside of `getLongestArgRelativeIssuePath()` it was trimming the prefix that matched the path prefix (e.g. `foo/`).

If you have the file structure:

  - foo/bar/baz.go
  - bur/bar/baz.go

There is no way to isolate `foo/bar` from `bur/baz` without strictly controlling both your `skip-dirs` configuration and the arguments to `run`.

The rest of the logic to skip files that don't match `run`'s argument is valid, however the regexp should be evaluated based on the `filepath.Dir()` of the input (e.g. `foo/bar`) and not the truncated version of the issue's filepath.

This fixed unexpected breakage resulting from a recent upgrade.

Fixes: #301

### DIFF
--- a/pkg/result/processors/skip_dirs.go
+++ b/pkg/result/processors/skip_dirs.go
@@ -85,9 +85,7 @@ func (p *SkipDirs) getLongestArgRelativeIssuePath(i *result.Issue) string {
 			continue
 		}
 
-		relPath := strings.TrimPrefix(issueAbsPath, arg)
-		relPath = strings.TrimPrefix(relPath, string(filepath.Separator))
-		return relPath
+		return i.FilePath()
 	}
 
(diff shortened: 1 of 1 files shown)


jirfag:
thank you! but tests don't pass, I will fix it in https://github.com/golangci/golangci-lint/pull/327

PR #315, opened 2018-12-06
Allow configure module download mode

nvartolomei:
Before go modules, `go list` worked flawlessly, using vendor directory first and then looking in `GOPATH` for packages.

Since modules were introduce, vendor directory is used differently, by default `vendor` directory seems to be ignored.

> Finer grain control on whether or not go.mod is allowed to be updated, how vendor directory is used, and whether or not network access is allowed (e.g., -mod=readonly, -mod=vendor, GOPROXY=off)

This may work ok locally, but for CI/CD purposes I want to always use files in the vendor directory, at least for the time being.

`golangci-lint` is executing `go list`, in order to make it work under the constraint described above it should be possible to pass `-mod=vendor` flag to `go list`.

Feel free to adjust the PR before merging.


nvartolomei:
Looks like this can be also achieved by setting `GOFLAGS` environment variable.

`export GOFLAGS="-mod=vendor"`

PR #317, opened 2018-12-07
Fix errors when specifying absolute path

dgolub:


jirfag:
thank you! I will add test for it

PR #319, opened 2018-12-12
Use go env GOPATH for installation

LK4D4:
People these days tend to use default GOPATH, so environment variable will be empty.

go env GOPATH will work for all cases


jirfag:
thank you!

PR #336, opened 2018-12-26
Add support for searching for errcheck exclude file

joshuarubin:
We were having an issue where we had to have uncommitted changes to our `.golangci.yml` file so that our errcheck exclude file could work properly in different environments. We had to use absolute paths to ensure that the linter could be run from anywhere. However, having different home directories and GOPATHs made this challenging.

This patch adds support for a few things:

1. `~` expansion of home directories
2. If the exclude name from the config is not absolute, then use the same logic as finding the `.golangci.yml` file to work up the path looking for it.

### DIFF
--- a/pkg/golinters/errcheck.go
+++ b/pkg/golinters/errcheck.go
@@ -5,6 +5,8 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"os/user"
+	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


jirfag:
thank you very much for this contribution. But I am in doubts because it's not obvious logic. Also, it's not the first issue when we need different configurations for different environments. I had an idea to allow `golangci.yml` overriding like [`docker-compose` does](https://docs.docker.com/compose/extends/#understanding-multiple-compose-files). If I implement it, does it solve your problem?

joshuarubin:
That would make what we need possible, but not very clean. Perhaps an easier solution would be to add a command line flag to set that file.

BTW, the logic used to find the file is the same used to find `.golangci.yml`. If it's not obvious, then the existing logic isn't either.

joshuarubin:
Any thoughts on my latest suggestion?

jirfag:
@joshuarubin you've convinced me, thank you :)

PR #342, opened 2018-12-28
gocritic: update default checks list && fix new added gocritic linting errs

### LINKED ISSUE #341: gocritic enabled by default check list
I don't know exactly what criteria is being followed to enable or not by default a check inside gocritic.

Since latest gocritic update on: https://github.com/golangci/golangci-lint/pull/329 a lot new stable checks were added to gocritic.

Maybe it makes sense to keep this:

https://github.com/golangci/golangci-lint/blob/master/pkg/config/config.go#L260

In sync witn this:

https://github.com/go-critic/go-critic/blob/master/checkers/checkers_test.go#L63

So any gocritic default run like in gocritic, as expained in https://github.com/go-critic/go-critic#usage:

```
# Run all stable checkers on current dir and all its childs recursively:
gocritic check ./...
```

And a default gocritic run inside golangci-lint are using the same checks.

A note should be added somewhere IMHO in the code until we find a way to get this list programatically from lintpack/gocritic.

lopezator:
* Update gocritic default checks list, add the current gocritic stable checks list.
* Fix gocritic new added linters detected problems.

Closes #341


lopezator:
CC/ @jirfag not sure why the the check is failing. It looks that `make readme` added a version flag that was missing (and I didn't changed). Maybe this could be the cause?

lopezator:
Thanks @jirfag! It looks like we've have now "gocritic" checks that are not passing. 

If you want to go forward with this I could update this with the needed changes.

pkg/lint/lintersdb/manager.go
```diff
@@ -237,18 +237,18 @@ func (m Manager) GetAllEnabledByDefaultLinters() []linter.Config {
 	return ret
 }
 
-func linterConfigsToMap(lcs []linter.Config) map[string]*linter.Config {
+func linterConfigsToMap(lcs []*linter.Config) map[string]*linter.Config {
 	ret := map[string]*linter.Config{}
 	for _, lc := range lcs {
 		lc := lc // local copy
```
lopezator: @jirfag Is this local copy necessary if we use pointers?

lopezator:
PTAL @jirfag

pkg/lint/lintersdb/manager.go
jirfag: it's for `scopelint` - it finds a lot of false-positives

jirfag:
I was looking too long because I had doubts do we really need to enable performance checks in go-critic by default. Some developers can say that fixing them is a premature optimization. And a lot of users just use `--enable-all` and `go get` and will get these performance warnings after merging this PR. But I decided it's more useful than not.
Thank you very much, great work, this pointers refactoring was long needed

lopezator:
@jirfag I understand, but IMHO the only way to stay safe is to disable all linters by default. Then define all your linters + checks manually.

I agree that this would change "enable-all" users current behaviour, but also it will the inclusion of a new default enabled check (gocritic was enabled) and the latest even more drastically.

Maybe this information should be added somewhere in the docs to warn new users.

lopezator:
Btw I have some ideas to face versioning and breaking changes. I could give more details or propose something when I have more time :)

jirfag:
I've already warned in README about it, it doesn't help :)
During an update of gocritic version I've found that it's behavior doesn't match our behavior: gocritic doesn't enable performance checkers by default. I decided that we should be in sync to not confuse users. Also, I've removed copy-pasting a list of checks. PTAL this https://github.com/golangci/golangci-lint/pull/353 pull request

PR #350, opened 2019-01-08
stop memory from blowing up on projects with compilation errors

### LINKED ISSUE #349: golangci-lint allocates gigabites of memory on compilation errors

Please include the following information:
1. Version of golangci-lint: `golangci-lint --version` (or git commit if you don't use binary distribution)
v1.2.3 & current master
2. Config file: `cat .golangci.yml`
 golangci-lint run --disable-all --enable gosimple max-issues-per-linter 0 --max-same-issues 0
3. Go environment: `go version && go env`
go 1.12, +e3eb2ff827

gopackage fails to load some packages correctly with the latest go, e.g. 
runtime/internal/atomic:/home/rski/tmp/memoryissues/goroot/src/runtime/internal/atomic/asm_amd64.s:8:1: illegal character U+0023 '#'

This is because some package that is checked by golangci-lint imports atomic. That's fine - it's not golangci-lint's problem.

The issue comes in that in this case golangi-lint starts continuously going through the checked package's dependencies in extractErrorsImpl, ballooning in memory use. 
I  don't have a specific reproduction case at hand, but in our repo, it got to 70Gigs of ram processing a   package before I killed it. yikes. 
I have a fix for this, I'll push a pr in a bit

rski:
In some case, a compilation failure in a package could cause
golangci-lint to repeatedly go through its dependencies, allocating
huge amounts of memory. Presumably at some point it would stop and
print the failure, but in my case it did not.

packages.Visit keeps track of the packages visited, so using that
instead prevents the memory allocation spiral.

Fixes #349

### DIFF
--- a/pkg/packages/util.go
+++ b/pkg/packages/util.go
@@ -49,21 +49,11 @@ func ExtractErrors(pkg *packages.Package, astCache *astcache.Cache) []packages.E
 }
 
 func extractErrorsImpl(pkg *packages.Package) []packages.Error {
-	if !pkg.IllTyped { // otherwise it may take hours to traverse all deps many times
-		return nil
-	}
-
-	if len(pkg.Errors) != 0 {
-		return pkg.Errors
-	}
-
 	var errors []packages.Error
-	for _, iPkg := range pkg.Imports {
-		iPkgErrors := extractErrorsImpl(iPkg)
-		if iPkgErrors != nil {
-			errors = append(errors, iPkgErrors...)
-		}
-	}
-
+	packages.Visit([]*packages.Package{pkg},
+		// only keep if pkg or its parents have errors,
+		// otherwise save the effort, there are no errors to be found.
+		func(pkg *packages.Package) bool { return pkg.IllTyped },
+		func(pkg *packages.Package) { errors = append(errors, pkg.Errors...) })
 	return errors
 }


pkg/packages/util.go
```diff
@@ -49,21 +49,15 @@ func ExtractErrors(pkg *packages.Package, astCache *astcache.Cache) []packages.E
...
-			errors = append(errors, iPkgErrors...)
-		}
-	}
-
+	packages.Visit([]*packages.Package{pkg},
+		// only keep if pkg or its parents have errors,
+		// otherwise save the effort, there are no errors to be found.
+		func(pkg *packages.Package) bool { return pkg.IllTyped },
+		func(pkg *packages.Package) {
+			for _, err := range pkg.Errors {
```
golangcibot: should replace loop with errors = append(errors, pkg.Errors...) (from `megacheck`)

jirfag:
@rski thank you! Please, sign CLA and I will merge it

rski:
@jirfag cool, I'm trying to get approval from my company for the CLA :)

jirfag:
@rski how can we improve CLA or licensing to ease such approvals?

rski:
no idea :)
I'm waiting for someone to review the CLA, which is required for all CLAs. I'll ask when they get back to me about it

jirfag:
I've made own fix in https://github.com/golangci/golangci-lint/pull/363 to speed up things. My fix is slightly different and doesn't conflict with your code on copyright. Closing it, please, report if my fix didn't help.

PR #354, opened 2019-01-08
Remove stylecheck from README.md

cflewis:
stylecheck is no longer supported:

```shell
$ golangci-lint --version
golangci-lint has version 1.12.5 built from 609de32 on 2018-12-23T09:42:24Z
$ golangci-lint help linters | grep stylecheck
$
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,6 @@ $ golangci-lint help linters
 ...
 Disabled by default linters:
 golint: Golint differs from gofmt. Gofmt reformats Go source code, whereas golint prints out style mistakes [fast: true]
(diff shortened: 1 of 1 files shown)


cflewis:
I don't think I am allowed to sign the CLA. Sorry!

jirfag:
hi! Anyway thank you :)
why it's no longer supported? where to read about it?

cflewis:
I don't know, it just isn't in the release as far as I can see. I would be happy for it to come back, but my guess was it was removed on purpose but the documentation was out of date.

Issue #344, opened 2019-01-01
golangci/go-misc: checksum mismatch


ml-:
When trying to build the project, there is a checksum mismatch when dependencies are retrieved.
I tried this on my own system, as well as on the golang:1.11.4 docker image.

**$** `go build -o golangci-lint ./cmd/golangci-lint/main.go`

**Error:**
```
go: verifying github.com/golangci/go-misc@v0.0.0-20180628070357-927a3d87b613: checksum mismatch
	downloaded: h1:9kfjN3AdxcbsZBf8NjltjWihK2QfBBBZuv91cMFfDHw=
	go.sum:     h1:WadunOE/TeHR8U7f0TXiJACHeU3cuFOXuKafw4rozqU=
```

**Info:**
System: Arch Linux
`go version go1.11.4 linux/amd64`

```
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/ml/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/ml/.local/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/lib/go"
GOTMPDIR=""
GOTOOLDIR="/usr/lib/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/ml/code/golangci-lint/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build893572642=/tmp/go-build -gno-record-gcc-switches"
```

mbana:
I am getting the **_same_** error but in a different context; I am trying to place all the tools that I use in `tools/tools.go` file.

**Sample Repo:**
https://github.com/banaio/go_samples/tree/master/how_can_i_track_tool_dependencies_for_a_module
```sh
git clone git@github.com:banaio/go_samples.git
cd go_samples/how_can_i_track_tool_dependencies_for_a_module
make
```

**Error Log:**
[...]

PR #378, opened 2019-01-26
Fix checksum error in go.sum

y0ssar1an:
`go.sum` contained the wrong checksum for the `go-misc` repo. I fixed it like this:
```sh
> rm go.sum
> GO111MODULE=on go mod tidy
```
I'm not sure what this means or how that happens. I'm using Go 1.11.5.


y0ssar1an:
I'm not sure what's causing this. Clearing out the module cache on my system (`go clean -modcache`)  and rerunning those commands changes the checksum too. Sorry for the noise!

jirfag:
see https://github.com/golangci/golangci-lint/issues/344 for details and links

Issue #367, opened 2019-01-21
unparam false positive since upgrading to 1.13


mastertinner:
`unparam` reports false positives since upgrading to 1.13. Probably they are related to https://github.com/mvdan/unparam/issues/36.

Please include the following information:
1. Version of golangci-lint: `golangci-lint --version` (or git commit if you don't use binary distribution)
```shell
golangci-lint has version 1.13 built from 2192097 on 2019-01-21T07:58:24Z
```
2. Config file: `cat .golangci.yml`
```yaml
linters:
  enable-all: true
  disable:
    - lll
    - dupl
    - scopelint

linters-settings:
  govet:
    check-shadowing: true
  maligned:
    suggest-new: true
  misspell:
    locale: US
```
3. Go environment: `go version && go env`
```shell
go version go1.11.4 darwin/amd64
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/tobi/Library/Caches/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOOS="darwin"
GOPATH="/Users/tobi/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/Cellar/go/1.11.4/libexec"
GOTMPDIR=""
GOTOOLDIR="/usr/local/Cellar/go/1.11.4/libexec/pkg/tool/darwin_amd64"
GCCGO="gccgo"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/Users/tobi/go/src/bitbucket.org/medisante/device-metrics/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=/var/folders/vl/n1hn_wj96dx0v7l8nm0cgbwr0000gn/T/go-build062638386=/tmp/go-build -gno-record-gcc-switches -fno-common"
```
4. Verbose output of running: `golangci-lint run -v`
```shell
[...]

PR #381, opened 2019-01-27
unparam: update unparam, and go mod vendor

chappjc:
This updates the unparam linter to the current version that has fewer
false positives. The go.{mod,sum} files and the vendor folder are
updated as follows:

export GO111MODULE=on
go get -u mvdan.cc/unparam
go mod tidy
go mod vendor

The unparam callgraph algorithm is no longer selectable. Ref:
https://github.com/mvdan/unparam/commit/e6a6d1c5
This is the reason for removing the `c.CallgraphAlgorithm(us.Algo)` line
from pkg/golinters/unparam.go.


chappjc:
I believe this resovles Issue https://github.com/golangci/golangci-lint/issues/367

jirfag:
thank you!

PR #390, opened 2019-02-09
fix #320: implement exclude rules

### LINKED ISSUE #320: Feature request: skip files for a specific linter.
It would be useful to have an option to skip files for a specific linter.

A use-case: do not run `unparam` for `*_test.go`, as stub functions defined in tests are triggering false positives.

Version: v1.12.2

`.golangci.yml`:
```
run:
  deadline: 2m

issues:
  max-issues-per-linter: 0
  max-same-issues: 0
  exclude:
    # This project does not have a rule to avoid unkeyed fields
    - composite literal uses unkeyed fields
    # This project does not require all parameters to be used
    - "`.*` - `.*` is unused"

linters:
  enable-all: true
  disable:
    # This project does not have a rule to have no globals
    - gochecknoglobals
    # This project does not have a rule to have no inits
    - gochecknoinits
    # This test has too many false positives
    - gocyclo
    # This project does not have a standard line length
    - lll
    # This project does not have a rule to forbid naked returns
    - nakedret
    # This project does not care about optimizing out few bytes of memory
    - maligned
```

ernado:
See #320 

Feature allows filtering issues by file path, issue text and linter(s) simultaneously.

Currently it is forbidden to use it for filtering by one parameter, because
it is already implemented as "exclude" or "disable" config.

I've tested this feature on one of my projects which I [can't migrate](https://github.com/gortc/stun/blob/f3548a08bb362e554b57288aa2f4c4cc59173171/Makefile#L30) from gometalinter because of this issue and it finally worked for me :)

Example configuration:
```yml
issues:
  exclude-rules:
    # Exclude some linters from running on tests files.
    - path: _test\.go
      linters:
        - gocyclo
        - errcheck
        - dupl
        - gosec

    # Ease some gocritic warnings on test files.
    - path: _test\.go
      text: "(unnamedResult|exitAfterDefer)"
      linters:
        - gocritic

    # Exclude known linters from partially hard-vendored code,
    # which is impossible to exclude via "nolint" comments.
    - path: internal/hmac/
      linters:
        - gosec
    - path: internal/hmac/
      text: "Write\\` is not checked"
      linters:
        - errcheck

    # Ease linting on benchmarking code.
    - path: cmd/stun-bench/
      linters:
        - gosec
        - errcheck
  enable-all: true
  disable:
    - prealloc
    - gochecknoglobals
```


jirfag:
wow, it's the cool configuration!
but I need some time to test it. Also, I need update docs for it

ernado:
Thank you!
If you need something to be changed - just leave a comment, I'll update the implementation and tests.

Also I'm still not sure about the way how it is configured now, my concerns are mostly about existing ways to exclude linters and issues.

jirfag:
thank you very much!

Issue #393, opened 2019-02-13
Docker image: would you consider moving the binary elsewhere?


fsouza:
Using the Docker image in a Drone CI pipeline, I noticed that due to the binary being written to ``$GOPATH/bin``, whenever using ``$GOPATH`` as the Drone workspace (to persist dependencies between steps), the binary becomes unavailable because Drone mounts $GOPATH as a volume.

One solution is to not use ``/go`` as the $GOPATH, but that would require either setting the environment variable in every step or customizing the images we use in every step. So the solution was to create a Docker image for golangci-lint that installs the binary to ``/bin/`` instead of ``$GOPATH/bin``. Would a PR with that change be accepted? :)

Some links:

- Public Drone pipeline (it's also used in a private Drone instance): https://github.com/fsouza/s3-upload-proxy/blob/master/.drone.yml
- Fork of the Docker image: https://github.com/fsouza/golangci-lint-docker/blob/master/Dockerfile

PR #396, opened 2019-02-15
Dockerfile: move binary to /bin/

fsouza:
This addresses #393.


jirfag:
@fsouza thank you!
@AnikHasibul thank you for suggestion I've fixed it and placed binary in `/usr/bin`

PR #417, opened 2019-03-04
Fix AST cache key consistency

tpounds:
After upgrading to v1.15.0 I started seeing some strange AST cache resolution errors.

```
Can't run linter golint: no AST for file /path/to/foo.go in cache: {...}

Can't process result by autogenerated_exclude processor: can't filter issue result.Issue{...}: can't parse file /path/to/foo.go: <nil>"

Can't process result by nolint processor: can't filter issue result.Issue{...}: can't parse file /path/to/foo.go: <nil>, astcache is [...]
```

I've narrowed it down to commit 96af9582052f62722593d4085f2357df508a6310 which changed the symlink resolution behavior. This change fixes a key consistency issue by ensuring all loads/stores accessing the AST cache use the normalized file path.

### DIFF
--- a/pkg/lint/astcache/astcache.go
+++ b/pkg/lint/astcache/astcache.go
@@ -129,10 +129,12 @@ func (c *Cache) loadFromPackage(pkg *packages.Package) {
 			continue
 		}
 
-		c.m[pos.Filename] = &File{
+		filePath := c.normalizeFilename(pos.Filename)
+
+		c.m[filePath] = &File{
 			F:    f,
 			Fset: pkg.Fset,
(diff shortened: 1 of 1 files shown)


tpounds:
The Go 1.12 build failure is unrelated to this change and appears to be related to removal of the errchk script mentioned here https://github.com/golang/go/issues/25669 and https://github.com/golang/go/commit/02fecd33f608e3a2f11fcee424d55232f08c28cd

jirfag:
@tpounds thank you!
I've fixed the issue with `errchk` in master

PR #435, opened 2019-03-13
Add --color flag to force colored output on/off

benpaxton-hf:
fatih/color disables color output automatically if the output is not a tty, and expects applications to override this if required. In Jenkins the output is not a tty, but it does understand color codes.


jirfag:
thank you @benpaxton-hf !

Issue #433, opened 2019-03-13
linter should ignore C/C++ sources


vitalyisaev2:
It seems like linter tries to analyze `*.cc` files that may occur in `cgo` projects. Since it treats them just like common `*.go` files, it causes false arams.

main.go:
```go
package main

import (
	"fmt"

	"gopkg.in/kothar/brotli-go.v0/enc"
)

func main() {
	params := enc.NewBrotliParams()
	fmt.Println(params)
}
```
This file can be compiled and run, but doesn't pass typecheck:
```bash
➜  linter-issue go run .
&{{0 11 22 0 true false false true}}
➜  linter-issue golangci-lint run .   
main.go:1: /home/isaev/go/src/gopkg.in/kothar/brotli-go.v0/enc/backward_references.cc:9:1: illegal character U+0023 '#' (typecheck)
package main
```

1. Version of golangci-lint: `golangci-lint --version` (or git commit if you don't use binary distribution)
```7514bf823931b8c777a68ba7a7af8e5217fff191```

2. Config file: `cat .golangci.yml`
No config file is used.

3. Go environment: `go version && go env`
```
go version go1.12 linux/amd64
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/isaev/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/isaev/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build596847704=/tmp/go-build -gno-record-gcc-switches"
```

4. Verbose output of running: `golangci-lint run -v`
```bash
golangci-lint run -v .
[...]

PR #438, opened 2019-03-14
Update golang.org/x/tools package to fix false IllTyped error

vitalyisaev2:
Hi, please consider this small PR fixing #433. Actually #433 was caused by outdated `golang.org/x/tools` package in `vendor`. To make sure: 
1. `go get gopkg.in/kothar/brotli-go.v0`
2. Put this small file somewhere inside the `github.com/golangci/golangci-lint` to make `vendor` work and run it:

```go
package main

import (
	"fmt"

	"golang.org/x/tools/go/packages"
)

func main() {
	conf := &packages.Config{
		Mode:  packages.LoadAllSyntax,
		Tests: true,
	}
	result, _ := packages.Load(conf, "gopkg.in/kothar/brotli-go.v0/enc")
	fmt.Println(result[0].IllTyped)
}
```

On the `master` branch `true` will be printed, while on the current branch the response is `false` which means that `gopkg.in/kothar/brotli-go.v0/enc`  is not considered as `IllTyped` anymore. This will prevent `typecheck` lnter from emitting false alarm about bad package syntax. Test can be added if necessary.


jirfag:
@vitalyisaev2 thank you!

PR #448, opened 2019-03-19
Update go version to 1.12 in Dockerfile

jumpeiMano:
Hi! I always use golangci-lint with the docker image.

I edited Dockerfile to update go version to 1.12
I would be glad if you reviewed and merged this request.

Thanks!


jirfag:
thank you!

Issue #458, opened 2019-03-23
Makefile: Add target build


teawater:
kata-containers's ci will re-build and re-install golangci-lint every
time it did the staic check.
So run test of golangci-lint with build will make it slow.

PR #459, opened 2019-03-23
Makefile: Add target build

teawater:
kata-containers's ci will re-build and re-install golangci-lint every
time it did the staic check.
So run test of golangci-lint with build will make it slow.
Add target build to make golangci-lint just build but not test.

Fixes: #458

Signed-off-by: Hui Zhu <teawater@hyper.sh>

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,11 @@
-test:
-	go build -o golangci-lint ./cmd/golangci-lint
+test: build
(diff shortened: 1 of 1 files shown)


jirfag:
I didn't understand what did you mean but I've merged it. I guess you need to install golangci-lint as a binary release.

PR #442, opened 2019-03-17
Support excluding issues by source line regexp

jirfag:
See issues.exclude-rules[i].source.
Also introduced file data and file lines cache.

PR #465, opened 2019-03-28
Lll: add regexp excludes

francoispqt:
Hi, 

I'd like to add excludes to `lll` so that if a line longer than the maxLineLen matches one of the excludes regexp, it won't be considered an issue.

Example config:
```yaml
lll:
    line-length: 80
    tab-width: 1
    excludes: 
    - "`[A-z0-9]+:\".*.\"`"
```

### DIFF
--- a/pkg/golinters/lll_test.go
+++ b/pkg/golinters/lll_test.go
@@ -1,0 +1,128 @@
+package golinters
+
+import (
+	"go/token"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/golangci/golangci-lint/pkg/result"
+	"github.com/stretchr/testify/assert"
+)
+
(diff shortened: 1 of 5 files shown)


pkg/golinters/lll_test.go
```diff
@@ -0,0 +1,128 @@
+package golinters
+
+import (
+	"go/token"
+	"io/ioutil"
+	"os"
+	"testing"
+
+	"github.com/golangci/golangci-lint/pkg/result"
```
golangcibot: File is not `goimports`-ed (from `goimports`)

francoispqt:
I just noticed that one https://github.com/golangci/golangci-lint/pull/442, I'm closing my PR.

Issue #446, opened 2019-03-18
go vet: allow disabling composites analysis


alessio:
I'd be nice to be able to disable `go vet`'s `-composites` analysis.

Thanks for considering.

----

1. Version of golangci-lint: `golangci-lint --version` (or git commit if you don't use binary distribution)

```
$ golangci-lint --version
golangci-lint has version 1.15.0 built from 901cf25 on 2019-02-18T08:24:43Z
```

2. Config file: `cat .golangci.yml`

```yaml
linters:
  disable-all: true
  enable:
    - errcheck
    - golint
    - vet
    - ineffassign
    - unconvert
    - misspell
linters-settings:
  gocyclo:
    min-complexity: 11
  errcheck:
    ignore: fmt:.*,io/ioutil:^Read.*,github.com/spf13/cobra:MarkFlagRequired,github.com/spf13/viper:BindPFlag
  golint:
    min-confidence: 1.1
run:
  tests: false
```

3. Go environment: `go version && go env`

```
$ go version && go env
go version go1.12.1 linux/amd64
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/alessio/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/alessio/work/tendermint"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/home/alessio/work/tendermint/src/github.com/cosmos/cosmos-sdk/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build889440970=/tmp/go-build -gno-record-gcc-switches"
```

4. Verbose output of running: `golangci-lint run -v`

```
$ golangci-lint run -v
[...]

PR #486, opened 2019-04-08
Add check-composites flag to Govet

alessio:
Ref #446


jirfag:
hi, thank you for the pull request!
I think we need a more generic way to enable/disable govet checks.  `CheckShadowing` flag exists there only for backward compatibility.

jirfag:
closing it because I would like to see a more generic support

Issue #493, opened 2019-04-11
Cannot exclude test files when using errcheck


alaypatel07:
1. Version of golangci-lint: `golangci-lint --version` (or git commit if you don't use binary distribution)
Commit: de1d1ad903cdfdfc577ecb08d5027576e70ed2f0
2. Config file: `cat /tmp/.golangci-lint-config.yml`
```
linters-settings:
  linters-settings:
    errcheck:
      without_tests: true
  dupl:
    threshold: 400
    min-len: 3
  lll:
    line-length: 170
    tab-width: 1'
```
3. Go environment: `go version && go env`
```go version go1.11.8 linux/amd64
GOARCH="amd64"
GOBIN=""
GOCACHE="/root/.cache/go-build"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/go"
GOPROXY=""
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build208835018=/tmp/go-build -gno-record-gcc-switches"
```
4. Verbose output of running: 
```golangci-lint run --disable-all --config /tmp/.golangci-lint-config.yml --deadline 5m --enable=misspell --enable=structcheck --enable=golint --enable=deadcode --enable=goimports --enable=errcheck --en
able=varcheck --enable=goconst --enable=unparam --enable=ineffassign --enable=nakedret --enable=inter
facer --enable=misspell --enable=gocyclo --enable=lll --enable=dupl --skip-dirs=atomic --enable=goimp
orts ./pkg/... ./examples/... .
pkg/builder/builder_suite_test.go:69:14: Error return value of `testenv.Stop` is not checked (errchec
k)
        testenv.Stop()
                    ^
[...]

PR #496, opened 2019-04-12
Add support to ignore test files when using errcheck

alaypatel07:
This PR solves #493  by adding `without-tests` field to `errcheck`. 

Depends on golangci/errcheck#2

### DIFF
--- a/README.md
+++ b/README.md
@@ -469,6 +469,7 @@ Flags:
       --no-config                   Don't read config
       --skip-dirs strings           Regexps of directories to skip
       --skip-files strings          Regexps of files to skip
+      --errcheck.without-tests      Boolean value to include tests in errcheck or not
   -E, --enable strings              Enable specific linter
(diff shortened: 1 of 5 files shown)


jirfag:
hi, thank you for the pull request!
we already have settings for ignoring test files for a specific linter:
```yaml
  exclude-rules:
    # Exclude some linters from running on tests files.
    - path: _test\.go
      linters:
        - errcheck
```

can you use this solution?

alaypatel07:
@jirfag Thanks for the alternative path, it worked for me.

I guess that makes this PR redundant, closing it and the issue. Thanks

mattbirman:
hey @jirfag, I am running into the same issue and this solution is not working for me.

in my .golangci.yml file I have the following:

```  
exclude-rules:
    - path: _test\.go
      linters:
        - gosec
```
but golangci-lint still fails on `gosec` lint errors in `*.test.go` files

PR #498, opened 2019-04-17
Fix max-issues-per-linter variable name in example

vovapi:
`.golangci.example.yml` has wrong name for `max-issues-per-linter` variable. 
So I've fixed it


jirfag:
thank you!

PR #501, opened 2019-04-19
Enable auto-fixing when running via pre-commit

aneeshusa:

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,7 +1,7 @@
 - id: golangci-lint
   name: golangci-lint
   description: Fast linters runner for Go.
-  entry: golangci-lint run
+  entry: golangci-lint run --fix
   types: [go]
   language: golang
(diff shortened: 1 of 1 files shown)


aneeshusa:
@diegs pointed out this great feature in v1.15.0 - we just updated everyone at our company to 1.16.0 and I'm super excited about `--fix`!

jirfag:
thank you!
I'm not sure is it ok to merge now because old versions of golangci-lint without `--fix` support will fail. Maybe we should make another target like `golangci-lint-autofix`?

aneeshusa:
That shouldn't be an issue as pre-commit will use the .pre-commit-hooks.yaml (in this repo) that exists in the golangci-lint version the user specifies in their .pre-commit-config.yaml. So, if they use an older version of golangci-lint, `--fix` won't be specified and it should work fine.

jirfag:
ok, thank you

PR #506, opened 2019-04-20
Add strict mode for go vet shadow check

hummerd:
Add "shadow-strict" configuration option for go vet.

### DIFF
--- a/pkg/golinters/govet.go
+++ b/pkg/golinters/govet.go
@@ -1,6 +1,8 @@
 package golinters
 
 import (
+	"log"
+
 	"golang.org/x/tools/go/analysis"
 
 	"github.com/golangci/golangci-lint/pkg/config"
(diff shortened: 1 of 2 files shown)


pkg/golinters/govet.go
```diff
@@ -73,6 +73,9 @@ func NewGovet(cfg *config.GovetSettings) *goanalysis.Linter {
 	if cfg != nil {
 		if cfg.CheckShadowing {
 			analyzers = append(analyzers, shadow.Analyzer)
+			if cfg.ShadowStrict {
+				shadow.Analyzer.Flags.Set("strict", "true")
```
golangcibot: Error return value of `shadow.Analyzer.Flags.Set` is not checked (from `errcheck`)

hummerd: fixed

jirfag:
But I don't like the idea of adding every possible configuration to our code. Please, let's do it in a generic way, like with go-critic. `CheckShadowing` is the legacy just for compatibility.

hummerd:
I am ready to make it in any way :) What do you mean by "generic way" ? enabled-checks options?  
```
 enabled-checks:
      - shadowstrict
```
Or something else?

jirfag:
I've found that you don't even need to change code, we already have:
```yaml
    # settings per analyzer
    settings:
      printf: # analyzer name, run `go tool vet help` to see all analyzers
        funcs: # run `go tool vet help printf` to see available settings for `printf` analyzer
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Infof
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Warnf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Errorf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
```

and it works in a generic way. So you need something like
```yaml
    # settings per analyzer
    settings:
      shadow:
        strict: true
```

PR #520, opened 2019-04-26
fix: openapi generated by detection

23doors:
https://github.com/OpenAPITools/openapi-generator doesn't follow existing conventions (sadly)


23doors:
Turns out when using `-DwithGoCodegenComment` it is picked up by linter correctly

PR #525, opened 2019-04-30
Make install script work on Windows.

### LINKED ISSUE #472: golangci-lint.sh: crit uname_os_check 'MINGW64_NT-10.0' got converted to 'mingw64_nt-10.0' which is not a GOOS value
Running the install script on Azure Pipelines (Windows) returns this error:

```
golangci/golangci-lint crit uname_os_check 'MINGW64_NT-10.0' got converted to 'mingw64_nt-10.0' which is not a GOOS value. Please file bug at https://github.com/client9/shlib
```

I know it says to file a bug over in shlib, but [similar issues are unresolved](https://github.com/client9/shlib/issues/13), and this code lives in this repo anyway, not shlib.

Would be nice to have this work, as writing more platform-specific code to install a go program doesn't sound awesome. I'll probably just fall back to using `go get` for now.

Update: Fortunately the `go get` works! Any idea why it is not recommended? Seems much simpler...

AlekSi:
Closes #472.

See https://github.com/client9/shlib/issues/13 and https://github.com/goreleaser/godownloader/issues/98 for background and other links.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -172,7 +172,8 @@ log_crit() {
 uname_os() {
   os=$(uname -s | tr '[:upper:]' '[:lower:]')
   case "$os" in
-    msys_nt) os="windows" ;;
+    msys_nt*) os="windows" ;;
+    mingw*) os="windows" ;;
   esac
   echo "$os"
 }


syntaqx:
What can we do to get this merged in? Looks like golangci just needs a good kickin' as it had some issues on its side?

Also, I'm now an added contributor for github.com/client9/shlib so please ping me for any Windows related issues this project might need so users from both projects can be happy.

jirfag:
thank you! merged

PR #533, opened 2019-05-10
 Added "bodyclose" support

gsora:
This commit adds full support for bodyclose linter
(https://github.com/timakin/bodyclose), which checks if an `http.Body`
element is correctly closed after usage.

Since it can be used via `go/analysis`, I followed the `govet` example
as suggested by https://github.com/golangci/golangci-lint/wiki/How-to-add-a-custom-linter.

This commit is fully tested, and contains a (flawed) test program which
calls `http.Get()` on `https://google.com` and does not closes its
corresponding `http.Body`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -204,6 +204,7 @@ and the following linters are disabled by default:
 $ golangci-lint help linters
 ...
 Disabled by default linters:
+bodyclose: bodyclose checks whether HTTP response body is closed successfully [fast: false, auto-fix: false]
 depguard: Go linter that checks if package imports are in a list of acceptable packages [fast: true, auto-fix: false]
 dupl: Tool for code clone detection [fast: true, auto-fix: false]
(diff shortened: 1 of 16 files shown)


gsora:
Uhm, I do run `goimports` through VSCode on save, I don't understand why those errors appear.

alexkohler:
@gsora - Perhaps you're using an older version of goimports? It could possibly be the import order. Typically I think goimports does standard library imports first.

gsora:
I checked and I'm using the latest version of `golang.org/x/tools/cmd/goimports`, which seems to be used by `https://github.com/golangci/gofmt/blob/master/goimports/golangci.go` too.

jirfag:
thank you for this contribution! it has some issues and I will fix them myself because it's my fault that it's being merged too long

PR #548, opened 2019-05-27
Typo GOCC to GOGC

Harsimran1:
Currently go garbage collector variable is written as GOCC in a link where it should be GOGC. Fixed the typo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -344,7 +344,7 @@ Read [this section](#internals) for details.
 
 ### Memory Usage of Golangci-lint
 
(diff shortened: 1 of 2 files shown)


Harsimran1:
Looks like most of the recent PR's are failing GolangCI check because of `.golangci.yml` config issue
`disabled checker dupImport doesn't exist` .

LukeShu:
Commit 692dacb773b703162c091c2d8c59f9cd2d6801db added `dupImport`, but since that commit isn't yet in a stable release, GolangCI isn't using it yet, so it complains.  I guess that's expected until the next release happens.

jirfag:
thank you!

PR #555, opened 2019-06-04
Makefile: Make it easier to re-generate everything

LukeShu:
When I start looking at contributing to a project, one of the first things I try to answer is "how do I re-build/generate everything that is generated; am I actually able to re-build this?"  This PR aims to lower that bar.

 - Modify `make check_generated` to check *all* generated files, not just `README.md`
 - Merge `make gen`, `make readme`, `make update_deps`, and `make assets` into a single `make generate`
 - Add `make clean` to remove non-commited (`.gitignore`d) files that the Makefile generates
 - Add `make maintainer-clean` to remove committed files that the Makefile generates

This mostly aims to document how everything was generated, and the steps necessary to re-generate the same files verbatim; the goal is mostly to figure out how to reproduce the files as they currently are.

To re-generate just an individual file, like `README.md`, you can just `make FILENAME`; `make README.md` will work as expected.

The Makefile now does more work; it downloads the "Dracula" iTerm2 theme for generating `demo.svg`, it installs the correct version of `mockgen` for generating `log_mock.go`, et cetera.  All of these "extra" things it puts in a `.gitignored` directory called `tools/`.  I tried to keep the Makefile readable and understandable, though.

I rolled back the version of `github.com/golang/mock` from v1.1.1 to v1.0.0, since that's the version used to generate `log_mock.go`, and mixing versions of `mockgen` outputs vs mock libraries seemed unwise.

I adjusted the test suite to work with `GO111MODULE=on`, which the Makefile sets.


LukeShu:
v3: Adjust the test suite to work with GO111MODULE=on

LukeShu:
The "GolangCI" check failure is "expected" and has nothing to do with the changes in this PR.

Commit 692dacb773b703162c091c2d8c59f9cd2d6801db added `dupImport`, but since that commit isn't yet in a stable release, GolangCI isn't using it yet, so it complains.  I guess that CI failure is expected until the next release happens.

jirfag:
the fantastic work, thank you very much!

PR #558, opened 2019-06-09
update x/tools

jirfag:
Don't merge until https://github.com/golang/go/issues/32457 will be fixed. After that need to recheck performance.


jirfag:
https://github.com/golang/go/issues/32538

jirfag:
I've used forked x/tools until https://github.com/golang/tools/pull/139 is merged

Issue #581, opened 2019-06-19
Failed to build golangci-lint v1.17.1 when enable GO111MODULE


liubog2008:
```
$ go build -v ./...
build github.com/golangci/golangci-lint/cmd/golangci-lint: cannot load github.com/go-critic/go-critic/checkers: zip for github.com/go-critic/go-critic@v0.0.0-20181204210945-1df300866540 has unexpected file github.com/go-critic/go-critic@v0.0.0-20190526074819-1df300866540/.gitignore
```

go version: go1.12.5 linux/amd64
golangci-lint version: v1.17.1

fhs:
Go 1.13 gives a slightly better error:
```
$ GOPROXY=direct go get github.com/golangci/golangci-lint@latest
go get: github.com/golangci/golangci-lint@v1.17.1 requires
	github.com/go-critic/go-critic@v0.0.0-20181204210945-1df300866540: invalid pseudo-version: does not match version-control timestamp (2019-05-26T07:48:19Z)
```

pierrre:
I've submitted a PR that fixes the issue
https://github.com/golangci/golangci-lint/pull/605

liubog2008:
I have fixed it long long time ago = = @pierrre See #583

PR #583, opened 2019-06-19
fix go mod problem

liubog2008:
See #581


Helcaraxan:
FWIW. There is a more recent PR (#605) that should be preferred here as it upgrades to a more recent version of `go-critic` that contains other relevant fixes as well.

jirfag:
sorry for the slow response and thank you for the pull request, I will merge #605 and close it

PR #585, opened 2019-06-23
Fix a false-positive from 'unused'

LukeShu:
Add failing testcase for false-positive from 'unused'. This false-positive is not present in the upstream stand-alone 'unused' 2019.1.1 program that golangci-lint uses.

I am not (yet?) sure how to fix the problem.

edit: I've managed to fix the issue:

pkg/lint.ContextLoader.filterPackages() did two things:
 1. It removed synthetic "testmain" packages (packages with .Name=="main" and .PkgPath ending with ".test")
 2. It removed pruned subsumed copies of packages; if a package with files "a.go" and "a_test.go", it results in packages.Load giving us two packages:
      - ID=".../a" GoFiles=[a.go]
      - ID=".../a [.../a.test]" GoFiles=[a.go a_test.go]
    The first package is subsumed in the second, and leaving it around results in duplicated work, and confuses the 'deadcode' linter.

However, the 'unused' linter relies on both the ".../a" and ".../a [.../a.test]" packages being present.  Pruning them causes it to panic in some situations, which lead to this workaround: https://github.com/golangci/go-tools/commit/af6baa5dc196960e7c24eea96da833727ed58f7f

While that workaround got it to not panic, it causes incorrect results.

So, split filterPackages() in to two functions: filterTestMainPackages() and filterDuplicatePackages().  The linter.Context.Packages list only gets filterTestMainPackages() called on it, while linter.Context.Program and linter.Context.SSAProgram get both filters applied.

With the source of the panic fixed, roll back a few now-unnecessary commits in go-tools.


LukeShu:
Ping

LukeShu:
Ping.

jirfag:
@LukeShu sorry for the slow response.
Thank you very much for this contribution, I'll merge it as it passes CI.
Also, I will make optimization in a separate commit for using deduplicated packages for `golint` and some other linters.

PR #589, opened 2019-06-26
Update depguard version to 1.0.0 (Performance improvements)

hbandura:
depguard was patched to improve performance on large projects, by caching checks for root packages from the import paths.

This PR merely updates depguard to version 1.0.0, and then `make vendor` was run.

/edit

`make test` was also run without issues


jirfag:
thank you, It's a super-important fix for golangci-lint to able to work with go modules.

Issue #595, opened 2019-07-01
go get cmd/golangci-lint fails on Go 1.13 (tip) which uses proxy.golang.org (HTTP error 410 Gone)


mewmew:
1. Go environment: `go version && go env`

`go version` output:

```
go version devel +bc593eac2d Mon Jul 1 14:45:49 2019 +0000 linux/amd64
```

`go env` output:

```
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/u/.cache/go-build"
GOENV="/home/u/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/u/goget:/home/u/Desktop/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/home/u/go1.13"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/home/u/go1.13/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/dev/null"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build741963370=/tmp/go-build -gno-record-gcc-switches"
```

On the latest version of Go (which uses `proxy.golang.org`) it seems like `go get` is resulting in a 410 HTTP error.

```
$ go get github.com/golangci/golangci-lint/cmd/golangci-lint
go: finding github.com/golangci/golangci-lint/cmd/golangci-lint latest
verifying github.com/go-critic/go-critic@v0.0.0-20181204210945-1df300866540/go.mod: github.com/go-critic/go-critic@v0.0.0-20181204210945-1df300866540/go.mod: reading https://sum.golang.org/lookup/github.com/go-critic/go-critic@v0.0.0-20181204210945-1df300866540: 410 Gone
```

mewmew:
[...]

PR #594, opened 2019-07-01
Fix panic in filecache

be-b10g:
Hi!

We've got panic happened while liniting our repository with golangci-lint. The reason is concurrent writes in filecache map. I tried to fix it via using sync.Map instead of map, so there must be no concurrent [read/]writes. Also there might be possible race in linecache because there is no any mutex either. 

<details>
  <summary>Panic</summary>

```
if [ "Darwin" = "Darwin" ]; then export PKG_CONFIG_PATH=/opt/X11/lib/pkgconfig/; else export PKG_CONFIG_PATH=/usr/local/lib/pkgconfig/; fi && \
	golangci-lint run -c golangci_extended.yml -v  ./src/mail/...
level=info msg="[config_reader] Used config file golangci_extended.yml"
level=info msg="[lintersdb] Active 29 linters: [bodyclose deadcode depguard dupl errcheck goconst gocritic gocyclo gofmt goimports golint gosec gosimple govet ineffassign interfacer lll misspell nakedret prealloc scopelint staticcheck structcheck stylecheck typecheck unconvert unparam unused varcheck]"
level=info msg="[lintersdb] Optimized sublinters [staticcheck gosimple unused stylecheck] into metalinter megacheck"
level=info msg="[loader] Go packages loading at mode load types and syntax took 19.223275966s"
level=info msg="[loader] SSA repr building timing: packages building 266.546786ms, total 6.847481739s"
src/mail/api/clickerproxy/vid.go:59:42: bodyclose: response body must be closed (bodyclose)
	resp, err := requests.BaseOnce().Request(ctx, req, requests.Params{
	                                        ^
fatal error: concurrent map writes

goroutine 10801 [running]:
runtime.throw(0x1a8048f, 0x15)
	/usr/local/go/src/runtime/panic.go:617 +0x72 fp=0xc0020916a0 sp=0xc002091670 pc=0x102e862
runtime.mapassign_faststr(0x18efd40, 0xc000468420, 0xc000517560, 0x51, 0x1742)
	/usr/local/go/src/runtime/map_faststr.go:211 +0x42a fp=0xc002091708 sp=0xc0020916a0 pc=0x1014e5a
github.com/golangci/golangci-lint/pkg/fsutils.(*FileCache).GetFileBytes(0xc00046e0c8, 0xc000517560, 0x51, 0xc166d77500, 0x991, 0xc034f46c80, 0x0, 0x8)
[...]

PR #605, opened 2019-07-10
fix invalid dependencies

### LINKED ISSUE #581: Failed to build golangci-lint v1.17.1 when enable GO111MODULE

pierrre:
Some dependencies have invalid date.

Fixes #581 #595


pierrre:
I don't know why some vendors have been deleted.

pierrre:
Strangely, the vendored dependencies are not the same if I'm using Go 1.13beta or 1.12.x.
Now they're added correctly.

pierrre:
Is someone available to review this PR ?

ww-daniel-mora:
I am also experiencing this issue

Helcaraxan:
Yes would be great if this could indeed be merged and a `1.17.2` patch release tagged (which would also contain the fix for #612 (i.e. #594).

FWIW. There are 2 other PRs (#583, #614) that address the same issue but I think this is the one that should be merged given it's the only one that actually updates `go-critic` and only `go-critic` to a recent commit.

pierrre:
@dajohi which version of Go are you using ?
I'm using 1.13beta1, and it was working fine.
Can someone else confirm ?

bcmills:
@pierrre, the version check is performed when downloading the module from version control. If an invalid version is already present in your module cache, it will continue to be available to builds using that cache.

To test this change, you'll probably want to start from a clean module cache (either run `go clean -modcache` first or set `GOPATH=$(mktemp -d)`) and set `GOPROXY=direct`.

pierrre:
Hi, sorry for the delay.
It should be fixed now.

pierrre:
The build doesn't pass, but I don't understand the error.
Can someone help me ?

pierrre:
It seems that `docs/demo.svg` is generated differently.
But it's not related to my PR.
I guess that `master` is also broken :man_shrugging:

Helcaraxan:
Just checking: did you run `go run scripts/gen_readme/main.go`? I've just run it on `master` and there does not seem to be a diff which would imply it is not broken.

CI seems to agree with me [as well](https://travis-ci.com/golangci/golangci-lint/branches).

pierrre:
```

**Comment by pierrre** [comment]
```
➜  golangci-lint git:(bug/fix-581) go run scripts/gen_readme/main.go
2019/07/26 10:35:10 Successfully generated README.md
➜  golangci-lint git:(bug/fix-581) git --no-pager diff
```

there is no diff on my computer

**Comment by Helcaraxan** [comment]
Yes. Apologies. Edited my message above as well after I tested your branch myself. 👍 

Having a look if I can figure out what's going on.

**Comment by Helcaraxan** [comment]
Ah got it, although not the root cause. You can regenerate the file locally by running:
```shell
make -B docs/demo.svg
```
The `-B` forces to regenerate the thing despite `make` thinking it's up-to-date. And it actually generates a diff.

My guess: given this relies on `npm` black magic I suppose some thing got updated somewhere and induced this hidden change. The `npm` environment is not checked / understood by `make` and thus is not picked-up as "changed".

pierrre:
I'm sorry, but my PR is only about broken Go modules dependencies.
It wouldn't be correct to submit a change related to `docs/demo.svg`.

Helcaraxan:
I'll have a look if I can make a sensible PR to fix it separately. 👍

Helcaraxan:
@pierrre (and maintainers) proposed fix is up in PR #625. Once that one is merged we can rebase this PR on top of `master` and the CI should pass.

dahu33:
Any update from maintainer?

Helcaraxan:
No. The maintainer has been inactive for 6 weeks now and there's no news available at the moment. 

See #647.

go.mod
```diff
@@ -1,11 +1,13 @@
 module github.com/golangci/golangci-lint
 
+go 1.11
```
bmhatfield: Why does this PR specify `go 1.11`? I actively don't want this.

Helcaraxan: Could you explain what issue this raises for you?

This statement is not a hard requirement. It's merely an indication used by the `go` toolchain and you should not need to worry about it.

Tl;dr: This tells the Go compiler that if the build fails and you are using a < 1.11 version it will tell you it might be due to a too-low Go version... however given that versions < 1.11 are not module aware they will not even bother reading this.

Helcaraxan: For reference see the [Go 1.12 release notes](https://golang.org/doc/go1.12#modules).

bmhatfield: My understanding is that this specifier causes the files in the module to be built with an older version of Go?

As mentioned here: https://golang.org/doc/go1.12#modules

> The go directive in a go.mod file now indicates the version of the language used by the files within that module.

Helcaraxan: It's actually a bit different.

> the version of the language used by the files within that module

refers to the version for which the code was written (as it might be using a new function introduced in that version... but not necessarily), not the one with which the files should be built. As the sentence following the one you quote indicates:

>  If the go directive for a module specifies a version newer than the toolchain in use, the go command will attempt to build the packages regardless, and will note the mismatch only if that build fails.

bmhatfield: I still feel that the writing is a bit ambiguous here, but I think I see what you're saying. I don't object if it's true that it's only a guideline. I just don't want tools like this to force-build on an old version, if that won't happen, I retract my objection :-)

Helcaraxan: You will always have full control over your build process. There is no version enforcement going, I can guarantee you that. 🙂 

I'll forward the point about the feeling of ambiguity around the phrasing. I can imagine myself in your place and understand it can introduce confusion for people not familiar with the relatively new syntax of `go.mod` files.

jirfag:
Guys sorry for the slow response. @pierrre thank you for the fix. I've merged it.
Some CI flaps, I will fix them later.

pierrre:
Thank you !

diovann:
Microsoft Windows [Version 6.2.9200]
(c) 2012 Microsoft Corporation. All rights reserved.

C:\Users\diovann>cd c:\go-work\HOME\core\

c:\go-work\HOME\core>go build
go: github.com/golangci/golangci-lint@v1.17.1 requires
        github.com/go-critic/go-critic@v0.0.0-20181204210945-1df300866540: inval
id pseudo-version: does not match version-control timestamp (2019-05-26T07:48:19
Z)

c:\go-work\HOME\core>



how to fix this error??

Helcaraxan:
@diovann you need to upgrade your dependency on `golangci-lint` to `v1.18.0`. That version contains the fix.

diovann:
how to upgrade golangci-lint to v1.18.0 ??

Helcaraxan:
It depends.

1. If you are simply trying to build `golangci-lint`, and supposing you have cloned it as a Git repository, in the same folder do `git fetch && git checkout v0.18.0` followed by `go build` or `go install`.
2. If `golangci-lint` is a dependency of your own project then `go get -u github.com/golangci/golangci-lint@v1.18.0 && go mod tidy` should do it.

NB: If it's someone else's project you can still do 2. but you will need to raise a pull request against the project in question, or you need to file an issue with the project for the maintainers to upgrade their dependency.

diovann:
@helcaraxan thanks for this bro. !!

PR #607, opened 2019-07-11
Provide pre-built binary for ARM arch

### LINKED ISSUE #491: Provide pre-built binary for ARM arch
Currently, the [list of releases](https://github.com/golangci/golangci-lint/releases) only supports binaries for Windows, Linux and MacOSX. It would like to use the tool with ARM architecture (e.g. Raspberry Pi). Does anything speak against creating a binary for it with new releases going forward. The script at https://install.goreleaser.com/github.com/golangci/golangci-lint.sh should also support this type of architecture.

techknowlogick:
Fix #491

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -14,7 +14,12 @@ builds:
       - linux
     goarch:
       - amd64
+      - arm64
+      - arm
       - 386
+    goarm:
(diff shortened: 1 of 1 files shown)


jirfag:
thank you!

PR #614, opened 2019-07-17
update deps; fix go 1.13

dajohi:


Helcaraxan:
FWIW. There are already 2 other PRs (#583, #605) that address the same issue but without updating all dependencies (to keep the diff minimal).

jirfag:
Hi, thank you for the pull request! I've already merged a few pull requests fixing the same things.

PR #625, opened 2019-07-26
Make generation of demo.svg deterministic

Helcaraxan:
While working on #605 we found out that the PR's CI revealed an unexpected diff in `docs/demo.svg`. CI on `master` looked fine as well but I could reproduce the failure nonetheless locally by running `make -B docs/demo.svg`.

The rootcause is as follows:

1. The generation of `demo.svg` depends on a Node tool that is installed via `npm install svg-term-cli`. Underneath this the `npm` command performs its own recursive dependency resolution and fetches the latest version (that is compatible with any existing version constraints) of all required direct & indirect dependencies.

2. The latest CI run on `master` dates back to 13/14 July and passed correctly. Since then the `svgo` NPM package, a direct dependency of `svg-term-cli`, released a new minor version making the most recent version `1.3.0` instead of `1.2.2` which is still compatible with the `>=1.0.3` constraint defined by `svg-term-cli`. The move from `1.2.2` to `1.3.0` apparently introduced a change that resulted in a diff in the generated `docs/demo.svg` file.

The solution here (and which will help prevent this occurring in the future as well) is to fix the versions of all (in)direct dependencies used to install the `svg-term-cli` package.

This is achieved by creating a fake NPM package in the `tools` folder using a `package.json` with `svg-term-cli` as a dependency and a `package-lock.json` file to hardcode the required version for each dependency. This also allows us to run `npm ci` instead of `npm install svg-term-cli` which is designed specifically for the automated runs of a CI environment.

This will make the generation of `docs/demo.svg` deterministic and allows the safe upgrade of any of the required package versions in the future via code-reviewed updates to the `package.json` and `package-lock.json` files.


.travis.yml
```diff
@@ -1,11 +1,15 @@
-sudo: false
```
Helcaraxan: Removing this as it's deprecated according [to Travis documentation](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration#timeline---its-happening-fast).

pierrre:
Any news from the maintainers ?

Helcaraxan:
Unfortunately not. I can't seem to find any trace of their activity on the project over the past 3 weeks. Probably due to it being the summer holiday season, so quite understandable. 🙂

jirfag:
thank you!

Helcaraxan:
You're welcome. 🙂 Happy to see this getting merged!

PR #626, opened 2019-07-30
Fix install script on Windows

Ruluk:
`adjust_format()` should use OS and not ARCH. In Windows, `$OS` will be calculated to be `"windows"` while `$ARCH` will be `"386"` or `"amd64"` (most of the time). `$ARCH` will surely never be `"windows"`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -97,8 +97,8 @@ tag_to_version() {
   VERSION=${TAG#v}
 }
 adjust_format() {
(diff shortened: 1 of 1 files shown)


jirfag:
thank you!

jirfag:
And I've decided to update the script in https://github.com/golangci/golangci-lint/pull/672

PR #629, opened 2019-08-04
go.mod: fix timestamp on module replacement for bodyclose

shuLhan:
Building on master (d2b1eea2) result in the following error,

  go: github.com/golangci/bodyclose@v0.0.0-20190714214026-65da19158fa2: invalid pseudo-version: does not match version-control timestamp (2019-07-14T14:40:26Z)

This may be caused by manual timestamp using local zone instead of UTC
when adding replace directive _or_ somethings get forced push previously
on bodyclose repository _or_ bug on Go tools itself.

While at it add missing dependency: github.com/Quasilyte/go-consistent
which added automatically by Go tools.


pierrre:
Already fixed in #605

jirfag:
hi, thank you! but closing because already fixed

PR #630, opened 2019-08-05
fix #416: Add skip-dirs-use-default

### LINKED ISSUE #416: Allow lint of `examples` directory
Hey @jirfag 

golangci-lint is amazing, switching to it has made such a huge difference for my project. It is so nice to not manage all these linters on my own (and wasn't aware of how bad my code was before)

My repo [pions/webrtc](https://github.com/pions/webrtc) has an examples directory, and I would like to lint it. It is ignored by default [here](https://github.com/golangci/golangci-lint/blob/master/pkg/packages/skip.go#L22) and I haven't found a way to override it.

Are there any hacks I can do to make this happen though? I noticed the quality of my examples slipping without this. For now I can exec twice, but that puts a bigger burden on contributors. 

thanks!

Sean-Der:
Hopefully this is ok @jirfag!

This will make big difference for Pion. The more we crank up the linter the easier it gets to be a maintainer :) thanks again for golangci-lint, it has made a huge difference.

### DIFF
--- a/pkg/lint/runner.go
+++ b/pkg/lint/runner.go
@@ -48,8 +48,10 @@ func NewRunner(astCache *astcache.Cache, cfg *config.Config, log logutils.Log, g
 		return nil, err
 	}
 
-	skipDirs := append([]string{}, packages.StdExcludeDirRegexps...)
-	skipDirs = append(skipDirs, cfg.Run.SkipDirs...)
+	skipDirs := cfg.Run.SkipDirs
+	if cfg.Run.UseDefaultSkipDirs {
+		skipDirs = append(skipDirs, packages.StdExcludeDirRegexps...)
+	}
 	skipDirsProcessor, err := processors.NewSkipDirs(skipDirs, log.Child("skip dirs"), cfg.Run.Args)
 	if err != nil {
 		return nil, err
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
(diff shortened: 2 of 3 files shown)


Sean-Der:
This is blocked/failing because of https://github.com/golangci/golangci-lint/pull/625

pkg/lint/runner.go
```diff
@@ -48,8 +48,10 @@ func NewRunner(astCache *astcache.Cache, cfg *config.Config, log logutils.Log, g
 		return nil, err
 	}
 
-	skipDirs := append([]string{}, packages.StdExcludeDirRegexps...)
-	skipDirs = append(skipDirs, cfg.Run.SkipDirs...)
+	skipDirs := cfg.Run.SkipDirs
+	if cfg.Run.UseDefaultSkipDirs {
+		skipDirs = append([]string{}, packages.StdExcludeDirRegexps...)
```
ernado: Hey @Sean-Der, great to meet you  :)

Does it mean that `skip-dirs-use-default: true` (which will be true by default) will replace the explicit `skip-dirs` setting?
```go
	skipDirs := cfg.Run.SkipDirs // ["e2e", "foo/bar"]
	if cfg.Run.UseDefaultSkipDirs {
		skipDirs = append([]string{}, packages.StdExcludeDirRegexps...)
	}
	skipDirs // is now equal to packages.StdExcludeDirRegexps
```
Probably we want the `skipDirs = append(skipDirs, packages.StdExcludeDirRegexps...)` here?

jirfag:
Hi, thank you!
`README.md` needs to be regenerated, I will do it in master after mergin

PR #557, opened 2019-06-06
junit-xml output

soniah:
Squashing commits and doing PR off master not dev branch (versus https://github.com/golangci/golangci-lint/pull/540)

PR #632, opened 2019-08-06
Update format of junit xml output to mark failures as such

glenjamin:
As noted in https://github.com/golangci/golangci-lint/pull/557#issuecomment-518152477 - the current Junit xml output treats each lint issue as a passing test.

This PR expands the details and shuffles things around so that it shows up as a failing test instead.

There didn't seem to be any automated tests for this functionality, so I haven't added any new ones.

Here's a sample of what that looks like with CircleCI's junit parsing
https://circleci.com/gh/glenjamin/golangci-lint/5

I don't know which other popular tools do junit parsing like this - I based this implementation loosely off https://github.com/eslint/eslint/blob/master/lib/cli-engine/formatters/junit.js

### DIFF
--- a/pkg/printers/junitxml.go
+++ b/pkg/printers/junitxml.go
@@ -21,9 +21,14 @@ type testSuiteXML struct {
 }
 
 type testCaseXML struct {
-	Name      string `xml:"name,attr"`
-	ClassName string `xml:"classname,attr"`
-	Status    string `xml:"status,attr"`
+	Name      string     `xml:"name,attr"`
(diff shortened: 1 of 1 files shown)


glenjamin:
I've inspected the SVG, and I can't see any difference - I'm not sure why this would have changed on my branch.

Helcaraxan:
@glenjamin - See #625 for the solution. Once that fix is merged all latests PR can be rebased on top of the fresh master and go green.

jirfag:
thank you!

PR #636, opened 2019-08-12
Add kubeedge to list of projects using golangci-lint

fisherxu:
Thanks for the great project golangci-lint which we have used and thank you for considering CNCF project kubeedge/kubeedge to your list of projects using golangci-lint. 

I have changed the `README.tmpl.md` and running `make README.md`.


jirfag:
thank you, the great project!

PR #640, opened 2019-08-19
Add support for bash completions

cblecker:
Cobra has built in support for bash completions. Expose these via a command.

These can be used with `source <(./golangci-lint completion)`

### DIFF
--- a/pkg/commands/completion.go
+++ b/pkg/commands/completion.go
@@ -1,0 +1,26 @@
+package commands
+
+import (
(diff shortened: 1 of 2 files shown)


jirfag:
Added to README and changed the command to `golangci-lint completion bash` in https://github.com/golangci/golangci-lint/commit/5fb91e6434fcdd7ae60c55783cbc2bf4979bfd89

PR #643, opened 2019-08-19
Fix unversioned module imports to work with GOPROXY.

mhutchinson:
All of the imports fixed here failed via proxy.golang.org with 'invalid pseudo-version: does not match version-control timestamp'. This change keeps the same versions but updates the timestamp to something the proxy is happy with as per https://tip.golang.org/doc/go1.13#version-validation.

One exception is that the bodyclose dep has been really updated to include https://github.com/timakin/bodyclose/pull/17 as per the comment. The previous fix pointed at PR 16.


mhutchinson:
The Travis tests are failing, but the diff fails for me at HEAD too. My guess is that the .svg generation is not hermetic and something else has changed since the last run. Do the control files need updating?

If this PR has changed something functional please let me know, this is unexpected.

mhutchinson:
This looks like it's blocked on PR #625, along with the other PRs outstanding at the moment.

bvwells:
Might want to add ```  - 1.13.x``` to the travis CI definition as go proxy is enabled by default in 1.13 and this would test things are working as expected.

jirfag:
Hi, thank you! And sorry for the slow response.
But already fixed in master, please, check

PR #660, opened 2019-09-04
fix the timestamp for golangci/bodyclose 65da19158fa2

laverya:
This pull request fixes a timestamp included in `go.mod`, specifically for the dep replacement `github.com/golangci/bodyclose`. Incorrect timestamps prevent go 1.13 from including the dep properly.


laverya:
It doesn't feel like the `docs/demo.svg` diff is related to this change (since it also failed [here](https://travis-ci.com/golangci/golangci-lint/jobs/231201826), and [that commit](https://github.com/golangci/golangci-lint/commit/f31a1bbf41079d5a47d85f0f85d2b2fd6aa6b72b) just added go 1.13 to the travis jobs list - which shouldn't impact the go 1.12 runs at all)

laverya:
dupe of https://github.com/golangci/golangci-lint/pull/605

PR #662, opened 2019-09-05
Add user supplied error messages in depguard issues

codyleyhan:
This adds the ability for a user to add an error message for specific packages when encountered in depguard issues.  This is really useful to point users to use another package. 

An example of what the new output looks like when a user supplies an error message
image

I have made an upstream PR to add similar functionality to depguard in https://github.com/OpenPeeDeeP/depguard/pull/12


jirfag:
thank you! I'll also add example to `.golangci.example.yml`

PR #672, opened 2019-09-10
dev: update install.sh

jirfag:


jirfag:
CI build fail will be fixed in a separate commit

Issue #687, opened 2019-09-13
Go module revision for tools broken


troyronda:
golangci-lint: 1.18

3540c026601b is not a valid revision. https://github.com/golangci/golangci-lint/blob/master/go.mod#L58

`go mod tidy`                                                                                                                                                                          
go: finding github.com/go-critic/go-critic v0.3.5-0.20190904082202-d79a9f0c64db
go: finding github.com/golangci/tools v0.0.0-20190910062050-3540c026601b
go: finding github.com/golangci/gosec v0.0.0-20190911143311-c62ea1e208f0
go: finding github.com/stretchr/testify v1.4.0
go: finding github.com/matoous/godox v0.0.0-20190910121045-032ad8106c86
go: finding github.com/ultraware/whitespace v0.0.2
go: finding gopkg.in/yaml.v2 v2.2.2
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: unknown revision 3540c026601b
go: error loading module requirements

--------

This causes a problem when using `gobin -run github.com/golangci/golangci-lint` on a clean go path.
failed to run go list -find -json github.com/golangci/golangci-lint/cmd/golangci-lint: exit status 1
go: finding github.com/golangci/tools v0.0.0-20190910062050-3540c026601b
go: github.com/golangci/tools@v0.0.0-20190910062050-3540c026601b: open $HOME/go/pkg/mod/cache/download/github.com/golangci/tools/@v/v0.0.0-20190910062050-3540c026601b.info: no such file or directory
[...]

PR #691, opened 2019-09-13
Fix replaced dependencies

tpounds:
Fixes replaced dependencies that have been merged upstream. This should also fix issue #687.


jirfag:
thank you!

geoah:
@jirfag could we get a new tag that includes this fix as v1.18.0 is older and thus broken because of #687 ?

Thank you :)

Issue #208, opened 2018-08-30
5x slower than plain "go vet" with just "govet" enabled


atombender:
golangci-lint seems to have become much slower recently. As a test case, I see that `golangci-lint` with `govet` takes about 6.4 seconds, compared to plain `go vet` at 1.25 seconds.

Just with `govet` enabled:

```
$ time /usr/local/bin/golangci-lint run --disable-all --enable=govet --tests -v
INFO [config_reader] Config search paths: [./ /Users/alex/Projects/Transparensee/pond /Users/alex/Projects/Transparensee /Users/alex/Projects /Users/alex /Users /]
INFO [config_reader] Config search paths: [./ /Users/alex/Projects/Transparensee/pond /Users/alex/Projects/Transparensee /Users/alex/Projects /Users/alex /Users /]
INFO [lintersdb] Active 1 linters: [govet]
INFO [load/path_resolver] Skipped dirs: [.git .vscode]
INFO [load/path_resolver] Paths resolving took 212.852477ms: [blobstore blobstore/filestore blobstore/gcs blobstore/mocks cmd/pond cmd/pondctl e2e pkg/client proto proto/mocks server server server/database server/database/mocks utils]
INFO [load] Can't get cur project path: currently no in gopath: "/Users/alex/.go" isn't a prefix of "/Users/alex/Projects/Transparensee/pond"
INFO [load] Program loading took 6.151314387s
INFO [load] Not compiling packages: [./server_test ./cmd/pondctl ./server ./proto github.com/stretchr/testify/require github.com/gogo/protobuf/types github.com/stretchr/testify/mock github.com/t11e/pond/server go.uber.org/zap github.com/gogo/protobuf/gogoproto google.golang.org/grpc github.com/spf13/cobra google.golang.org/grpc/status github.com/jackc/pgx cloud.google.com/go/storage google.golang.org/grpc/metadata github.com/stretchr/testify/assert github.com/t11e/pond/proto]
INFO [runner] worker.1 took 152.927µs
INFO [runner] worker.6 took 247.58µs
INFO [runner] worker.2 took 267.053µs
INFO [runner] worker.5 took 246.144µs
INFO [runner] worker.8 took 287.753µs
INFO [runner] worker.3 took 237.41µs
INFO [runner] worker.4 took 328.038µs
INFO [runner] worker.7 took 14.485719ms with stages: govet: 14.195536ms
[...]

PR #693, opened 2019-09-14
Speed up packages loading

jirfag:
Don't perform extra go env calls in go/packages.
Load only needed go env vars in golangci-lint.
Stay in sync by enabled analyzers in go vet: remove nilness and
atomicalign analyzers, add errorsas analyzer.
Don't build SSA for govet.

Standalone govet runs 25% faster than before. All runs can be 5-10% faster
than before.
Relates: #208


ernado:
Is there a way to enable the `atomicalign` back?

PR #694, opened 2019-09-14
Update to latest securego/gosec.

tpounds:
Updates securego/gosec to include fix for https://github.com/securego/gosec/pull/366 to use gosec API directly without a patch.


jirfag:
thank you, cool

PR #697, opened 2019-09-14
govet: implement analyzers config

### LINKED ISSUE #696: govet: allow analyzers suite configuration
Enabling or disabling go vet plug-ins would be very cool.
Seems like implementation is pretty straight-forward:
https://github.com/golangci/golangci-lint/blob/c9a92552385e742555fe341f813e6beba93defae/pkg/golinters/govet.go#L36-L61

We can filter analyzers in `NewGovet` function, but I'm not sure about config.
I think that there should be something like we have for `linters` section:
```yml
linters:
  enable:
    - megacheck
    - govet
  enable-all: false
  disable:
    - maligned
    - prealloc
  disable-all: false
```
But in `linters-settings.govet`.

Rel: #446

ernado:
Fix #696, #446

Please feel free to request any changes.

### DIFF
--- a/pkg/golinters/govet.go
+++ b/pkg/golinters/govet.go
@@ -11,6 +11,7 @@ import (
 	"golang.org/x/tools/go/analysis/passes/asmdecl"
 	"golang.org/x/tools/go/analysis/passes/assign"
 	"golang.org/x/tools/go/analysis/passes/atomic"
+	"golang.org/x/tools/go/analysis/passes/atomicalign"
 	"golang.org/x/tools/go/analysis/passes/bools"
 	"golang.org/x/tools/go/analysis/passes/buildtag"
 	"golang.org/x/tools/go/analysis/passes/cgocall"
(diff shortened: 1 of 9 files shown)


pkg/golinters/govet.go
```diff
@@ -59,20 +88,64 @@ func NewGovet(cfg *config.GovetSettings) *goanalysis.Linter {
...
+	}
+	// Raw for loops should be OK on small slice lengths.
+	for _, n := range cfg.Enable {
+		if n == name {
+			return true
+		}
+	}
+	for _, n := range cfg.Disable {
+		if n == name {
+			return true
```
jirfag: I guess it should be
```suggestion
			return false
```

pkg/golinters/govet.go
```diff
@@ -59,20 +88,64 @@ func NewGovet(cfg *config.GovetSettings) *goanalysis.Linter {
...
+	}
+	for _, a := range defaultAnalyzers {
+		if a.Name == name {
+			return true
+		}
+	}
+	return false
+}
+
+func analyzersFromConfig(cfg *config.GovetSettings) []*analysis.Analyzer {
```
jirfag: please, add minimal validation to `(r *FileReader) validateConfig()` by calling something like:
```go
func (cfg *config.GovetSettings) Validate() error {
    if cfg.EnableAll && cfg.DisableAll {
        return errors.New("enable-all and disable-all can't be combined")
    }
    if cfg.EnableAll && len(cfg.Enable) != 0) {
        return errors.New("enable-all and enable can't be combined")
    }
    if cfg.DisableAll && len(cfg.Disable) != 0) {
        return errors.New("disable-all and disable can't be combined")
    }
    return nil
}
```

pkg/golinters/govet.go
ernado: You are right, nice catch!
I've also added a corresponding test for that function.

pkg/golinters/govet.go
ernado: Done, thank you.
We probably want to validate that `cfg.Enable` contains valid analyzer names, but I'm not sure how to do it in that scope.

PR #698, opened 2019-09-15
Upgrade depguard to fix performance issue with go modules

dixonwille:
Little insight on what I did. I timed depguard's `Run` method using depguard's cli alone. Before the fix it took on average a minute to go through golangci-lint repository. This does not include the time it took to build the loader. With the new fix in place it takes on average 3 milliseconds.

Before it was using go's build's context and the `Import` method on that. This killed time as I think it was doing a lot more than what I really needed. Instead, I am using build's context `GOROOT` and running `iotuil.ReadDir` to get the prefix for all known root packages. And if there is one thing depguard was good at, it was efficiently looking up prefixes from a slice.

Would love to hear your numbers as well.


jirfag:
very cool, thank you!

PR #704, opened 2019-09-16
Update to valyala/quicktemplate v1.2.0

tpounds:
Updates Update to valyala/quicktemplate v1.2.0. to latest version and cleanup some obsolete `go get` commands that were bypassing go.mod dependency management in CI.


jirfag:
thank you!

PR #705, opened 2019-09-16
Add dogsled linter

tpounds:
Adds https://github.com/alexkohler/dogsled linter to detect assignments with too many blank identifiers.

e.g. `x, _, _, _, := f()`

The logic was ported similar to what was done for `nakedret` and `prealloc` since the upstream source cannot be integrated directly.


jirfag:
great, thank you!

PR #714, opened 2019-09-19
Update to shirou/gopsutil v2.19.8.

tpounds:
Updates to latest shirou/gopsutil (v2.19.8). Note, dependency has to be pinned to a specific release version commit hashes since it doesn't follow Go module v2+ guidelines.

see:
https://github.com/shirou/gopsutil/issues/663
https://github.com/golang/go/issues/34402


tpounds:
@jirfag Test failures here are caused by a concurrent linter race and are fixed by #716. I'll rebase on `master` once it's merged.

jirfag:
thank you!

PR #716, opened 2019-09-20
Fix flaky cgo test failures

tpounds:
Fixes spurious test failures caused by overlapping printf format checks in `staticcheck` and `go vet`. In general, this is a concurrency issue which can happen with any linters that detect similar issues with different reporting formats.

see:
https://travis-ci.com/golangci/golangci-lint/jobs/236978712
https://travis-ci.com/golangci/golangci-lint/jobs/237582672


jirfag:
thank you!

PR #717, opened 2019-09-20
Enable Go module support across all commands

tpounds:
Misc. cleanup to dedupe GO111MODULE configurations and enable Go modules across all make targets.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 .DEFAULT_GOAL = test
 .PHONY: FORCE
 
+# Enable Go module support across all commands.
+export GO111MODULE = on
+
 # Build
 
 fast_build: FORCE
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,9 @@
 .DEFAULT_GOAL = test
 .PHONY: FORCE
 
+# Enable Go module support across all commands.
+export GO111MODULE = on
```
tariq1890: Is this required? Go 1.13 has this on by default.

tpounds: @tariq1890 Go 1.13 defaults it to `auto` which implies module-awareness by default (see: https://golang.org/doc/go1.13#modules). However, the primary reason for this change is to consolidate all of the places it was being set and to force consistent module `on` behavior for both Go 1.12 and 1.13 builds in CI (see: https://github.com/golangci/golangci-lint/blob/master/.travis.yml#L2). FWIW, I also tested this change against Go 1.11 locally and it seemed to work fine. It may be possible to re-enable Go 1.11 in CI if there aren't any internal Go runtime/interface issues to be wary of. @jirfag Thoughts?

jirfag:
thank you, I agree, good change

PR #718, opened 2019-09-21
Whitelist regexp and fmt.Errorf global vars

korjavin:
I got some good exception from:
https://github.com/bombsimon/gochecknoglobals/tree/whitelisted-selectors

### DIFF
--- a/pkg/golinters/gochecknoglobals.go
+++ b/pkg/golinters/gochecknoglobals.go
@@ -60,8 +60,56 @@ func (lint Gochecknoglobals) checkFile(f *ast.File, fset *token.FileSet) []resul
 	return res
 }
 
-func isWhitelisted(i *ast.Ident) bool {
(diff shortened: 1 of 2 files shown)


tpounds:
@korjavin This is a patch that deviates from the upstream repository which means it will make it harder to source directly from that in the future. I'm not sure what @jirfag's policy has been in this regard but my opinion would always be to stay in sync with upstream linters to maintain compatibility with their behavior.

tpounds:
@korjavin One additional thought is to provide a way to configure additional whitelist rules within `.golangci.yml` and/or command line options. This provides more flexibility to all projects without relying on hardcoded rules within the linter. Thoughts?

jirfag:
@korjavin thank you, but I wouldn't like to import patches not from the upstream repository. Why can't you merge it into upstream first?

PR #719, opened 2019-09-21
Enable Go 1.11.x testing

tpounds:
Re-enables testing Go 1.11.x in CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 language: go
 go:
+  - 1.11.x
   - 1.12.x
   - 1.13.x
 


tpounds:
@jirfag I wasn't sure this is something that should be done but I figured I'd put it for review to see what you thought.

jirfag:
Hi, I think it will slow down CI too much. It's common practice in Go to support only two latest versions. Why do you think it's needed?

tpounds:
@jirfag I think you are correct. According to the Go [wiki](https://github.com/golang/go/wiki/MinorReleases) minor release are only cut for the past two major releases which I assume to mean any Go release <= 1.11 is effectively EOL. The only upside I see to re-enabling 1.11 in CI is to force broader compatibility of golangci-lint since it may be used on code bases using older Go versions. I'm not sure that matters too much though. Any additional thoughts before I discard this?

jirfag:
I think development speed for newest versions is more important so I would like to not slow down CI

tpounds:
@jirfag Agreed. I've updated go.mod and README to require Go 1.12+ (see: https://github.com/golangci/golangci-lint/pull/728).

PR #723, opened 2019-09-23
Update whitespace

NiseVoid:
I added a new setting to whitespace and cleaned up the code for whitespace's autofix


tpounds:
@NiseVoid Thanks!

.golangci.example.yml
```diff
@@ -205,6 +205,9 @@ linters-settings:
     # checks assignments with too many blank identifiers; default is 2
     max-blank-identifiers: 2
 
+  whitespace:
+    multi-if: false
```
jirfag: please, document it - you can see that this file is full of comments :)

PR #745, opened 2019-09-25
Add Alpine Linux images

tpounds:
Adds building Alpine images to goreleaser config.


Dockerfile.alpine
```diff
@@ -0,0 +1,5 @@
+FROM golang:1.13-alpine
+
+# don't place it into $GOPATH/bin because Drone mounts $GOPATH as volume
+COPY golangci-lint /usr/bin/
+CMD ["golangci-lint"]
```
jirfag: did you try this image on `cgo` project? I guess it will fail without `gcc` tools installed

tpounds: I haven't performed any rigorous testing. Is there an easy way to test `cgo` support?

jirfag: yep, see `test/testdata/cgo*`

tpounds: I tried the `cgo` testdata and it seems to be working as expected after I added the `gcc` and `musl-dev` dependencies to the image.

see: https://github.com/golangci/golangci-lint/pull/745/files#diff-39bd4a9d0958b0d526ad2585d0738b26R4

Dockerfile.alpine
```diff
@@ -0,0 +1,8 @@
+FROM golang:1.13-alpine
```
jirfag: there too many files in the project root. Please, move both of Dockerfile-s into a new `/build` directory.

tpounds: Moved the Dockerfiles to top-level build directory.

jirfag:
merged, thank you!

PR #747, opened 2019-09-25
Fix broken dominikh/go-tools URLs

tpounds:
Fixes some broken dominikh/go-tools linter URLs.


jirfag:
We don't need to change `README.tmp.md` here

PR #759, opened 2019-09-28
govet: add more analyzers

ernado:
This allows enabling more analyzers.
No defaults were changed.

Command that helped me to produce full list:
```bash
# in golang.org/x/tools
ls go/analysis/passes/ | grep -vP '(README|internal)' | awk '$0=$0".Analyzer,"'
```


pkg/golinters/govet.go
```diff
@@ -41,18 +49,26 @@ func getAllAnalyzers() []*analysis.Analyzer {
 		atomic.Analyzer,
 		atomicalign.Analyzer,
 		bools.Analyzer,
+		buildssa.Analyzer,
```
jirfag: but this analyzer in internal and exists just for building SSA: it reports no issues

pkg/golinters/govet.go
```diff
@@ -41,18 +49,26 @@ func getAllAnalyzers() []*analysis.Analyzer {
...
 		buildtag.Analyzer,
 		cgocall.Analyzer,
 		composite.Analyzer,
 		copylock.Analyzer,
+		ctrlflow.Analyzer,
```
jirfag: the same

pkg/golinters/govet.go
```diff
@@ -41,18 +49,26 @@ func getAllAnalyzers() []*analysis.Analyzer {
...
+		deepequalerrors.Analyzer,
 		errorsas.Analyzer,
+		findcall.Analyzer,
 		httpresponse.Analyzer,
+		inspect.Analyzer,
```
jirfag: I guess it's internal too

pkg/golinters/govet.go
ernado: Can we automatically filter internal analyzers by some condition?

jirfag: I don't know how to do it

ernado: Seems like the following should work:
```go
func isInternal(a *analysis.Analyzer) bool {
	return a.ResultType != nil
}
```

ernado: I've created the #763 to address this issue. 
I still want to import every analyzer and explicitly external ones, seems like it is most maintainable approach.

Issue #337, opened 2018-12-26
Lower or control memory usage


jirfag:
Users complain about too much memory usage exceeding 8gb

lopezator:
Related to #342 and pointers usage on latest commit. 

We could dig more, sure, but it's a start.

jirfag:
Hi!
A current solution may be to configure `GOGC`: it can trade memory to time and decrease memory usage in ~2 times. I've shown some results [here](https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint).

jirfag:
[mem.pdf](https://github.com/golangci/golangci-lint/files/3027069/mem.pdf)
Also, I've profiled memory of golangci-lint: PDF with results is attached. We can see that type checking consumes the most of memory. I have some ideas on how to fix it but they are all too costly to implement.

alexec:
We have regular intermittent build failures due to high memory usage. It would be good to reduce usage.

Currently: we're using...

```
LINT_GOGC=20 LINT_CONCURRENCY=1 LINT_DEADLINE=3m0s
...
golangci-lint run --fix --verbose --concurrency $(LINT_CONCURRENCY) --deadline $(LINT_DEADLINE)
```

See https://github.com/golangci/golangci-lint#memory-usage-of-golangci-lint

howardjohn:
istio/istio repo uses 32gb to lint: `INFO Memory: 218 samples, avg is 18237.6MB, max is 31795.7MB`

daxmc99:
👍 
Looking forward to testing out master

jirfag:
I've merged a fix reducing memory usage of go/analysis linters (staticcheck, gosimple, stylecheck, bodyclose, govet) 5x. Please, check it in a master branch.

gracenoah:
It's still using 29 GB of memory when running with all linters enabled rather than enabling 1 linter at a time. Is that expected? Is it just a side effect of Go's GC and it wouldn't use that much memory if it wasn't available?

jirfag:
@gracenoah thank you for comparison,
I'am continuously reducing memory consumption: it's my top priority now. Is the repo rhere are you testing open-source?

gracenoah:
It's not, sorry. It's an internal monorepo. If you want me to test some proposed performance changes, you can email me at gracenoahgh@gmail.com but that's all I can do.

jirfag:
I've investigated more high memory consumption.
[...]

PR #764, opened 2019-09-30
reduce 1.5x memory usage on large repos on repeated runs

jirfag:
Get rid of AST cache: load AST when needed. Optimize memory allocations
for go/analysis actions.

Relates: #337


jirfag:
@ernado did you try disabling `unused`?

ernado:
@jirfag I've tried right now, and it is much better:
```
172.86user 3.05system 0:28.17elapsed 624%CPU (0avgtext+0avgdata 1581020maxresident)k
226720inputs+42408outputs (5major+594733minor)pagefaults 0swaps
```
So just 1.5Gb RSS, very cool.

jirfag:
It's good. We need to add information about `unused` and `GOMAXPROCS` into README into the memory section. I will look at it at the weekend

PR #758, opened 2019-09-28
Dramatically reduce memory usage

jirfag:
Run all linters per package. It allows unloading package data when it's
processed. It dramatically reduces memory (and CPU because of GC) usage.

Relates: #337

PR #766, opened 2019-09-30
Fix incorrect description

tpounds:
Fixes incorrect description from accidental copy/paste in https://github.com/golangci/golangci-lint/pull/758.


jirfag:
Thank you :)

Issue #255, opened 2018-10-25
goimports linter works incorrectly


derElektrobesen:
```bash
$ golangci-lint --version
golangci-lint has version 1.10.2 built from dbfcf63 on 2018-09-02T12:30:29Z

$ go version && go env
go version go1.10.2 linux/amd64
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/p.berezhnoy/.cache/go-build"
GOEXE=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOOS="linux"
GOPATH="/home/p.berezhnoy/go"
GORACE=""
GOROOT="/usr/local/go"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build415063740=/tmp/go-build -gno-record-gcc-switches"

$ golangci-lint run --build-tags="testing,integration" --new-from-rev=30a6969eb9 -v
INFO [config_reader] Config search paths: [./ /home/p.berezhnoy/go/src/gitlab.corp.mail.ru/swa/totem /home/p.berezhnoy/go/src/gitlab.corp.mail.ru/swa /home/p.berezhnoy/go/src/gitlab.corp.mail.ru /home/p.berezhnoy/go/src /home/p.berezhnoy/go /home/p.berezhnoy /home /]
INFO [config_reader] Used config file .golangci.yaml
INFO [lintersdb] Active 23 linters: [deadcode depguard dupl errcheck goconst gocyclo gofmt goimports golint gosec govet ineffassign interfacer lll megacheck misspell nakedret prealloc structcheck typecheck unconvert unparam varcheck]
INFO [load/path_resolver] Skipped dirs: [.git contrib/tarantool/third_party pkg/mapi/testdata testdata vendor]
[...]

PR #771, opened 2019-10-02
Add WSL linter

bombsimon:
Added linter [WSL](https://github.com/bombsimon/wsl). Tests and linter is working with `golangci-lint run`, however I'm experiencing an old issue noted in #255.

This is with `golangci-lint` built from this commit (master + `08de1f8`) and `goimports` installed with `go get` today (`GO111MODULE=off go get -u golang.org/x/tools/cmd/goimports`). No difference with `-local` flag or not.

```sh
golangci-lint run pkg/golinters/wsl.go
pkg/golinters/wsl.go:6: File is not `goimports`-ed (goimports)
        "github.com/bombsimon/wsl"

goimports -w pkg/golinters/wsl.go
gofmt -w -s pkg/golinters/wsl.go

diff \
  <(goimports pkg/golinters/wsl.go) \
  <(cat pkg/golinters/wsl.go) | wc -l
       0

golangci-lint run pkg/golinters/wsl.go
pkg/golinters/wsl.go:6: File is not `goimports`-ed (goimports)
        "github.com/bombsimon/wsl"
```


pkg/golinters/wsl.go
```diff
@@ -0,0 +1,67 @@
+package golinters
+
+import (
+	"strings"
+
+	"github.com/bombsimon/wsl"
```
golangcibot: File is not `goimports`-ed (from `goimports`)
```suggestion
	"github.com/bombsimon/wsl"
	"golang.org/x/tools/go/analysis"

```

tpounds:
@bombsimon Thanks for the submission. How does your WSL linter differ from https://github.com/ultraware/whitespace which was recently added?

bombsimon:
@tpounds WSL has a bunch more features other than checking leading and trailing whitespaces in blocks. It checks for variable usages and different node types to determine if there should be a newline or not. Please give the [WSL readme](https://github.com/bombsimon/wsl/blob/master/README.md) a quick look to get a better understanding of how WSL works, there should be a lot of examples describing how the linter works.

bombsimon:
Other than the `goimports` issue I see I have a problem with `go mod` and `go1.13`. Since I've had some problems with `go mod` I've set `GOSUMDB=off` in my go environment and thus didn't notice the issue with my package.

I can't believe I'm having such ha hard time getting go mod to work, I feel stupid.

I've added a tag `v1.0.0` to `master` and I've created a release named `v1.0.0`. Apparently this is not enough and result in `go mod` not working:

```sh
go: github.com/bombsimon/wsl@v1.0.0: unexpected status (https://proxy.golang.org/github.com/bombsimon/wsl/@v/v1.0.0.info): 410 Gone
```

Doesn't seem to ba any problem when trying said URL manually:

```sh
http https://proxy.golang.org/github.com/bombsimon/wsl/@v/v1.0.0.info
HTTP/1.1 200 OK
Accept-Ranges: bytes
Age: 64
Alt-Svc: quic=":443"; ma=2592000; v="46,43",h3-Q048=":443"; ma=2592000,h3-Q046=":443"; ma=2592000,h3-Q043=":443"; ma=2592000
Cache-Control: public, max-age=10800
Content-Length: 50
Content-Type: application/json
Date: Wed, 02 Oct 2019 14:50:55 GMT
Expires: Wed, 02 Oct 2019 17:50:55 GMT
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Xss-Protection: 0

{
    "Time": "2019-10-02T13:06:13Z",
    "Version": "v1.0.0"
}
```

`GOPROXY` and `GOSUMDB` are set to `""` in the Travis CI environment and if I do the same in a new project and run `go get -u github.com/bombsimon/wsl@v1.0.0` everything is working.

Any help or pointers in the right direction is much appreciated!

.golangci.example.yml
```diff
@@ -208,6 +208,10 @@ linters-settings:
   whitespace:
     multi-if: false
 
+  wsl:
+    # Skip test files
+    no-test: false
```
jirfag: 1. please, don't use negative naming
2. please, don't add such duplicating option: any linter can be disabled for tests by `exclude-rules`. It's not performance effective, I agree, but it should be done in a general way for all linters, not per-linter

pkg/golinters/wsl.go
```diff
@@ -0,0 +1,67 @@
...
+
+				fileNames = append(fileNames, pos.Filename)
+			}
+
+			wslErrors, _ := wsl.ProcessFiles(fileNames)
+			if len(wslErrors) == 0 {
+				return nil, nil
+			}
+
+			for _, err := range wslErrors {
```
jirfag: this block should be run only with mutex held: go/analysis passes are run in parallel

.golangci.example.yml
bombsimon: Sorry about that, I'll remote this since the configuration isn't needed!

pkg/golinters/wsl.go
bombsimon: I'll add a mutex, thanks!

bombsimon:
Ok so now the Travis pipeline fails with the lint error `File is not goimports-ed (from goimports)`. I have no idea how I should solve that, see my previous comments about what I've tried with `goimports [-local]` and `gofmt -s`. Any ideas how I could resolve this?

jirfag:
@bombsimon did you updated your `goimports`?
also, you can fix it by committing suggestion by `golangcibot` or by running `golangci-lint run --fix`

bombsimon:
@jirfag Just fixed them with `golangci-lint` instead of `goimports` itself. Don't know why `goimports -w` doesn't fix the issue but I guess I'm running a bad version. `goimports -w` doesn't make any changes before or after the fix with `golangci-lint`.

Seems like GolangCI is happy now at least! :)

bombsimon:
@tpounds Absolutely, will add more tests!

tpounds:
@bombsimon Thanks for the additional tests! It looks like CI is still failing due to a `go.mod`/`go.sum` mismatch.

https://travis-ci.com/golangci/golangci-lint/jobs/242161917#L699

Can you run `go mod tidy` and/or `make vendor` to fix this issue?

bombsimon:
@tpounds Thank you for your feedback, help and the merge! What's common practice regarding changes (new errors, new/changed line markings etc)? Should I bump major or minor if/when I do changes to those? It's not really a common interface/API used by other's so I might as well ask. Anything else I should think about now that it's a part of `golangci-lint`?

tpounds:
@bombsimon In general I think following standard semantic versioning is reasonable. In the case of third party linters it's fairly inconsistent across the board. That said, I think these are general interpretations to the semver guidlines that seem reasonable to me:

1. Bump major version when you add a new detection rule that is on by default or flip an existing detection rule from on to off or off to on.
2. Bump minor version when you add a new detection rule that is off by default but can be opted into.
3. Bump patch version when you fix a detection bug (e.g. fix false positives/negatives), update deps, etc.

/cc @jirfag

PR #779, opened 2019-10-03
Cleanup nolint directives

tpounds:
Misc. cleanup of obsolete `nolint` directives.


jirfag:
it would be nice if we could find these obsolete nolints automatically

PR #781, opened 2019-10-04
Exposed modules-download-mode as a command line option

pstibrany:
This tiny change exposes modules-download-mode as a command line option, allowing us to easily control it from Makefile.


pkg/commands/run.go
```diff
@@ -82,6 +82,7 @@ func initFlagSet(fs *pflag.FlagSet, cfg *config.Config, m *lintersdb.Manager, is
 
 	// Run config
 	rc := &cfg.Run
+	fs.StringVar(&rc.ModulesDownloadMode, "modules-download-mode", "", "Modules download mode. If not empty, passed as -mod=<mode> to go tools")
```
golangcibot: line is 141 characters (from `lll`)

jirfag:
Hi, why do you need to control it from Makefile?

pstibrany:
Hi, we use go modules, but also keep vendor up to date for now. Our Makefile uses vendoring by default, but when I'm developing locally, I need to point some modules to my local forks, and that's easier done without vendoring (by using replace). I have modified our Makefile so that I can turn vendoring off simply by running `VENDOR=false make`, but it doesn't apply to the linting step, which uses golangcli-lint. Hence the extra flag.

Thanks.

tpounds:
@pstibrany, @jirfag Does it make sense to just have golangci-lint use the same flags and resolution strategy as the Go tool chain `-mod` options? That is, users can specify `-mod=X` on the golangci-lint command line or set the value via `GOFLAGS` environment variable?

pstibrany:
It seems to me that all golang-cli does is to pass the configured value to -mod parameter when running go tools. I'm fine with renaming the parameter to -mod. I have tried to use GOFLAGS, but that didn't work, and exposing the parameter seemed like smaller change than making GOFLAGS work :)

jirfag:
1. Naming inside `.golangci.yml` should be the same as in command-line
2. If we name it `mod` in `.golangci.yml` it will look too unclear.
3. I think it's also unclear for a lot of people even in the command-line. I would prefer clarity over uniformity

jimen0:
What about `-mod-mode`? I too want this feature to be added 🎉

jirfag:
I like your ‘modules-download-mode’

pstibrany:
I don't mind using either, it's just a name. More important to me is whether I can quickly find out what it does. "-mod" would be more familiar to people since it's the same as used by Go, "-modules-download-mode" is consistent with .golangci.yml. But grepping `golangci-lint -help` for `-mod` already shows this option.

I was wondering more about passing GOFLAGS env when calling go tools. Do you want to do that as part of this PR?

PR #783, opened 2019-10-04
gocognit: fix default threshold from 0 to 30

jirfag:


uudashr:
Sonarqube use 15 actually, with no reason just a magic number.

uudashr:
But I think since this is Golang, 30 would be good

PR #792, opened 2019-10-07
Update to latest securego/gosec

tpounds:
Update to latest [securego/gosec](https://github.com/securego/gosec) and fix `G204` warnings.


jirfag:
thank you!

PR #793, opened 2019-10-07
Rename deadline option to timeout and mark deadline as deprecated

### LINKED ISSUE #655: --deadline is a misnomer
`--deadline` should be called `--timeout`. A timeout is a relative time like what we're supposed to pass to the `--deadline` flag. An actual deadline is an absolute time.

> Timeout: finish in 10 seconds.
> Deadline: finish at 7:15:35PM.

https://news.ycombinator.com/item?id=9948203
https://golang.org/pkg/context/#WithDeadline

This was confusing to me when I was reading the docs.

titusjaka:
fixes #655

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -7,7 +7,7 @@ run:
   concurrency: 4
 
   # timeout for analysis, e.g. 30s, 5m, default is 1m
-  deadline: 1m
+  timeout: 1m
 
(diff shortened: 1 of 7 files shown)


jirfag:
thank you! I was also embarrassed, but I did not change since it was the same in the `gometalinter`.

PR #794, opened 2019-10-07
Update WSL to v1.2.1

bombsimon:
After WSL v1.0.1 got merged last week I found a few false positives which is resolved in this version. I've also added support for some configuration which is bundled in this commit.

I aimed to use non-negative configuration names so defaulting them to `true` required me to create a config struct, hope this is OK.

I sent an email to Denis yesterday with a few questions regarding this linter but decided to create this PR even though I haven't gotten any response yet since I don't want to risk miss this update if a new release of `golangci-lint` is being created.

### DIFF
--- a/pkg/golinters/wsl.go
+++ b/pkg/golinters/wsl.go
@@ -33,13 +33,23 @@ func NewWSL() *goanalysis.Linter {
 		nil,
 	).WithContextSetter(func(lintCtx *linter.Context) {
 		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
-			files := []string{}
+			var (
+				files        = []string{}
+				linterCfg    = lintCtx.Cfg.LintersSettings.WSL
+				processorCfg = wsl.DefaultConfig()
+			)
+
+			processorCfg.StrictAppend = linterCfg.StrictAppend
+			processorCfg.AllowAssignAndCallCuddle = linterCfg.AllowAssignAndCallCuddle
+			processorCfg.AllowMultiLineAssignCuddle = linterCfg.AllowMultiLineAssignCuddle
 
 			for _, file := range pass.Files {
 				files = append(files, pass.Fset.Position(file.Pos()).Filename)
 			}
 
-			wslErrors, _ := wsl.ProcessFiles(files)
+			wslErrors, _ := wsl.NewProcessorWithConfig(processorCfg).
+				ProcessFiles(files)
+
 			if len(wslErrors) == 0 {
 				return nil, nil
 			}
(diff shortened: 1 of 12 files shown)


pkg/golinters/wsl.go
```diff
@@ -33,13 +33,23 @@ func NewWSL() *goanalysis.Linter {
 		nil,
 	).WithContextSetter(func(lintCtx *linter.Context) {
 		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
-			files := []string{}
+			var (
+				files        = []string{}
+				linterCfg    = lintCtx.Cfg.LintersSettings.WSL
+				processorCfg = wsl.DefaultConfig()
```
tpounds: Is this necessary given the defaults added to `config.go`. We probably only want one place the default values can be set and sourced from.

tpounds:
@bombsimon Thanks for the updated version! You mentioned this fixes some false positives. Would it be possible to add those as additional test cases? This will help ensure `golangci-lint` tests against the expected behavior of upstream deps.

bombsimon:
@tpounds You are absolutely right! I'll add a test case which is what I resolved.

pkg/golinters/wsl.go
bombsimon: I agree that there should be one place for the defaults and right now they're all true which is kind of bad if you want to marshal the configuration from `.golangci.yml` but good if you don't want negative names. :)

How do you propose I should attack this issue? I want the values to be `true` by default and therefore I don't want them in the `.golangci.yaml`. However if they're not present and the the configuration is being unmarshalled their default values will be false. Because of this I added the `WSLSettings` which is set in `config.go`. So the only way to set them to `false` right now is to explicitly specify that in the configuration, otherwize the defaults will be made.

Do you think I should re-write how this is handled in the linter itself (don't expose `DefaultConfig()` or am I missing something obvious about how the defaults are set in `golangci-lint`? I still think there might be use-cases for WSL outside golangci-lint.

tpounds: I don't have a strong recommendation but if you want `golangci-lint`'s setting to derive from the linter directly I would set that within the `config.go` file. However, my general feeling is it might be better to just have `golangci-lint` have a copy of its own defaults that way the defaults can't change when upgrading the linter in the future. This implies removing the `wsl.DefaultConfig()` assignment here and mapping the struct fields 1:1.

```Go
ctxCfg := lintCtx.Cfg.LintersSettings.WSL
wslCfg := wsl.Configuration{
    StrictAppend:               ctxCfg.StrictAppend,
    AllowAssignAndCallCuddle:   ctxCfg.AllowAssignAndCallCuddle,
    AllowMultiLineAssignCuddle: ctxCfg.AllowMultiLineAssignCuddle,
}
```

PR #811, opened 2019-10-13
Update WSL to v1.2.5

bombsimon:
* Fix false positive multiline case (test added)
* Fix false positive slice expression (test added)
* Fix false positive index expression (test added)
* Support to configure/allow cuddle declarations (example config added)
* Support to configure/allow case blocks to end with whitespace (example config added)
* Support cuddle defer http body close

Note: Had to update the test runner with `--max-same-issues=10` to allow my tests to cover added false-positives which yields the same error.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -259,7 +259,7 @@ type WSLSettings struct {
 	AllowAssignAndCallCuddle    bool `mapstructure:"allow-assign-and-call"`
 	AllowMultiLineAssignCuddle  bool `mapstructure:"allow-multiline-assign"`
 	AllowCaseTrailingWhitespace bool `mapstructure:"allow-case-trailing-whitespace"`
-	AllowCuddleDeclaration      bool `mapstructure:"allow-declarations"`
+	AllowCuddleDeclaration      bool `mapstructure:"allow-cuddle-declarations"`
 }
 
 var defaultLintersSettings = LintersSettings{
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -255,9 +255,11 @@ type GocognitSettings struct {
 }
 
 type WSLSettings struct {
-	StrictAppend               bool `mapstructure:"strict-append"`
-	AllowAssignAndCallCuddle   bool `mapstructure:"allow-assign-and-call"`
-	AllowMultiLineAssignCuddle bool `mapstructure:"allow-multiline-assign"`
+	StrictAppend                bool `mapstructure:"strict-append"`
+	AllowAssignAndCallCuddle    bool `mapstructure:"allow-assign-and-call"`
+	AllowMultiLineAssignCuddle  bool `mapstructure:"allow-multiline-assign"`
+	AllowCaseTrailingWhitespace bool `mapstructure:"allow-case-trailing-whitespace"`
+	AllowCuddleDeclaration      bool `mapstructure:"allow-declarations"`
 }
 
 var defaultLintersSettings = LintersSettings{
@@ -289,9 +291,11 @@ var defaultLintersSettings = LintersSettings{
 		MinComplexity: 30,
 	},
 	WSL: WSLSettings{
-		StrictAppend:               true,
-		AllowAssignAndCallCuddle:   true,
-		AllowMultiLineAssignCuddle: true,
+		StrictAppend:                true,
+		AllowAssignAndCallCuddle:    true,
+		AllowMultiLineAssignCuddle:  true,
+		AllowCaseTrailingWhitespace: true,
+		AllowCuddleDeclaration:      false,
 	},
 }
 
--- a/test/testdata/wsl.go
+++ b/test/testdata/wsl.go
@@ -164,3 +164,11 @@ func allowTrailing(i int) {
 		fmt.Println("three")
 	}
 }
+
+// ExampleSomeOutput simulates an example function.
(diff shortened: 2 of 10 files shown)


bombsimon:
Build for 1.13 OK but not for 1.12. Can reproduce with `make test` (or just the step `go test`). Seems to be something with `TestConcurrentReadsAndWrites` from `internal/renameio`, an intermittent error:

```sh
$ for _ in {1..5}; do go test ./internal/renameio/... -count=1; done

ok      github.com/golangci/golangci-lint/internal/renameio     16.203s
ok      github.com/golangci/golangci-lint/internal/renameio     15.782s
ok      github.com/golangci/golangci-lint/internal/renameio     16.256s
--- FAIL: TestConcurrentReadsAndWrites (15.79s)
    renameio_test.go:75: unexpected error: rename /var/folders/q3/bf76j42160j2nt4s2tz1_fd40000gn/T/renameio152466423/blob.bin614503871.tmp /var/folders/q3/bf76j42160j2nt4s2tz1_fd40000gn/T/renameio152466423/blob.bin: no such file or directory
    renameio_test.go:127: 2047 (of 2048) writes succeeded; want ≥ 2048
    renameio_test.go:142: 2048 (of 2048) reads succeeded (ok: ≥ 2048)
FAIL
FAIL    github.com/golangci/golangci-lint/internal/renameio     15.964s
FAIL
ok      github.com/golangci/golangci-lint/internal/renameio     16.014s
```

I'll try to see i I can figure anything out if this isn't already a known issue.

pkg/config/config.go
```diff
@@ -255,9 +255,11 @@ type GocognitSettings struct {
...
 
 type WSLSettings struct {
-	StrictAppend               bool `mapstructure:"strict-append"`
-	AllowAssignAndCallCuddle   bool `mapstructure:"allow-assign-and-call"`
-	AllowMultiLineAssignCuddle bool `mapstructure:"allow-multiline-assign"`
+	StrictAppend                bool `mapstructure:"strict-append"`
+	AllowAssignAndCallCuddle    bool `mapstructure:"allow-assign-and-call"`
+	AllowMultiLineAssignCuddle  bool `mapstructure:"allow-multiline-assign"`
+	AllowCaseTrailingWhitespace bool `mapstructure:"allow-case-trailing-whitespace"`
+	AllowCuddleDeclaration      bool `mapstructure:"allow-declarations"`
```
jirfag:
```suggestion
	AllowCuddleDeclaration      bool `mapstructure:"allow-cuddle-declarations"`
```

bombsimon: 🤦🏼‍♀️Thanks!

bombsimon:
I forgot that I tried the fix/commit button from GitHub after the requested change and had rebased on master locally to avoid conflicts. Did a force push (with lease) to get back to original state and had to re-do the commit for bad field tag.

Since I had a PR with a bug merged I added a test and renamed this to v1.2.5.

test/testdata/wsl.go
```diff
@@ -100,3 +100,75 @@ func f3() int {
...
+func indexExpr() {
+	var aMap = map[string]struct{}{"key": {}}
+
+	key := "key"
+	if k, ok := aMap[key]; ok {
+		fmt.Println(k)
+	}
+
+	xxx := "xxx"
+	if _, ok := aMap[key]; ok { // ERROR "if statements should only be cuddled with assignments used in the if statement itself"
```
jirfag: but how to rewrite this code properly? it's the common coding pattern

bombsimon: I see how that's a question for the linter itself and there's an issue [here](https://github.com/bombsimon/wsl/issues/249) to allow the variable to be cuddled if it's used anywhere in the block body. Before this PR the linter gave an error even if `xxx` was used as the index expression. :(

The way to rewrite to satisfy the linter:

```go
notUsedInIf := "separated from the statement"

if _, ok := aMap[key]; ok {
    fmt.Println("Not used here either")
    // Many lines later...
    fmt.Println(notUsedInIf)
}
```

And if the variable is used it may be cuddled.

```go
key := "used in if"
if _, ok := aMap[key]; ok {
    fmt.Println("ok")
}
```

Or if it's used after the condition.

```go
xxx := "not in if but first in block"
if _, ok := aMap[key]; ok {
    fmt.Println(xxx) // Ok
}
```

bombsimon: I got a bit confused after writing my comment and had a reviset to [Effective Go](https://golang.org/doc/effective_go.html). I see that a lot of the code uses a pattern that WSL discourages in favour of readability. Although I don't see newlines being discussed in the documentation and the examples doesn't state it's the only (or preferred way) to handle empty lines.

The reason for me adding it to this project is that I got asked by two of the places I've worked who uses `golangci-lint` and uses this kind of style to improve readability. I got a bit anxious about it since it's mainly based on personal preferences and sent and e-mail to you but I didn't get any response so I didn't revert anything.

I know it's not much but since this got merged in v1.20.0 the linter have had 10 stars and a few issues added so there seems to be a bit of an interest. Although maybe not enough to be a part of this project?

Do you think this kind of linter/styling is out of scope for golangci-lint? If so, do you wan't me to revert it being added?

If you wan't more details and thoughts behind this linter please give the README from the repository a few minutes!

jirfag: got it, thank you!

PR #822, opened 2019-10-15
deadline is now deprecated, but should be taking its value from the configuration if set

dcaba:
Hi,

  I'm afraid the PR https://github.com/golangci/golangci-lint/pull/793 generated a side effect: while deadline still works and becomes the effective timeout setting when supplying it as a flag/cmdline arg, we've realized deadline in the configuration file is no longer being honored (ok, its deprecated, but probably we wanted this to become a breaking change, forcing all consumers to immediately change their config): see the test we are incorporating, that clearly reproduces the situation.

 Proposing a fix (but not prod of it), so deadline is still considered (no matter where it comes from) if its value is different of the default timeout (that becomes a constant). There may be alternative implementations or better places to put this logic, but I didnt want to make the config package aware of the default value...

  Sorry if the explanation is not clear: let me know if that is the case and I will do my best to clarify.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -112,6 +112,6 @@ issues:
 # golangci.com configuration
 # https://github.com/golangci/golangci/wiki/Configuration
 service:
-  golangci-lint-version: 1.19.x # use the fixed version to not introduce new linters unexpectedly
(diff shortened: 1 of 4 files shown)


jirfag:
thank you!

PR #831, opened 2019-10-21
Fix autogen exclude for comments under package

hypnoglow:
The commit https://github.com/golangci/golangci-lint/commit/df4f6766baff8f2ce10ae7a6a4d81fe37b729989 broke the exclusion of autogenerated files that have appropriate comments below the `package` statement.

Some tools put `DO NOT EDIT` comment not in the beginning of the file but below the `package` statement, for example minimock: https://github.com/gojuno/minimock/blob/master/tests/formatter_mock.go

For such autogenerated files detection stopped working since the commit mentioned above.

This commit fixes the issue by allow scanning below the `package` statement.


jirfag:
thank you for this fix!

Issue #832, opened 2019-10-21
Proposal: check for unused //nolint comments


vearutop:
Sometimes the code needs a`//nolint` suppressor to disable check(s), but as the code evolves the `//nolint`may become redundant.

For example such comment may become obsolete when the symbol is actually used:
```go
// MyConst is something special.
// nolint:unused // To be used in subsequent PR.
const MyConst = "my-const"
```

I would like to check that `//nolint` comment was relevant to at least one linting issue and emit lint error otherwise.

tpounds:
@vearutop There was pr #740 that was recently closed to implement similar functionality.

tpounds:
cc @ashanbrown @jirfag

jirfag:
I like this idea and it's not difficult to implement in `nolint` filter

ldez:
`nolintlint` handle this case now:

```
  nolintlint:
    # Enable to ensure that nolint directives are all used. Default is true.
    allow-unused: false
```

PR #837, opened 2019-10-28
Add nolintlint linter as internal linter

ashanbrown:
This is a rework of https://github.com/golangci/golangci-lint/pull/740 with the `nolintlint` package moved into golangci-lint.  This also tries to address https://github.com/golangci/golangci-lint/issues/832.  It works by creating possible unused issues in the `nolintlint` linter and then filters them out in the `nolint` processor.  I realize this is a debatable approach to the problem but it does allow the decoupling of analysis and the nolint processor (and inadvertently allows caching of the nolintlint analysis).

The alternate approach to this is to move all the logic into the nolint processor.  I'd be happy to do that if that is the preferred approach.  Please let me know your thoughts.

Finally, this PR removes all the unused nolint statements in golangci-lint.  That could be done in another PR but it is interesting to see what the effect of this is.


ashanbrown:
Greetings, was wondering if this is still something you would consider?  I’d be happy to update if you are.  Please let me know.  Thank you.

ernado:
Hi, sorry for lack of feedback, seems like no maintainer is currently free to review new features :(
I hope that it will change soon, but currently we are (or at least myself) trying to focus on bugfixes.

Thank you for your PR and sorry again.

PR #849, opened 2019-11-09
Add rows.Err weather checked linter

jingyugao:
Check weather `rows.Err` is checked.
For example:
```golang
rows, _ := db.Query("select id from tb") // Wrong case
if err != nil {
	// handle error
}
for rows.Next(){
	// handle rows
}
// if rows.Err()!=nil{
	// forget to check rows.Err()
// }
```
If we forget to check `rows.Err`,and unluckly an invaliad connection error happend when fetch data from database, we will get an incomplete data. This is really dangerous.
Emm, yes, my company just make a big mistake because of this.


jingyugao:
```shell   
 cache_test.go:139: go: internal cache error: cache verify failed: id=0100000000000000000000000000000000000000000000000000000000000000 changed:<<<
823        
824        >>>
825        old: cb8379ac2098aa165029e3938a51da0bcecfc008fd6795f401178647f96c5b34 3
826        new: ba7816bf8f01cfea414140de5dae2223b00361a396177a9cb410ff61f20015ad 3
```

What's wrong with it?  It seems I need clear some cache?

jingyugao:
By the way, how can I ignore some bad directories? These contain some bad code and cant be compiled.

jingyugao:
@jirfag PTAL

dajohi:
it fails this:
```
$ go vet -vettool=/home/dhill/go/bin/rowserrcheck  
# a
./a.go:13:15: rows err must be checked
```

```go
package a

import (
	"database/sql"
)

func get(db *sql.DB) (*sql.Rows, error) {
	rows, err := db.Query("")
	return rows, err
}

func main() {
	_, err := get(new(sql.DB))
	if err != nil {
		panic(err)
	}
}
```

jingyugao:
> it fails this:
> 
> ```
> $ go vet -vettool=/home/dhill/go/bin/rowserrcheck  
> # a
> ./a.go:13:15: rows err must be checked
> ```
> 
> ```go
> package a
> 
> import (
> 	"database/sql"
> )
> 
> func get(db *sql.DB) (*sql.Rows, error) {
> 	rows, err := db.Query("")
> 	return rows, err
> }
> 
> func main() {
> 	_, err := get(new(sql.DB))
> 	if err != nil {
> 		panic(err)
> 	}
> }
> ```



> it fails this:
> 
> ```
> $ go vet -vettool=/home/dhill/go/bin/rowserrcheck  
> # a
> ./a.go:13:15: rows err must be checked
> ```
> 
> ```go
> package a
> 
> import (
> 	"database/sql"
> )
> 
> func get(db *sql.DB) (*sql.Rows, error) {
> 	rows, err := db.Query("")
> 	return rows, err
> }
> 
> func main() {
> 	_, err := get(new(sql.DB))
> 	if err != nil {
> 		panic(err)
> 	}
> }
> ```

1. I dont think this code is rigorous, if we ignore the rows and err is not nil, then the connection will be leaked until keepalive find it.
2. If we treated it right, it seems no good way to fix it.I cant find how to check var name is "_" in ssa.

dajohi:
> > it fails this:
> > ```
> > $ go vet -vettool=/home/dhill/go/bin/rowserrcheck  
> > # a
> > ./a.go:13:15: rows err must be checked
> > ```
> > 
> > 
> > ```go
> > package a
> > 
> > import (
> > 	"database/sql"
> > )
> > 
> > func get(db *sql.DB) (*sql.Rows, error) {
> > 	rows, err := db.Query("")
> > 	return rows, err
> > }
> > 
> > func main() {
> > 	_, err := get(new(sql.DB))
> > 	if err != nil {
> > 		panic(err)
> > 	}
> > }
> > ```
> 
>     1. I dont think this code is rigorous, if we ignore the rows and err is not nil, then the connection will be leaked until keepalive find it.
> 
>     2. If we treated it right, it seems no good way to fix it.I cant find how to check var name is "_" in ssa.

yes, i misunderstood the error; i retract that test :)

What do you think about changing the output from `rows err must be checked` to `rows.Err() must be checked` or possibly `varname.Err() must be checked` ?

jingyugao:
@dajohi How about use `rows.Err must be checked`,the `()` seems not correctly recognized by `analysistest`

jingyugao:
@dajohi PTAL

jirfag:
thank you!

PR #852, opened 2019-11-10
Add linter testpackage

maratori:
Add linter [testpackage](https://github.com/maratori/testpackage)

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -222,6 +222,9 @@ linters-settings:
   rowserrcheck:
     packages:
       - github.com/jmoiron/sqlx
+  testpackage:
+    # regexp pattern to skip files
+    skip-regexp: (export|internal)_test\.go
   unparam:
     # Inspect exported functions, default is false. Set to true if no external program/library imports your code.
     # XXX: if you enable this setting, unparam will report a lot of false-positives in text editors:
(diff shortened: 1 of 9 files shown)


.golangci.example.yml
```diff
@@ -213,7 +213,9 @@ linters-settings:
   dogsled:
     # checks assignments with too many blank identifiers; default is 2
     max-blank-identifiers: 2
-
+  testpackage:
+    # regexp pattern to skip files
+    skip-regexp: (export|internal)_test\.go
```
ernado, line 218: Can we do same with `exlude-rules`? Seems like this option is excessive, what do you think?

vendor/modules.txt
```diff
@@ -108,6 +108,8 @@ github.com/kisielk/gotool/internal/load
 github.com/konsorten/go-windows-terminal-sequences
 # github.com/magiconair/properties v1.8.1
 github.com/magiconair/properties
+# github.com/maratori/testpackage v1.0.0
```
ernado, line 111: Please rebase and remove vendoring :slightly_smiling_face:

.golangci.example.yml
maratori: `(export|internal)_test\.go` is default skip pattern for testpackage linter. And this default is good enough for most projects. You should change it in very rare cases.

Correct me if I'm wrong. If we replace it with `exclude-rules`, we have to define default pattern for each new project explicitly.

That's why I've added this setting.

ernado: Oh, now I see, thank you.

vendor/modules.txt
maratori: Done

PR #856, opened 2019-11-14
Use correct semver for wsl

bombsimon:
PR according to e-mail with @tpounds and the issue bombsimon/wsl#60

The wsl v1.2.7 release didn't properly follow semver and had breaking changes. I resolved this by making a new v1.2.8 release backwards compatible (but not correct). To have a proper long term solution I've released v2.0.0 and with this PR I'll revert the merge of the faulty release tag v1.2.7 and instead jump to a proper v2.0.0 release.

Please see linked issue or let me know for more details.

I hope this PR makes sense and that I've understood how to use semver and go modules in a proper way. Sorry for the extra work and inconvenience.


bombsimon:
Maybe I shouldn't revert the v1.2.7 merge and just go straight to v2.0.0? The bad release as a patch (and the fix up to v1.2.8) would still be used with the previous v.1.2.5 release.

Since this is the only project where wsl is used however I could actually remove the v1.2.7 release from wsl if this gets reverted so I guess it won't harm and it also makes it more clear that the release is unwanted.

tpounds:
@bombsimon I would just leave the faulty tag release so as to not erase history (even though it's incorrect). I think it's best to just move forward at this point.

PR #877, opened 2019-12-03
Add errcheck option WithoutTests

grongor:
I'm sorry but I didn't figure out how to write a test for this ...

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -82,6 +82,9 @@ linters-settings:
     # see https://github.com/kisielk/errcheck#excluding-functions for details
     exclude: /path/to/file.txt
 
+    # if true, checking of _test.go files is disabled
+    without-tests: false
+
   funlen:
     lines: 60
     statements: 40
--- a/README.md
+++ b/README.md
@@ -685,6 +685,9 @@ linters-settings:
     # see https://github.com/kisielk/errcheck#excluding-functions for details
     exclude: /path/to/file.txt
 
+    # if true, checking of _test.go files is disabled
+    without-tests: false
+
   funlen:
     lines: 60
     statements: 40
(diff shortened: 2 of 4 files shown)


tpounds:
@grongor Thanks for the pr! It looks like we need to update the fork of `errcheck`before we can support the newer options. @jirfag Any thoughts?

ernado:
This can be done with `issues.exclude-rules` config for any linter and any file name regex:
```yml
issues:
  exclude-rules:
    - path: _test\.go
      linters: [errcheck]
```

grongor:
@ernado nice, didn't think about that - thanks :)

jirfag:
Hi!
I agree with @ernado - let's not create different ways to do one thing. Closing it, sorry.

tpounds:
@jirfag It's probably ok to support the equivalent functionality when it exists in the dependency in order to maintain feature parity. We also need to make a pass at all of the golangci forked linters such as errcheck to bring them up to date.

jirfag:
1. it will confuse users - how to disable some linter from tests. Is it more correct to disable it in linter settings or in `exclude-rules` or by `--tests`. We have 3 ways to do it. And `exclude-rules` and `--test` work differently: the first just excluded issues but loads/parses all test files, the second doesn't load/parse test files. We need to document it better. 
2. I see no benefit from this feature parity. If the user would like to disable `errcheck` for tests - they can use `exclude-rules`. Maybe we can document it better in `README` or `.golangci.yml`.
3. this pull request doesn't work: `WithoutTests` works only for [packages loading](https://github.com/kisielk/errcheck/blob/master/internal/errcheck/errcheck.go#L183). golangci-lint does packages loading itself.

ernado:
Please use `exclude-rules` for this and open issue if other approach for handling such cases is needed.
Closing for now.

PR #878, opened 2019-12-03
pkg/golinters: add erris

### LINKED ISSUE #872: add erris linter
Since this is unlikely to be added to `vet`, it would be nice to add the `github.com/romanyx/erris` linter; happy to assist.

arnottcr:
This change adds a pkg/golinters.NewXxx func for the erris linter. The
configuration file has been updated to enforce this linter internally
too. As erris is now an external dependency and its MIT license has
been vendored into third_party.

Fix #872.


ernado:
Now we have https://github.com/Djarvur/go-err113 and I'm not sure that we should add linter with intersecting features.

@arnottcr are there any reasons to add them both?
Sorry that it took so long to came back with feedback, there were a lot of PR-s and issues to address.

arnottcr:
slick, it looks like err13 is a superset of erris, happy to use the more inclusive of the two

PR #880, opened 2019-12-03
Fix depguard config in yml

pohang:
This PR fixes the following issues:

* `packages-with-error-messages` should be `packages-with-error-message` based on the `mapstructure` tag from https://github.com/golangci/golangci-lint/blob/master/pkg/config/config.go#L165

* Due to https://github.com/spf13/viper/issues/348, after fixing above typo, it'll fail with
```
* 'linters-settings.Depguard.packages-with-error-message[github]' expected type 'string', got unconvertible type 'map[string]interface {}'
```
since alot of package names will have dots in them. A workaround is to specify as a list of maps, which will get merged by `mapstructure` (see [config](https://github.com/mitchellh/mapstructure/blob/master/mapstructure.go#L76)). Viper does have `WeaklyTypedInput` [enabled by default](https://github.com/spf13/viper/blob/master/viper.go#L874).

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -138,9 +138,9 @@ linters-settings:
     include-go-root: false
     packages:
       - github.com/sirupsen/logrus
-    packages-with-error-messages:
+    packages-with-error-message:
       # specify an error message to output when a blacklisted package is used
-      github.com/sirupsen/logrus: "logging is allowed only by logutils.Log"
+      - github.com/sirupsen/logrus: "logging is allowed only by logutils.Log"
   misspell:
     # Correct spellings using locale preferences for US or UK.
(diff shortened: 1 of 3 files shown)


tpounds:
@pohang Thanks for the contribution! Can you make your updates to `README.tmpl.md` and rerun with `make README.md` and then add both files to the pr?

pohang:
@tpounds All the changes to README.tmpl.md come from updating the two `.yml` files and then running `make README.md`. I don't think that updating `README.tmpl.md` is necessary here.

tpounds:
@pohang My mistake. It looks like the tests were failing for an unrelated issue. I've merged in your commit. Thanks!

PR #917, opened 2020-01-10
Add zsh completion to homebrew

darklore:
zsh completion has been implemented at https://github.com/golangci/golangci-lint/pull/903 .
So this PR installs it in homebrew.


darklore:
I think I've signed the CLA.

PR #920, opened 2020-01-12
make uniq-by-line configurable

to6ka:
hi,
sometimes it's usefull to see all linters, that was triggered by the line
so i have added new param for this


jirfag:
thank you!

Issue #897, opened 2019-12-25
docker run failed


DarthPestilane:
run this

```
docker run --rm -v $(pwd):/app -w /app  golangci/golangci-lint:v1.21.0 golangci-lint run
```

as readme.md said.

and error happend:

```
level=error msg="Running error: context loading failed: failed to load program with go/packages: could not determine GOARCH and Go compiler"
```

tested on macOS

tpounds:
@DarthPestilane Does this only happen on macOS with Docker? Do you get the same error with the binary release?

ernado:
Seems like same as #884

Luzifer:
- Docker: `19.03.5-ce`
- OS: Archlinux

```console
# tags=(latest latest-alpine v1.21-alpine v1.22-alpine v1.22.2-alpine)
# for tag in "${tags[@]}"; do echo "==== $tag"; docker run --rm -ti -v $(pwd):$(pwd) -w $(pwd) golangci/golangci-lint:$tag golangci-lint run; done
==== latest
ERRO Running error: context loading failed: failed to load program with go/packages: could not determine GOARCH and Go compiler 
==== latest-alpine
ERRO Running error: context loading failed: failed to load program with go/packages: could not determine GOARCH and Go compiler 
==== v1.21-alpine
ERRO Running error: context loading failed: failed to load program with go/packages: could not determine GOARCH and Go compiler 
==== v1.22-alpine
ERRO Running error: context loading failed: failed to load program with go/packages: could not determine GOARCH and Go compiler 
==== v1.22.2-alpine
ERRO Running error: context loading failed: failed to load program with go/packages: could not determine GOARCH and Go compiler 
```

Not experiencing the same issue when running `golangci-lint run` locally with version built from `master` on 2020-01-02.

Luzifer:
Follow-up on my last comment:

When executing the same commands with `golangci/golangci-lint:v1.22.2-alpine` and enabled `--modules-download-mode=vendor` the error does **not** occur.

ClaudiaJ:
We found in https://github.com/reviewdog/action-golangci-lint/issues/21 this error occurs with projects importing private module dependencies.

I've tested further by downloading modules to host, mounting the host's `${GOPATH}/pkg/mod` into the container, and passing `--modules-download-mode=readonly`:
```
go mod download
docker run --rm \
        -v ${GOPATH}/pkg/mod:/go/pkg/mod \
        -v $(pwd):/app \
        -w /app \
[...]

PR #922, opened 2020-01-13
Dockerfile.alpine: add Git & Mercurial

Dominik-K:
Fixes partly #897

### DIFF
--- a/build/Dockerfile.alpine
+++ b/build/Dockerfile.alpine
@@ -3,6 +3,9 @@ FROM golang:1.13-alpine
 # required to support cgo
 RUN apk --no-cache add gcc musl-dev
 
+# needed most times for `go get`, etc. See https://github.com/docker-library/golang/issues/80
+RUN apk add --no-cache git mercurial
+
 # don't place it into $GOPATH/bin because Drone mounts $GOPATH as volume
 COPY golangci-lint /usr/bin/
 CMD ["golangci-lint"]


build/Dockerfile.alpine
```diff
@@ -1,7 +1,7 @@
 FROM golang:1.13-alpine
 
-# required to support cgo
-RUN apk --no-cache add gcc musl-dev
+# needed most times for `go get`, etc. See https://github.com/docker-library/golang/issues/80
+RUN apk add --no-cache git mercurial
```
jirfag: why did you remove `gcc` and `musl-dev`?

Dominik-K: @jirfag The `golang:alpine` images already install them. See e.g. https://github.com/docker-library/golang/blob/7b69aab8c4a9117743168ca8b19b314c0307e415/1.13/alpine3.11/Dockerfile#L16-L17
(noted it also in the commit message, though)

shogo82148: `gcc` and `musl-dev` are only used for building Go, and they are removed from the `golang:alpine` image at https://github.com/docker-library/golang/blob/7b69aab8c4a9117743168ca8b19b314c0307e415/1.13/alpine3.11/Dockerfile#L55

Actually, running `gcc` fails with the `golang:alpine` image.

```
$ docker run golang:1.13-alpine gcc
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"gcc\": executable file not found in $PATH": unknown.
ERRO[0000] error waiting for container: context canceled
$ docker run golang:1.13-alpine bash
docker: Error response from daemon: OCI runtime create failed: container_linux.go:346: starting container process caused "exec: \"bash\": executable file not found in $PATH": unknown.
ERRO[0000] error waiting for container: context canceled
```

Dominik-K: @shogo82148 Thanks for the clarification. 
@jirfag I've undone this commit. Sorry about my mistake.

PR #936, opened 2020-01-22
Add Intuit to README.md

alexec:
Note that we Open Source code under the argoproj org as well as intuit org.

### DIFF
--- a/README.tmpl.md
+++ b/README.tmpl.md
@@ -125,7 +125,7 @@ The following companies/products use golangci-lint:
 * [Red Hat OpenShift](https://github.com/openshift/telemeter)
 * [Yahoo](https://github.com/yahoo/yfuzz)
 * [IBM](https://github.com/ibm-developer/ibm-cloud-env-golang)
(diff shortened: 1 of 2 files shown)


README.tmpl.md
```diff
@@ -125,6 +125,7 @@ The following companies/products use golangci-lint:
 * [Red Hat OpenShift](https://github.com/openshift/telemeter)
 * [Yahoo](https://github.com/yahoo/yfuzz)
 * [IBM](https://github.com/ibm-developer/ibm-cloud-env-golang)
+* [Intuit](https://github.com/argoproj)
```
tpounds: It probably makes sense to say `Argo Project` rather than `Intuit` since that's what it links to.

alexec: interesting one - should we list both? we use it at Intuit on our OSS projects, which are published in various places including https://github.com/intuit

tpounds: Adding both links would be fine. In general, I think it's important for the descriptions to be precise about what they link to.

PR #937, opened 2020-01-22
Change writeFixedFile to replace the right line-range.

csilvers:
The function assumed, previously, that issue.Line always equals
issue.LineRange.From.  But this needn't be true, and the code needn't
assume it.  Now we actually replace the specified line-range.

Please make sure you didn't directly change `README.md`: it should be changed only by changing `README.tmpl.md` and running `make README.md`.


csilvers:
I couldn't find a good place to add any tests for this.  I'm happy to do so if you can give me any pointers, but maybe the change is "obvious" enough not to need it.

PR #939, opened 2020-01-24
Update go-mnd to v1.2.0 and make it ignore 0 and 1 numbers entirely

vitalyisaev2:
The author of `go-mnd` has just implemented ignore list for numbers (https://github.com/tommy-muehle/go-mnd/issues/3). I think that `0` and `1` should not be considered as magic numbers by default, so I've fixed both `.golangci.example.yml` and `.golangci.yml`.


vitalyisaev2:
@tpounds @jirfag please have a look, I've fixed the build

ernado:
Also seems like we've updated go-mnd to v2.0.0 so I'm not sure if this PR is still relevant.

PR #983, opened 2020-03-07
deps: upgrade mnd from v1.1.1 to 2.0.0

markus-wa:
I had to directly reference the commit SHA instead of `@v2.0.0` because of https://github.com/tommy-muehle/go-mnd/pull/6

### new features
- **Add "exclude" option and exclude test files by default**
- **Add 1 to be not considered as magic number**
- Add new option to specify further ignored numbers

see also #939 which proposed the upgrade to `v1.2.0`.


bombsimon:
Since `1` is no longer considered a magic number I suppose [this test](https://github.com/golangci/golangci-lint/blob/master/test/testdata/gomnd.go#L12) should be updated/removed for the pipeline to succeed.

thaJeztah:
looks like you need to run `go mod vendor` to update the vendor directory

markus-wa:
should be fixed, let's see what Travis thinks :slightly_smiling_face:

PR #989, opened 2020-03-11
vet: Support disable rule when enable-all is true

ferhatelmas:
Similar to the behavior of command line flags:
enable-all to enable existing linters then disable unwanted ones.
But for vet rules.

For example

```yaml
govet:
  enable-all: true
  disable:
    - nilfunc
```

### DIFF
--- a/pkg/golinters/govet.go
+++ b/pkg/golinters/govet.go
@@ -101,6 +101,11 @@ var (
 
 func isAnalyzerEnabled(name string, cfg *config.GovetSettings, defaultAnalyzers []*analysis.Analyzer) bool {
 	if cfg.EnableAll {
+		for _, n := range cfg.Disable {
+			if n == name {
+				return false
+			}
+		}
 		return true
 	}
 	// Raw for loops should be OK on small slice lengths.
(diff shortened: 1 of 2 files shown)


ernado:
Thank you!
This looks OK, but are we sure that we want to use "enable-all" pattern that retrospectively is prone to unexpected failures after upgrade? 
Currently we are discouraging usage of "enable-all" pattern and suggesting to explicitly enable linters.

Is there any difference with "vet"?

ferhatelmas:
> This looks OK
Good to hear.

> are we sure that we want to use "enable-all" pattern that retrospectively is prone to unexpected failures after upgrade
Change in behavior is possible but failure shouldn't happen since it will disable more rules so will be running less. 

However, clear mention in changelog should be enough I think.

> Is there any difference with "vet"?
I think yes.

`go vet ./...` runs everything
`go vet -nilfunc=false ./...` disables `nilfunc` rule but runs every other rule. 

This behavior wasn't supported with golangci-lint so far but with this PR, example config in the description enables it.

ernado:
Oh, now I see, the difference, thank you!

PR #997, opened 2020-03-16
Remove reference to golangci.com in README

evalexpr:
This project is shutting down on the 15th April. See the blog below:
https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e

### DIFF
--- a/README.md
+++ b/README.md
@@ -1195,10 +1195,7 @@ By doing this you won't create new issues in your code and can choose fix existi
 
 **How to use `golangci-lint` in CI (Continuous Integration)?**
 
-You have 2 choices:
-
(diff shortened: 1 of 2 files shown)


lopezator:
Thank you for this, we ❤️  doc contributions.

Apart from the changes proposed here there a more mentions in the README like the `golangci.com` configuration, for example.

This possibly also will need some change changes.

@jirfag do you prefer to do all at once, or are you ok with merging this and continue the remaining needed changes in future PRs?

evalexpr:
I’m happy to make further changes if we want to remove everything in one PR. I’ll grep for the obvious stuff tomorrow and see if it’s easily removable.

ernado:
Now we have a https://github.com/golangci/golangci-lint-action which is doing pretty similar what golangci did.

Probably we can promote this method as alternative.

jirfag:
thank you,
I'm ok to remove it iteratively. In 1-2 weeks I would like to promote our GitHub action in README.

PR #1019, opened 2020-04-12
Add go-err113 linter support

tdakkota:
Adds https://github.com/Djarvur/go-err113 linter support

### DIFF
--- a/pkg/golinters/goerr113.go
+++ b/pkg/golinters/goerr113.go
@@ -1,0 +1,19 @@
+package golinters
+
+import (
+	"github.com/Djarvur/go-err113"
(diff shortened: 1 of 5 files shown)


pkg/golinters/goerr113.go
```diff
@@ -0,0 +1,18 @@
+package golinters
+
+import (
+	"github.com/Djarvur/go-err113"
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
```
golangcibot, line 5: File is not `goimports`-ed with -local github.com/golangci/golangci-lint (from `goimports`)
```suggestion

```

PR #1032, opened 2020-04-21
doc: provide instructions for from-source installation

### LINKED ISSUE #1006: request: make source builds a first class approach to installation
The issue template is not appropriate for this issue, so I have removed it.

In the README's installation section there is a direction to not install golangci-lint from source. Given the support for reliable builds with Go modules, this seems like an extremely odd approach.

It would be good if there were added support for source install of golang-ci lint. 

Addressing the directive's points in the README:

> Please, do not install `golangci-lint` by `go get`:

> 1. [`go.mod`](https://github.com/golangci/golangci-lint/blob/master/go.mod) replacement directive doesn't apply. It means you will be using patched version of `golangci-lint`.
    **There are no `replace` directives in that file, and if there were that would be breaking the rationale for the `replace` directive.**
> 2. it's much slower than binary installation
    **Surely this is up to the user to decide, being a balance between speed of install and the capacity to audit code that is running on the user's machine. Personally, I do not install from binary on my hardware, though I do allow this for travis builds.**
> 3. its stability depends on your Go version (e.g. on [this compiler Go <= 1.12 bug](https://github.com/golang/go/issues/29612)).
    **This can be avoided by only supporting source installs for Go versions greater than 1.13. This could easily be stated as building from source is only supported for currently supported Go releases.**
> 4. it's not guaranteed to work: e.g. we've encountered a lot of issues with Go modules hashes.
    **Has this been reported anywhere?**
> 5. it allows installation from `master` branch which can't be considered stable.
    **If modules are used correctly, this is not the case.**

kortschak:
Please take a look.

Closes #1006.

### DIFF
--- a/README.tmpl.md
+++ b/README.tmpl.md
@@ -111,7 +111,7 @@ docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:{{.LatestVersion}}
 Go source installations are supported for the two most recent Go releases.
 
 ```bash
-GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint
+GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@{{.LatestVersion}}
 ```
 
 ## Trusted By
--- a/README.tmpl.md
+++ b/README.tmpl.md
@@ -108,13 +108,11 @@ docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:{{.LatestVersion}}
 
 ### Go
 
-Please, do not install `golangci-lint` by `go get`:
+Go source installations are supported for the two most recent Go releases.
 
-1. [`go.mod`](https://github.com/golangci/golangci-lint/blob/master/go.mod) replacement directive doesn't apply. It means you will be using patched version of `golangci-lint`.
-2. it's much slower than binary installation
-3. its stability depends on your Go version (e.g. on [this compiler Go <= 1.12 bug](https://github.com/golang/go/issues/29612)).
-4. it's not guaranteed to work: e.g. we've encountered a lot of issues with Go modules hashes.
-5. it allows installation from `master` branch which can't be considered stable.
+```bash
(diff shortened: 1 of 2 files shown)


README.tmpl.md
```diff
@@ -108,13 +108,11 @@ docker run --rm -v $(pwd):/app -w /app golangci/golangci-lint:{{.LatestVersion}}
...
-Please, do not install `golangci-lint` by `go get`:
+Go source installations are supported for the two most recent Go releases.
 
-1. [`go.mod`](https://github.com/golangci/golangci-lint/blob/master/go.mod) replacement directive doesn't apply. It means you will be using patched version of `golangci-lint`.
-2. it's much slower than binary installation
-3. its stability depends on your Go version (e.g. on [this compiler Go <= 1.12 bug](https://github.com/golang/go/issues/29612)).
-4. it's not guaranteed to work: e.g. we've encountered a lot of issues with Go modules hashes.
-5. it allows installation from `master` branch which can't be considered stable.
+```bash
+GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint
```
ernado, line 114: Please use `@{{.LatestVersion}}`, this command will <kbd>Ctrl</kbd> + <kbd>C</kbd>  <kbd>Ctrl</kbd> + <kbd>V</kbd> and we should use version pinning.

kortschak: It's not clear to me how pinning a dynamically defined string is better than using the `go get` behaviour which would do the same thing. But OK.

ernado: > It's not clear to me how pinning a dynamically defined string is better than using the go get behaviour which would do the same thing. But OK.

Sorry for being unclear. I'm talking about running this command in CI, like that:
```yaml
lint:
  run: GO111MODULE=on go get github.com/golangci/golangci-lint/cmd/golangci-lint@1.24.0
```

You are right, there is no difference outside this case.

PR #1036, opened 2020-04-22
Adding case-sensitive exclude processor

iwankgb:
In some scenarios (such us introducing `golangci-lint` to an existing project) it's useful to be able to exclude rules in case-sensitive manner so that exported fields, types, function etc are ignored. Such an approach allows to improve existing codebase iteratively without introducing backward incompatible changes.


ernado:
Is it possible to accomplish via exclude-rules?

iwankgb:
`processor.ExcludeRules` relies on case-insensitive regular expressions for [text and source](https://github.com/golangci/golangci-lint/blob/cbc2134fafd906a634c124679b3c61f55758250b/pkg/result/processors/exclude_rules.go#L45). It leaves me with same issue of telling exported form unexported. It could be resonable to apply case sensitivity/insensitivity logic to these expressions, though.

ernado:
Looks like adding `(?i)` in exclude rules were my mistake.
Probably the right choice is to remove it, but it will be breaking change.

What do you think about adding `case-sensitive` parameter to `ExcludeRule` which will be `false` by default?
I known that default value is bad, but we will change it to `true` in v2 to preserve semantic versioning rules.

Example:
```yaml
---
issues:
  exclude-rules:
    - path: internal/hmac/
      case-sensitive: true
```

If  `case-sensitive` is true, the `(?i)` will not be appended to  regexp.

iwankgb:
Sounds reasonable to me. Should there be two parameters (for exclude and exclude rules) or single one changing behaviours of both of them?

ernado:
> Should there be two parameters (for exclude and exclude rules) or single one changing behaviours of both of them?

I'm not sure.
```yml
---
issues:
   excludes-case-sensitive: true
```
You are talking about something like that?

The per-rule option is pretty obvious and easy to implement, but one global option that change both `exclude` and `exclude-rules` behavior will be more ergonomical.

iwankgb:
@ernado let me know what you think :)

pkg/result/processors/exclude_rules.go
```diff
@@ -118,3 +118,36 @@ func (ExcludeRules) Name() string { return "exclude-rules" }
 func (ExcludeRules) Finish()      {}
 
 var _ Processor = ExcludeRules{}
+
+type ExcludeRulesCaseSensitive struct {
+	*ExcludeRules
+}
+
+func NewExcludeRulesCaseSensitive(rules []ExcludeRule, lineCache *fsutils.LineCache, log logutils.Log) *ExcludeRulesCaseSensitive {
```
ernado, line 126: I've expected a `caseSensitive bool` parameter here to aviod code duplication.
Boolean params is anti-pattern, but I think that we can make an exception here.

iwankgb:
I reuse the code anyway and this approach keeps processors cleaner. It follows open/close principle too by extending existing functionality rather then changing it. Perhaps `New*` functions could take additional argument and configure same `Processor` differently. It would be the best solution, I think.

iwankgb:
It would be safe to do unless you expect someone to use `golangci-lint` as a library. We should be careful about changing signatures of existing functions it it may be the case.

ernado:
Seems like we are not maintaining backward compatibility for programmatic API, so it is safe to change signatures.

I can't expect someone to use `golangci-lint` as a library, never heard of such a use-case.

pkg/result/processors/exclude_rules.go
```diff
@@ -118,3 +118,36 @@ func (ExcludeRules) Name() string { return "exclude-rules" }
...
+			parsedRule.text = regexp.MustCompile(rule.Text)
+		}
+		if rule.Source != "" {
+			parsedRule.source = regexp.MustCompile(rule.Source)
+		}
+		if rule.Path != "" {
+			parsedRule.path = regexp.MustCompile(rule.Path)
+		}
+		r.rules = append(r.rules, parsedRule)
+	}
```
ernado, line 146: Actually the only thing that confuses me in this PR is code duplication in this place.
Can you please extract this range loop into a private function and use in both `ExcludeRulesCaseSensitive` and `ExcludeRules`?

pkg/result/processors/exclude_rules.go
iwankgb: I think we do not need boolean parameter if we have separate constructors.

pkg/result/processors/exclude_rules.go
iwankgb: The loop has been extracted. :bomb:

PR #1044, opened 2020-04-25
update gomodguard

ryancurrah:
Update `gomodguard` to version `v1.0.4` to not block a module import when the linted module is a recommended alternative

> If the linted module imports a blocked module but the linted module is in the recommended modules list the blocked module is ignored. Usually, this means the linted module wraps that blocked module for use by other modules, therefore the import of the blocked module should not be blocked.


ernado:
The killer feature would be license-based blocks, e.g. forbid AGPL :)

Issue #456, opened 2019-03-22
please reconsider ignoring missing doc comments


bhcleek:
We are in the process of adding support to [vim-go](https://github.com/fatih/vim-go) for `golangci-lint` and plan to very soon make it the default to replace `gometaliner` when running vim-go's `:GoMetaLinter`.

Please reconsider excluding missing doc comments from golint by default. The README.md justifies the decision with

> The rare codebase has such comments

But that rationale is false _and_ goes against Go recommended best practices. Even if the rationale were true, appealing to such a common situation alone is insufficient to justify perpetuating bad practices.

[Effective Go](https://golang.org/doc/effective_go.html#commentary) specifically states

>  Every exported (capitalized) name in a program should have a doc comment.

And the only reason we have such great tools as godoc.org is _because_ so many packages implement that guideline.

pseidemann:
+1
I was wondering why golint is disabled by default. I think this is one of the most important linters which also probably never has false positives

gchaincl:
+1

gchaincl:
I managed to enable it using these parameters:
`golangci-lint run --exclude-use-default=false --disable-all --enable=golint`

thaney071:
> I managed to enable it using these parameters:
> `golangci-lint run --exclude-use-default=false --disable-all --enable=golint`

that turn back on more than just the doc comment requirements.

It would be nice if this particular one was more configurable.

matoous:
@jirfag can I get your comment on this? I think @bhcleek made a valid point and if we agree on it I will fix this issue.

jirfag:
Hi!
In my opinion, 80+% of projects analyzed by `golangci-lint` have >= 30 identifiers without comments. If we will enable `golint` warnings it will be annoying for 80% of users. If it's right it doesn't matter what's right or what's not by Go guidelines: this project has the key feature of convenience (no false-positives, easy use, easy and flexible configuration) for users. I really try to maximize the performance and convenience of `golangci-lint`. Anyone who would like to force comments can disable this default behavior by `--exclude-use-default=false`.

So, let's talk about numbers. If you are really interested in changing the default behavior please help with calculating the following metrics:
1. share of projects with `--exclude-use-default=false`
2. share of projects with >=30/>=50/>=100/==0 identifiers without comments

lopezator:
Even if we ever consider doing this, IMHO this is a "breaking" change, so I would delay it to 2.0.

matoous:
[...]

PR #1045, opened 2020-04-25
Re-enable default excludes by ID

JohnStarich:
Adds IDs to default exclusions so they can be re-included on a case-by-case basis.

Backward-compatible part of https://github.com/golangci/golangci-lint/issues/456

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -101,7 +101,11 @@ var DefaultExcludePatterns = []ExcludePattern{
 	},
 }
 
-func GetDefaultExcludePatternsStrings(include []string) []string {
+func GetDefaultExcludePatternsStrings() []string {
+	return GetExcludePatternsStrings(nil)
+}
+
+func GetExcludePatternsStrings(include []string) []string {
 	includeMap := make(map[string]bool, len(include))
 	for _, inc := range include {
 		includeMap[inc] = true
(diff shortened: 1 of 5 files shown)


pkg/config/config.go
```diff
@@ -30,70 +30,88 @@ var OutFormats = []string{
...
 	{
+		ID:      "EXC0010",
 		Pattern: "Potential file inclusion via variable",
 		Linter:  "gosec",
 		Why:     "False positive is triggered by 'src, err := ioutil.ReadFile(filename)'",
 	},
 }
 
-func GetDefaultExcludePatternsStrings() []string {
+func GetDefaultExcludePatternsStrings(include []string) []string {
```
ernado, line 104: Can you please create a new function so we can maintain backward compatibility in API?
E..g. `GetExcludePatternsStrings(include []string)` that will be called from `GetDefaultExcludePatternsStrings()`.

JohnStarich:
Thanks for the lightning-fast review! Updated 👍

JohnStarich:
Thank you @ernado! Easily the fastest turn-around on a PR I've had in a while 😄 

Thoughts on when the next release might be cut?

ernado:
> Thoughts on when the next release might be cut?

I'm planning to do it next week, currently busy with github action for golangci-lint :)
Also trying not to rush with new versions and should test it for a while.

JohnStarich:
Ok, no worries! Sounds fun 🙂

PR #1047, opened 2020-04-27
Exclude "go:generate" from 'lll' linter

landru29:
All comments concerning `go:generate` should be excluded from `lll`linter


ernado:
I'm not sure that it will fix the issue, effectively it is same as doing it via `exclude-rules`.

ptman:
Please don't close this even if nothing has happened unless it has been fixed.

PR #1050, opened 2020-04-28
Update godot linter

tetafro:
Fix checking `#gosec` in comments


tetafro:
Finally fixed it, but with a [dirty hack](https://github.com/tetafro/godot/blob/4432b5a4a9a7b4c1f24811c2246774411bbbc09a/godot.go#L104-L108).

@ernado hi, maybe you can help. I have a problems with detecting cgo code blocks, and I don't even see `import "C"` inside `analysis.Pass.File[].Imports`. Do know any trick to get this info?

ernado:
Seems like your solution for cgo is good, so let's stick with it.

PR #1062, opened 2020-05-03
Update gosec linter to v2; add .test files to .gitignore

### LINKED ISSUE #1052: Update gosec linter
Hi There,

Gosec has released v2 version and there  are many updates in code, can we please update golangci-lint code to support v2 version of gosec.

Thanks in advance.

Regards,
Nikhil

theckman:
This change updates the `gosec` linter to its v2 version. It seems like v2
maintained API compatibility for the functionality we use, so we only needed to
update the dependency and imports.

Please note, this does not use a tagged version of `gosec` and instead is the
latest `master` commit due to this issue: https://github.com/securego/gosec/issues/470.
Once this issue is resolved we should update `gosec` to a tagged release.

This also adds a line to the `.gitignore` file, to ignore the `.test` output
files that get generated by `go test`. This could accidentally get left behind,
so no use committing binary files.

Fixes #1052

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@
 /tools/node_modules
 /tools/svg-term
 /.vscode/
+*.test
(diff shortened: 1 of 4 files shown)


jirfag:
thank you!

Issue #1011, opened 2020-04-03
Can't run linter goanalysis_metalinter: rowserrcheck: failed prerequisites


sagikazarmark:
Thank you for creating the issue!

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
golangci-lint has version 1.24.0 built from 6fd4383 on 2020-03-15T11:30:52Z
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
linters-settings:
    golint:
        min-confidence: 0.1
    goimports:
        local-prefixes: logur.dev/logur

linters:
    enable-all: true
    disable:
        - funlen
        - wsl
        - gomnd
```

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
Go 1.14 on GitHub Actions
```

</details>

<details><summary>Verbose output of running</summary>

```console
$ golangci-lint run -v
[...]

Issue #995, opened 2020-03-16
Can't run linter goanalysis_metalinter 1.24.x: failed prerequisites


marcelloh:
I still have the same issue (but you guys closed it, which you shouldn't, because it is not fixed)

Using the new 1.24.0 with Go 1.14

<details><summary>Config file</summary>

```console
linters-settings:
  depguard:
    list-type: blacklist
    packages:
      # logging is allowed only by logutils.Log, logrus
      # is allowed to use only in logutils package
      - github.com/sirupsen/logrus
  dogsled:
    # checks assignments with too many blank identifiers; default is 2
    max-blank-identifiers: 2
  dupl:
    threshold: 100
  funlen:
    lines: 100
    statements: 50
  gocognit:
    # minimal code complexity to report, 30 by default (but we recommend 10-20)
    min-complexity: 10
  goconst:
    min-len: 2
    min-occurrences: 5
  gocritic:
    enabled-tags:
      - diagnostic
      - experimental
      - opinionated
      - performance
      - style
    disabled-checks:
      - wrapperFunc
      - dupImport # https://github.com/go-critic/go-critic/issues/845
      - ifElseChain
      - octalLiteral
      - commentedOutCode
      - commentFormatting
  gocyclo:
    min-complexity: 12
  godox:
    # report any comments starting with keywords, this is useful for TODO or FIXME comments that
    # might be left in the code accidentally and should be resolved before merging
    keywords: # default keywords are TODO, BUG, and FIXME, these can be overwritten by this setting
      - NOTE
      - OPTIMIZE # marks code that should be optimized before merging
      - HACK # marks hack-arounds that should be removed before merging
  gofmt:
    # simplify code: gofmt with `-s` option, true by default
    simplify: true
  goimports:
    # put imports beginning with prefix after 3rd-party packages;
    # it's a comma-separated list of prefixes
    local-prefixes: github.com/org/project
  golint:
    min-confidence: 0.8
  govet:
    # report about shadowed variables
    check-shadowing: true
    settings:
      printf:
        funcs:
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Infof
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Warnf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Errorf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
  lll:
    line-length: 120
  maligned:
    suggest-new: true
  misspell:
[...]

Issue #994, opened 2020-03-15
Increased memory usage in 1.24.0 causing out-of-memory


invidian:
It seems that after upgrading `golangci-lint` for my project to 1.24.0, Travis CI is no longer able to lint the project, as it fails with `out-of-memory` errors: https://travis-ci.org/github/flexkube/libflexkube/jobs/662847029#L1086

After reverting the same PR to 1.23.8, linting passes: https://travis-ci.org/github/flexkube/libflexkube/jobs/662848310?utm_medium=notification&utm_source=github_status.

PR affected by the issue: https://github.com/flexkube/libflexkube/pull/82

Offending patch (now reverted):
```diff
diff --git Makefile Makefile
index bb0b33c..caa47b4 100644
--- Makefile
+++ Makefile
@@ -11,7 +11,7 @@ GORUN=$(GOCMD) run
 GOBUILD=CGO_ENABLED=$(CGO_ENABLED) $(GOCMD) build -v -buildmode=exe -ldflags $(LD_FLAGS)
 
 CC_TEST_REPORTER_ID=6e107e510c5479f40b0ce9166a254f3f1ee0bc547b3e48281bada1a5a32bb56d
-GOLANGCI_LINT_VERSION=v1.23.8
+GOLANGCI_LINT_VERSION=v1.24.0
 BIN_PATH=$$HOME/bin
 
 GO_PACKAGES=./...
```

Thank you for creating the issue!

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
golangci-lint has version 1.24.0 built from 6fd4383 on 2020-03-15T11:38:02Z
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
cat: .golangci.yml: No such file or directory```

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
go version go1.14 linux/amd64
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/invidian/.cache/go-build"
GOENV="/home/invidian/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
[...]

PR #987, opened 2020-03-10
update staticcheck to v2020.1.3

thaJeztah:
full diff: https://github.com/dominikh/go-tools/compare/2019.2.3...2020.1.3

PR #1063, opened 2020-05-03
reduce mem. usage of unused and update staticcheck

jirfag:
The primary improvement is in early clearing of
analyzed package's TypeInfo, facts, etc for
whole program analyzers (`unused`). Clear it when it
becomes unused and GC collects them early. Initially this
clearing was performed for all analyzers except `unused`.

Update staticcheck from v0.0.1-2019.2.3 to v0.0.1-2020.1.4

Also in this commit:
  * speed up loading packages from export data (2.5s -> 2.1s for std)
    by not using mutex for export data since it was allowed in
    x/tools#07722704da13
  * make an order of execution of linters stable
  * update renameio and robustio
  * use robustio in caching

Relates: #987, #994, #995, #1011


internal/renameio/renameio_test.go
```diff
@@ -130,10 +128,18 @@ func TestConcurrentReadsAndWrites(t *testing.T) {
...
+
+	switch runtime.GOOS {
+	case "windows":
 		// Windows produces frequent "Access is denied" errors under heavy rename load.
-		// As long as those are the only errors and *some* of the writes succeed, we're happy.
+		// As long as those are the only errors and *some* of the reads succeed, we're happy.
 		minReadSuccesses = attempts / 4
+
+	case "darwin":
+		// The filesystem on macOS 10.14 occasionally fails with "no such file or
```
thaJeztah, line 139: 😞 why can't we have nice things?

I like the "robustio" package though (interesting)

(Also, sorry for the noise; just glancing over the code changes, and learned about this macOS bug)

jirfag: Yep.
Also, it's copy-pasted code from [Go source](https://github.com/golang/go/tree/master/src/cmd/go/internal/renameio)

PR #1067, opened 2020-05-04
Fix godot false positive for cgo export comments

tetafro:
One more cgo-related false positive for godot linter: https://github.com/tetafro/godot/issues/3

Also run `go mod tidy`


jirfag:
thank you

PR #1068, opened 2020-05-04
Fix CI

tetafro:
Add unshallow to get tags in CI for `check_generated` test.

See example of failed test here: https://github.com/golangci/golangci-lint/pull/1067/checks?check_run_id=642619832


tetafro:
But I have no idea what's wrong with macos tests.

PR #1073, opened 2020-05-07
dev: generate assets/github-action-config.json

jirfag:
It will be used by GitHub action `golangci-lint-action`.

Relates: golangci/golangci-lint-action#11


scripts/gen_github_action_config/main.go
```diff
@@ -0,0 +1,231 @@
...
+	if err = enc.Encode(cfg); err != nil {
+		return fmt.Errorf("failed to json encode config: %w", err)
+	}
+
+	return nil
+}
+
+type logInfo struct {
+	Warning string `json:",omitempty"`
+	Info    string `json:"omitempty"`
```
ernado, line 53: Seems like a typo, probably `,omitempty` is expected here

scripts/gen_github_action_config/main.go
```diff
@@ -0,0 +1,231 @@
...
+	}
+
+	if len(os.Args) != 2 { //nolint:gomnd
+		return fmt.Errorf("usage: go run .../main.go out-path.json")
+	}
+	outFile, err := os.Create(os.Args[1])
+	if err != nil {
+		return fmt.Errorf("failed to create output config file: %w", err)
+	}
+	defer outFile.Close()
```
ernado, line 41: Probably we still want to check this error.

jirfag: it's just a script, we can visually validate if it's output is corrupted.

scripts/gen_github_action_config/main.go
jirfag: you are right, thank you, we need a linter for that :)

PR #1077, opened 2020-05-08
Fix nil pointer in codeclimate.go

elliots:
Some issues don't have any source lines


jirfag:
hi,
thank you,
what do you think about #1078 fix?

elliots:
Seems like both could be accepted without a problem?

PR #1079, opened 2020-05-08
Update gosec2 dependency to a tagged version (v2.3.0)

theckman:
When raising #1062 I noted that the gosec dependency had been upgraded to v2,
but needed to be pulled from `master` due to some issues with the v2.2.0 tagged
release. That was tracked in https://github.com/securego/gosec/issues/470.

They've since cut v2.3.0, so let's update to that tagged release before we
release a new version of golangci-lint.


theckman:
@jirfag I assume we'll want this before v1.27.0.

jirfag:
thank you

Issue #1082, opened 2020-05-12
Godot removes comments not ending with dot when fix option is used


doriandekoning:
Thank you for creating the issue!

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint version                                                                                                                                                                                                                                                                                 
golangci-lint has version 1.26.0 built from 6bd10d0 on 2020-05-01T15:26:22Z
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
run:
  concurrency: 3
  timeout: 15m
  issues-exit-code: 1
  tests: true

  skip-dirs:
    - design
    - \.go
  skip-files:
    #TODO add generated models files
    # - ".*\\.my\\.go$

  modules-download-mode: readonly


# output configuration options
output:
  # colored-line-number|line-number|json|tab|checkstyle|code-climate, default is "colored-line-number"
  format: code-climate

  # print lines of code with issue, default is true
  print-issued-lines: true

  # print linter name in the end of issue text, default is true
  print-linter-name: true


# all available settings of specific linters
linters-settings:
  errcheck:
    check-type-assertions: false

    check-blank: true


    # path to a file containing a list of functions to exclude from checking
    # see https://github.com/kisielk/errcheck#excluding-functions for details
    # exclude: /path/to/file.txt
  govet:
    # report about shadowed variables
    check-shadowing: false

  golint:
    min-confidence: 0.8

  gofmt:
    simplify: true
  gocyclo:
    min-complexity: 18
  maligned:
    suggest-new: true
  dupl:
    threshold: 175
  goconst:
    min-len: 3
    min-occurrences: 3
  funlen:
    lines: 80
    statements: 50
  godox:
    keywords:
      - HACK
      - OPTIMIZE
      - TODO
      - BUG
  misspell:
    locale: US
    ignore-words:
  lll:
    line-length: 120
    tab-width: 2
  unused:
    check-exported: false

  unparam:
    check-exported: false

  nakedret:
    max-func-lines: 0
  prealloc:
    simple: true
    range-loops: true # Report preallocation suggestions on range loops, true by default
    for-loops: false # Report preallocation suggestions on for loops, false by default

  gocritic:
    enabled-checks:

    disabled-checks:
[...]

PR #1084, opened 2020-05-13
godot: fix issue with `--fix` flag

tetafro:
Fix https://github.com/golangci/golangci-lint/issues/1082


ernado:
Thank you!

The `Replacement` field is slightly confusing and seemed like there is no validation that "can fix" option is in sync with fixing behavior.

PR #1094, opened 2020-05-16
docs: migrate README.md to a website

jirfag:
Make a website https://golangci-lint.run
for golangci-lint documentation.
The website is backed by Gatsby static site generator
and deployed into Netlify.


jirfag:
we need to exclude fossa from checking `docs/`, I can't do it right now because it doesn't work now.

Issue #1087, opened 2020-05-13
Cannot install - gomodguard fails


dmlyons:
Thank you for creating the issue!

- [ x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
golangci-lint has version v1.26.0 built from (unknown, mod sum: "h1:CLLGRSA9BLMiNvsWPXHioYAdfIx9tkgdVWyA6bIdYCo=") on (unknown)
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
no config file
```

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
09:43 $ go version && go env
go version go1.14.2 linux/amd64
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/dl/.cache/go-build"
GOENV="/home/dl/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/dl/go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build381143871=/tmp/go-build -gno-record-gcc-switches"
```

</details>

<details><summary>Verbose output of running</summary>

```console
09:43 $ GO111MODULE=on go get -u github.com/golangci/golangci-lint/cmd/golangci-lint
[...]

PR #1140, opened 2020-05-17
bump gomodguard version

ryancurrah:
adds a new feature to allow blocking modules based on version constraints

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -221,6 +221,10 @@ type LintersSettings struct {
 				Recommendations []string `mapstructure:"recommendations"`
 				Reason          string   `mapstructure:"reason"`
 			} `mapstructure:"modules"`
+			Versions []map[string]struct {
+				Version string `mapstructure:"version"`
+				Reason  string `mapstructure:"reason"`
+			} `mapstructure:"versions"`
 		} `mapstructure:"blocked"`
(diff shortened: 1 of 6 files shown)


pkg/config/config.go
```diff
@@ -221,6 +221,10 @@ type LintersSettings struct {
 				Recommendations []string `mapstructure:"recommendations"`
 				Reason          string   `mapstructure:"reason"`
 			} `mapstructure:"modules"`
+			Versions []map[string]struct {
```
jirfag, line 224: please, add new params to docs - currently, it's `.golangci.example.yml`

jirfag: Also, could you please document somewhere (e.g. in description of a linter) what are differences from [depguard](https://github.com/OpenPeeDeeP/depguard) linter?

jirfag:
I've disabled mac os tests: they are too slow and result in frequent timeouts.

pkg/config/config.go
ryancurrah: Done, also added a reminder for contributors to this in the docs

jirfag: thank you!

ryancurrah:
Resolves https://github.com/golangci/golangci-lint/issues/1087

PR #1150, opened 2020-05-18
Add autofix for godot linter

tetafro:
Now a period can be added automatically.


pkg/lint/lintersdb/manager.go
```diff
@@ -263,6 +263,7 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithURL("https://github.com/ryancurrah/gomodguard"),
 		linter.NewConfig(golinters.NewGodot()).
 			WithPresets(linter.PresetStyle).
+			WithAutoFix().
```
jirfag, line 266: could you add at least one test case to `test/testdata/fix/`, please, to ensure that integration works ok?

tetafro: Missed it, thanks. There is a test now.

Issue #825, opened 2019-10-16
context loading failed: no go files to analyze


rmilejcz:
I have regularly run `golangci-lint run` from the root of my directory without issue for the last several weeks. Today, after successfully running `golangci-lint run`, I fixed all the reported issues and then ran the command in the same terminal but I got the output:

```
ERRO Running error: context loading failed: no go files to analyze
```
even though nothing seems to have changed. The project structure is basically 60 or so folders, each having a `main.go`, a `main_test.go` and two auto generated files which are ignored by `golangci-lint`. If I specify a package it lints without issue e.g.:

```
golangci-lint run ActivityLog
```
runs the linter on that package without any error.

<details><summary>Version of golangci-lint</summary>

```bash
$ golangci-lint --version
golangci-lint has version 1.21.0 built from 645e794 on 2019-10-15T18:15:04Z
```

</details>

<details><summary>Config file</summary>

```bash
$ cat .golangci.yml
run:
  concurrency: 4
  deadline: 2m
  issues-exit-code: 1
  tests: true

output:
  format: colored-line-number
  print-issued-lines: true
  print-linter-name: true

linter-settings:
  errcheck:
    check-type-assertions: true
    check-blank: true

  govet:
    check-shadowing: true
    settings:
      printf:
        funcs:
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Infof
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Warnf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Errorf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
  golint:
    min-confidence: 0.8

  gofmt:
    simplify: true

  goimports:
    local-prefixes: kalos-core

  gocyclo:
    min-complexity: 20

  maligned:
    suggest-new: true

  goconst:
    min-len: 3
    min-occurrences: 3

  misspell:
    locale: US
    ignore-words:
      - supercalifragilisticexpialidocious

  lll:
    line-length: 120
    tab-width: 2

  unused:
    check-exported: false

  unparam:
    check-exported: false

  prealloc:
    simple: true
    range-loops: true
    for-loops: false

  gocritic:
    enabled-checks:
      - appendAssign
      - argOrder
      - assignOp
      - badCond
      - boolExprSimplify
      - builtinShadow
      - captLocal
      - caseOrder
      - commendOutCode
      - commentFormatting
      - commentedOutImport
[...]

PR #1154, opened 2020-05-18
fix invalid error message "no go files to analyze"

jirfag:
In case of timeouts of go/packages loading
we could return such error.

Relates: #825

### DIFF
--- a/pkg/lint/load.go
+++ b/pkg/lint/load.go
@@ -199,7 +199,14 @@ func (cl *ContextLoader) loadPackages(ctx context.Context, loadMode packages.Loa
 	cl.debugf("Built loader args are %s", args)
 	pkgs, err := packages.Load(conf, args...)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to load program with go/packages")
+		return nil, errors.Wrap(err, "failed to load with go/packages")
+	}
+
+	// Currently, go/packages doesn't guarantee that error will be returned
+	// if context was canceled. See
+	// https://github.com/golang/tools/commit/c5cec6710e927457c3c29d6c156415e8539a5111#r39261855
+	if ctx.Err() != nil {
+		return nil, errors.Wrap(ctx.Err(), "timed out to load packages")
 	}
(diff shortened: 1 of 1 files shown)


pkg/lint/load.go
```diff
@@ -199,7 +199,14 @@ func (cl *ContextLoader) loadPackages(ctx context.Context, loadMode packages.Loa
...
 	pkgs, err := packages.Load(conf, args...)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to load program with go/packages")
+		return nil, errors.Wrap(err, "failed to load with go/packages")
+	}
+
+	// Currently, go/packages doesn't guarantee that error will be returned
+	// if context was cancelled. See
+	// https://github.com/golang/tools/commit/c5cec6710e927457c3c29d6c156415e8539a5111#r39261855
+	if ctx.Err() != nil {
```
iwankgb, line 208: Perhaps:
```golang
if err := ctx.Err(); err != nil {
```
would do? It would allow to avoid calling `ctx.Err()` twice.

pkg/lint/load.go
```diff
@@ -199,7 +199,14 @@ func (cl *ContextLoader) loadPackages(ctx context.Context, loadMode packages.Loa
...
+		return nil, errors.Wrap(ctx.Err(), "timeouted to load packages")
```
ryancurrah, line 209: `timeout` is mispelled

jirfag: thank you, fixed

pkg/lint/load.go
jirfag: I don't see any profit to save nanoseconds here :)

jirfag: upate: maybe microseconds

```
func (c *cancelCtx) Err() error {
	c.mu.Lock()
	err := c.err
	c.mu.Unlock()
	return err
}
```

It doesn't make sense to waste time on such micro-optimization, there are a lot of things to save seconds of time

PR #1155, opened 2020-05-18
add ability to set issue severity

ryancurrah:
**User facing changes**:
- added new `severity` config to allow for custom issue severity for out formats that support it based on severity rules. Case sensitive severity rules are supported as well
- added `severity` support to `checkstyle`, `code-climate` and `github` out formats
- pretty print `checkstyle` out format so humans can read it as well

**Internal changes**:
- added a new field to `Issue` struct named `Severity`. This allows setting severity for issues
- added a new method to `Issue` struct named `Description`. This allows reusing logic to create a issue description for printers. It's a combination of linter name and issue text
- added a new method to `Issue` struct named `Fingerprint`. This allows reusing logic to create a issue md5 hash for printers. The fingerprint is made up of issue filename, text and if available the first line in the file
- a new `BaseRule` struct has been added that allows us to extend the rule code beyond just using it for `ExcludeRules`
- `BaseRule` methods have 100% test coverage
- made `NewRunner` method easier to read by moving some logic out of it for creating new processors
- made `codeclimate` code easier to read


ryancurrah:
@jirfag or @ernado what are you thoughts on this feature? do you think it is a feasible way to solve this?

ernado:
I have no objections, design looks good to me.

ryancurrah:
One thought I had where this could get complicated is if multiple out formats are supported.  It would be difficult to change this to accommodate multiple out formats, atleast not without breaking changes.

ernado:
cc: @jirfag

We can avoid breaking changes via keeping default behavior, so format will change only with explicit configuration.

So I'm pretty positive for this feature.

ryancurrah:
@ernado sounds fair. I will add some tests to this PR

ryancurrah:
Tests have been added

pkg/config/config.go
```diff
@@ -452,13 +452,55 @@ func (e ExcludeRule) Validate() error {
...
 
+type SeverityRule struct {
+	Severity string
+	Linters  []string
+	Path     string
+	Text     string
+	Source   string
+}
+
+func (s *SeverityRule) Validate() error { // nolint:dupl
```
jirfag, line 463: maybe we can generalize it to avoid copy-paste?

.golangci.example.yml
```diff
@@ -382,3 +382,27 @@ issues:
...
+
+  # Default value is empty string.
+  # Set the default severity for issues. If severity rules are defined and the issues 
+  # do not match or no severity is provided to the rule this will be the default 
+  # severity applied. Severities should match the supported severity names of the 
+  # selected out format.
+  # - Code climate: https://docs.codeclimate.com/docs/issues#issue-severity
+  # -   Checkstyle: https://checkstyle.sourceforge.io/property_types.html#severity
+  # -       Github: https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message
+  severity-default: error
```
jirfag, line 394: I think it should be in the nested object similar to:

```yaml
severity:
  default-severity: error
  case-sensitive: false
  rules:
    - ...
```

pkg/lint/runner.go
```diff
@@ -29,6 +29,7 @@ type Runner struct {
 	Log        logutils.Log
 }
 
+//nolint:funlen
```
jirfag, line 32: the function is really too long, why not split it?

pkg/config/config.go
ryancurrah: validate logic was deduplicated

.golangci.example.yml
ryancurrah: Severity was moved to it's own parent key

pkg/lint/runner.go
ryancurrah: Moved some of the logic for creating new processors out of this function

ryancurrah:
@jirfag thanks for the review. Updated based on feedback and added more detail to the PR description

emas80:
hi, any plans for releasing this?

PR #1160, opened 2020-05-22
dev: fix Windows asset name

### LINKED ISSUE #1157: Change in release-asset names breaks automation

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Between 1.26.0 and 1.27.0 the name of the Windows release assets has changed from:
```
golangci-lint-1.26.0-windows-amd64.zip
```
to
```
golangci-lint.exe-1.27.0-windows-amd64.zip
```

This change in format is a pain to manage in automated install scripts as it deviates from what existed before and from the format used for Linux and OSX assets.

It appears that the asset names are generated via `goreleaser` so this might be due to some change in that tool?

It would be very much appreciated if this change was reverted.

ernado:
Fix #1157


jirfag:
hm, it looks like this bot checked that you aren't in a `team` (because you are in the `core team`).

Issue #1083, opened 2020-05-13
Improve generated files detection


rayjlinden:
golangci-lint is not ignoring auto generated files like it is supposed to.  The autogenerated files are made by open-api generator.  Here is a part of the top of one of the generated files:

```
/*
 * webhooksio client
 *
 * Hack because Ray is too lazy to write a client by hand.
 *
 * API version: 1.0.0
 */

// Code generated by OpenAPI Generator (https://openapi-generator.tech); DO NOT EDIT.
```

Thank you for creating the issue!

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
# golangci-lint has version 1.26.0 built from 6bd10d0 on 2020-05-01T15:33:57Z
```

</details>

<details><summary>Config file</summary>

```console
$ golangci-lint run --fix --no-config --disable-all -E gofmt -E govet -E errcheck -E staticcheck -E unused -E gosimple -E structcheck -E varcheck -E ineffassign -E deadcode -E typecheck -E gosec ./...
```

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
go version go1.13 linux/amd64
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/rayj/.cache/go-build"
GOENV="/home/rayj/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GONOPROXY="github.com/lindenlab/*"
GONOSUMDB="github.com/lindenlab/*"
GOOS="linux"
GOPATH="/var/tmp/rayj/gocode"
GOPRIVATE="github.com/lindenlab/*"
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/home/rayj/.gvm/gos/go1.13"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/home/rayj/.gvm/gos/go1.13/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/var/tmp/rayj/gocode/src/github.com/lindenlab/ecom-webhooks/go.mod"
[...]

PR #1161, opened 2020-05-23
handle some block comment to detect generated files

ksoichiro:
I changed autogenerated_exclude.go to handle block comments partially to detect generated comment inside or after the block comments.

I avoided to change to read whole file, because it might affect to performance.
Although the generated comment text may appear anywhere in the file, this change still only see around the top of the file.
It will quit reading when the line is other than comments, package statement, and empty.

I'm not sure it can be accepted because it's not a perfect solution, but I think it will be improved than it is now.

Related: #1083


pkg/result/processors/autogenerated_exclude.go
```diff
@@ -130,11 +130,32 @@ func getDoc(filePath string) (string, error) {
 	scanner.Buffer(make([]byte, initialSize), maxSize)
 
 	var docLines []string
+	inBlockComment := false
 	for scanner.Scan() {
 		line := strings.TrimSpace(scanner.Text())
-		if strings.HasPrefix(line, "//") {
+		if inBlockComment {
```
jirfag, line 136: sorry, I don't think it's a good idea to manually parse comments.
Maybe we can use [ast.ParseFile](https://golang.org/pkg/go/parser/#ParseFile) with flag like `PackageClauseOnly` or similar to stop early? I'm sure that performance won't be an issue here if we implement it properly.

ksoichiro: Thank you for your advice.
I was actually on the fence about using AST, but I'll try to do that.

ksoichiro: I've changed to stop parsing manually by using `parser.ParseFile()`.
Although `parser.ParseFile()` reads entire file regardless of mode, with `PackageClauseOnly` mode (and `ParseComments` mode), it stops parsing files after import statements, as you pointed out.

pkg/result/processors/autogenerated_exclude_test.go
```diff
@@ -92,3 +103,11 @@ and this text also`,
 		assert.Equal(t, tc.doc, doc)
 	}
 }
+
+// Issue 954: Some lines can be very long, e.g. auto-generated
+// embedded resources. Reported on file of 86.2KB.
+func TestGetDocFileWithLongLine(t *testing.T) {
```
ksoichiro, line 109: This change replaces the whole implementation of `getDoc()`, so I also added a test for confirming previous issue (#954).

pkg/result/processors/autogenerated_exclude.go
ernado: > The text may appear anywhere in the file.

https://github.com/golang/go/issues/13560#issuecomment-288457920

Are we sure that avoiding possible performance regression has higher priority that compliance with convention?

jirfag: I agree that it looks more important. I guess time still will be <100ms total. We can easily check it: `golangci-lint run -v` shows all processors' times.

ksoichiro: Do you mean compliance with convention is more important?

Suppose you think so.. performance regression depends on how many issues will be reported on generated files, so I think it's difficult to choose which repository to measure for this purpose. But as you explain in [usage](https://golangci-lint.run/usage/performance/), parsing file seems not so much cost.

> Parsing one source file takes 200 us on average. Parsing of all files in $GOROOT/src takes 2 seconds. Currently we parse each file more than once because it's not the bottleneck.

However, you also write like this:

> We're planning to parse each file only once.

If we check entire file for comments, should we use existing AST instead of parsing file here again?
I haven't confirmed if it's possible yet, I guess we can do so since this function is called on reporting phase.

Issue #925, opened 2020-01-16
Frequent "failed to get data from low-level cache" warning


oschwald:
I have been seeing the following warning quite frequently in both our CI environment and on my local development machine:

```
level=warning msg="[runner/goanalysis_metalinter/goanalysis] Failed to get persisted facts: failed to get data from low-level cache by key printf/facts for package trace: failed to stat file /home/greg/.cache/golangci-lint/af/aff20f6cd29342c0f0dcc2ea1acf50b374a6cf1954a9287664287e2d313d7244-d: stat /home/greg/.cache/golangci-lint/af/aff20f6cd29342c0f0dcc2ea1acf50b374a6cf1954a9287664287e2d313d7244-d: no such file or directory"
```

There isn't anything else running that should be deleting these files and we aren't running multiple instances of `golangci-lint` concurrently.

This has been going on for a month or so, likely starting around the time I upgraded to 1.22.0.

Please let me know if I can supply any additional information.

---

Thank you for creating the issue!

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
golangci-lint has version 1.22.2 built from cb2f8ba on 2019-12-30T19:26:28Z
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.tonl
[run]
  deadline = "10m"

  modules-download-mode = "vendor"

  skip-files = [
    "_easyjson\\.go$",
    "_easyjson_test\\.go$",
  ]

  tests = true

[linters]
  disable-all = true
  enable = [
    "deadcode",
    "depguard",
    "errcheck",
    "goconst",
    "gocyclo",
    "gocritic",
    "gofmt",
    "golint",
    "gosec",
    "gosimple",
    "ineffassign",
    "maligned",
    "misspell",
    "nakedret",
    # https://github.com/golangci/golangci-lint/issues/287
    # "safesql",
    "staticcheck",
    "structcheck",
    "typecheck",
    "unconvert",
    "unparam",
    "varcheck",
    "vet",
    "vetshadow",
  ]

[linters-settings.depguard]
  list-type = "blacklist"
  include-go-root = true
  packages = [
    # This library panics frequently on invalid input. Please use
    # github.com/pelletier/go-toml instead.
[...]

PR #1162, opened 2020-05-23
cache: fix warning

jirfag:
Fix warning about not existing cache file.

Fixes: #925

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -81,7 +81,7 @@ func (c *Cache) fileName(id [HashSize]byte, key string) string {
 var errMissing = errors.New("cache entry not found")
 
 func IsErrMissing(err error) bool {
-	return err == errMissing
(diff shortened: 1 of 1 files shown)


internal/cache/cache.go
```diff
@@ -81,7 +81,7 @@ func (c *Cache) fileName(id [HashSize]byte, key string) string {
 var errMissing = errors.New("cache entry not found")
 
 func IsErrMissing(err error) bool {
-	return err == errMissing
+	return errors.Cause(err) == errMissing
```
mcristina422, line 84: Shouldnt this be https://godoc.org/github.com/pkg/errors#Is

jirfag: thank you, you are right. We will fix when will be replacing `pkg/errors` by `errors`. I guess it can happen with go1.15.

Issue #1041, opened 2020-04-23
update scopelint with looppointer or exportloopref


agnivade:
Hi there,

I am totally aware that the project is going through some rough times. So there's no pressure to work on this right now. Nevertheless, I wanted to file this to just to keep a record.

The author of scopelint has deprecated it for other better linters: https://github.com/kyoh86/scopelint#obsoleted.

Let's use either of them to remain up to date.

Thanks for all the work.

ernado:
Should we just add both `looppointer` and `exportloopref`?

agnivade:
I think so. Let the user choose what they want. And document it adequately mentioning which is to be used when.

tamalsaha:
```
If you want to find lints as nervous as possible (with some false-positives), you should use looppointer.

If you want to find lints as accurately as possible (with some lints ignored), you should use exportloopref.
```

If you are going to include only one, please include `exportloopref`

sayboras:
Agreed with comment from @tamalsaha above, maybe _exportloopref_ is good enough :)

kyoh86:
`looppointer` can be an option.
Maybe this is not so important, but I'd like to mention it.

Since `exportloopref` cannot find some diagnostics like below.

Case 1: pass the pointer to the function to export.

Case 2: pass the pointer to the local variable, and export it.

```go
package main

type List []*int

func (l *List) AppendP(p *int) {
	*l = append(*l, p)
}

func main() {
	var slice []*int
	list := List{}

	println("loop expect exporting 10, 11, 12, 13")
	for _, v := range []int{10, 11, 12, 13} {
		list.AppendP(&v) // Case 1: wanted "exporting a pointer for the loop variable v", but cannot be found

		p := &v                  // p is the local variable
		slice = append(slice, p) // Case 2: wanted "exporting a pointer for the loop variable v", but cannot be found
	}

	println(`slice expecting "10, 11, 12, 13" but "13, 13, 13, 13"`)
	for _, p := range slice {
		printp(p)
	}
	println(`array expecting "10, 11, 12, 13" but "13, 13, 13, 13"`)
	for _, p := range ([]*int)(list) {
		printp(p)
	}
}

func printp(p *int) {
	println(*p)
}
```

So we may use `looppointer` instead of `exportloopref` if we want to find them all (with some false-positives).

https://github.com/kyoh86/exportloopref#known-false-negatives

kyoh86:
[...]

PR #1163, opened 2020-05-23
Add exportloopref linter

sayboras:
Fixes https://github.com/golangci/golangci-lint/issues/1041


pkg/lint/lintersdb/manager.go
```diff
@@ -272,6 +272,9 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 		linter.NewConfig(golinters.NewNestif()).
 			WithPresets(linter.PresetComplexity).
 			WithURL("https://github.com/nakabonne/nestif"),
+		linter.NewConfig(golinters.NewExportLoopRef()).
+			WithPresets(linter.PresetStyle).
```
jirfag, line 276: isn't it `bugs` preset?

pkg/lint/lintersdb/manager.go
```diff
@@ -272,6 +272,9 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
...
+			WithURL("github.com/kyoh86/exportloopref"),
```
ernado, line 277: Please add `https://`

sayboras: Thanks :100:

pkg/lint/lintersdb/manager.go
sayboras: you are right, bugs present is more appropriate :100:

kyoh86:
`looppointer` can be an option.
Maybe this is not so important, but I'd like to mention it.

Since `exportloopref` cannot find some diagnostics like below.

Case 1: pass the pointer to the function to export.

Case 2: pass the pointer to the local variable, and export it.

```go
package main

type List []*int

func (l *List) AppendP(p *int) {
	*l = append(*l, p)
}

func main() {
	var slice []*int
	list := List{}

	println("loop expect exporting 10, 11, 12, 13")
	for _, v := range []int{10, 11, 12, 13} {
		list.AppendP(&v) // Case 1: wanted "exporting a pointer for the loop variable v", but cannot be found

		p := &v                  // p is the local variable
		slice = append(slice, p) // Case 2: wanted "exporting a pointer for the loop variable v", but cannot be found
	}

	println(`slice expecting "10, 11, 12, 13" but "13, 13, 13, 13"`)
	for _, p := range slice {
		printp(p)
	}
	println(`array expecting "10, 11, 12, 13" but "13, 13, 13, 13"`)
	for _, p := range ([]*int)(list) {
		printp(p)
	}
}

func printp(p *int) {
	println(*p)
}
```

So we may use `looppointer` instead of `exportloopref` if we want to find them all (with some false-positives).

https://github.com/kyoh86/exportloopref#known-false-negatives

PR #1166, opened 2020-05-25
add exhaustive linter

nishanths:
This adds a linter for the exhaustive static analysis program found at https://github.com/nishanths/exhaustive. The program checks for exhaustiveness of enum switch statements. 

I've disabled the linter in `.golangci.yml`. (I'm happy to change this behavior if necessary.)

Thanks for taking a look.


test/testdata/exhaustive/exhaustive.go
```diff
@@ -0,0 +1,10 @@
+package exhaustive
```
nishanths, line 1: I used a separate package here to check that the linter is integrates correctly for cross-package checks (ie., switch statement and enum definitions in different packages).

pkg/golinters/goanalysis/runner.go
```diff
@@ -935,7 +935,8 @@ func sizeOfReflectValueTreeBytes(rv reflect.Value, visitedPtrs map[uintptr]struc
 		return rv.Len()
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64,
 		reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64,
-		reflect.Uintptr, reflect.Bool, reflect.Float32, reflect.Float64, reflect.UnsafePointer:
+		reflect.Uintptr, reflect.Bool, reflect.Float32, reflect.Float64,
+		reflect.Complex64, reflect.Complex128, reflect.Func, reflect.UnsafePointer:
```
nishanths, line 939: These missing cases
```
reflect.Complex64, reflect.Complex128, reflect.Func
```
were reported by the exhaustive linter when I temporarily enabled it for a run.

test/testdata/exhaustive/exhaustive.go
nishanths: This is no longer true. I've reverted to using the same file because importing the testdata package did not seem to work.

pkg/lint/lintersdb/manager.go
```diff
@@ -275,6 +277,9 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 		linter.NewConfig(golinters.NewExportLoopRef()).
 			WithPresets(linter.PresetBugs).
 			WithURL("https://github.com/kyoh86/exportloopref"),
+		linter.NewConfig(golinters.NewExhaustive(exhaustiveCfg)).
+			WithPresets(linter.PresetBugs).
+			WithURL("https://github.com/nishanths/exhaustive"),
```
jirfag, line 282: add `WithLoadForGoAnalysis()` please - otherwise if only this linter is enabled it won't work

jirfag: @all also, we should automate this check: it's really very common fault.

nishanths: Thanks for explaining, I've added it.

.golangci.yml
```diff
@@ -104,6 +106,7 @@ linters:
 
   # don't enable:
   # - asciicheck
+  # - exhaustive
```
jirfag, line 109: why not enable it?

nishanths: Sounds good, I can enable it.

I did not enable it originally because I was unsure if users would find the default behavior of the analyzer to be too strict. (The default behavior of the analyzer is to point out missing enum members in the switch even if a `default` case is present.)

nishanths:
It looks like a test is failing because a prebuilt binary is being used in the test that doesn't know about the `exhaustive` analyzer? Any suggestion appreciated.

test/testdata/exhaustive.go
```diff
@@ -0,0 +1,17 @@
+//args: -Eexhaustive
```
mcristina422, line 1: Should another test case including enabling `DefaultSignifiesExhaustive` be added?

nishanths: Yes, that would help. I'll add one.

test/testdata/exhaustive_default.go
```diff
@@ -0,0 +1,25 @@
...
+
+const (
+	North Direction = iota
+	East
+	South
+	West
+)
+
+func processDirection(d Direction) {
+	switch d { // ERROR "missing cases in switch of type Direction: East, West"
```
nishanths, line 15: Needed a testcase in this file that writes _something_ to stderr. Otherwise these lines in `runGoErrchk()` in linters_test.go fail with the following error:

```
An error is expected but got nil.
```

```go
func runGoErrchk(c *exec.Cmd, files []string, t *testing.T) {
	output, err := c.CombinedOutput()
	assert.Error(t, err)
	...
}
```

mcristina422:
@nishanths The failing test is because it's only running with the last released version. Maybe it should only be added as a default after it is released? 

https://github.com/golangci/golangci-lint/blob/master/.github/workflows/pr.yml#L21

test/testdata/exhaustive_default.go
nishanths: I adjusted `runGoErrchk()` to account for this instead of having this extra "dummy" test case.

nishanths:
> @nishanths The failing test is because it's only running with the last released version. Maybe it should only be added as a default after it is released?

Thank you. I've made it disabled by default. I can try to revisit after the next release. cc @jirfag

nishanths:
> @nishanths The failing test is because it's only running with the last released version. Maybe it should only be added as a default after it is released?

On second thought, this appears to be a bottleneck to dogfooding new linters. What is the reason the CI tests use a released version instead of `HEAD`? I'm happy to help improve this process in any way.

ernado:
> What is the reason the CI tests use a released version instead of HEAD?

It is problem of chicken and egg, because we are using github action, which is using a released version.

bombsimon:
Isn't it possible to just disable the official action and rely on `make test` which will build and run `golangci-lint` from `HEAD` in the PR?

nishanths:
@bombsimon I did a little more digging. `pr.yml` has a comment mentioning that testing the latest stable release is intentional, in addition to testing using `make test`.
```
 # We already run the current golangci-lint in tests, but here we test
 # our GitHub action with the latest stable golangci-lint.
```
https://github.com/golangci/golangci-lint/blob/master/.github/workflows/pr.yml#L11-L12

But I wonder what the reason is for testing new PRs on an already released version?

bombsimon:
@nishanths Oh, nice findings. I thought about something similar but came to the conclusion that it wasn't needed. I guess I was wrong. 😸 

I kind of see the reason, to ensure that any bugs in the action will be detected as soon as possible, but it puts this project in a weird state with sometimes hackish workarounds to be able to release changes.

Maybe we could use different configurations for the action and the project itself. I guess it will be a mess to maintain but it might be better than the alternatives. I had a similar issue in https://github.com/golangci/golangci-lint/pull/1148 after a linter introduced a false positive in a minor release.

ernado:
Can we merge this?

nishanths:
Yes, from me for merging. :) I feel the rest of the discussion is separate from this PR. Maybe we track it in a separate issue if necessary.

ernado:
Great!

PR #1167, opened 2020-05-25
update new-linters doc

nishanths:
This updates the new-linters doc. I've fixed a filename (`lintersdb.go` -> `manager.go`), and remove outdated line numbers from the links. 

If preferred, I can add correct line numbers using a GitHub link at a fixed commit instead of master — to avoid the line numbers being outdated again.


jirfag:
thank you!

PR #1171, opened 2020-05-30
Update not found url in new-linters.mdx

momotaro98:
I just encountered not found page after I clicked the URL link.

I updated to correct URL link by following other URLs in this page.


ryancurrah:
Thanks the URL you corrected is working.

PR #1175, opened 2020-06-03
Replace '{.LatestVersion}' with 'latest' as install.sh parameter

chs2:
install.sh expected `latest` as a parameter instead of `{.LatestVersion}`

    # wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s {.LatestVersion}
    2020-06-03 10:31:26 URL:https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh [10739/10739] -> "-" [1]
    golangci/golangci-lint info checking GitHub for tag '{.LatestVersion}'
    golangci/golangci-lint crit unable to find '{.LatestVersion}' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details

### DIFF
--- a/docs/src/docs/usage/install/index.mdx
+++ b/docs/src/docs/usage/install/index.mdx
@@ -28,13 +28,13 @@ Here is the recommended way to install golangci-lint {.LatestVersion}:
 
 ```sh
 # binary will be $(go env GOPATH)/bin/golangci-lint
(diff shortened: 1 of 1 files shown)


ernado:
Pinning specific version is intended.

Issue #490, opened 2019-04-10
Add gofumpt


andrewrynhard:
Include https://github.com/mvdan/gofumpt in set of linters.

jirfag:
Hi, nice tool, thank you!
Pull requests [are welcome](https://github.com/golangci/golangci-lint/wiki/How-to-add-a-custom-linter)

dackroyd:
Given `gofumpt`/`gofumports` does everything that `gofmt` / `goimports` would do, what is the preference for applying this?

> Enforce a stricter format than `gofmt`, while being backwards compatible. That is, `gofumpt` is happy with a subset of the formats that `gofmt` is happy with.
> The tool is a modified fork of `gofmt`, so it can be used as a drop-in replacement. Running `gofmt` after `gofumpt` should be a no-op.

Would `gofumpt` (and `gofumports`) be added as an entirely new linter, or options for `gofmt`/`goimports` ?

gonzaloserrano:
Hey there, I'm also interested in adding gofumpt/goimports.

> Would gofumpt (and gofumports) be added as an entirely new linter, or options for gofmt/goimports ?

About new linter vs gofmt option is see pros and cons:

- new linter pros: more clean to add/remove the linter if it gets deprecated
- cons: does not make sense to run gofmt/goimports if gofumpt/gofumports are enabled since they are a superset of gofmt/goimports, so how to handle this?

ernado:
Hey, sorry for long responses, we are currently trying to deal with bugs and preparing to go1.14, so very low of capacity.

I'm also not sure how to deal with linters that superset other linters, e.g. we can't just auto-ignore them.

gonzaloserrano:
Maybe we can just start adding it as a new linter and disable gofmt/goimports manually.

teivah:
I created a PR for it: https://github.com/golangci/golangci-lint/pull/1177

dnwe:
@teivah thanks for working on the PR, pleased to see that it was merged

A quick scan of the PR suggests it doesn't currently map the golangci-lint "Autofix" functionality to a `gofumpt -s -w` equivalent, like it does for the `gofmt` linter. It would be good if that could be added, as then we could switch our .golanci-lint.yml to disable the fmt linter and switch to gofumpt wholesale

cc @mvdan as he'd probably be interested in knowing that this work is taking place to add gofumpt to golangci-lint

SVilgelm:
[...]

PR #1177, opened 2020-06-05
Adding gofumpt

teivah:
Hey!
I was also interested in https://github.com/golangci/golangci-lint/issues/490 (adding gofumpt).
Here is a PR for it.

### DIFF
--- a/pkg/golinters/gofumpt.go
+++ b/pkg/golinters/gofumpt.go
@@ -1,0 +1,76 @@
+package golinters
+
+import (
+	"bytes"
+	"fmt"
+	"go/token"
+	"io/ioutil"
+	"sync"
+
+	"golang.org/x/tools/go/analysis"
+	"mvdan.cc/gofumpt/format"
+
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
+	"github.com/golangci/golangci-lint/pkg/lint/linter"
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+const gofumptName = "gofumpt"
+
+func NewGofumpt() *goanalysis.Linter {
+	var mu sync.Mutex
+	var resIssues []goanalysis.Issue
+
+	analyzer := &analysis.Analyzer{
+		Name: gofumptName,
+		Doc:  goanalysis.TheOnlyanalyzerDoc,
+	}
+	return goanalysis.NewLinter(
+		gofumptName,
+		"Gofumpt checks whether code was gofumpt-ed.",
+		[]*analysis.Analyzer{analyzer},
+		nil,
+	).WithContextSetter(func(lintCtx *linter.Context) {
+		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
+			var fileNames []string
+			for _, f := range pass.Files {
+				pos := pass.Fset.PositionFor(f.Pos(), false)
+				fileNames = append(fileNames, pos.Filename)
+			}
+
+			var issues []goanalysis.Issue
+
+			for _, f := range fileNames {
+				input, err := ioutil.ReadFile(f)
+				if err != nil {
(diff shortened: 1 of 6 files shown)


pkg/golinters/gofumpt.go
```diff
@@ -0,0 +1,76 @@
...
+				}
+				output, err := format.Source(input, "")
+				if err != nil {
+					return nil, fmt.Errorf("error while running gofumpt: %w", err)
+				}
+				if !bytes.Equal(input, output) {
+					issues = append(issues, goanalysis.NewIssue(&result.Issue{
+						FromLinter: gofumptName,
+						Text:       "File is not `gofumpt`-ed",
+						Pos: token.Position{
```
tetafro, line 56: There is a `Replacement` parameter to automatically fix issues, which is very useful for gofumpt.

teivah: If I'm not mistaken, gofumpt provides only as a public function a way to get the gofumpt-ed version of the file (as a `[]byte`).
With `Replacement` you mean I should include the gofumpt-ed version of the whole file in the issue?

teivah: @tetafro Am I correct?

tetafro: > gofumpt provides only as a public function a way to get the gofumpt-ed version of the file

Yeah, sorry, I didn't think of it.

> With Replacement you mean I should include the gofumpt-ed version of the whole file in the issue?

I'm not sure if it's a right thing to do. Someone more experienced with golangci-lint architecture should answer.

---

Anyway, if there's no easy way to do this, just skip my first comment.

pkg/lint/lintersdb/manager.go
```diff
@@ -191,6 +191,10 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithPresets(linter.PresetFormatting).
 			WithAutoFix().
 			WithURL("https://golang.org/cmd/gofmt/"),
+		linter.NewConfig(golinters.NewGofumpt()).
+			WithPresets(linter.PresetFormatting).
+			WithAutoFix().
```
tetafro, line 196: If there is no autofix, you should remove this line

tetafro:
Can you please add test to tests/testdata?

teivah:
> Can you please add test to tests/testdata?

@tetafro I've just added a test. Just something to mention, previously I was setting the [error line](https://github.com/teivah/golangci-lint/blob/dffb7b4140ab4ee1d8d06c9e2fbfcd7fbb642e2c/pkg/golinters/gofumpt.go#L69) to 0. I wanted to change it to line 1 but in the test file, I can't do this:
```
// ERROR "File is not `gofumpt`-ed"
//args: -Egofumpt
package testdata
```
Hence, I set it to the last line number and it's working correctly.

tetafro:
> Hence, I set it to the last line number and it's working correctly.

I've looked how gofmt linter is implemented. Seems like the right thing to do here is to re-use diff code from [gofmt linter](#https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/gofmt_common.go#L50).

But I think using the last line is ok for the first iteration. Maybe just add `TODO` in the code.

teivah:
> Thanks! This looks good to me.

@tetafro Any chance to get it merged then? :)

tetafro:
> Any chance to get it merged then? :)

Yeah, sorry for delay

teivah:
Thanks @tetafro

gonzaloserrano:
Hi, how's the doc updated once a new linter is merged?

teivah:
@tetafro Just out of curiosity, when do you plan roughly to make a new release?
Thanks

tetafro:
@teivah this is a question for guys from core team. I just help with pr reviews from time to time.

Issue #1180, opened 2020-06-07
Add goheader linter


denis-tingajkin:
Hello guys,

Do you interesting include [this linter ](https://github.com/denis-tingajkin/go-header) to the set of linters? 

The linter just checks copyright headers for the passed set of `.go` files.
I think this linter should be disabled by default and can be enabled by users who want to use headers checking in their projects.

PR #1181, opened 2020-06-07
Add go-header linter

denis-tingajkin:
Signed-off-by: denis-tingajkin <denis.tingajkin@xored.com>

https://github.com/golangci/golangci-lint/issues/1180

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -16,6 +16,7 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
 github.com/Djarvur/go-err113 v0.0.0-20200511133814-5174e21577d5 h1:XTrzB+F8+SpRmbhAH8HLxhiiG6nYNwaBZjrFps1oWEk=
 github.com/Djarvur/go-err113 v0.0.0-20200511133814-5174e21577d5/go.mod h1:4UJr5HIiMZrwgkSPdsjy2uOQExX/WEILpIrO9UPGuXs=
+github.com/Djarvur/go-err113 v0.1.0/go.mod h1:4UJr5HIiMZrwgkSPdsjy2uOQExX/WEILpIrO9UPGuXs=
 github.com/Masterminds/semver v1.5.0 h1:H65muMkzWKEuNDnfl9d70GUjFniHKHRbFPGBuZ3QEww=
 github.com/Masterminds/semver v1.5.0/go.mod h1:MB6lktGJrhw8PrUyiEoblNEGEQ+RzHPF078ddwwvV3Y=
 github.com/OneOfOne/xxhash v1.2.2/go.mod h1:HSdplMjZKSmBqAxg5vPj2TmRDmfkzw+cTzAElWljhcU=
@@ -23,6 +24,7 @@ github.com/OpenPeeDeeP/depguard v1.0.1 h1:VlW4R6jmBIv3/u1JNlawEvJMM4J+dPORPaZasQ
 github.com/OpenPeeDeeP/depguard v1.0.1/go.mod h1:xsIw86fROiiwelg+jB2uM9PiKihMMmUx/1V+TNhjQvM=
 github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6 h1:fLjPD/aNc3UIOA6tDi6QXUemppXK3P9BI7mr2hd6gx8=
 github.com/StackExchange/wmi v0.0.0-20180116203802-5d049714c4a6/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
+github.com/StackExchange/wmi v0.0.0-20190523213315-cbe66965904d/go.mod h1:3eOhrUMpNV+6aFIbp5/iudMxNCF27Vw2OZgy4xEx0Fg=
 github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuyumcjzFXgccqObfd/Ljyb9UuFJ6TxHnclSeseNhc=
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
 github.com/armon/circbuf v0.0.0-20150827004946-bbbad097214e/go.mod h1:3U/XgcO3hCbHZ8TKRvWD2dDTCfh9M9ya+I9JpbB7O8o=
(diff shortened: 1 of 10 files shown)


test/linters_test.go
```diff
@@ -183,10 +183,9 @@ func extractRunContextFromComments(t *testing.T, sourcePath string) *runContext
 	scanner := bufio.NewScanner(f)
 	for scanner.Scan() {
 		line := scanner.Text()
-		if !strings.HasPrefix(line, "//") {
+		if strings.HasPrefix(strings.TrimSpace(line), "package") {
```
denis-tingajkin, line 186: Here I did a not typical change for linter contributing:
Before this change test parameters can be only in first comment lines.
After this change test parameters can be set in any comments line before key work `packgage`

denis-tingajkin: This needed to add possible to test go-header with golangci testing approach.

go.sum
```diff
@@ -369,11 +380,14 @@ github.com/subosito/gotenv v1.2.0 h1:Slr1R9HxAlEKefgq5jn9U+DnETlIUa6HfgEzj0g5d7s
...
 github.com/tdakkota/asciicheck v0.0.0-20200416190851-d7f85be797a2 h1:Xr9gkxfOP0KQWXKNqmwe8vEeSUiUj4Rlee9CMVX2ZUQ=
 github.com/tdakkota/asciicheck v0.0.0-20200416190851-d7f85be797a2/go.mod h1:yHp0ai0Z9gUljN3o0xMhYJnH/IcvkdTBOX2fmJ93JEM=
+github.com/tdakkota/asciicheck v0.0.0-20200416200610-e657995f937b/go.mod h1:yHp0ai0Z9gUljN3o0xMhYJnH/IcvkdTBOX2fmJ93JEM=
 github.com/tetafro/godot v0.4.2 h1:Dib7un+rYJFUi8vN0Bk6EHheKy6fv6ZzFURHw75g6m8=
 github.com/tetafro/godot v0.4.2/go.mod h1:/7NLHhv08H1+8DNj0MElpAACw1ajsCuf3TKNQxA5S+0=
 github.com/timakin/bodyclose v0.0.0-20190930140734-f7f2e9bca95e h1:RumXZ56IrCj4CL+g1b9OL/oH0QnsF976bC8xQFYUD5Q=
 github.com/timakin/bodyclose v0.0.0-20190930140734-f7f2e9bca95e/go.mod h1:Qimiffbc6q9tBWlVV6x0P9sat/ao1xEkREYPPj9hphk=
+github.com/timakin/bodyclose v0.0.0-20200424151742-cb6215831a94/go.mod h1:Qimiffbc6q9tBWlVV6x0P9sat/ao1xEkREYPPj9hphk=
 github.com/tmc/grpc-websocket-proxy v0.0.0-20190109142713-0ad062ec5ee5/go.mod h1:ncp9v5uamzpCO7NfCPTXjqaC+bZgJeR0sMTm6dMHP7U=
+github.com/tommy-muehle/go-mnd v1.3.0/go.mod h1:dSUh0FtTP8VhvkL1S+gUR1OKd9ZnSaozuI6r3m6wOig=
```
sayboras, line 390: not sure why this entry got added (it's old version). I have run _go mod tidy_ on this branch, and there were a bunch of git dirty changes. Not sure if it's my problem.

```
$ go mod tidy
$ git status   
On branch pull/1181
  (use "git add <file>..." to update what will be committed)
	modified:   go.sum
$ go version   
go version go1.14.4 linux/amd64
```

denis-tingajkin: thanks fixed

pkg/config/config.go
```diff
@@ -247,6 +248,12 @@ type LintersSettings struct {
 	Custom map[string]CustomLinterSettings
 }
 
+type GoHeaderSettings struct {
```
sayboras, line 251: Can we have some example for settings in https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml ? 

This will help for quick reference, compared to checking the original repo or actual implementation.

denis-tingajkin: >Can we have some example for settings in https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml ?

Sure, fixed

test/linters_test.go
```diff
@@ -212,8 +211,6 @@ func extractRunContextFromComments(t *testing.T, sourcePath string) *runContext
 			rc.configPath = configPath
 			continue
 		}
-
-		assert.Fail(t, "invalid prefix of comment line %s", line)
```
iwankgb, line 216: Why have you removed this line and why do you think it's safe to do so?

denis-tingajkin: > Why have you removed this line

I think this line can be removed to allow put comments before `//args`, `//config_path`

> why do you think it's safe to do so?

Why do you think that user comments before  `//args`, `//config_path` is not safe?

iwankgb: I would prefer to keep this line and improve condition in line 186. I don't think that comments before two mentioned  prefixes are unsafe but I want this test to fail when it is necessary - and in my opinion you removed rather critical assertion as we expect this code to be never reached (unless tests fail).

denis-tingajkin: @iwankgb I do not see any scenario when this assert will really necessary, can you please explain why we need to keep this assert? 

As I know before section `package` for `go` files we can have only comments. So If the source file will have something other before section `package` then it should fail on compile time.

denis-tingajkin:
@iwankgb please take a look at comment https://github.com/golangci/golangci-lint/pull/1181#discussion_r449781494

test/linters_test.go
iwankgb: If I understand code correctly, then original intention was to `return rc` on multi-line comment that starts with `/*` and hits original line 186. After your change such line will be ignored and would ultimately hit the assertion you deleted. As far as I understand such failure occurs in `test/testdata/go-header.go` and it's perfectly fine with me that multi-line comment would be used for file header although I think that it would be *much* better to add multi-line comment handling to the loot rather then remove `assert.Fail()`.

denis-tingajkin: > If I understand code correctly, then original intention was to return rc on multi-line comment that starts with /* and hits original line 

Interesting... Can you point where is using something that starts with multi-line comment? I've checked all testdata/.*\.go files and did not find using this.

iwankgb: Here: https://github.com/golangci/golangci-lint/pull/1181/files#diff-787479a653c2e5d026e44d51f6d64580R1

denis-tingajkin: > Here: https://github.com/golangci/golangci-lint/pull/1181/files#diff-787479a653c2e5d026e44d51f6d64580R1

Sorry, I thought you meant that multiline comments are used in golangci tests...


>  although I think that it would be much better to add multi-line comment handling to the loop rather then remove assert.Fail().

Yes, it sounds like a solution but I did not quite understand what the case this assert should catch a problem.

iwankgb: I'm pretty conservative when it comes to removing testing code and I would appreciate if you handled multi-line comment as all other cases are handled. I do admit that I can't think of non-comment case but I do not know what author had in mind.

denis-tingajkin: ok, finally used a more conservative solution.

test/linters_test.go
```diff
@@ -183,10 +183,12 @@ func extractRunContextFromComments(t *testing.T, sourcePath string) *runContext
 	scanner := bufio.NewScanner(f)
 	for scanner.Scan() {
 		line := scanner.Text()
+		if strings.HasPrefix(line, "/*") {
```
iwankgb, line 186: I'm afraid you have to take more cases into consideration:
* Lines between end and beginning of multi-line comment
* End of multi-line comment

denis-tingajkin: Sure, fixed!

PR #1184, opened 2020-06-08
Add instruction to install globally

pararang:
Add instruction to install globally when using pre-commit on Linux.

### DIFF
--- a/docs/src/docs/usage/install/index.mdx
+++ b/docs/src/docs/usage/install/index.mdx
@@ -70,6 +70,19 @@ golangci-lint --version
 
 On Windows, you can run the above commands with Git Bash, which comes with [Git for Windows](https://git-scm.com/download/win).
 
+#### Command Not Found on Linux with Pre-Commit
+
+```sh
+- exit code: 127
+/home/user/.cache/pre-commit/som3reP0/run-golangci-lint.sh: 2: exec: golangci-lint: not found 
+```
+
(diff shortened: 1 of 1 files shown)


docs/src/docs/usage/install/index.mdx
```diff
@@ -70,6 +70,19 @@ golangci-lint --version
...
+
+```sh
+- exit code: 127
+/home/user/.cache/pre-commit/som3reP0/run-golangci-lint.sh: 2: exec: golangci-lint: not found 
+```
+
+If youn thos error, make sure golangci-lint installed globally. The indicator is command `golangci-lint --version` returned nothing when you run it on the outside of directory `$(go env GOPATH)/bin`. To fix this, move the downloaded binari from `$(go env GOPATH)/bin` to `/usr/bin`.
+
+```sh
+sudo mv $(go env GOPATH)/bin/golangci-lint /usr/bin/golangci-lint
```
iwankgb, line 83: I think that we should leave `/usr/bin` under control of distribution developers. In your case I would recommend updating `PATH` to include `$HOME/go/bin` (or whatever `go env GOBIN` shows) and keep installing your dependencies locally.

tetafro: One more tip - the right directory for global installation is `/usr/local/bin`.

But as mentioned above - you should just add your bin directory to `$PATH`. It will work with git hooks as well.

pararang: noted. thanks @iwankgb , @tetafro

Issue #612, opened 2019-07-17
fatal error: concurrent map writes


yaches:
Running linter in gilab ci
golangci-lint:v1.17
go 1.12 

err log:
[golangci.log](https://github.com/golangci/golangci-lint/files/3402158/golangci.log)

.golangci.yml:
```linters:
  enable-all: true
  disable:
    # Globals are OK
    - gochecknoglobals
    # So are inits
    - gochecknoinits
    # Cant disable some of go vet checks => disable all of them
    - govet
    # Useless
    - interfacer
    # Long lines are fine
    - lll
    # Using "weak" primitivies is fine when they're not used for crypto
    - gosec

run:
  tests: false
  issues-exit-code: 0
  skip-dirs:
    - vendor/
  skip-files:
    - _test.go$
```

Helcaraxan:
I just encountered exactly the same error with an identical [stack-trace](https://github.com/golangci/golangci-lint/files/3402978/concurrent-writes.log).

Helcaraxan:
@yaches looks like this was actually fixed 4 days ago in `master` (#594) and hence was not part of `1.17.1` so we should be able to close this issue. Although having a `1.17.2` patch release with this fix would be _really_ nice.

Helcaraxan:
As suggested in #622, given this has been addressed it could be worth closing the issue. 🙂

jirfag:
Hi, thank you for reporting, it's fixed in [v1.18.0 release](https://github.com/golangci/golangci-lint/releases/tag/v1.18.0)

gaurav-dalvi:
I am seeing same issue on 
```
golangci-lint version
golangci-lint has version 1.20.1 built from 849044b on 2019-10-15T19:11:27Z
```

jirfag:
@gaurav-dalvi can you attach new logs/output please?

ConorNevin:
@jirfag I'm also having the same issue intermittently on v1.23.8

error log: 
[golangci.log](https://github.com/golangci/golangci-lint/files/4426389/golangci.log)

michaljemala:
[...]

PR #1192, opened 2020-06-21
Protect NewFilenameUnadjuster from concurrent map writes

Helcaraxan:
As signalled in #612 there has been another occurrence of concurrent map writes.

This should address the issue by passing a mutex alongside the map and locking it when writing. Given that the concurrent writes are only occurring while the map is being filled we do not need to pass the mutex back as part of the `FilenameUnadjuster` struct. Consumers of this type should protect against concurrency issues themselves when operating on the enclosed map field.

### DIFF
--- a/pkg/result/processors/filename_unadjuster.go
+++ b/pkg/result/processors/filename_unadjuster.go
@@ -16,6 +16,11 @@ import (
 
 type posMapper func(pos token.Position) token.Position
 
+type adjustMap struct {
+	sync.Mutex
+	m map[string]posMapper
+}
+
 // FilenameUnadjuster is needed because a lot of linters use fset.Position(f.Pos())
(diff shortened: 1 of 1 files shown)


Helcaraxan:
@jirfag do you need anything from me in order to merge this? Happy to do so if required, just want to make we're not blocked by anything that I can help with. :slightly_smiling_face:

michaljemala:
@Helcaraxan Could we pls get this merged soon-ish? :-)

Helcaraxan:
@michaljemala I do not have write access on the repository so I can't merge it unfortunately, hence my previous question to @jirfag. :disappointed:

bombsimon:
I'll merge this, although I don't know when we will get a new release.

Helcaraxan:
Ah great. Many thanks @bombsimon ! Now I can contribute better. My previous contribution pre-dates the existence of the automator. :slightly_smiling_face:

Issue #1208, opened 2020-07-05
Automate update for github-action-config.json


sayboras:
Need to investigate on how to automate this

_Originally posted by @sayboras in https://github.com/golangci/golangci-lint/pull/1207#issuecomment-653829660_

Currently we have Make target `make assets/github-action-config.json` to update github-action-config.json, will be better to have it run as part of release automatically.

Updated: Just found the below steps in https://github.com/golangci/golangci-lint/blob/master/docs/src/docs/contributing/workflow.mdx

After making a release you need to update (TODO: automate partially):

1. GitHub [action config](https://github.com/golangci/golangci-lint/blob/master/assets/github-action-config.json) by running:

```sh
make assets/github-action-config.json
```

2. Update netlify state hash

```sh
make update_netlify_state
```

3. Contributors list

```sh
make update_contributors_list # may take 15 min
```
`

SVilgelm:
So, actually we need to run these steps after a release. It means, that when the release workflow is completed, next one should run those steps, create a commit an push to the master branch

sayboras:
> So, actually we need to run these steps after a release. It means, that when the release workflow is completed, next one should run those steps, create a commit an push to the master branch

Yes, you are right. Currently, I think we just wait till the issue with gorelease for bew gets resolved first, then the above steps can be done in the same job (after gorelease)

denis-tingajkin:
@sayboras I feel to commit changes automatically we need a user account/bot. 

Question: Have `golangci` a user for automatization?

ernado:
AFAIK changes can be commited via default github actions user, but I can be wrong

sayboras:
@denis-tingajkin @ernado I also think github action user can do so, similar to goreleaser is able to create the release. Might need to test it out. I did have dependabot commit directly into master in my repo before.

SVilgelm:
@sayboras it's easy, except on part. The master branch is protected, we need to create the Pull Request from an action.
1. the `release` job is completed
2. git checkout -b update-docs
3. `make update_contributors_list`, `make assets/github-action-config.json`, `make update_netlify_state`
4. git add -u
5. git push origin update-docs
6. magic script to create a pull request using GitHub API

SVilgelm:
I know that the default GITHUB_TOKEN does not allow creation of the pull requests, but we can use `secrets.GOLANGCI_LINT_TOKEN`

SVilgelm:
[...]

PR #1207, opened 2020-07-05
To enable v1.28 in github action

sayboras:
Update github-action.json and README.md by running below make target.

```
make assets/github-action-config.json
make update_contributors_list
```

PR #1209, opened 2020-07-05
Fix: goheader linter can throw nil pointer exception in case of a source file has not issues

denis-tingajkin:
## Description

During local testing latest golangci, we've found that go-header linter can throw nil pointer exception in case of a source file is valid.

## What is this patch doing?
- [x] Fix issue.
- [x] Add test to cover.
- [x] Allow using empty new lines in `testdata/.*\.go` files.

## Steps to reproduce the issue

1. Use next go-header config
```
linters-settings:
  goheader:
    template: "test"
```
2. Run golangci linter on file with valid header, for example
```
//test

//Package mypkg ...
package mypkg
```
Actual: nil pointer exception
Expected: Template has matched no issues should be found

### DIFF
--- a/pkg/golinters/goheader.go
+++ b/pkg/golinters/goheader.go
@@ -51,6 +51,9 @@ func NewGoHeader() *goanalysis.Linter {
 			var res []goanalysis.Issue
 			for _, file := range pass.Files {
 				i := a.Analyze(file)
+				if i == nil {
+					continue
+				}
 				issue := result.Issue{
 					Pos: token.Position{
 						Line:     i.Location().Line + 1,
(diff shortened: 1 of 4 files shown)


denis-tingajkin:
Here is failed one check that is fixed in https://github.com/golangci/golangci-lint/pull/1207

denis-tingajkin:
@sayboras, @iwankgb, I interested to release a new version with this patch dfa0013583b1a0eef9c00e512fd6c48bbe7033a9, Do you have any idea what I need to do? :)

iwankgb:
This is bugfix release so:
* Push `v1.28.1` tag.
* follow steps outlined in #1208

denis-tingajkin:
@iwankgb thanks! Done! Also I got a question where we should put generated docs/.*mdx files?

iwankgb:
@denis-tingajkin I don't think I understand what you mean. `mdx` files gets published automatically to the docs site.

denis-tingajkin:
@iwankgb I meant generated files after `make expand_website_templates`. Should I also create PR to the website repository?

PR #1215, opened 2020-07-06
Explicitly pass GITHUB_TOKEN env to goreleaser action

SVilgelm:
Fix a failing release workflow

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -23,3 +23,5 @@ jobs:
         uses: goreleaser/goreleaser-action@v1
         with:
           args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GOLANGCI_LINT_TOKEN }}
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -12,7 +12,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install Go
-        uses: actions/setup-go@v1
+        uses: actions/setup-go@v2
         with:
           go-version: 1.14
       - name: Unshallow
@@ -20,8 +20,9 @@ jobs:
       - name: Login do docker.io
         run: docker login -u golangci -p ${{ secrets.GOLANGCI_LINT_DOCKER_TOKEN }}
       - name: Create release
-        uses: goreleaser/goreleaser-action@v1
+        uses: goreleaser/goreleaser-action@v2
         with:
+          version: latest
           args: release --rm-dist
         env:
           GITHUB_TOKEN: ${{ secrets.GOLANGCI_LINT_TOKEN }}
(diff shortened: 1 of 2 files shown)


.github/workflows/tag.yml
```diff
@@ -23,3 +23,5 @@ jobs:
         uses: goreleaser/goreleaser-action@v1
         with:
           args: release --rm-dist
+        env:
```
sayboras, line 26: :100: 

Just curious on how to re-run, or one more release is required.

SVilgelm: Actually it's impossible to re run this action because it  triggers on a tag event:
```
on:		on:
  push:		  push:
    tags:		    tags:
      - v*
```
You can re-run failed actions, but they will use an old code.
Only one way to avoid a new release is  run goreleaser manually

.github/workflows/tag.yml
```diff
@@ -23,3 +23,5 @@ jobs:
         uses: goreleaser/goreleaser-action@v1
```
sayboras, line 23: not related to this PR, but we can also bump to v2 for this action as well.

SVilgelm: Should I change it? it will take 2 minutes

sayboras: yes please :)

SVilgelm: ok

.github/workflows/tag.yml
sayboras: seems like only brew release was failing, I think we can re-run manually for brew only. /cc @golangci/team @golangci/core-team

.github/workflows/tag.yml
```diff
@@ -12,14 +12,17 @@ jobs:
...
       - name: Login do docker.io
         run: docker login -u golangci -p ${{ secrets.GOLANGCI_LINT_DOCKER_TOKEN }}
       - name: Create release
-        uses: goreleaser/goreleaser-action@v1
+        uses: goreleaser/goreleaser-action@v2
         with:
+          version: latest
           args: release --rm-dist
+        env:
+          GITHUB_TOKEN: ${{ secrets.GOLANGCI_LINT_TOKEN }}
```
sayboras, line 28: nits we can refer to env define in job level as well, something like `${{ env.GITHUB_TOKEN }}`, but I think it's fine to use secrets.GOLANGCI_LINT_TOKEN here :)

SVilgelm:
@sayboras @golangci/team @golangci/core-team 
Can we bump v1.28.2 to update the brew formulae? Or I can create a pr manually for tap repo

sayboras:
Thanks for your reminder. I just push new tag _v1.28.2_ to kick off the new release again. Sadly, it's still failing https://github.com/golangci/golangci-lint/runs/856010530?check_suite_focus=true.

Might need to wait for @golangci/core-team to help to verify GITHUB_TOKEN

SVilgelm:
yeah, I see, I checked the `.goreleaser.yml` and everything looks correct

SVilgelm:
@golangci/core-team Could you check please that the `GITHUB_TOKEN` has write permissions for the `homebrew-tap` repo?

Issue #1217, opened 2020-07-07
Sorting issues 


butuzov:
**Is your feature request related to a problem? Please describe.**
It would be great to have sorted results by file/line/column. In case of similar checks, we can track issue as a group (see similarities in `unused` and `deadcode` example ), in case of different issues its can be useful to see what else linter found on/around same lines.

**Describe the solution you'd like**
Implement `sort` flag into the output config structure.

**Additional context**

Here is an example of the output without sorting (it's a little bit cut, but you got the idea).
```
main.go:19:6: `deadcode` is unused (deadcode)
main.go:10:2: S1021: should merge variable declaration with assignment on next line (gosimple)
main.go:19:6: func `deadcode` is unused (unused)
```

Here is an example of the output with sorting.

```
main.go:10:2: S1021: should merge variable declaration with assignment on next line (gosimple)
main.go:19:6: `deadcode` is unused (deadcode)
main.go:19:6: func `deadcode` is unused (unused)
```

butuzov:
If everything OK, I can add documentation and examples (once figure out how to do that).

PR #1218, opened 2020-07-07
Sorting result.Issues implementation (golangci/golangci-lint#1217)

butuzov:
Implements https://github.com/golangci/golangci-lint/issues/1217

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -80,6 +80,7 @@ func initFlagSet(fs *pflag.FlagSet, cfg *config.Config, m *lintersdb.Manager, is
 	fs.BoolVar(&oc.PrintIssuedLine, "print-issued-lines", true, wh("Print lines of code with issue"))
 	fs.BoolVar(&oc.PrintLinterName, "print-linter-name", true, wh("Print linter name in issue line"))
 	fs.BoolVar(&oc.UniqByLine, "uniq-by-line", true, wh("Make issues output unique by line"))
+	fs.BoolVar(&oc.SortResults, "sort-results", false, wh("Sort linter results"))
 	fs.BoolVar(&oc.PrintWelcomeMessage, "print-welcome", false, wh("Print welcome message"))
 	fs.StringVar(&oc.PathPrefix, "path-prefix", "", wh("Path prefix to add to output"))
 	hideFlag("print-welcome") // no longer used
--- a/pkg/lint/runner.go
+++ b/pkg/lint/runner.go
@@ -80,6 +80,7 @@ func NewRunner(cfg *config.Config, log logutils.Log, goenv *goutil.Env, es *lint
 			processors.NewPathShortener(),
 			getSeverityRulesProcessor(&cfg.Severity, log, lineCache),
 			processors.NewPathPrefixer(cfg.Output.PathPrefix),
+			processors.NewSortResults(cfg),
 		},
 		Log: log,
 	}, nil
--- a/pkg/result/processors/sort_results.go
+++ b/pkg/result/processors/sort_results.go
@@ -1,0 +1,173 @@
+package processors
+
+import (
+	"sort"
+	"strings"
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+// Base propose of this functionality to sort results (issues)
+// produced by various linters by analyzing code. We achieving this
+// by sorting results.Issues using processor step, and chain based
+// rules that can compare different properties of the Issues struct.
+
+var _ Processor = (*SortResults)(nil)
+
+type SortResults struct {
+	cmp comparator
+	cfg *config.Config
+}
+
+func NewSortResults(cfg *config.Config) *SortResults {
+	// For sorting we are comparing (in next order): file names, line numbers,
+	// position, and finally - giving up.
+	return &SortResults{
+		cmp: ByName{
+			next: ByLine{
+				next: ByColumn{},
+			},
+		},
+		cfg: cfg,
+	}
+}
+
+// Process is performing sorting of the result issues.
+func (sr SortResults) Process(issues []result.Issue) ([]result.Issue, error) {
+	if !sr.cfg.Output.SortResults {
+		return issues, nil
+	}
+
+	sort.Slice(issues, func(i, j int) bool {
+		return sr.cmp.Compare(&issues[i], &issues[j]) == Less
+	})
+
+	return issues, nil
+}
+
+func (sr SortResults) Name() string { return "sort_results" }
+func (sr SortResults) Finish()      {}
+
+type compareResult int
+
+const (
+	Less compareResult = iota - 1
+	Equal
+	Greater
+	None
+)
+
+func (c compareResult) isNeutral() bool {
+	// return true if compare result is incomparable or equal.
+	return c == None || c == Equal
+}
+
+//nolint:exhaustive
+func (c compareResult) String() string {
+	switch c {
+	case Less:
+		return "Less"
+	case Equal:
+		return "Equal"
+	case Greater:
+		return "Greater"
+	}
+
+	return "None"
+}
+
+// comparator describe how to implement compare for two "issues" lexicographically
+type comparator interface {
+	Compare(a, b *result.Issue) compareResult
+	Next() comparator
+}
+
+var (
+	_ comparator = (*ByName)(nil)
+	_ comparator = (*ByLine)(nil)
+	_ comparator = (*ByColumn)(nil)
+)
+
+type ByName struct{ next comparator }
+
+//nolint:golint
+func (cmp ByName) Next() comparator { return cmp.next }
+
+//nolint:golint
+func (cmp ByName) Compare(a, b *result.Issue) compareResult {
+	var res compareResult
+
+	if res = compareResult(strings.Compare(a.FilePath(), b.FilePath())); !res.isNeutral() {
+		return res
+	}
+
+	if next := cmp.Next(); next != nil {
+		return next.Compare(a, b)
+	}
+
+	return res
+}
+
+type ByLine struct{ next comparator }
+
+//nolint:golint
+func (cmp ByLine) Next() comparator { return cmp.next }
+
+//nolint:golint
+func (cmp ByLine) Compare(a, b *result.Issue) compareResult {
+	var res compareResult
+
+	if res = numericCompare(a.Line(), b.Line()); !res.isNeutral() {
+		return res
+	}
+
+	if next := cmp.Next(); next != nil {
+		return next.Compare(a, b)
+	}
+
+	return res
+}
+
+type ByColumn struct{ next comparator }
+
+//nolint:golint
+func (cmp ByColumn) Next() comparator { return cmp.next }
+
+//nolint:golint
+func (cmp ByColumn) Compare(a, b *result.Issue) compareResult {
+	var res compareResult
+
+	if res = numericCompare(a.Column(), b.Column()); !res.isNeutral() {
+		return res
+	}
+
+	if next := cmp.Next(); next != nil {
+		return next.Compare(a, b)
+	}
+
+	return res
+}
+
+func numericCompare(a, b int) compareResult {
+	var (
+		isValuesInvalid  = a < 0 || b < 0
+		isZeroValuesBoth = a == 0 && b == 0
+		isEqual          = a == b
+		isZeroValueInA   = b > 0 && a == 0
+		isZeroValueInB   = a > 0 && b == 0
+	)
+
+	switch {
+	case isZeroValuesBoth || isEqual:
+		return Equal
+	case isValuesInvalid || isZeroValueInA || isZeroValueInB:
+		return None
+	case a > b:
+		return Greater
+	case a < b:
+		return Less
+	}
+
+	return Equal
+}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -519,6 +519,7 @@ type Config struct {
 		PrintIssuedLine     bool   `mapstructure:"print-issued-lines"`
 		PrintLinterName     bool   `mapstructure:"print-linter-name"`
 		UniqByLine          bool   `mapstructure:"uniq-by-line"`
+		SortResults         bool   `mapstructure:"sort-results"`
 		PrintWelcomeMessage bool   `mapstructure:"print-welcome"`
 		PathPrefix          string `mapstructure:"path-prefix"`
 	}
--- a/pkg/result/processors/sort_results_test.go
+++ b/pkg/result/processors/sort_results_test.go
@@ -1,0 +1,182 @@
+package processors
+
+import (
+	"fmt"
+	"go/token"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+var issues = []result.Issue{
+	{
+		Pos: token.Position{
+			Filename: "file_windows.go",
+			Column:   80,
+			Line:     10,
+		},
+	},
+	{
+		Pos: token.Position{
+			Filename: "file_linux.go",
+			Column:   70,
+			Line:     11,
+		},
+	},
+	{
+		Pos: token.Position{
+			Filename: "file_darwin.go",
+			Line:     12,
+		},
+	},
+	{
+		Pos: token.Position{
+			Filename: "file_darwin.go",
+			Column:   60,
+			Line:     10,
+		},
+	},
+}
+
+type compareTestCase struct {
+	a, b     result.Issue
+	expected compareResult
+}
+
+func testCompareValues(t *testing.T, cmp comparator, name string, tests []compareTestCase) {
+	t.Parallel()
+
+	for i := 0; i < len(tests); i++ {
+		test := tests[i]
+		t.Run(fmt.Sprintf("%s(%d)", name, i), func(t *testing.T) {
+			res := cmp.Compare(&test.a, &test.b)
+			assert.Equal(t, test.expected.String(), res.String())
+		})
+	}
+}
+
+func TestCompareByLine(t *testing.T) {
+	testCompareValues(t, ByLine{}, "Compare By Line", []compareTestCase{
+		{issues[0], issues[1], Less},    // 10 vs 11
+		{issues[0], issues[0], Equal},   // 10 vs 10
+		{issues[3], issues[3], Equal},   // 10 vs 10
+		{issues[0], issues[3], Equal},   // 10 vs 10
+		{issues[3], issues[2], Less},    // 10 vs 12
+		{issues[1], issues[1], Equal},   // 11 vs 11
+		{issues[1], issues[0], Greater}, // 11 vs 10
+		{issues[1], issues[2], Less},    // 11 vs 12
+		{issues[2], issues[3], Greater}, // 12 vs 10
+		{issues[2], issues[1], Greater}, // 12 vs 11
+		{issues[2], issues[2], Equal},   // 12 vs 12
+	})
+}
+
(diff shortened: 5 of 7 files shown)


pkg/lint/runner.go
```diff
@@ -72,6 +72,7 @@ func NewRunner(cfg *config.Config, log logutils.Log, goenv *goutil.Env, es *lint
 			processors.NewNolint(log.Child("nolint"), dbManager, enabledLinters),
 
 			processors.NewUniqByLine(cfg),
+			processors.NewSortResults(cfg),
```
sayboras, line 75: any reason sorting processor is added in this order? I was thinking about having it last

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,177 @@
...
+
+// Base propose of this functionality to sort results (issues)
+// produced by various linters by analizing code. We achieving this
+// by sorting results.Issues using processor step, and chain based
+// rules that can compare different properties of the Issues struct.
+
+var _ Processor = (*SortResults)(nil)
+
+type SortResults struct {
+	cmp Comparer
```
sayboras, line 19: Could be named as Comparator ?

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Process is performing sorting of the result issues.
+func (sr SortResults) Process(issues []result.Issue) ([]result.Issue, error) {
+	if !sr.cfg.Output.SortResults {
+		return issues, nil
+	}
+
+	sort.Slice(issues, func(i, j int) bool {
+		return sr.cmp.Compare(&issues[i], &issues[j]) == Less
```
sayboras, line 43: Common way is just comparing to zero. We can simplify consts as well.

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,177 @@
...
+
+type ByName struct{ next Comparer }
+
+func (cmp ByName) Next() Comparer { return cmp.next }
+func (cmp ByName) Compare(a, b *result.Issue) CompareResult {
+	var res CompareResult
+
+	if res = CompareResult(strings.Compare(a.FilePath(), b.FilePath())); !res.isNeutral() {
+		return res
+	} else if next := cmp.Next(); res.isNeutral() && next != nil {
```
sayboras, line 100: res.isNeutral()  is required ?

pkg/result/processors/sort_results.go
butuzov: Yeah, it could!

pkg/lint/runner.go
butuzov: That makes sense!

pkg/result/processors/sort_results.go
butuzov: I would keep it, its simplify if/else branches. Having `res.isNeutral()` vs `res == Less || res == Greater` and `!res.IsNeutral()` vs `res == Equal || res == None`, imho better for understanding, live width and general complexity. If you are insisting - I will change, but I  prefer not to.

pkg/result/processors/sort_results.go
butuzov: Same reasons as in `res.isNeutral()` related comment. Read it as a story: If the results of comparing `a` and `b` are state `less`, compare function receives true.  I would keep names constants as it really simplifies understanding of the code for a small price.

pkg/result/processors/sort_results.go
sayboras: oh what I mean is that !res.isNeutral() will return right away in previous if. I am thinking about something like

```suggestion
	if res = CompareResult(strings.Compare(a.FilePath(), b.FilePath())); !res.isNeutral() {
		return res
	}
	
	if next := cmp.Next(); next != nil {
		return next.Compare(a, b)
	}
	return res
```

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,177 @@
...
+import (
+	"sort"
+	"strings"
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+// Base propose of this functionality to sort results (issues)
+// produced by various linters by analizing code. We achieving this
```
sayboras, line 12: nits: analizing might be analyzing

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,177 @@
...
+var _ Processor = (*SortResults)(nil)
+
+type SortResults struct {
+	cmp Comparator
+	cfg *config.Config
+}
+
+func NewSortResults(cfg *config.Config) *SortResults {
+	// For sorting we are comparing (in next order): file names, line numbers,
+	// position, range numbers (if position not available), and finnally - giving up.
```
sayboras, line 25: nits: finnally -> finally

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,177 @@
...
+	return &SortResults{
+		cmp: ByName{
```
sayboras, line 27: understand that your intention is to compose the sorting rules. However, if I looked at the Comparer we have now _ByName_ along. 

```
               ByName{
			next: ByLine{
                        ...
			},
		}
```

Comparing ByName implicitly checking Line number due to composition, but will ByName comparator should _only_ deal with names, it doesn't require to know the next rule should be.

daixiang0: +1

pkg/result/processors/sort_results.go
butuzov: fair.

pkg/result/processors/sort_results.go
butuzov: fair.

pkg/result/processors/sort_results.go
butuzov: damn! my miss.

pkg/result/processors/sort_results.go
butuzov: This is `Chain of Responsibility` pattern, current resolver only needs to know about the exitance of the next one (`res.isNeutral()` we calling defines should we proceed to the next rule or not, but rules are different, strings compare, single numbers compare (line and column), two numbers(`range`) ).  Guess we suppose to see the `Comparator` interface, instead of chain of its implementations. `ByName` or any other implementation of `Comparator` needs to know only that there is the next rule in the chain.

sayboras: Not sure if chain of responsibility pattern is correct one to use here. Maybe I am mixing two different things (comparator interface + design pattern), but if we can interface as Comparator, it should just to compare two entities and nothing else. 

The logic of calling _next_ comparator is kind of repetitive in currently (e.g. check res is neutral or not, if not check the next one is there, then call next one), so adding new Comparator will require a fairly some level of understanding of existing one. 

Few snippets (based on your code) to illustrate my thinking:

```golang
type SortResults struct {
	compareFuncs []compareFunc
	cfg          *config.Config
}

type compareFunc func(result.Issue, result.Issue) compareResult

func NewSortResults(cfg *config.Config) *SortResults {
        // single responsibility function  
	byName := func(first result.Issue, second result.Issue) compareResult {
		return compareResult(strings.Compare(first.FilePath(), second.FilePath()))
	}

	byLine := func(first result.Issue, second result.Issue) compareResult {
		return compareResult(first.Line() - second.Line())
	}

	byCol := func(first result.Issue, second result.Issue) compareResult {
		return compareResult(first.Column() - second.Column())
	}

	byRange := func(first result.Issue, second result.Issue) compareResult {
		return compareResult(first.GetLineRange().From - second.GetLineRange().From)
	}

	return &SortResults{
		compareFuncs: []compareFunc{
			byName,
			byLine,
			byCol,
			byRange,
                        ... --> we can add more type of compareFunc here
		},
		cfg: cfg,
	}
}

// Process is performing sorting of the result issues.
func (sr SortResults) Process(issues []result.Issue) ([]result.Issue, error) {
	if !sr.cfg.Output.SortResults {
		return issues, nil
	}

	sort.Slice(issues, func(i, j int) bool {
		for _, c := range sr.compareFuncs {
			if res := c(issues[i], issues[j]); !res.isNeutral() {
				return res <= less
			}
		}
		return false
	})
	return issues, nil
}

```

butuzov: I had to use CoR in order to avoid making logic complicated (cmon who what to see 3-5 blocks of different if statements ?), CoR helps to pass the request along a chain of handlers.  More of that - I can test each of them. And logic isn't really simple here (what if linter don't provide column? or line?) I did try to cover it with tests.

```go
func numericCompare(a, b int) CompareResult {
	var (
		isValuesInvalid  = a < 0 || b < 0
		isZeroValuesBoth = a == 0 && b == 0
		isEqual          = a == b
		isZeroValueInA   = b > 0 && a == 0
		isZeroValueInB   = a > 0 && b == 0
	)

	switch {
	case isZeroValuesBoth || isEqual:
		return Equal
	case isValuesInvalid || isZeroValueInA || isZeroValueInB:
		return None
	case a > b:
		return Greater
	case a < b:
		return Less
	}

	return Equal
}
```


I see what you have written, and got the idea, and I wish I have more time for opensource, to rethink the structure and implement all checks, write new tests, cover edge cases to pleasure community, and maintainers in order to keep code quality high.  But my family wouldn't agree with me.  I am really sorry for wasting your time with this PR. Please reject this PR if you think code quality is way lower the level you will accept.

sayboras: > I see what you have written, and got the idea, and I wish I have more time for opensource, to rethink the structure and implement all checks, write new tests, cover edge cases to pleasure community, and maintainers in order to keep code quality high. But my family wouldn't agree with me. I am really sorry for wasting your time with this PR. Please reject this PR if you think code quality is way lower the level you will accept.

Appreciate that you spend time to send PR with a very good coverage :100: . Yeah, I personally understand your point about OSS involvement. While only few are lucky to have OSS as paid job, the rest of us (including myself) are voluntarily doing it either during night, weekends or our own spare time. In my opinion, OSS PR process is a chance to have open discussion, and in turn learn from each other. Sometimes it can take a little longer than we expect.

I could be a bit opinionated, or even wrong on this subject. So, let me park it right here, we can see if someone else from @golangci/team has any comment/idea in general.

In the meantime, I can continue reviewing on the basis that CoR is used. Hope it is fine to you.

butuzov: Appreciate this. Using your way to solve the issue will require additional time to think about implementation and tests, and actually require to throw my code out almost entirely.

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,179 @@
...
+	if res = numericCompare(a.Column(), b.Column()); !res.isNeutral() {
+		return res
+	} else if next := cmp.Next(); res.isNeutral() && next != nil {
+		return next.Compare(a, b)
+	}
+
+	return res
+}
+
+type ByRange struct{ next Comparator }
```
sayboras, line 138: Just want to confirm if this one is in used.

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,179 @@
...
+		res = to
+	}
+
+	return res
+}
+
+func numericCompare(a, b int) CompareResult {
+	var (
+		isValuesInvalid  = a < 0 || b < 0
+		isZeroValuesBoth = a == 0 && b == 0
```
sayboras, line 161: nit: this case is same as _isEqual_ case, the compare result is same for both as well

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,179 @@
...
+
+type ByLine struct{ next Comparator }
+
+func (cmp ByLine) Next() Comparator { return cmp.next }
+func (cmp ByLine) Compare(a, b *result.Issue) CompareResult {
+	var res CompareResult
+
+	if res = numericCompare(a.Line(), b.Line()); !res.isNeutral() {
+		return res
+	} else if next := cmp.Next(); res.isNeutral() && next != nil {
```
sayboras, line 117: nits: similar to previous comment. The check for res.isNeutral() can be removed like you have done for ByName

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,179 @@
...
+
+type ByColumn struct{ next Comparator }
+
+func (cmp ByColumn) Next() Comparator { return cmp.next }
+func (cmp ByColumn) Compare(a, b *result.Issue) CompareResult {
+	var res CompareResult
+
+	if res = numericCompare(a.Column(), b.Column()); !res.isNeutral() {
+		return res
+	} else if next := cmp.Next(); res.isNeutral() && next != nil {
```
sayboras, line 131: nits: similar to previous comment. The check for res.isNeutral() can be removed like you have done for ByName

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,179 @@
...
+	return "None"
+}
+
+// Comparator describe how to implement compare for two "issues" lexicographically
+type Comparator interface {
+	Compare(a, b *result.Issue) CompareResult
+	Next() Comparator
+}
+
+var (
```
bombsimon, line 85: Why is an empty interface instance created as nil but thrown away for each type? I know you can use this method to always ensure interface casting will work but I'm not familiar with doing this without ever storing the value. I removed these lines and line 16 and the tests still works just fine.

Is there any planned conversion going on or used elsewhere outside this PR? Even if so, isn't the "comma-ok" notation preferred (`byName, ok := someInterace.(*ByName)` and check for `ok`)?

pkg/result/processors/sort_results.go
butuzov: It is. But I have trouble to run the demo (not sure is that a `golangci-lint` bug, or just my configuration).

The base idea, that linters can generate different output, in purchase fallback to `ByRange` can happen if there a duplication code in the few copy/paste files in different packages, so output can looks like this.

```
examples/dupl1.go:1: 1-15 lines are duplicate of `examples/dupl2.go:1-15` (dupl)
examples/dupl1.go:1: 1-14 lines are duplicate of `examples/dupl3.go:1-14` (dupl)
examples/dupl2.go:1: 1-15 lines are duplicate of `examples/dupl3.go:1-15` (dupl)
examples/dupl3.go:1: 1-15 lines are duplicate of `examples/dupl1.go:1-15` (dupl)
...
```

Correct me if I am wrong and such output is impossible.

pkg/result/processors/sort_results.go
butuzov: You are right, however, what I was trying to show with his compare (maybe I need to add comments regarding this comparisons): `isZeroValuesBoth` and `isEqual` are correct comparisons (like itєs same column or no column at all `wsl`/`dup`), while `isValuesInvalid` (should not happen at all as lines/columns (ref `go/token`) starts from 1 ) and `isZeroValueIn(A|B)` should indicate incompatible comparison.

I just need to list all possible cases, even if they are impossibly looking ones.

pkg/result/processors/sort_results.go
butuzov:
```go
var (
	_ Comparator = (*ByName)(nil)
	_ Comparator = (*ByLine)(nil)
	_ Comparator = (*ByColumn)(nil)
	_ Comparator = (*ByRange)(nil)
)
```

Commenting code on github sometimes can be confusing =). This is nil allocations interface checksum that will block compilation, you can find (almost) same looking code in almost every preprocessor, the only difference  `var _ Processor = SkipFiles{}` will escape to the heap.

pkg/result/processors/sort_results.go
sayboras: This is tricky one. Will it be "<=" instead of "==" ?

I didn't scan for all Comparator implementation to make sure that negative result will _only_ be -1

pkg/result/processors/sort_results.go
sayboras: Oh, so you tried to add _ByRange_ but it's not working as expected right? The current rules are initialized as below only

```
	return &SortResults{
		cmp: ByName{
			next: ByLine{
				next: ByColumn{},
			},
		},
		cfg: cfg,
	}
```

I am ok to remove this ByRange comparator, as sorting till Line number is good enough for me though. It's best effort anyway right now, one benefit is that we can simplify numericCompare or remove this method.

butuzov: I did delete ByRange till the moment I can confirm the results of `dupl` linter are parsed correctly (or not) - that's the only case its can be useful afaik.

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,164 @@
...
+		return sr.cmp.Compare(&issues[i], &issues[j]) == Less
+	})
+
+	return issues, nil
+}
+
+func (sr SortResults) Name() string { return "sort_results" }
+func (sr SortResults) Finish()      {}
+
+type CompareResult int
```
sayboras, line 52: this could be private

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,164 @@
...
+
+const (
+	Less CompareResult = iota - 1
```
sayboras, line 55: this could be private

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,164 @@
...
+		return "Equal"
+	case Greater:
+		return "Greater"
+	}
+
+	return "None"
+}
+
+// Comparator describe how to implement compare for two "issues" lexicographically
+type Comparator interface {
```
sayboras, line 81: nit: this could be private

pkg/result/processors/sort_results.go
```diff
@@ -0,0 +1,164 @@
...
+	Next() Comparator
+}
+
+var (
+	_ Comparator = (*ByName)(nil)
+	_ Comparator = (*ByLine)(nil)
+	_ Comparator = (*ByColumn)(nil)
+)
+
+type ByName struct{ next Comparator }
```
sayboras, line 92: nit: these comparator can be changed to private

pkg/commands/run.go
```diff
@@ -80,6 +80,7 @@ func initFlagSet(fs *pflag.FlagSet, cfg *config.Config, m *lintersdb.Manager, is
 	fs.BoolVar(&oc.PrintIssuedLine, "print-issued-lines", true, wh("Print lines of code with issue"))
 	fs.BoolVar(&oc.PrintLinterName, "print-linter-name", true, wh("Print linter name in issue line"))
 	fs.BoolVar(&oc.UniqByLine, "uniq-by-line", true, wh("Make issues output unique by line"))
+	fs.BoolVar(&oc.SortResults, "sort-results", false, wh("Sort issues by line number"))
```
sayboras, line 83: can you add the test for this option? Can expand description a little bit as well.

SVilgelm: I think, @sayboras means to add the tests in the `test/run_test.go` because of the new config option

pkg/result/processors/sort_results.go
butuzov: done

pkg/result/processors/sort_results.go
butuzov: I know it's controversial, but can we keep constants at least Uppercased?

pkg/result/processors/sort_results.go
butuzov: done

pkg/result/processors/sort_results.go
butuzov: done

pkg/commands/run.go
butuzov: @SVilgelm @sayboras added, in addition, changed help text.

butuzov:
Thank everyone, that was quite a ride =)

sayboras:
> Thank everyone, that was quite a ride =)

haha hope you enjoy it :sweat_smile:. Looking forward to your next ride then :+1:

PR #1224, opened 2020-07-10
Fix deprecation warning in goreleaser

SVilgelm:
DEPRECATED: `brews.github` should not be used anymore, check https://goreleaser.com/deprecations#brewsgithub for more info.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -80,11 +80,11 @@ dockers:
       - "golangci/golangci-lint:v{{ .Major }}.{{ .Minor }}-alpine"
 
 brews:
-  - github:
+  - tap:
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -80,11 +80,11 @@ dockers:
       - "golangci/golangci-lint:v{{ .Major }}.{{ .Minor }}-alpine"
 
 brews:
-  - github:
+  - tap:
```
sayboras, line 83: LGTM.

Need to update a couple of repos personally as well :)

SVilgelm: yeah, I just did it on my repos

PR #1225, opened 2020-07-10
Add ability to specify path prefix in output

jwilner:
## problem

It's often convenient in monorepos to run a tool from a different directory but still report issues relative to the root of the repository.

This is exactly the problem in the github action https://github.com/golangci/golangci-lint-action/issues/31

## proposed solution

Add a flag for specifying a path prefix in output. Default behavior remains the same.

---

Thanks for a great tool!

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -6,6 +6,7 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path"
 	"runtime"
 	"strings"
 	"time"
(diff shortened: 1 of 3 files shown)


pkg/commands/run.go
```diff
@@ -413,6 +417,14 @@ func (e *Executor) createPrinter() (printers.Printer, error) {
 	return p, nil
 }
 
+func (e *Executor) postProcess(issues []result.Issue) {
```
sayboras, line 420: I was about to add the comment for Processor pattern to make it more composable, but then saw your #1226 

Which approach do you prefer?

SVilgelm: +1 for the #1226 
It looks more appropriate

jwilner:
@sayboras @svilgelm closed in deference to the other. Thanks y'all.

PR #1201, opened 2020-06-30
Add support for fish completion

sayboras:
Add completion support for fish shell

image

PR #1226, opened 2020-07-10
Configure path prefix via processor abstraction

jwilner:
Meant as an alternative to https://github.com/golangci/golangci-lint/pull/1225 which uses the standard processor abstraction

## problem

It's often convenient in monorepos to run a tool from a different directory but still report issues relative to the root of the repository.

This is exactly the problem in the github action https://github.com/golangci/golangci-lint-action/issues/31

## proposed solution

Add a flag for specifying a path prefix in output. Default behavior remains the same.

---

Thanks for a great tool!

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -81,6 +81,7 @@ func initFlagSet(fs *pflag.FlagSet, cfg *config.Config, m *lintersdb.Manager, is
 	fs.BoolVar(&oc.PrintLinterName, "print-linter-name", true, wh("Print linter name in issue line"))
 	fs.BoolVar(&oc.UniqByLine, "uniq-by-line", true, wh("Make issues output unique by line"))
 	fs.BoolVar(&oc.PrintWelcomeMessage, "print-welcome", false, wh("Print welcome message"))
+	fs.StringVar(&oc.PathPrefix, "path-prefix", "", wh("Path prefix to add to output"))
 	hideFlag("print-welcome") // no longer used
 
 	// Run config
--- a/pkg/lint/runner.go
+++ b/pkg/lint/runner.go
@@ -79,6 +79,7 @@ func NewRunner(cfg *config.Config, log logutils.Log, goenv *goutil.Env, es *lint
 			processors.NewSourceCode(lineCache, log.Child("source_code")),
 			processors.NewPathShortener(),
 			getSeverityRulesProcessor(&cfg.Severity, log, lineCache),
+			processors.NewPathPrefixer(cfg.Output.PathPrefix),
 		},
 		Log: log,
 	}, nil
--- a/pkg/result/processors/path_prefixer.go
+++ b/pkg/result/processors/path_prefixer.go
@@ -1,0 +1,37 @@
+package processors
+
+import (
+	"path"
+
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+// PathPrefixer adds a customizable prefix to every output path
+type PathPrefixer struct {
+	prefix string
+}
+
+var _ Processor = new(PathPrefixer)
+
+// NewPathPrefixer returns a new path prefixer for the provided string
+func NewPathPrefixer(prefix string) *PathPrefixer {
+	return &PathPrefixer{prefix: prefix}
+}
+
+// Name returns the name of this processor
+func (*PathPrefixer) Name() string {
+	return "path_prefixer"
+}
+
+// Process adds the prefix to each path
+func (p *PathPrefixer) Process(issues []result.Issue) ([]result.Issue, error) {
+	if p.prefix != "" {
+		for i := range issues {
+			issues[i].Pos.Filename = path.Join(p.prefix, issues[i].Pos.Filename)
+		}
+	}
+	return issues, nil
+}
+
+// Finish is implemented to satisfy the Processor interface
+func (*PathPrefixer) Finish() {}
--- a/pkg/result/processors/path_prefixer_test.go
+++ b/pkg/result/processors/path_prefixer_test.go
@@ -1,0 +1,37 @@
+package processors
+
+import (
+	"go/token"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+func TestPathPrefixer_Process(t *testing.T) {
+	paths := func(ps ...string) (issues []result.Issue) {
+		for _, p := range ps {
+			issues = append(issues, result.Issue{Pos: token.Position{Filename: p}})
+		}
+		return
+	}
+	for _, tt := range []struct {
+		name, prefix string
+		issues, want []result.Issue
+	}{
+		{"empty prefix", "", paths("some/path", "cool"), paths("some/path", "cool")},
+		{"prefix", "ok", paths("some/path", "cool"), paths("ok/some/path", "ok/cool")},
+		{"prefix slashed", "ok/", paths("some/path", "cool"), paths("ok/some/path", "ok/cool")},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			r := require.New(t)
+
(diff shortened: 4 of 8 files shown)


pkg/result/processors/path_prefixer.go
```diff
@@ -0,0 +1,37 @@
...
+	return &PathPrefixer{prefix: prefix}
+}
+
+// Name returns the name of this processor
+func (*PathPrefixer) Name() string {
+	return "path_prefixer"
+}
+
+// Process adds the prefix to each path
+func (p *PathPrefixer) Process(issues []result.Issue) ([]result.Issue, error) {
```
sayboras, line 27: can you add the test for this method? Shout out if you need any help.

pkg/lint/runner.go
```diff
@@ -79,6 +79,7 @@ func NewRunner(cfg *config.Config, log logutils.Log, goenv *goutil.Env, es *lint
 			processors.NewSourceCode(lineCache, log.Child("source_code")),
 			processors.NewPathShortener(),
 			getSeverityRulesProcessor(&cfg.Severity, log, lineCache),
+			processors.NewPathPrefixer(cfg.Output.PathPrefix),
```
sayboras, line 82: Should this processor before or after NewPathShortener processor ?

pkg/result/processors/path_prefixer.go
jwilner: Added a test

pkg/commands/run.go
```diff
@@ -81,6 +81,7 @@ func initFlagSet(fs *pflag.FlagSet, cfg *config.Config, m *lintersdb.Manager, is
 	fs.BoolVar(&oc.PrintLinterName, "print-linter-name", true, wh("Print linter name in issue line"))
 	fs.BoolVar(&oc.UniqByLine, "uniq-by-line", true, wh("Make issues output unique by line"))
 	fs.BoolVar(&oc.PrintWelcomeMessage, "print-welcome", false, wh("Print welcome message"))
+	fs.StringVar(&oc.PathPrefix, "path-prefix", "", wh("Path prefix to add to output"))
```
sayboras, line 84: will it require to have / at the end for this (e.g. /my/prefix vs /my/prefix/) ?

pkg/lint/runner.go
jwilner: I don't believe so. The prefix is essentially a user override and user overrides should be preferred over default behavior like the shortener.

Put another way -- if the shortener comes after the prefixer and it removes the prefix or puts something in front of the prefix, it will have counteracted the user's express wishes.

pkg/commands/run.go
jwilner: No, the desired behavior will always be for this to be a directory and it uses path joining.

jwilner:
Will squash merge away noise

SVilgelm:
> Will squash merge away noise

no worries, it's only one allowed option to merge any pull requests here.

SVilgelm:
Since it adds a new config option, could you also add a test in the `test/run_test.go`?

pkg/lint/runner.go
sayboras: oh right, thanks

pkg/result/processors/path_prefixer.go
```diff
@@ -0,0 +1,37 @@
...
+	if p.prefix != "" {
```
sayboras, line 28: The current logic seems fine to me. 

Just a note on immutability, I just take a look at existing implementation of processors. Seems like https://github.com/golangci/golangci-lint/blob/master/pkg/result/processors/utils.go#L40 is widely used (this function create a copy of slides).

What do you think ?

pkg/result/processors/path_prefixer_test.go
```diff
@@ -0,0 +1,37 @@
...
+		issues, want []result.Issue
+	}{
+		{"empty prefix", "", paths("some/path", "cool"), paths("some/path", "cool")},
+		{"prefix", "ok", paths("some/path", "cool"), paths("ok/some/path", "ok/cool")},
+		{"prefix slashed", "ok/", paths("some/path", "cool"), paths("ok/some/path", "ok/cool")},
+	} {
+		t.Run(tt.name, func(t *testing.T) {
+			r := require.New(t)
+
+			p := NewPathPrefixer(tt.prefix)  //nolint:scopelint
```
sayboras, line 30: My goland auto generated tabular test is exactly the same here, seems like this was discussed here https://github.com/kyoh86/scopelint/issues/4

We can just skip this scopelint here.

pkg/result/processors/path_prefixer.go
jwilner: Well, it's safe for this to be mutating the passed in slice because, by definition, the processors happen serially, so there's no risk of a race on the mutated memory.

I would guess the transformSlices function exists because it's easy to mess up working with a slice of values -- every value in a range loop is a copy, etc, etc.

As it is, I figured this is a memory-sensitive application and the Issue struct is non-trivial in its size, so I'd avoid the unnecessary copies and alloc.

Without running any sort of measurements, I'm guessing there's a lot of fat to trim around this processor chain's memory usage.

pkg/result/processors/path_prefixer_test.go
jwilner: Yeah; the closure is only a problem if it's executed after the iteration of the loop, which it's not -- but it'd be very tough for the scopelint tool to recognize that closures are executing more or less immediately here.

As it is, people passing closures out of loops and messing up the bindings is such a common toe stub, it makes sense to me to have a CI tool that will yell, even if it is with a lot of false positives.

jwilner:
> > Will squash merge away noise
> 
> no worries, it's only one allowed option to merge any pull requests here.

The only civilized option! :)

jwilner:
@SVilgelm added to `run_test.go` per your request.

I'm impressed with this project's test coverage. 👍

jwilner:
🥳 looking forward to using it :)

jwilner:
Oh I don't have the ability to merge btw, if that was ambiguous.

SVilgelm:
Let's wait a feedback from @sayboras, he also revised this PR

sayboras:
It's wierd that test failed in master https://github.com/golangci/golangci-lint/runs/859811449 and another one https://github.com/golangci/golangci-lint/pull/1201

@jwilner Do you have any idea or pointer ?

jwilner:
@sayboras whoa, huh, one of the tests I added!? I'd be happy to look -- but I can't tell which test it is from the output. Can you point me in the right direction?

SVilgelm:
@jwilner Here it is: https://github.com/golangci/golangci-lint/runs/859880992?check_suite_focus=true#step:5:881

jwilner:
So something about the test runner is failing on the command

`../golangci-lint run --allow-parallel-runners --path-prefix=cool testdata/withTests`

Isn't the test dir `testdata/withTests` used within other tests? Hmmm. Digging deeper.

```
=== RUN   TestPathPrefix/prefixed
level=info msg="[test] stderr: "
level=info msg="[test] ran [../golangci-lint run --allow-parallel-runners --path-prefix=cool testdata/withTests] in 358.830342ms"
    TestPathPrefix/prefixed: testshared.go:71: 
        	Error Trace:	testshared.go:71
        	            				run_test.go:308
        	Error:      	Expect "level=error msg="Running error: context loading failed: failed to load packages: cannot find package \".\" in:\n\t/home/runner/work/golangci-lint/golangci-lint/test/testdata/withTests: failed to analyze"
        	            	" to match "^cool/testdata/withTests"
        	Test:       	TestPathPrefix/prefixed
        	Messages:   	exit code is 3
```

jwilner:
@SVilgelm @sayboras I've opened up a fix here if you want to check it out: https://github.com/golangci/golangci-lint/pull/1229

SVilgelm:
@jwilner thank you, I merged it to unblock the CI

PR #1227, opened 2020-07-10
Update contributing workflow document

sonatard:
Updated doc with reference to v1.28 release.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -106,6 +106,8 @@ linters:
     - varcheck
     - whitespace
 
+  # Enable in next release
+
   # don't enable:
   # - asciicheck
   # - gochecknoglobals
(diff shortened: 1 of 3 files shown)


iwankgb:
In my opinion it requires more then PR author decision to enable a linter for this project. @golangci/team, can we have another opinion(s) here?

bombsimon:
I agree. I think we should be sparse with the enabled list and only enable common, well tested linters which isn't considered opinionated and encourage the users to use their own configuration file for desired linters. I think all new linters should be disabled and leave the decision to the core team to enable new linters. I guess changing default enabled linters could be seen as a breaking change which is a bit of a hassle to deal with.

sayboras:
Agreed with above points. Just want to share my experience, the current default enabled linters for me now seems a little too much for my normal usage as well. Most of the time, I need to disable _structcheck_ and _errcheck_. On the other side, I added gofmt, goimports into enabled config.

Another point I think we can do is to scan all linters supported by now, and deprecate those linters which are no longer maintainer, or already covered by another more active one.

sonatard:
It's better.

I'll fix document in this PR.

1. Update new-linters.mdx for new linter creator.
Creator can't enable your linter to lint golangci-lint codebase.

2. Update contributing/workflow.mdx 
Releaser decide to choose enable linter from new linters.

3. Created new issue #1230 - Which linter to enable for golangci-lint codebase

SVilgelm:
@sonatard is this ready to be merged or you want to add something else?
If it still in progress, let's add `WIP` prefix in the title, to block the PR from merging

PR #1234, opened 2020-07-11
Update assets and README

SVilgelm:
```shell
make assets/github-action-config.json
make update_contributors_list
make update_netlify_state
```

Added `update_netlify_state ` target in Makefile to run the `go run ./scripts/expand_website_templates/main.go -only-state` command. It updates the state, that triggers Netlify build


docs/src/docs/usage/install/index.mdx
```diff
@@ -24,17 +24,17 @@ use deprecated option `--enable-all` and a new linter is added or even without `
 
 It's highly recommended to install a specific version of golangci-lint available on the [releases page](https://github.com/golangci/golangci-lint/releases).
 
-Here is the recommended way to install golangci-lint {.LatestVersion}:
```
sayboras, line 27: I am not so sure if we are supposed to change these files. As I think these are templates. Might work for _this commit_ only, but the next run will not change anything right ?

docs/template_data.state
```diff
@@ -1,2 +1,2 @@
 This file stores hash of website templates to trigger Netlify rebuild when something changes, e.g. new linter is added.
-4896b58a5fb540cfca8bbcec4ef437b426bac0cbb1a4bd8dda3fe3a2a60f0c29
\ No newline at end of file
+5e1b1118687c421b50605b619e5497b4058abe0991eb345f9b450f3e1de76163
```
sayboras, line 2: I think netlify will check this hash and re-build the website, but not 100% sure it will be triggered automatically.

SVilgelm: Netlify triggers rebuild only if anything has changes in `docs/` folder: https://golangci-lint.run/contributing/website/#trigger-website-rebuild

sayboras: yeah, I saw this condition as well. But seems like latest version is not available in public website

docs/src/docs/usage/install/index.mdx
SVilgelm: I reverted this changes, otherwise the `Netlify` fails

Makefile
```diff
@@ -109,4 +109,7 @@ expand_website_templates:
 .PHONY: expand_website_templates
 
 update_contributors_list:
-	cd .github/contributors && npm run all
\ No newline at end of file
+	cd .github/contributors && npm run all
+
+trigger-netlify:
```
sayboras, line 114: nits: trigger-netlify might imply that running this make target will kick off netlify  deployment.

How about just _update_netlify_state_ or _updaste_netlify_hash?

docs/src/docs/usage/install/index.mdx
SVilgelm: I run ` make expand_website_templates` and it changed the templates.
The problem here is that the Netlify build the documentation with old version, but I think this happens because the assets is not merged yet. Let merge this PR and check what will happen :-)

Makefile
SVilgelm: I like `update_netlify_state`. let me change it

SVilgelm: Done

PR #1235, opened 2020-07-12
fix incompatible call for gofumpt

slixurd:
Couple days ago, there is incompatible change method declaration for format.Source()

Check this commit for the change
https://github.com/mvdan/gofumpt/commit/42bb51311e31092b352bf753ea246ec904c6d4d5

Simply run `go get -u github.com/golangci/golangci-lint/cmd/golangci-lint` will fail 

`golangci-lint@v1.28.3/pkg/golinters/gofumpt.go:49:41: cannot use "" (type string) as type format.Options in argument to format.Source`

### DIFF
--- a/pkg/golinters/gofumpt.go
+++ b/pkg/golinters/gofumpt.go
@@ -46,7 +46,7 @@ func NewGofumpt() *goanalysis.Linter {
 				if err != nil {
 					return nil, fmt.Errorf("unable to open file %s: %w", f, err)
 				}
-				output, err := format.Source(input, "")
+				output, err := format.Source(input, format.Options{})
 				if err != nil {
 					return nil, fmt.Errorf("error while running gofumpt: %w", err)
 				}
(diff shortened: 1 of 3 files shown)


pkg/golinters/gofumpt.go
```diff
@@ -46,7 +46,7 @@ func NewGofumpt() *goanalysis.Linter {
 				if err != nil {
 					return nil, fmt.Errorf("unable to open file %s: %w", f, err)
 				}
-				output, err := format.Source(input, "")
+				output, err := format.Source(input, format.Options{})
```
sayboras, line 49: `go get`  with `-u` will internally update dependencies if I am not wrong. Additionally, `go get` is _not_ recommended way to install golangci-lint as per https://golangci-lint.run/usage/install/

> Note: such go get installation aren't guaranteed to work. We recommend using binary installation.

Without changing version in go.mod, your PR will cause compilation error.

SVilgelm:
This bug is already fixed in #1128
Closing as a duplicate

PR #1239, opened 2020-07-13
Support --fix for gofumpt

SVilgelm:
Add autofix functionality (`-w`)
Add fix and linter tests.
Closes: #490

Since the `gofumpt/format.Source` returns already formatted body of a file, but `golangci-lint` expects a diff to support auto-fixing, so I use `github.com/shazow/go-diff/difflib` library to make a diff of input and output. I expect, that it can be slower than `gofmt` linter.

### DIFF
--- a/pkg/golinters/gofmt_common.go
+++ b/pkg/golinters/gofmt_common.go
@@ -207,7 +207,29 @@ func (p *hunkChangesParser) parse(h *diffpkg.Hunk) []Change {
 	return p.ret
 }
 
-func extractIssuesFromPatch(patch string, log logutils.Log, lintCtx *linter.Context, isGoimports bool) ([]result.Issue, error) {
+func getErrorTextForLinter(lintCtx *linter.Context, linterName string) string {
+	text := "File is not formatted"
+	switch linterName {
+	case gofumptName:
+		text = "File is not `gofumpt`-ed"
+		if lintCtx.Settings().Gofumpt.ExtraRules {
+			text += " with `-extra`"
+		}
+	case gofmtName:
+		text = "File is not `gofmt`-ed"
+		if lintCtx.Settings().Gofmt.Simplify {
+			text += " with `-s`"
+		}
+	case goimportsName:
+		text = "File is not `goimports`-ed"
+		if lintCtx.Settings().Goimports.LocalPrefixes != "" {
+			text += " with -local " + lintCtx.Settings().Goimports.LocalPrefixes
+		}
+	}
+	return text
+}
+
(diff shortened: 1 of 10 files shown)


andrewrynhard:
`gofumpt` also has `gofumports`. Any way we can sneak that in here?

SVilgelm:
> `gofumpt` also has `gofumports`. Any way we can sneak that in here?

I created an issue to add it as a new linter

SVilgelm:
@golangci/team Please review this

pkg/golinters/gofmt_common.go
```diff
@@ -207,7 +207,29 @@ func (p *hunkChangesParser) parse(h *diffpkg.Hunk) []Change {
 	return p.ret
 }
 
-func extractIssuesFromPatch(patch string, log logutils.Log, lintCtx *linter.Context, isGoimports bool) ([]result.Issue, error) {
+func getErrorTextForLinter(lintCtx *linter.Context, linterName string) string {
+	text := "File is not formatted"
+	switch linterName {
+	case gofumptName:
+		text = "File is not `gofumpt`-ed"
```
daixiang0, line 214: "File is not `gofumpt` -ed"

daixiang0: Same below

SVilgelm: it is same style as it was: https://github.com/golangci/golangci-lint/pull/1239/files/79726260904207cbef41c72458a1d8d415d47f80#diff-4d48c424c434d446b09abeaa31ab7e91L230

Let's keep it in this way, we cannot change the error messages for the `gofmt` and `goimports` because it's backward incompatible changes. Maybe we can do it in `v2`, but not now

SVilgelm:
@golangci/team Could anyone please review it?

SVilgelm:
@golangci/team Could anyone review it? Approve?
@sayboras @tetafro @daixiang0 ???

arvenil:
This new linter broke `goimports` sorting with `-local` and `--fix`. I first thought it's `gci` linter as it seems to do similar thing as `goimports`, but in fact gci doesn't sort `local-prefixes` at all (broken?). I've disabled `gofumpt` and now `goimports` works correctly. I think it's a general issue that there is no single well working import sorter and some linters support partial import sorting and then they fight with other linters for which one has the final word on the order. I've noticed that every run I got different order of packages. Just FYI.

daixiang0:
`gci` foucs on import group by setting `local-prefixes`, it is expected to be run after `goimports`.

PR #1242, opened 2020-07-14
Move dependabot config to .github folder

SVilgelm:
Replace existent dependabot config with v2 and move it to the `.github` folder
Enable version updates for GitHub actions


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
+version: 2
```
sayboras, line 1: awesome

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+updates:
+  - package-ecosystem: gomod
```
sayboras, line 3: nit: you can add _assignees_ so that new PR will notify the team if you want.

.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    directory: "/"
+    schedule:
+      interval: weekly
+    ignore:
+      # ignore forked linters because of their versioning issues. TODO: fix it.
+      - dependency-name: "github.com/golangci/*"
+  - package-ecosystem: github-actions
```
sayboras, line 10: so great to have this one to update github-actions as well.

I just take a look at the all supported package-ecosystem. Docker might be another one we can enable.

.github/dependabot.yml
SVilgelm: probably reviewers

.github/dependabot.yml
SVilgelm: let me check it

SVilgelm: done

PR #1243, opened 2020-07-14
Run nancy validation for all dependencies

SVilgelm:


.github/workflows/pr.yml
```diff
@@ -15,6 +15,11 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
```
sayboras, line 18: should I be in another job as this job is focusing on golangci-lint ? I think golangci-lint and nancy are two independent tasks, they can be running in parallel as well :)

.github/workflows/pr.yml
```diff
@@ -15,6 +15,11 @@ jobs:
...
+      # We cannot use nancy-github-action because it is outdated, so it's better to use the latest
+      # docker image for the validation
+      - name: nancy
+        run: go list -m all | docker run -i sonatypecommunity/nancy:latest
```
sayboras, line 22: We might need to test it out for negative case to make sure it failed here. I faced a few cases github action didn't fail as expected, so just a precautious step.

SVilgelm: I'm using it in several projects, an at work and it's failing :)

SVilgelm: recently I had to disable it on a weekend because the server was unavailable

.github/workflows/pr.yml
SVilgelm: ok, let me move it

.github/workflows/pr.yml
sayboras: oh I see, we can just let it run as non-required job for sometime first, then change settings later once we see fit.

SVilgelm: then I need to move it to a separate file

SVilgelm: like `extra`

SVilgelm: done

SVilgelm:
> /lgtm

then approve? ;-)

.github/workflows/pr-extra.yml
```diff
@@ -0,0 +1,19 @@
...
+  vulns:
+    name: Vulnerability scanner
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+      # We cannot use nancy-github-action because it is outdated, so it's better to use the latest
+      # docker image for the validation
+      - name: nancy
+        run: go list -m all | docker run -i sonatypecommunity/nancy:latest
```
iwankgb, line 19: You should rather use:
```
go list -json -m all | nancy
```
(see [deprecation notice](https://github.com/sonatype-nexus-community/nancy#what-is-the-best-usage-of-nancy))

.github/workflows/pr.yml
sayboras: oh actually, you can still have it in the same pr workflow, mandatory setting is per `job`. 

Anyway, no harm to keep it in another file.

.github/workflows/pr-extra.yml
SVilgelm: thanks I didn't know

SVilgelm: done

PR #1244, opened 2020-07-14
Improve 'no such linter' error message

denis-tingajkin:
Signed-off-by: denis-tingajkin <denis.tingajkin@xored.com>

## Steps to reproduce

1. Use new golangci-linter config file
2. Use old linter version (for example 1.26)

**Actual**: Printed err msg 'no such linter %v' for all new linters which was added since 1.26. The error message looks is not clear for new users. The message can confuse the user like 'Should I download missed linter?'. 

I think we need to point that the user can run  `golangci-lint linters` to see supported linter list of his `golangci-linter` tool.

### DIFF
--- a/pkg/lint/lintersdb/validator.go
+++ b/pkg/lint/lintersdb/validator.go
@@ -17,7 +17,7 @@ func NewValidator(m *Manager) *Validator {
 	}
 }
 
-func (v *Validator) validateLinterConfigNames(cfg *config.Linters) error {
+func (v Validator) validateLintersNames(cfg *config.Linters) error {
 	allNames := append([]string{}, cfg.Enable...)
 	allNames = append(allNames, cfg.Disable...)
 	for _, name := range allNames {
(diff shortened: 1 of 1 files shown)


denis-tingajkin:
@iwankgb actually here is no difference at this moment. But if for Validator struct will be added some struct fields then pointer receivers will be better :)

iwankgb:
We should not anticipate future code changes, I believe. Would you be OK with reverting to value receivers?

pkg/lint/lintersdb/validator.go
```diff
@@ -17,19 +17,19 @@ func NewValidator(m *Manager) *Validator {
 	}
 }
 
-func (v Validator) validateLintersNames(cfg *config.Linters) error {
+func (v *Validator) validateLinterConfigNames(cfg *config.Linters) error {
```
sayboras, line 20: agreed that changing to pointer is not related to the scope of this PR.

I don't think it's required as well.

sayboras, line 20: just curious why the method got changed to `validateLinterConfigNames` ? 

AFAIK this validation is applicable for `golangci-lint run -E 121231` as well.

denis-tingajkin: ok, fixed.

denis-tingajkin: ok, fixed.

Issue #1257, opened 2020-07-22
add new linter to upport repairing import grouping/ordering


daixiang0:
**Is your feature request related to a problem? Please describe.**

See this [discuss](https://github.com/golang/go/issues/20818#issue-238956416).

**Describe the solution you'd like**

Introduce [gci](https://github.com/daixiang0/gci) as a new linter.

**Additional context**

Golang community has little inseret about imporing it, but it would be useful for both contributors and maintainers.

Now gci implement basic functions，i think add it as new linter would be very cool.

WDYT? @golangci/team

SVilgelm:
Looks like the gci is a very useful linter, but 
1. it should support go/analysis interface
2. the gci implementation contains only `main` package that prevents the other applications to import that tool

daixiang0:
I will refactor it.

daixiang0:
Done.

PR #1266, opened 2020-07-24
Introduce gci as new linter

daixiang0:
Related to #1257 

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -109,6 +109,10 @@ linters-settings:
   funlen:
     lines: 60
     statements: 40
+  gci:
+    # put imports beginning with prefix after 3rd-party packages;
+    # only support one prefix
+    local-prefixes: github.com/org/project
   gocognit:
     # minimal code complexity to report, 30 by default (but we recommend 10-20)
     min-complexity: 10
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -112,6 +112,7 @@ linters-settings:
   gci:
     # put imports beginning with prefix after 3rd-party packages;
     # only support one prefix
+    # if not set, use goimports.local-prefixes
     local-prefixes: github.com/org/project
   gocognit:
     # minimal code complexity to report, 30 by default (but we recommend 10-20)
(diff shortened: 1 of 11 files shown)


daixiang0:
please review @golangci/core-team

.golangci.example.yml
```diff
@@ -109,6 +109,10 @@ linters-settings:
   funlen:
     lines: 60
     statements: 40
+  gci:
+    # put imports beginning with prefix after 3rd-party packages;
+    # only support one prefix
+    local-prefixes: github.com/org/project
```
ernado, line 115: Is it intended? Probably singular form `local-prefix` can be better here, but it will be inconsistent with goimports.

Also it would be great to pick prefix automatically from goimports settings, but seems like it is out of scope.

test/testdata/gci.go
```diff
@@ -0,0 +1,17 @@
+//args: -Egoimports
```
sayboras, line 1: is it goimports ?

test/testdata/gci.go
```diff
@@ -0,0 +1,17 @@
...
+//config: linters-settings.goimports.local-prefixes=github.com/golangci/golangci-lint
+package goimports
+
+import (
```
sayboras, line 5: will be great if we can have negative test case as well

test/testdata/gci.go
daixiang0: Done.

.golangci.example.yml
daixiang0: Yep, it is intended, since i want to support multi-prefixes in feature, so keep same with `goimports` here.

`gci` handles empty lines between import groups and re-split them into groups aby setting, `goimports` can not do that.

test/testdata/gci.go
daixiang0: I see dirs under `test/testdata` while for errcheck it has some test data files, so i can do like errcheck does?

.golangci.example.yml
ernado: > how about use goimports settings if no settings for gci?

I like this

PR #1267, opened 2020-07-24
new nlreturn linter

ssgreg:
Add nlreturn linter.
https://github.com/ssgreg/nlreturn


julienschmidt:
Does this differ from https://github.com/bombsimon/wsl or is it simply a subset of the rules enforced by wsl?

ssgreg:
> Does this differ from https://github.com/bombsimon/wsl or is it simply a subset of the rules enforced by wsl?
WSL is too strict linter. We tried to add more configurable options but author did not accept our changes and suggested to make an alternative implementation saying that WSL was started as a joke.
https://github.com/bombsimon/wsl/pull/89

We think that empty lines are really matter before return and branch statements. They greatly improve code readability. Besides that we're open to add any other configurable WSL like checks if anyone need them.

bombsimon:
I think it's fine to have overlapping linters in `golangci-lint` so I see no reason not to merge this. It's a huge difference between WSL and this linter just like the difference between WSL and [whitespace](https://github.com/ultraware/whitespace) which is also overlapping. The same goes for [gofumpt](https://github.com/mvdan/gofumpt) regarding empty lines, grouping `var` etc.

I just wanted to say that I very much appreciate changes and the PR never got merge since I never got any response to my questions or proposals on how to use keep development of WSL. And to be clear, it's no longer a joke but something I want to be useful. I just don't think configuring a linter to turn of all checks is the right way to go so again, I think this will be a good contribution to `golangci-lint`.

I'll approve this but I will let someone else from @golangci/team merge this so we agree with each other.

SVilgelm:
IMO if it's not possible to achieve same result with other linters, then I don't see any reasons to not merge it

SVilgelm:
@ssgreg Could please resolve conflicts and we will merge it

PR #1279, opened 2020-08-02
Add tests-on-macos

SVilgelm:
IMO we need to run the tests on macOS in same way we do for windows, so I added tests-on-macos as a parallel job with the go v1.14, so only ubuntu job will use different go version for testing.


SVilgelm:
@golangci/core-team @jirfag @ ernado Could you please review this and if you are OK, then please modify the branch protection

SVilgelm:
@ernado Could you make same changes here for the branch protection?

ernado:
I've updated protected branch rules.

PR #1315, opened 2020-08-18
Support short and json formats for version cmd

### LINKED ISSUE #1314: Provide machine-parseable version output

Automatically checking the available tool version is hacky and fragile, as seen in the examples below. This is because the output of `golangci-lint --version` is mean for human rather than machine consumption.

Something easy to parse in a shell script. The piece that would make us happy would be:

```
$ golangci-lint --version-string
v1.777.0
```

If other pieces are also needed, one could also imagine:
```
$ golangci-lint --version-mod-sum
h1:HxAxpR8Z0M8omihvQdsD3PF0qPjlqYqp2vMJzstoKeI=
```

One could also imagine `--version --json` that would produce structured output, but that is not so easy to use in a shell script without additional tools (`jq`).

The output for `--version` could be changed without introducing another flag, but that would break backwards compatibility.

**Additional context**

Some projects try to ensure that the version of `golangci-lint` they use is consistent for all developers ([example](https://github.com/kudobuilder/kudo/blob/0be3e0d0122e90b397cd1b32e1b547f5774e2aa7/Makefile#L80), [another](https://github.com/kudobuilder/kuttl/blob/cbc4eba01d813b954b362f845f75b1fb817e4195/Makefile#L34)).

SVilgelm:
Closes #1314 

Support printing the version I different formats:
* short: prints only version
* json: prints in json format

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -234,6 +234,7 @@ func (e *Executor) getConfigForCommandLine() (*config.Config, error) {
 	// Use another config variable here, not e.cfg, to not
 	// affect main parsing by this parsing of only config option.
 	initFlagSet(fs, &cfg, e.DBManager, false)
+	initVersionFlagSet(fs, &cfg)
 
 	// Parse max options, even force version option: don't want
 	// to get access to Executor here: it's error-prone to use
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -521,6 +521,10 @@ type Severity struct {
 	Rules         []SeverityRule `mapstructure:"rules"`
 }
 
+type Version struct {
+	Format string `mapstructure:"format"`
+}
+
 type Config struct {
 	Run Run
 
@@ -539,6 +543,7 @@ type Config struct {
 	Linters         Linters
 	Issues          Issues
 	Severity        Severity
+	Version         Version
 
 	InternalTest bool // Option is used only for testing golangci-lint code, don't use it
 }
--- a/pkg/commands/version.go
+++ b/pkg/commands/version.go
@@ -1,17 +1,59 @@
 package commands
 
 import (
+	"encoding/json"
+	"strings"
+
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+
+	"github.com/golangci/golangci-lint/pkg/config"
 )
+
+type jsonVersion struct {
+	Version string `json:"version"`
+	Commit  string `json:"commit"`
+	Date    string `json:"date"`
+}
+
+func (e *Executor) initVersionConfiguration(cmd *cobra.Command) {
+	fs := cmd.Flags()
+	fs.SortFlags = false // sort them as they are defined here
+	initVersionFlagSet(fs, e.cfg)
+}
+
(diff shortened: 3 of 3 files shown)


SVilgelm:
`Nancy` fails because of `etcd`, it's unrelated to this PR

pkg/commands/run.go
```diff
@@ -234,6 +234,7 @@ func (e *Executor) getConfigForCommandLine() (*config.Config, error) {
 	// Use another config variable here, not e.cfg, to not
 	// affect main parsing by this parsing of only config option.
 	initFlagSet(fs, &cfg, e.DBManager, false)
+	initVersionFlagSet(fs, &cfg)
```
sayboras, line 237: can you explain rationale behind this line ?

pkg/config/config.go
```diff
@@ -539,6 +543,7 @@ type Config struct {
 	Linters         Linters
 	Issues          Issues
 	Severity        Severity
+	Version         Version
```
sayboras, line 546: so this one will come in .golangci.yaml file as well ?

pkg/commands/run.go
SVilgelm: Sure. We need to refactor this code. Now the executer uses same set of the flags for all commands, and when I defined the flag for version command only it's failed with an error with unknown `--format` arg, here: https://github.com/golangci/golangci-lint/blob/f7417e6ff048c50b89364a074881a0dddec5a5ac/pkg/commands/run.go#L244

This is kind of workaround, I think we need to have a hard refactoring of all this code and do not mix the flags for `run` command with other command, ideally every command must have independent sets of flags

pkg/config/config.go
SVilgelm: yes, it does.
```shell
20-08-18 10:28 % ./golangci-lint version
golangci-lint has version (devel) built from (unknown, mod sum: "") on (unknown)


20-08-18 10:28 % git diff
diff --git a/.golangci.yml b/.golangci.yml
index 5e9e1e7..fa899f6 100644
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -146,3 +146,6 @@ service:
   golangci-lint-version: 1.23.x # use the fixed version to not introduce new linters unexpectedly
   prepare:
     - echo "here I can run custom commands, but no preparation needed for this repo"
+
+version:
+  format: json
20-08-18 10:28 % ./golangci-lint version
{"version":"(devel)","commit":"(unknown, mod sum: \"\")","date":"(unknown)"}
```

pkg/commands/run.go
sayboras: Oh I see :+1:

pkg/config/config.go
sayboras: My original concern was that if this should be part of configuration file. Then, I think it's not much a big deal as long as we don't break declarative understanding of config file (e.g. it will be very weird to specify actual version in config file).

sayboras (approved):

pkg/config/config.go
SVilgelm: oh no, we don't specify a version in the config file, just the output format

PR #1337, opened 2020-08-28
Gci: update the way to call Run()

daixiang0:
Fix [issue](https://github.com/daixiang0/gci/issues/8).

It is helpful to run in a small os without `diff` cmd.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,7 +6,7 @@ require (
 	github.com/Djarvur/go-err113 v0.0.0-20200511133814-5174e21577d5
 	github.com/OpenPeeDeeP/depguard v1.0.1
 	github.com/bombsimon/wsl/v3 v3.1.0
-	github.com/daixiang0/gci v0.2.2
+	github.com/daixiang0/gci v0.2.4
 	github.com/denis-tingajkin/go-header v0.3.1
 	github.com/fatih/color v1.9.0
 	github.com/go-critic/go-critic v0.5.2
--- a/pkg/golinters/gci.go
+++ b/pkg/golinters/gci.go
@@ -1,10 +1,13 @@
 package golinters
 
 import (
+	"bytes"
+	"fmt"
 	"sync"
 
 	"github.com/daixiang0/gci/pkg/gci"
 	"github.com/pkg/errors"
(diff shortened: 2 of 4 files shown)


pkg/golinters/gci.go
```diff
@@ -43,13 +47,17 @@ func NewGci() *goanalysis.Linter {
 			var issues []goanalysis.Issue
 
 			for _, f := range fileNames {
-				diff, err := gci.Run(f, &gci.FlagSet{LocalFlag: localFlag})
+				source, result, err := gci.Run(f, &gci.FlagSet{LocalFlag: localFlag})
 				if err != nil {
 					return nil, err
 				}
-				if diff == nil {
+				if bytes.Equal(source, result) {
```
SVilgelm, line 54: It's not necessary, since you are doing it here: https://github.com/daixiang0/gci/pull/9/files#diff-1554c40a9d7ee0dfba770905afabc360R362

SVilgelm:
```suggestion
				if source == nil || result == nil {
```

go.mod
```diff
@@ -6,7 +6,7 @@ require (
 	github.com/Djarvur/go-err113 v0.0.0-20200511133814-5174e21577d5
 	github.com/OpenPeeDeeP/depguard v1.0.1
 	github.com/bombsimon/wsl/v3 v3.1.0
-	github.com/daixiang0/gci v0.2.2
+	github.com/daixiang0/gci v0.2.3
```
SVilgelm, line 9: Could you run `go mod tidy`? I believe that it should also update the `go.sum` file

daixiang0:
Wait for v0.2.4 since v0.2.3 has checksum mismatch issue.

ashokponkumar:
@daixiang0 @SVilgelm  Since go seems to automatically go for the latest minor version, even though go.sum is pointing to 0.2.2 this seems to still break golangcli-lint. Can we have the 0.2.4 released soon and merged to prevents builds from breaking while using golangcli-lint? Thanks!

daixiang0:
@ashokponkumar you can use release binary instead.

Now v0.2.4 releases, we can continue.

PR #1364, opened 2020-09-13
Added `.golangci.yaml` to list of configuration files searched

magnetikonline:
Just a very minor documentation change.

Noted that `.golangci.yaml` is also searched for on startup as a config file - which is great, as I prefer to use `.yaml` over `.yml` as a file extension - as suggested here: https://yaml.org/faq.html


SVilgelm (approved):
Looks good to me. Thank you

magnetikonline:
Thanks @SVilgelm 👍

sayboras (approved):
LGTM :100:

PR #1376, opened 2020-09-19
chore(dep): Update unparam/gofumpt to latest commit hash

### LINKED ISSUE #1375: Update unparam to the lastest version
Hi. I have a project and when I run golangci-lint It prints errors from [unparam](https://github.com/mvdan/unparam).
I consider them as false positive. Here is the code of my project: https://github.com/georgysavva/news-app/tree/unparam-false-positive
And this is the output of golangci-lint:
```
golangci-lint run
pkg/inmemory/inmemory.go:18:83: (*Storage).ReplaceArticles - result 0 (error) is always nil (unparam)
func (s *Storage) ReplaceArticles(_ context.Context, articles []*article.Article) error {
                                                                                  ^
pkg/inmemory/inmemory.go:37:103: (*Storage).GetArticles - result 1 (error) is always nil (unparam)
func (s *Storage) GetArticles(_ context.Context, categories, providers []string) ([]*article.Article, error) {
```
When I run the latest version of `unparam` directly, it doesn't print any error. I think golangci-lint should update its `unparam` version to the latest one to eliminate those false positives. 
BTW I am using the latest golangci-lint ` v1.31.0`

sayboras:
This PR is to update the unparam/gofumpt to latest hash
    
Fixes #1375
    
PS: Another reason to use tag version when possible as dependent bot might not be able to handle hash update.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -61,7 +61,7 @@ require (
 	golang.org/x/tools v0.0.0-20200918232735-d647fc253266
 	gopkg.in/yaml.v2 v2.3.0
 	honnef.co/go/tools v0.0.1-2020.1.5
-	mvdan.cc/gofumpt v0.0.0-20200709182408-4fd085cb6d5f
+	mvdan.cc/gofumpt v0.0.0-20200802201014-ab5a8192947d
 	mvdan.cc/interfacer v0.0.0-20180901003855-c20040233aed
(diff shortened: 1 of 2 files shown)


SVilgelm (approved):
Looks good to me. Thank you

PR #1378, opened 2020-09-20
build(dep): Ignore known dependency failure in nancy

sayboras:
Currently nancy is always failed, and we seem to ignore it completely.
This reduces the value of having security scanning significantly.
Ideally, the underlying issue should be fixed, however it will require
long time for external collaboration.

This commit is to ignore two known dependency failures.

Note: ignoring any dependency must be consious choice.

### DIFF
--- a/.github/workflows/pr-extra.yml
+++ b/.github/workflows/pr-extra.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
-      # We cannot use nancy-github-action because it is outdated, so it's better to use the latest
-      # docker image for the validation
-      - name: nancy
-        run: go list -json -m all | docker run -i sonatypecommunity/nancy:v0.3
+      - name: Run go list
+        run: go list -json -m all > go.list
+      - name: Nancy
+        uses: sonatype-nexus-community/nancy-github-action@master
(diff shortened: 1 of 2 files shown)


.github/workflows/pr-extra.yml
```diff
@@ -13,7 +13,7 @@ jobs:
...
       - uses: actions/checkout@v2
       - uses: actions/setup-go@v2
-      # We cannot use nancy-github-action because it is outdated, so it's better to use the latest
-      # docker image for the validation
-      - name: nancy
-        run: go list -json -m all | docker run -i sonatypecommunity/nancy:v0.3
+      - name: Run go list
+        run: go list -json -m all > go.list
+      - name: Nancy
+        uses: sonatype-nexus-community/nancy-github-action@master
```
sayboras, line 19: changed to github action as it's updated to the latest now, another benefit is that we don't need to mount volume for .nancy-ignore file.

.nancy-ignore
```diff
@@ -0,0 +1,11 @@
+# Skip for golang/golang.org/x/net@v0.0.0-20200904194848-62affa334b73
```
sayboras, line 1: latest x/net package is still having some vulnerbilities

Nivl (approved):
Looks good to me, Thanks!

PR #1380, opened 2020-09-21
Add tparallel linter

Nivl:
This PR adds the [tparallel](https://github.com/moricho/tparallel) linter. This linter checks for inefficient usage of `t.Parallel()` in the tests.

Closes #1371 (cc @moricho)

~The Nancy failure is being taken care of in #1378 and #1293~ Merged

~Note: After doing more manual tests with the linter, I found a false positive: https://github.com/moricho/tparallel/issues/8. That shouldn't prevent reviewing though, I'll update the linter's version once it's been fixed.~ Done

PR #1381, opened 2020-09-21
ci: prevent macos to be marked as passing upon failure

Nivl:
Our tests for macOS and Windows are using the `continue-on-error` option, making them always showing up as passing even when they fail. [Here's an example of a failing job being marked as passing](https://github.com/golangci/golangci-lint/pull/1380/checks?check_run_id=1141753724). This PR removes this option for macos because [Windows is broken](https://github.com/golangci/golangci-lint/pull/1381/checks?check_run_id=1141826954).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -44,7 +44,6 @@ jobs:
           go-version: 1.15 # test only the latest go version to speed up CI
       - name: Run tests
         run: make.exe test
-        continue-on-error: true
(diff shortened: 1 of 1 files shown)


Nivl:
I opened an issue for Windows https://github.com/golangci/golangci-lint/issues/1382

cblecker (approved):

sayboras (approved):

PR #1395, opened 2020-09-25
Ignore crlf on windows

Nivl:


Nivl:
@golangci/team anyone here with a Windows computer that can run some tests? Basically, most of the tests are failing because the `golangci-lint` binary is not found. Maybe the path is wrong, maybe something needs a `.exe` somewhere (like make or golangci-lint itself), etc. But it's a pain to try to debug by pushing experiments to the CI. Can someone run `make build` and show me the output as well as making sure the binary is where we expect it to be and is named the way we expect it to be named (and does it needs to be `make.exe`)?

daixiang0:
@Nivl traivs CI support debug mode, you can try it.

Nivl:
@daixiang0 That's a good idea, thanks. It just sucks because that would require setting up a new CI 😕 I'm gonna wait a bit longer in case someone shows up with a Windows computer 🤞

Nivl:
I installed Windows on my mac earlier today, so I did a few tests:

The CI fails with:

```
    testshared.go:118: can't get error code from exec: "..\\golangci-lint": file does not exist
level=info msg="[test] ran [..\\golangci-lint linters -c C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\golangci_lint_test042139755.yml --verbose --fast] in 999.1µs"
    testshared.go:77: 
        	Error Trace:	testshared.go:77
        	            				enabled_linters_test.go:190
        	Error:      	"exec: \"..\\\\golangci-lint\": file does not exist" does not contain "Active 8 linters: [deadcode errcheck gosimple ineffassign staticcheck structcheck typecheck varcheck]"
        	Test:       	TestEnabledLinters
        	Messages:   	exit code is -1
```

The good thing is that I get the same error locally. The bad thing now is that the file does exist. I tried to change the path from relative to absolute, but it still fails saying the binary doesn't exist, even though I can run `ls` on the path with no problem.

I don't really have much time to dig deeper for now, but maybe for some reason the path we provide is not being treated as a path, and Go/Windows is looking for a binary named `../golangci-lint` rather than for a binary named `golangci-lint` that is in the parent directory. but that's just a wild guess for now.

ldez:
The implementation of `Cmd.CombinedOutput`/`Cmd.Run` on Window have some constraints:

https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/os/exec/lp_windows.go#L37-L52

In short, the executable must have an extension (`.com`, `.exe`, `.bat`, `.cmd`)

https://github.com/golang/go/blob/3dd58676054223962cd915bb0934d1f9f489d4d2/src/os/exec/lp_windows.go#L74

Nivl:
@ldez Good catch, it's not even documented afaik!

Nivl:
Thanks @ldez that did solve the issue, sadly now there are 20k more 😀. Some issues around differences in how Windows deals with file descriptors, some because we have hardcoded `/` everywhere in the tests, etc. Nothing that looks too complicated, I'll try to give it another look by EOM.

ldez:
@Nivl I recommend using `filepath.FromPath`

ldez:
@Nivl take a look to https://github.com/golangci/golangci-lint/pull/3134 :wink:

PR #1396, opened 2020-09-25
Added support for only specifying default severity

ac-rappi:
Default severity is only used when rules are not empty

Make sure the rules transformations for modifying severity are skipped if there are not rules AND no default severity specified


alejandrocragnolini (approved):
+1!

andres-loyola:
+1

andres-loyola:
+1

daixiang0 (approved):
lgtm

PR #1409, opened 2020-10-02
ci(cleanup): Remove go 1.13 run as part of github action

sayboras:
As per https://golang.org/doc/devel/release.html, only the last golang
versions are supported i.e. golang 1.13 is end of line, so there is no
point running golang 1.13 in CI.

This commit is to remove 1.13 in github action matrix

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -62,7 +62,6 @@ jobs:
     strategy:
       matrix:
         golang:
-          - 1.13
           - 1.14
           - 1.15
     steps:


Nivl (approved):

Nivl:
I think we'll need someone from @golangci/core-team to remove the requirement for 1.13 and add the requirement on 1.15 instead

bombsimon (approved):
LGTM, also nice to speed up test. I think we can merge this without waiting from @golangci/core-team since this will only affect what we test.

sayboras:
Thanks both for reviewing. 

I think @Nivl 's point is to remove required check for 1.13 in github action, even now this PR is approved now, but I can't merge :)

bombsimon:
> Thanks both for reviewing.
> 
> I think @Nivl 's point is to remove required check for 1.13 in github action, even now this PR is approved now, but I can't merge :)

Ah, of course! I was thinking the check was for the job, not the specific targets.

saschagrunert (approved):
LGTM

PR #1421, opened 2020-10-09
docs: Add description of Homebrew's official formula

rinsuki:
`golangci-lint` formula [looks available in Homebrew's official formula](https://github.com/Homebrew/homebrew-core/blob/HEAD/Formula/golangci-lint.rb), so I added description of Homebrew's official formula.


docs/src/docs/usage/install/index.mdx
```diff
@@ -53,6 +53,13 @@ brew install golangci/tap/golangci-lint
 brew upgrade golangci/tap/golangci-lint
 ```
 
+If you don't want to use tap, official formula also available. but it requires install Go compiler from Homebrew, so be careful.
```
iwankgb, line 56: This is valid change, indeed but could you:
* delete section that mentions the tap above (it isn't maintained anymore)
* do not mention obsolete tap in your change?

iwankgb (changes requested):

docs/src/docs/usage/install/index.mdx
rinsuki: It looks like the tap is still being maintained, is there any information anywhere that the tap is obsolete?
https://github.com/golangci/homebrew-tap/commits/master

iwankgb: Huh, I guess we keep it for backward compatibility reason or because we missed the fact that tap is still being upgraded. Removing it from the docs sounds like a good idea to me anyway.

iwankgb (approved):

iwankgb:
There are some conflicts, apparently.

PR #1422, opened 2020-10-10
Update gochecknoglobals, use source analyzer

### LINKED ISSUE #1393: Update gochecknoglobals
It looks like in response to #199 [gochecknoglobals](https://github.com/leighmcculloch/gochecknoglobals) was ported into golangci-lint, but it's also out-of-date, as the latest gochecknoglobals makes some new exceptions.

Would it be possible for golandci-lint to call out to gochecknoglobals to make updating easier here, or is there a pattern I can follow to make it possible to hook into so you can simply import and call?

Alternatively the implementation in golangci-lint needs updating to pickup the new changes to maintain consistency.

bombsimon:
This will use the latest version of [`gochecknoglobals`](https://github.com/leighmcculloch/gochecknoglobals) and the analyzer that's been created. Closes #1393 

This bump includes two changes:
1. The error message goes from `'foo' is a global variable` to `foo is a global variable`.
2. Global variables assigned from `regexp.MustCompile()` are now allowed

I'm a bit confused regarding 1 since the old code doesn't add backticks and the source which this is copied from didn't do so previously either.


vovapi (approved):
LGTM!

bombsimon:
Regarding 1, I totally missed the call to `formatCode()`. Seems like `golangci-lint` does this for al ot of linters. Anyone knows why? Would it be a problem to stop putting the variable name in backticks? I guess this will affect the output when used in the GitHub action.

Also noticed that this will never lint test files so I will add a patch for this asap.

pkg/golinters/gochecknoglobals.go
```diff
@@ -1,100 +1,30 @@
...
-				}
-			}
-			if len(res) == 0 {
-				return nil, nil
-			}
+	// gochecknoglobals only lints test files if the `-t` flag is passed so we
+	// set up our own FlagSet and add it to the analyzer before running it. This
+	// can be turned of by using the regular golangci-lint flags such as
+	// `--tests` or `--skip-files`.
+	flags := flag.NewFlagSet("", flag.ContinueOnError)
```
vovapi, line 18: I think overwriting flag set is bad idea. In my opinion it will cause hard-to-spot issues on when flags will be updated.

You should pass a config to `go analysis.NewLinter` instead

vovapi (changes requested):

pkg/golinters/gochecknoglobals.go
```diff
@@ -1,100 +1,30 @@
...
-		},
-	}
 	return goanalysis.NewLinter(
-		gochecknoglobalsName,
-		"Checks that no globals are present in Go code",
-		[]*analysis.Analyzer{analyzer},
+		gochecknoglobals.Name,
+		gochecknoglobals.Doc,
+		[]*analysis.Analyzer{gochecknoglobals},
 		nil,
```
vovapi, line 28:
```suggestion
		map[string]map[string]interface{}{gochecknoglobals.Name: {"t": true}},
```

pkg/golinters/gochecknoglobals.go
sayboras: agreed. If required, we can add configuration params for this linter  in .golangci.yaml file as well.

pkg/golinters/gochecknoglobals.go
```diff
@@ -1,100 +1,30 @@
...
-			}
-			if len(res) == 0 {
-				return nil, nil
-			}
+	// gochecknoglobals only lints test files if the `-t` flag is passed so we
+	// set up our own FlagSet and add it to the analyzer before running it. This
+	// can be turned of by using the regular golangci-lint flags such as
+	// `--tests` or `--skip-files`.
+	flags := flag.NewFlagSet("", flag.ContinueOnError)
+	flags.Bool("t", true, "Include tests")
```
sayboras, line 19: same as above point. Globally, we have include-test flag in golang.yaml config file.

vovapi: Global flag affects files that are passed to an Analyzer through `*analysis.Pass`
Analyzer could filter out test files as part of it's analysis (based on provided flags). `Gochecknoglobals` filters out tests by default.
I agree with @bombsimon that test filtering should be turned off (`-t` flag should be turned on). But this should be done by passing config, not by overriding flag set.

sayboras: yes, it's making sense to me, thanks :100:

pkg/golinters/gochecknoglobals.go
bombsimon: This is the way to do it! It's bee na while since I looked at this code so I missed this simple solution. Will submit patch removing the flag overwriting in favour of this later today.

bombsimon:
Thanks for pointing out the faulty way to enforce the flag! Will patch this, are you OK with the change around `formatCode()`?

bombsimon:
Ping @vovapi @sayboras

vovapi:
> Are you OK with the change around `formatCode()`

I'm ok with it.

vovapi (approved):

sayboras (approved):
LGTM :100:

PR #1424, opened 2020-10-11
ci(dependabot): Change interval for npm to monthly

sayboras:
npm packages are used for website only in this repo. Consider that
golangci-lint normally releases once a month, this commit is to change
interval for npm dependabot from weekly to monthly.

Main benefits could be:

- Reduce the noise for maintainers.
- Improve generated CHANGELOG. These changes are not useful for 
end users.


bombsimon (approved):
LGTM

PR #1458, opened 2020-10-18
chore(dep): Use tag version for cobra

sayboras:
This commit is to use tag version for cobra (e.g. 1.1.1). The main
benefit is that this can be managed and upgraded by dependabot later.

This is follow-up item from PR #1201.


bombsimon (approved):
🏷️

PR #1460, opened 2020-10-19
Update Docs and Assets Github Actions

### LINKED ISSUE #1208: Automate update for github-action-config.json

neha-viswanathan:
Closes #1208


.github/workflows/tag.yml
```diff
@@ -61,3 +61,19 @@ jobs:
             golangci/golangci-lint:${{ steps.prepare.outputs.tag_name }}-alpine
             golangci/golangci-lint:${{ steps.prepare.outputs.major_tag }}-alpine
             golangci/golangci-lint:latest-alpine
+  update_github_action:
```
sayboras, line 64: the related make target will make changes in respective file, we need to add the step to check in changes as well.

sayboras: GITHUB_TOKEN is already there, you can use checkout@v2 as mentioned in this [link](https://github.community/t/how-does-one-commit-from-an-action/16127/6)

sayboras (changes requested):
Thanks for your PR, I have one comment below. Another point to think about is pre-release and post-release, currently, I think it's ok to do all the step in post-release.

.github/workflows/tag.yml
sayboras: Another point is about pushing to master
- pushing to master might not be allowed
- even it's allowed, it's kind of risky :thinking: 

So automate this work might be just sending PR (like dependabot), and let other member to review and merge.

neha-viswanathan: @sayboras Thanks for reviewing this MR. I have a few questions to clarify

1. Do you propose I save the changed files as artifacts?
2. Should I move these into another workflow after the release?
3. Can we create a dummy release and test?

sayboras: > Do you propose I save the changed files as artifacts?
I am thinking about sending changed files in another PR _automatically_

> Should I move these into another workflow after the release?
We can keep it in same workflow, but only run it if the gorelease + docker jobs completed successfully (e.g. use `need` attribute in github action)

> Can we create a dummy release and test?
I normally test this kind of changes in forked repo, in which you will have _full permission_. Then just share details of your testing here.

Hope it's clear, feel free to slack me in gopher channel #golangci-lint as well :+1:

neha-viswanathan:
Test on my fork by creating a tag

image

sayboras:
@neha-viswanathan first glance looks good, will do proper review later. I have added team for another :eyes: as well.

sayboras:
@neha-viswanathan correct me if I am wrong, based on your screen shot, the PR got created, but I can't find it in your forked repo. Can you share the link here ?

neha-viswanathan:
@sayboras I test this on a dummy tag I created on my fork - https://github.com/neha-viswanathan/golangci-lint/tree/v100
https://github.com/neha-viswanathan/golangci-lint/actions/runs/315233729

.github/workflows/tag.yml
```diff
@@ -61,3 +61,26 @@ jobs:
...
+    needs: [release]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Install Go
+      uses: actions/setup-go@v2
+      with:
+         go-version: 1.15
+    - name: Update GitHub action config
```
sayboras, line 73: >  I test this on a dummy tag I created on my fork - https://github.com/neha-viswanathan/golangci-lint/tree/v100
https://github.com/neha-viswanathan/golangci-lint/actions/runs/315233729

Thanks, the above github action was kicked off successfully, but I don't think that we can confirm things will run as per expectation due to 2 reasons:
- update_github_action was skipped due to _release_ failure. For testing, you might need to relax this dependency.
- check-in-files step didn't do anything due to no changes in previous step.

Another point is that GITHUB_TOKEN environment might be required for these two steps.

sayboras:
Seems like you are making some changes, please share link to auto-generated PR in your forked repo as part of your testing.

neha-viswanathan:
@sayboras This is the link to the actions - i'm unable to run `make update_contributors_list` successfully
https://github.com/neha-viswanathan/golangci-lint/runs/1301414065?check_suite_focus=true

sayboras:
> This is the link to the actions - i'm unable to run `make update_contributors_list` successfully
> https://github.com/neha-viswanathan/golangci-lint/runs/1306304385

As this one never runs in github action before, I am not sure what the reason could be. The best you can do is to replicate it locally by cross checking npm/node/etc  version in github action. This target is working fine for me. If you don't have time or can't figure it out, feel free to skip this `make update_contributors_list` step and carry on with your testing. The goal we want to achieve is having auto generated PR :)

SVilgelm:
Fixed all issues and tested on my fork:
Here is the action's log: https://github.com/SVilgelm/golangci-lint/runs/2005819581?check_suite_focus=true
And a PR created by new action: https://github.com/SVilgelm/golangci-lint/pull/102

SVilgelm:
@golangci/team @sayboras @ldez please review

.github/workflows/tag.yml
```diff
@@ -48,23 +48,23 @@ jobs:
...
+          echo ::set-output name=date::${DATE}
+          if [[ ${{ matrix.target.Dockerfile }} == *"alpine"* ]]; then
+            echo ::set-output name=full_tag_name::${TAG}-alpine
+            echo ::set-output name=full_major_tag::${MAJOR}-alpine
+            echo ::set-output name=latest_tag::latest-alpine
+          else
+            echo ::set-output name=full_tag_name::${TAG}
+            echo ::set-output name=full_major_tag::${MAJOR}
+            echo ::set-output name=latest_tag::latest
+          fi
```
SVilgelm, line 67: auto formatting changes

ldez:
@SVilgelm `secrets.GOLANGCI_LINT_TOKEN` has the good rights?


scopes:
- `user:email`
- `read:user`
- `repo:public_repo`
- `admin:org read:org`

SVilgelm:
> @SVilgelm `secrets.GOLANGCI_LINT_TOKEN` has the good rights?
> 
> scopes:
> 
> * `user:email`
> * `read:user`
> * `repo:public_repo`
> * `admin:org read:org`

yes, it has write permissions
anyway I cannot confirm it without testing on this repo

the new action is independent and will not fail the mai release, so it's safe to merge

ldez:
I think that the `write permissions` are not enough

SVilgelm:
> I think that the `write permissions` are not enough

it needs the permissions to create a branch and create a pull request, so should be enough I think

SVilgelm:
the action creates a pull request, it is not pushing to master branch

ldez:
It's not enough to run the commands

SVilgelm:
> It's not enough to run the commands

what commands ?

ldez:
```
run: make assets/github-action-config.json

run: make update_contributors_list
```

SVilgelm:
but how it works on my fork? I also provide a simple toke with write permissions

anyway we don't have another token, the default `secrets.GITHUB_TOKEN` is read only, if the permissions is not enough then we will ask @golangci/core-team to give more

ldez:
I'm not able to run the following commands without the rights that I provided the list in my previous message.
```
make assets/github-action-config.json
make update_contributors_list
```

SVilgelm:
@ldez I see, you are right just public_repo is not enough but the `GOLANGCI_LINT_TOKEN` has more rights, because we are using it to create the releases and and push changes to homebrew tap repo.
I think it has enough rights

ldez:
I'm not sure because `admin:org read:org` is a not needed to publish on a repo.

SVilgelm:
as I said, we cannot fully test it in forks, if the GOLANGCI_LINT_TOKEN has no enough rights, then we will request them from @golangci/core-team

SVilgelm:
@golangci/core-team 
@jirfag Could you please add this permissions to the `GOLANGCI_LINT_TOKEN`:
```
user -> user:email, read:user
admin:org -> read:org
```

@ernado Could you please ping @jirfag ?

We tested, the `GOLANGCI_LINT_TOKEN` has only these permissions: `repo, workflow, write:packages`

ernado:
I'll ping @jirfag

ernado:
@SVilgelm please check

ldez:
The permissions are good now.

.github/workflows/tag.yml
```diff
@@ -90,3 +90,41 @@ jobs:
...
+      GITHUB_TOKEN: ${{ secrets.GOLANGCI_LINT_TOKEN }}
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install Go
+        uses: actions/setup-go@v2
+        with:
+          go-version: 1.16
+
+      - name: Update GitHub action config
+        run: make assets/github-action-config.json
```
ldez, line 107: This step requires that the GitHub release was done, so I think that we have to handle all the post-release on another event that a tag

ldez (changes requested):

.github/workflows/tag.yml
SVilgelm: actually it's non needed since the `need` parameter was used. But it's OK

SVilgelm (approved):
@ldez  let's merge it and release a new version

SVilgelm:
@idez it's failed, I think that the @golangci-releaser user should be a member of the @golangci/team, according to the documentation https://docs.github.com/en/rest/reference/teams#members


@jirfag @ernado  Could you please add the https://github.com/golangci-releaser user to the team?

PR #1464, opened 2020-10-23
feat: support http and https url in custom linters

draveness:
Feature Support: Custom linters can download and open dynamic libraries with HTTP and HTTPs.

### DIFF
--- a/pkg/lint/lintersdb/manager.go
+++ b/pkg/lint/lintersdb/manager.go
@@ -2,6 +2,10 @@ package lintersdb
 
 import (
 	"fmt"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"net/url"
 	"os"
 	"plugin"
 
@@ -409,6 +413,27 @@ type AnalyzerPlugin interface {
 }
 
 func (m Manager) getAnalyzerPlugin(path string) (AnalyzerPlugin, error) {
+	pathURL, err := url.Parse(path)
+	if err != nil {
+		return nil, fmt.Errorf("plugin %s has invalid path", path)
+	}
+
+	switch pathURL.Scheme {
+	case "http", "https":
+		pluginPath, downloadErr := downloadFile(path)
+		if downloadErr != nil {
+			return nil, fmt.Errorf("download plugin %s error, err: %w", path, err)
+		}
+		defer os.Remove(pluginPath)
+
+		path = pluginPath
+
+	case "":
+
+	default:
+		return nil, fmt.Errorf("plugin %s with invalid scheme", path)
+	}
+
 	plug, err := plugin.Open(path)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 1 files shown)


pkg/lint/lintersdb/manager.go
```diff
@@ -426,3 +451,26 @@ func (m Manager) getAnalyzerPlugin(path string) (AnalyzerPlugin, error) {
...
 	return analyzerPlugin, nil
 }
+
+func downloadFile(path string) (string, error) {
+	temfile, err := ioutil.TempFile("", "plugin")
+	if err != nil {
+		return "", err
+	}
+
+	resp, err := http.Get(path)
```
sayboras, line 461: please correct lint error here

pkg/lint/lintersdb/manager.go
```diff
@@ -409,6 +413,27 @@ type AnalyzerPlugin interface {
...
+	pathURL, err := url.Parse(path)
+	if err != nil {
+		return nil, fmt.Errorf("plugin %s has invalid path", path)
+	}
+
+	switch pathURL.Scheme {
+	case "http", "https":
+		pluginPath, downloadErr := downloadFile(path)
+		if downloadErr != nil {
+			return nil, fmt.Errorf("download plugin %s error, err: %s", path, err.Error())
```
sayboras, line 425: %w for error to wrap it

pkg/lint/lintersdb/manager.go
```diff
@@ -409,6 +413,27 @@ type AnalyzerPlugin interface {
 }
 
 func (m Manager) getAnalyzerPlugin(path string) (AnalyzerPlugin, error) {
+	pathURL, err := url.Parse(path)
+	if err != nil {
+		return nil, fmt.Errorf("plugin %s has invalid path", path)
```
sayboras, line 418: if `path` is not valid path (e.g. maybe just local file system), will it fail and return here right away ?

sayboras (commented):
Seems like this PR is still WIP, I am marking it as draft. Feel free to mark it as ready once all CI jobs are passed.

pkg/lint/lintersdb/manager.go
draveness: by valid path, what do you mean here? maybe you can refer to https://play.golang.org/p/d5poeeQM8US

pkg/lint/lintersdb/manager.go
draveness: add a nolint here

draveness:
could you help to take a look at the failed CI? i don't think it's related to this change. @sayboras

sayboras:
> could you help to take a look at the failed CI? i don't think it's related to this change.

Please rebase with master to fix check_generated failure. Other jobs are passing now. Feel free to send for review once you think it's ready.

PS: Some description in PR will be nice, and make it easier for reviewer.

blaquez (approved):

blaquez (approved):

blaquez (approved):

blaquez (approved):

blaquez (approved):

ernado:
1. Is it safe to allow downloading code in such way by default? Looks like there will be no validation, e.g. no checksums. I think it would be better to add some feature flag to enable this logic.
2. Probably this can be improved via caching (we already have golangci lint cache) and retries, but current implementation is fine as experiment.

PR #1472, opened 2020-10-26
Update exhaustivestruct

mbilski:
fixes:
- ignore private fields
- report multiple missing fields at once

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -33,7 +33,7 @@ require (
 	github.com/maratori/testpackage v1.0.1
 	github.com/matoous/godox v0.0.0-20190911065817-5d6d842e92eb // v1.0
 	github.com/mattn/go-colorable v0.1.8
-	github.com/mbilski/exhaustivestruct v1.0.1
+	github.com/mbilski/exhaustivestruct v1.1.0
 	github.com/mitchellh/go-homedir v1.1.0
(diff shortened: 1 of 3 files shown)


bombsimon (commented):
Generally looks good but how do you feel about tests? Of course the heavy lifting and majority of testing should be a part of the linter but I think it's good to cover at least one of each pass in this repo too. The reason being that a too big change to the linter hidden behind a change to only `go.mod` could have unexpected affect to the end user. By always ensuring the testdata adds or removes passes it's easier to see how big the change is.

Wdyt? Should we add tests for the changes you did such as ignoring private fields or do you think the single pass in the current testdata is enough?

mbilski:
> Generally looks good but how do you feel about tests? Of course the heavy lifting and majority of testing should be a part of the linter but I think it's good to cover at least one of each pass in this repo too. The reason being that a too big change to the linter hidden behind a change to only `go.mod` could have unexpected affect to the end user. By always ensuring the testdata adds or removes passes it's easier to see how big the change is.
> 
> Wdyt? Should we add tests for the changes you did such as ignoring private fields or do you think the single pass in the current testdata is enough?

makes sense, I'll add some tests

bombsimon (approved):
Thanks, LGTM!

PR #1493, opened 2020-11-05
DefaultExcludePatterns should only be used for specified linter

### LINKED ISSUE #1474: DefaultExcludePatterns should only be used for specified linter
golangci-lint set `--exclude-use-default=true` by default, but the exclude pattern is used for global match but not for specified linter(we can see the linter in `pkg/config/config.go`). I develop a new linter, one rule just check whether there is at least one package comment in a Go package, but the issue match `EXC0002` for golint, so it is ignored by default.

I think we can make some changes to the  processor in `pkg/result/processors/exclude_rules.go`, so that it can handle both `exclude global match` and `exclude specified match for some linters`.

zhangyunhao116:
Fix https://github.com/golangci/golangci-lint/issues/1474

This PR add "EXC0011", because most of package in golangci-lint does not having a package comment, also fix testdata.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -127,6 +127,22 @@ func GetExcludePatternsStrings(include []string) []string {
 	return ret
 }
 
+func GetExcludePatterns(include []string) []ExcludePattern {
+	includeMap := make(map[string]bool, len(include))
+	for _, inc := range include {
+		includeMap[inc] = true
+	}
+
+	var ret []ExcludePattern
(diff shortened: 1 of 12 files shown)


ZYunH:
Look like the test failed because the golangci-lint itself has changed(with "EXC0011"), if without `// nolint:dupl,stylecheck`(hint nolint), the test will be failed in local.

ZYunH:
Closed. Update to https://github.com/golangci/golangci-lint/pull/1494

PR #1494, opened 2020-11-06
DefaultExcludePatterns should only be used for specified linter

ZYunH:
Fixes https://github.com/golangci/golangci-lint/issues/1474

~This PR add "EXC0011", because most of package in golangci-lint does not having a package comment~. `make test` passed.

---------------------------
updated, for now, "EXC0011" same with "EXC0002" but with different linter field for keeping backward compatible.

### DIFF
--- a/pkg/config/config_test.go
+++ b/pkg/config/config_test.go
@@ -1,0 +1,36 @@
+package config
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestGetExcludePatterns(t *testing.T) {
+	assert.Equal(t, GetExcludePatterns(nil), DefaultExcludePatterns)
+
+	include := make([]string, 2)
+	include[0], include[1] = DefaultExcludePatterns[0].ID, DefaultExcludePatterns[1].ID
+
+	exclude := GetExcludePatterns(include)
+	assert.Equal(t, len(exclude), len(DefaultExcludePatterns)-len(include))
+
+	for _, p := range exclude {
+		// Not in include.
+		for _, i := range include {
+			if i == p.ID {
+				t.Fatalf("%s can't appear inside include.", p.ID)
+			}
+		}
+		// Must in DefaultExcludePatterns.
+		var inDefaultExc bool
+		for _, i := range DefaultExcludePatterns {
+			if i == p {
+				inDefaultExc = true
+				break
+			}
+		}
+		assert.True(t, inDefaultExc, fmt.Sprintf("%s must appear inside DefaultExcludePatterns.", p.ID))
+	}
+}
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -99,22 +99,33 @@ var DefaultExcludePatterns = []ExcludePattern{
 		Linter:  "gosec",
 		Why:     "False positive is triggered by 'src, err := ioutil.ReadFile(filename)'",
 	},
+	{
+		ID: "EXC0011",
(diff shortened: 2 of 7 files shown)


ZYunH:
CI has failed, because 
```
Error: directive `// nolint:dupl,stylecheck` is unused for linter stylecheck (nolintlint)
Error: directive `// nolint:dupl,stylecheck` is unused for linter stylecheck (nolintlint)
```

If these two lines removed, the local test will be failed, output
```
pkg/golinters/gocognit.go:1:1: ST1000: package comment should be of the form "Package golinters ..." (stylecheck)
// nolint:dupl
^
pkg/golinters/gocyclo.go:1:1: ST1000: package comment should be of the form "Package golinters ..." (stylecheck)
// nolint:dupl
^
```

maybe something wrong in `stylecheck`? since even having a package comment in `pkg/golinters/asciicheck.go`,  the test still failed, and the output is same as above.

iwankgb:
I think that what you are doing here (changing semantics of an existing flag) is not a good idea. I would rather expect new flag to be introduced and handle the new case. Adding `// nolint:stylecheck` in quite a few places seems to be a prove that this change is not backward compatible and in the same time it does not seem to be a bug fix.

ZYunH:
The `// nolint:stylecheck` is added because `// nolint:dupl` is not a valid package comment, and makes the test failed(even add a package comment in `package golinters`). This problem will be fixed if https://github.com/dominikh/go-tools/pull/871 is merged, then we can remove these `// nolint:stylecheck`, and add a valid package comment to `package golinters`.

ZYunH:
@iwankgb we can keep backward compatible by adding a new rule like 
```
	{
		ID: "EXC0012",
		Pattern: "(comment on exported (method|function|type|const)|" +
			"should have( a package)? comment|comment should be of the form)",
		Linter: "stylecheck",
		Why:    "Annoying issue about not having a comment. The rare codebase has such comments",
	},
```
Then we can remove all these `// nolint:stylecheck`. But i don't know which is better, maybe someone from the team can give some ideas @golangci-releaser

ZYunH:
After put some thought to it, i think keeping backward compatible makes this PR more easily to understand, but it cover some mistakes in testdata and stylecheck.

iwankgb:
@ZYunH now I understand what you mean. I believe that your PR and issue are basically a bugfix and bugreport. If we provide linter information then we should not apply default exclude rules to all the linters. Taking above into consideration I have only one request: provide tests for `GetExcludePatterns()` and exclude processor use-case.
First test should be pretty straightforward. In case of the second one you will have to dig into [testdata directory](https://github.com/golangci/golangci-lint/tree/master/test/testdata) and prepare a file with matching violations that would fail on master and pass on your branch. Let me know if you have any issues, I will try to help :)

ZYunH:
@iwankgb PTAL, thanks!

pkg/config/config_test.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+		// Must in DefaultExcludePatterns.
+		var inDefaultExc bool
+		for _, i := range DefaultExcludePatterns {
+			if i == p {
+				inDefaultExc = true
+				break
+			}
+		}
+		if !inDefaultExc {
```
iwankgb, line 33: I would rather use `assert.True()`

iwankgb (changes requested):
This is last final piece of change and I will be more than happy to merge the PR.

pkg/config/config_test.go
ZYunH: Done

iwankgb (approved):

PR #1497, opened 2020-11-06
Unknown linter breaks //nolint

### LINKED ISSUE #1450: Treat "Found unknown linters" warning as an error
Currently, if I use a `nolint` directive with an unknown linter (regardless on whether the line has any linter errors or not), the result is a warning that looks like this:
```
WARN [runner/nolint] Found unknown linters in //nolint directives: foobar
```
The problem is that if such a directive is added by mistake (e.g., a typo on the linter name), this will masquerade the actual linter error (since the exit status of `golangci-lint` is `0` with this warning). I couldn't find an obvious configuration parameter to handle these scenarios properly.

I would expect the output of `golangci-lint` to be non-zero for these scenarios, or somehow configure warnings to be treated as errors. Another alternative would be for the `nolintlint` linter to catch these (I tried playing with `allow-unused`, but did not get what I expected)

N/A

**Additional context**
N/A

iwankgb:
Fixes #1450 by applying `//nolint` directive to a non-existent linter only.

### DIFF
--- a/pkg/result/processors/nolint.go
+++ b/pkg/result/processors/nolint.go
@@ -250,24 +250,20 @@ func (p *Nolint) extractInlineRangeFromComment(text string, g ast.Node, fset *to
 	var linters []string
 	text = strings.Split(text, "//")[0] // allow another comment after this comment
 	linterItems := strings.Split(strings.TrimPrefix(text, "nolint:"), ",")
-	var gotUnknownLinters bool
 	for _, linter := range linterItems {
 		linterName := strings.ToLower(strings.TrimSpace(linter))
 
 		lcs := p.dbManager.GetLinterConfigs(linterName)
 		if lcs == nil {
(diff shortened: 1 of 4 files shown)


iwankgb:
@golangci/team this PR is a breaking change, at least formally: at the moment when `//nolint` directive refers to non-existing linter, then:
* exit status will be 0.
* no violations from other linters will be reported for the line.

Is such change fine with us? I agree that it might force some people to do some extra work but I strongly believe that it is a valid bugfix too.

invidian (commented):
For me, if one specifies invalid linter in `//nolint:`, it should throw an error (or, a linter warning), instead of be silent and do nothing.

iwankgb:
Warning has always been displayed and I am not going to change it.

saschagrunert (approved):
LGTM

PR #1498, opened 2020-11-07
Update godot to 1.3.0

tetafro:
Update for godot linter. Now it can check any comment, not only top-level comments. Also it can check capital letter at the beginning of sentences.

There is a breaking change in config. I'm not sure what's the right way to introduce it. I removed old setting `check-all` from config example, but didn't remove it from `GodotSettings` struct to avoid breaking existing configs. Old setting `check-all` is silently translated to new `scope`.


bombsimon (approved):
LGTM

PR #1500, opened 2020-11-08
Using upstrem goconst

### LINKED ISSUE #1495: More configurable goconst
Hello!
I have a question :)

I needed the flags in `golangci-lint` that are in the original `goconst`:
https://github.com/jgautheron/goconst/issues/11

I wanted to do a PR and got confused where:
- https://github.com/golangci/goconst (outdated)
- https://github.com/jgautheron/goconst

What is the difference?

What is the best way to proceed? Try to update the `golangci/goconst` fork and then updating `golangci-lint` or update `golangci-lint` so that it no longer links to the fork, but works with the original linter?

Judging by the issues at https://github.com/golangci/goconst, I'm not the only one :)

iwankgb:
Thanks to https://github.com/jgautheron/goconst/issues/11 we should be able to get rid of goconst fork. Closes #1495.


iwankgb:
https://github.com/jgautheron/goconst/pull/13 must be merged and new version released before we proceed.

iwankgb:
@golangci/team, it seems that problem is more complicated here. Some time ago (our fork was based on 3-year old commit) new feature was introduced to `goconst`: 
https://github.com/jgautheron/goconst/blob/b58d7cf68fc68b9ce6a932bb4f5e48f0a26f19d6/visitor.go#L101
It cause `goconst` to notice repeated strings in function calls. With current settings in `.golangcilint.yml` there are 50 places that would have to be fixed. When `min-length` and `min-occurrences` are set to 3 (default values) there is only 11 of them.

Taking above into consideration we have 2 options, I think:
1. Fix violations and publish clear warning to the users stating that update in `goconst` may force them to do additional work.
2. Reach out to @jgautheron and try to make checks configurable, with same default config in our project.

Second option sounds much more reasonable to me and I would file all the necessary PRs. 

Third option (that I consider a non-option) would be to keep maintaining our fork and deviate from the upstream. This is probably the worst possible idea. In case of no feedback I will proceed with option 2 (if @jgautheron is open to such a solution).

ernado:
Can we use default excludes to mitigate (1) or error messages are same in both cases?

iwankgb:
@ernado it should be possible, I will investigate.

iwankgb:
@ernado message is the same for all types of violation and I can't really tell one from another.

iwankgb:
I have file a PR against upstream `goconst` repository. If it's merged we should be able to start using upstream seamlessly.

iwankgb:
@ernado, a quick look would be appreciated :)

ernado (approved):
LGTM, thank you!

saschagrunert (approved):

PR #1503, opened 2020-11-12
Parallel test linter

kunwardeep:
This PR adds the [paralleltest](https://github.com/kunwardeep/paralleltest) linter. This linter checks for incorrect usage of t.Parallel() in the tests. If the liner is enabled it will complain about each and every place we can use t.Parallel(). It also ensures that if we are using a range over test cases, we reinitialised the test case variable ([More Info](https://gist.github.com/kunwardeep/80c2e9f3d3256c894898bae82d9f75d0))


tetafro:
JFYI there is [scopelint](https://github.com/kyoh86/scopelint) that checks [incorrect usage of loop variables](https://github.com/kunwardeep/paralleltest#tparallel-is-called-in-the-range-method-and-test-case-variable-tc-being-used-but-is-not-reinitialised-more-info), but not only in tests.

go.sum
```diff
@@ -29,6 +29,7 @@ github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc/go.mod h1:LOuy
 github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf/go.mod h1:ybxpYRFXyAe+OPACYpWeL0wqObRcbAqCMya13uyzqw0=
```
tetafro, line 29: Can you move tools.go in your project to a separate directory with a separate go.mod file? Because otherwise your linter [depends on golangci-lint package](https://github.com/kunwardeep/paralleltest/blob/master/go.mod#L6) and it brings a lot of obsolete packages here because of a circle dependency:
```
golangci-lint@latest -> paralleltest@1.0.1 -> golangci-lint@v1.31.0
```

test/testdata/paralleltest.go
```diff
@@ -0,0 +1,140 @@
+//args: -Eparalleltest
```
tetafro, line 1: You don't have to bring all the tests from [your project](https://github.com/kunwardeep/paralleltest/blob/master/pkg/paralleltest/testdata/src/t/t_test.go) here. Since your code is already tested, it's enough to do a couple a simple test here just to check integration. [Example](https://github.com/golangci/golangci-lint/blob/master/test/testdata/bodyclose.go), [example](https://github.com/golangci/golangci-lint/blob/master/test/testdata/goimports/goimports.go).

tetafro:
And don't forget to squash commits ;)

test/testdata/paralleltest.go
kunwardeep: removed few tests

go.sum
kunwardeep: done 👍

tetafro (approved):
LGTM

kunwardeep:
Umm, how do get it merged? Is there a separate process for that?

bombsimon:
> Umm, how do get it merged? Is there a separate process for that?

Nope, nothing else. But probably good to have a few more approvals when merging a new linter.

bombsimon (approved):
LGTM

wyardley:
I don't see this in https://golangci-lint.run/usage/linters/ yet -- does it need to be added there?

bombsimon:
Not manually, but the `make` targe to re-generate the documentation and commit those changes needs to be done. I think it should be made automatically once a week with GitHub Actions so it will end up there.

wyardley:
@bombsimon thanks! Assumed it was automagic, just wanted to confirm (didn't know there was a delay).

kunwardeep:
yeah.. I thought so too, but after reading [this](https://github.com/golangci/golangci-lint/issues/1208), this looks like a manual step. so here is another PR- https://github.com/golangci/golangci-lint/pull/1526/files

Issue #1509, opened 2020-11-16
Discuss about switch to upstream Golint


daixiang0:
Thank you for creating the issue!

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

Please include the following information:

Now at the [website](https://golangci-lint.run/usage/linters/), we say we use upstream [golint](https://github.com/golang/lint), in fact, we use [our own golint](github.com/golangci/lint-1), and the test result is totally different.

Test code:
```
$ cat main.go
package main

import "fmt"

const TestString = "test"

func main() {
  fmt.Print(TestString)
}
```

Upstream Golint result:
```
$ golint main.go
main.go:5:7: exported const TestString should have comment or be unexported
```

Golangci-lint Golint result:
```
$ golangci-lint run -v --disable-all -E golint main.go
INFO [config_reader] Config search paths: [./root /]
INFO [lintersdb] Active 1 linters: [golint]
INFO [loader] Go packages loading at mode 575 (types_sizes|compiled_files|deps|name|exports_file|files|imports) took 86.063386ms
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 112.804µs
INFO [linters context/goanalysis] analyzers took 0s with no stages
INFO [runner] Issues before processing: 1, after processing: 0
INFO [runner] Processors filtering stat (out/in): path_prettifier: 1/1, identifier_marker: 1/1, skip_files: 1/1, cgo: 1/1, skip_dirs: 1/1, filename_unadjuster: 1/1, autogenerated_exclude: 1/1, exclude: 0/1
INFO [runner] processing took 278.777µs with stages: exclude: 143.284µs, identifier_marker: 77.659µs, autogenerated_exclude: 25.754µs, path_prettifier: 17.058µs, skip_dirs: 6.836µs, cgo: 1.638µs, nolint: 1.097µs, filename_unadjuster: 1.071µs, max_same_issues: 765ns, severity-rules: 601ns, uniq_by_line: 470ns, source_code: 460ns, skip_files: 336ns, diff: 320ns, max_from_linter: 306ns, exclude-rules: 273ns, path_shortener: 243ns, sort_results: 217ns, max_per_file_from_linter: 213ns, path_prefixer: 176ns
INFO [runner] linters took 21.666984ms with stages: golint: 21.26085ms
[...]

PR #1511, opened 2020-11-17
Update golint link

daixiang0:
Ref #1509 


ernado:
We have multiple forked linters, but we still using original url in `WithURL` for them.

Should we change the only one linter?
How it will help with #1509? Our fork documentation does not contain default excludes, it is in golangci-lint.

daixiang0:
@ernado it would be better, where is the good place to comment them?

ldez:
Following the discussion in #1509, this PR can be closed.

PR #1520, opened 2020-11-22
Add makezero linter

ashanbrown:
Would love to hear some thoughts on this linter.  In the interest of full disclosure, it was rejected by gocritic (see https://github.com/go-critic/go-critic/issues/875).  In practice, it has been immensely useful to my team.  Thanks for your consideration.  Details follow:

Makezero discourages the creation of slices with nonzero capacity but zero length.  Its primary goal is to prevent appending to an initialized slice with nonzero length, but it can be used to disable uninitialized slices with nonzero length altogether.  A common error among go programmers, particular when egged on by the useful `prealloc` linter, is this one:

```
x := make([]int, len(y))
for _, v := range y {
  x = append(x, v)
}
```

In this case, the programmer has forgotten to set the initial length of the array to 0 with `make([]int, 0, len(y)`.  In practice, my company has found it useful to any `make` calls with nonzero length, because the need to optimize by avoiding `append` statements is limited.  We also like seeing `append` statements because they discourage the additional indirection introduced by indices needed to access elements in a preallocated non-zero length array (i.e. we don't see the index variable `i` all over the place like you'd see in C-language programs).  We favorable readability of code over performance in almost all cases, adding `nolint` directives when we favor performance.

See https://github.com/ashanbrown/makezero for more details.


bombsimon (approved):
Thanks for the contribution. `golangci-lint` uses a bunch of opinionated linters and I'm not sure if the idea was to ever enable all. I see no reason to not add linters that could help with these things without the motive being that everyone should always use them. See [exhaustivestruct](https://github.com/mbilski/exhaustivestruct) for example, I don't think this is usable in general for most projects.

I see no reason to not merge this PR so I will approve it. However, if you're interested in feedback, I would stop using the terminology `size` since Go  refers to this as `capacity` both in the documentation ([see here](https://blog.golang.org/slices-intro)) and in the AST.

> The make function takes a type, a length, and an optional capacity.

ashanbrown:
Just noticed that this panics when running against golangci-lint itself so I'll have to look into that before this is merged.

sayboras:
I am sure that is existing linter to check this one, as Goland is showing similar warning for such cases for me. Give me some time to check settings.

It's ok to have some overlapping between linters, but if there is no additional value compared to existing one, I find it confusing to end users. Recently, there is one good question about differences between linters (e.g. deadcuode vs unused) in golangci-lint slack channel.

sayboras:
>I think that there might be existing linter/rule to check this one, as Goland is showing similar warning for such cases for me, it could be Goland's thingy only thinking. Give me sometime to check on this.

My bad, I was confused with another check

ashanbrown:
I made a small change to makezero that fixed parsing of golangci-lint itself, so I think this is in a state to be merged.

davidnewhall:
The discussion in this contribution about what this linter is supposed to do makes me think this is a bug. Is this expected? If so, how am I supposed to fix this?
```
pkg/logs/logs.go:77:3: use of `fmt.Println` forbidden by pattern `^fmt\.Print(|f|ln)$` (makezero)
512		fmt.Println("Logger Error:", err)
513		^
514pkg/logs/logs.go:85:3: use of `fmt.Println` forbidden by pattern `^fmt\.Print(|f|ln)$` (makezero)
515		fmt.Println("Logger Error:", err)
516		^
517pkg/logs/logs.go:93:3: use of `fmt.Println` forbidden by pattern `^fmt\.Print(|f|ln)$` (makezero)
518		fmt.Println("Logger Error:", err)
519		^
```

The three lines of code that got caught are in methods that all look like this:
```
 const callDepth = 2

 func (l *Logger) Printf(msg string, v ...interface{}) {
 	err := l.InfoLog.Output(callDepth, fmt.Sprintf(msg, v...))
 	if err != nil {
 		fmt.Println("Logger Error:", err)
 	}
 }
```

Thanks! If you want me to open a new issue I will, but this seems fresh and I'm hopeful my comment attracts some eyeballs.

davidnewhall:
I found the bug introduced in https://github.com/golangci/golangci-lint/pull/1569 and left a comment there. @ashanbrown - Please consider fixing that. Thank you!

Issue #1537, opened 2020-12-04
Can not detect issue when check dir


daixiang0:
In https://github.com/daixiang0/kubeedge/tree/test-ci, `edge/pkg/servicebus/config/config.go` has an obvious issue, while do not detect it.

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
golangci-lint has version 1.30.0 built from 45b90f6 on 2020-08-03T03:10:34Z

```

</details>

<details><summary>Config file</summary>

https://github.com/daixiang0/kubeedge/blob/test-ci/.golangci.yml

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
go version go1.14.4 linux/amd64
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/root/.cache/go-build"
GOENV="/root/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/root/gopkg"
GOPRIVATE=""
GOPROXY="https://goproxy.cn,https://proxy.golang.org,https://goproxy.io,direct"
GOROOT="/root/go"
GOSUMDB="gosum.io+ce6e7565+AY5qEHUk/qmHc5btzW45JVoENfazw8LielDsaI+lEbq6"
GOTMPDIR=""
GOTOOLDIR="/root/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD="/root/gopkg/src/github.com/daixiang0/222/go.mod"
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build843064397=/tmp/go-build -gno-record-gcc-switches"

```

</details>

[...]

PR #1540, opened 2020-12-07
fix: throw an error on panic.

ldez:
Related to:

- https://github.com/golangci/golangci-lint/issues/1537#issuecomment-739614904
- https://github.com/golangci/golangci-lint/issues/1537#issuecomment-739617883

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/golangci/golangci-lint
 
-go 1.12
+go 1.13
 
 require (
 	4d63.com/gochecknoglobals v0.0.0-20201008074935-acfc0b28355a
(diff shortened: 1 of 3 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/golangci/golangci-lint
 
-go 1.12
+go 1.13
```
daixiang0, line 3: Really upgrade go version?

ldez: - the current min go version in the GitHub Action is go1.13
- in the code, the `%w` is used (available since go1.13)

iwankgb (approved):

PR #1541, opened 2020-12-08
Add thelper linter

kulti:
Hi, I'm the author of [thelper](https://github.com/kulti/thelper) linter.

Hope you find it useful and accept this PR.


ldez:
Thanks! Would you mind adding your linter and the available settings, even though default, to `.golangci.example.yml`?  This will be used to generate documentation at [golangci-lint.run](https://golangci-lint.run/usage/configuration/#config-file).

ldez (changes requested):
Thanks! Would you mind adding your linter and the available settings, even though default, to `.golangci.example.yml`?  This will be used to generate documentation at [golangci-lint.run](https://golangci-lint.run/usage/configuration/#config-file).

Could you also add tests for the rules `t_first` and `t_name` in `test/testdata/thelper.go`?

kulti:
@ldez, thank you for your suggestions.

I've added config examples and tests for all available checks.

ldez:
I would like to suggest to you some changes to improve the configuration UX.
The value of the checks can be hard to find and it can be error-prone because it is a string.

I see 2 approaches:
- a slice-based configuration
- a boolean-based configuration

I would prefer a boolean-based configuration because I don't have to know the name of the options.

### Slice based configuration

```yaml
  # .golangci.example.yml

  thelper:
    checks:
      - t_first
      - t_name
      - t_begin
      - b_first
      - b_name
      - b_begin
```

```go
// pkg/config/config.go


type ThelperSettings struct {
	Checks []string
}
```

```go
	// pkg/golinters/thelper.go

	cfgMap := map[string]map[string]interface{}{}
	if cfg != nil {
		cfgMap[a.Name] = map[string]interface{}{
			"checks": strings.Join(cfg.Checks, ","),
		}
	}
```

### Boolean based configuration

```yml
  thelper:
    test:
      first: true
      name: true
      begin: true
    benchmark:
      first: true
      name: true
      begin: true
```

```go
type ThelperSettings struct {
	Test   struct {
		First bool `mapstructure:"first"`
		Name  bool `mapstructure:"name"`
		Begin bool `mapstructure:"begin"`
	} `mapstructure:"test"`
	Benchmark struct {
		First bool `mapstructure:"first"`
		Name  bool `mapstructure:"name"`
		Begin bool `mapstructure:"begin"`
	} `mapstructure:"benchmark"`
}
```

```go
	var opts []string

	if cfg.Test.Name {
		opts = append(opts, "t_name")
	}
	if cfg.Test.Begin {
		opts = append(opts, "t_begin")
	}
	if cfg.Test.First {
		opts = append(opts, "t_first")
	}

	if cfg.Benchmark.Name {
		opts = append(opts, "b_name")
	}
	if cfg.Benchmark.Begin {
		opts = append(opts, "b_begin")
	}
	if cfg.Benchmark.First {
		opts = append(opts, "b_first")
	}

	cfgMap := map[string]map[string]interface{}{}
	if cfg != nil {
		cfgMap[a.Name] = map[string]interface{}{
			"checks": strings.Join(opts, ","),
		}
	}
```

kulti:
I think boolean-based configuration looks great 👍 I've copy-pasted it with a few modifications.

ldez:
Did you plan to use tag/version?

There are several advantages:
- it's more "Go module idiomatic"
- that allow the dependabot to update automatically golangci-lint when you release a new version

ldez (approved):
Thanks :+1:

PR #1552, opened 2020-12-12
Using upstream dupl

### LINKED ISSUE #1551: We shall not user for of mibk/dupl
We shall not use fork of [mibk/dupl](https://github.com/mibk/dupl).

iwankgb:
Fixes #1551

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,6 @@ require (
 	github.com/go-xmlfmt/xmlfmt v0.0.0-20191208150333-d5b6f63a941b
 	github.com/gofrs/flock v0.8.0
 	github.com/golangci/check v0.0.0-20180506172741-cfe4005ccda2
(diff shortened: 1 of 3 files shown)


Nivl (commented):
My main concern here is that we're not removing a fork, we're switching from our own (outdated) fork to yours.

I think we should wait for https://github.com/mibk/dupl/pull/18 to get merged upstream before doing anything (which might be the reason why you put the PR back as Draft after requesting a review?).

iwankgb:
@Nivl, yes this is why PR is in draft. If the dupl PR is not merged then there is no point in swapping a fork for another one :)

ldez:
I will close this PR for now because #1551 already tracks this topic.

And the PR https://github.com/mibk/dupl/pull/18 is opened for more than a year.

PR #1563, opened 2020-12-17
Validate gocritic settings. Return error if settings includes a unsupported gocritic checker

sebastien-rosset:
This PR performs additional gocritic settings validation. The `settings` attribute is supposed to be a map where each key is the name of a supported gocritic checker. 

The existing validation already warns if the config data contains a settings for a disabled checker. But it does not validate the checker is supported. There is also an existing validation that the specific parameter is supported or not.

For example, as shown in the yaml config data below, `ruleguard` is a supported gocritic checker, but `rangeValueCopy` is not (the actual checker is `rangeValCopy`).

```
  gocritic:
    settings:
      rangeValueCopy:
        sizeThreshold: 32
      ruleguard:
        rules: myrules.go
```

### DIFF
--- a/pkg/config/config_gocritic.go
+++ b/pkg/config/config_gocritic.go
@@ -242,13 +242,7 @@ func (s *GocriticSettings) Validate(log logutils.Log) error {
 		return errors.Wrap(err, "validate disabled checks")
 	}
 
-	for checkName := range s.SettingsPerCheck {
-		if !s.IsCheckEnabled(checkName) {
-			log.Warnf("Gocritic settings were provided for not enabled check %q", checkName)
-		}
-	}
-
-	if err := s.validateCheckerNames(); err != nil {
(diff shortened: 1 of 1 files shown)


sebastien-rosset:
Note: I've noticed a separate but related problem: if the input ruleguard rules.go has a syntax error, the error is silently ignored and the entire ruleguard file is ignored. This is unlike the `gocritic` and `ruleguard` behavior which print an error message and exit with error status.

sebastien-rosset:
@quasilyte, would you be able to review this PR? thank you.

quasilyte (approved):
Looks good.

sebastien-rosset:
@ldez , I see you have removed two reviewers. I had added them because they seemed to be the most recent committers for this file. Are you the reviewer?

ldez (approved):
From my point of view, it's better to let people add themselves as reviewers if they want to review.

PR #1569, opened 2020-12-21
Add forbidigo linter

ashanbrown:
Forbidigo forbids identifiers.  It can be used to forbid code that is useful for debugging but should not be merged to the master branch.  I've found it particularly useful to make sure I don't commit `fmt.Print.*` statements or `ginkgo.F.*` commands that I've introduced during local debugging.

Not sure if there is another linter that already does this.  Please let me know if there is.

Base linter is at https://github.com/ashanbrown/forbidigo

### DIFF
--- a/pkg/golinters/forbidigo.go
+++ b/pkg/golinters/forbidigo.go
@@ -1,0 +1,66 @@
+package golinters
+
+import (
+	"sync"
+
+	"github.com/ashanbrown/forbidigo/forbidigo"
+	"github.com/pkg/errors"
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
+	"github.com/golangci/golangci-lint/pkg/lint/linter"
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+const forbidigoName = "forbidigo"
+
+func NewForbidigo() *goanalysis.Linter {
+	var mu sync.Mutex
+	var resIssues []goanalysis.Issue
+
+	analyzer := &analysis.Analyzer{
+		Name: forbidigoName,
+		Doc:  goanalysis.TheOnlyanalyzerDoc,
+	}
+	return goanalysis.NewLinter(
+		forbidigoName,
+		"Forbids identifiers",
+		[]*analysis.Analyzer{analyzer},
+		nil,
+	).WithContextSetter(func(lintCtx *linter.Context) {
+		s := &lintCtx.Settings().Forbidigo
+
+		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
+			var res []goanalysis.Issue
+			forbid, err := forbidigo.NewLinter(s.Forbid)
+			if err != nil {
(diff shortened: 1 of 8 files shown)


pkg/golinters/forbidigo.go
```diff
@@ -0,0 +1,63 @@
...
+	).WithContextSetter(func(lintCtx *linter.Context) {
+		s := &lintCtx.Settings().Forbidigo
+
+		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
+			var res []goanalysis.Issue
+			linter, err := forbidigo.NewLinter(s.Forbid)
+			if err != nil {
+				return nil, errors.Wrapf(err, "failed to create linter %q", forbidigoName)
+			}
+			for _, file := range pass.Files {
```
ldez, line 39:
```suggestion
			linter, err := forbidigo.NewLinter(s.Forbid)
			if err != nil {
				return nil, errors.Wrapf(err, "failed to create linter %q", forbidigoName)
			}
			
			var res []goanalysis.Issue
			for _, file := range pass.Files {
```

pkg/golinters/forbidigo.go
```diff
@@ -0,0 +1,63 @@
...
+		forbidigoName,
+		"Forbids identifiers",
+		[]*analysis.Analyzer{analyzer},
+		nil,
+	).WithContextSetter(func(lintCtx *linter.Context) {
+		s := &lintCtx.Settings().Forbidigo
+
+		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
+			var res []goanalysis.Issue
+			linter, err := forbidigo.NewLinter(s.Forbid)
```
ldez, line 35: Could you rename `linter` to `forbid` to avoid that the variable 'linter' collides with the imported package name?

ldez (changes requested):
Could you also run a `go mod tidy`?

Thank you.

ashanbrown:
Should we be verifying that `go mod tidy` produces no changes in CI?  Also, I'm wondering if we should consider linter checks for the other items you mentioned (whitespace and shadowed imports).

ldez:
@ashanbrown there was an RP to change this behavior #1397. I think that this PR can be revert since dependabot is now able to manage `go.sum` file.

ldez (approved):
LGTM

pkg/golinters/forbidigo.go
```diff
@@ -0,0 +1,66 @@
...
+			for _, file := range pass.Files {
+				hints, err := forbid.Run(pass.Fset, file)
+				if err != nil {
+					return nil, errors.Wrapf(err, "forbidigo linter failed on file %q", file.Name.String())
+				}
+				for _, hint := range hints {
+					res = append(res, goanalysis.NewIssue(&result.Issue{
+						Pos:        hint.Position(),
+						Text:       hint.Details(),
+						FromLinter: makezeroName,
```
davidnewhall, line 49: THIS IS A BUG.

SVilgelm: It's already fixed in #1590

PR #1570, opened 2020-12-21
Allow excluding unused nolint checks on a per-linter or per-line basis

### LINKED ISSUE #1542: Different behavior of govet && nolinitlint on Windows and Linux

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
golangci-lint has version 1.33.0 built from b90551c on 2020-11-23T05:15:36Z
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
run:
  concurrency: 4
  timeout: 15m
  issues-exit-code: 1
  tests: true
  build-tags: []
  skip-dirs: []
  skip-files: []
  #    - ".*pb\\.go$"
  #    - "bindata.go"
  modules-download-mode: readonly

output:
  format: colored-line-number
  print-issued-lines: true
  print-linter-name: true
  uniq-by-line: true

linters-settings:
  govet:
    check-shadowing: true
  golint:
    min-confidence: 0.8
  gocyclo:
    min-complexity: 30
  maligned:
    suggest-new: true
  goconst:
    min-len: 3
    min-occurrences: 3
  gocritic: {}
  lll:
    line-length: 120
    tab-width: 4
  misspell:
    locale: US
    ignore-words: []

linters:
  enable:
    - goimports
    - govet
    - golint
    - gocyclo
    - gocognit
    - maligned
    - goconst
    - gocritic
    - staticcheck
    - errcheck
    - lll
    - whitespace
    - nolintlint
    - unused
  disable-all: true
  fast: true

issues:
  exclude-use-default: false
  exclude-rules:
    - path: _test\.go
      linters:
        - gocognit
        - gocyclo

    - linters:
        - govet
      text: 'shadow: declaration of "err"'

    - linters:
        - govet
      text: "composite literal uses unkeyed fields$"

    - linters:
        - lll
      source: "^//go:generate "
```

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
go version go1.15.6 windows/amd64
set GO111MODULE=on
set GOARCH=amd64
set GOBIN=c:\dev\go\bin
set GOCACHE=C:\Users\webta\AppData\Local\go-build
set GOENV=C:\Users\webta\AppData\Roaming\go\env
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOINSECURE=
set GOMODCACHE=C:\dev\go\pkg\mod
set GONOPROXY=*****.com
set GONOSUMDB=*****.com
set GOOS=windows
set GOPATH=C:\dev\go
set GOPRIVATE=*****.com
set GOPROXY=https://proxy.golang.org,direct
set GOROOT=c:\go
set GOSUMDB=sum.golang.org
set GOTMPDIR=
set GOTOOLDIR=c:\go\pkg\tool\windows_amd64
set GCCGO=gccgo
set AR=ar
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=D:\Users\webta\Development\*****\go.mod
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -mthreads -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -fdebug-prefix-map=C:\Users\webta\AppData\Local\Temp\go-build143508486=/tmp/go-build -gno-record-gcc-switches
mi
```

</details>

<details><summary>Verbose output of running</summary>

```console
$ golangci-lint cache clean
$ golangci-lint run -v
# paste output here
```

</details>

Under Windows (WSL) if nolint-directive is presented, I'm getting
```
directive `// nolint:govet // func has predefined signature and job passes sync.Mutex by value` is unused for linter govet (nolintlint)
                                // nolint:govet // func has predefined signature and job passes sync.Mutex by value
                                ^
```
So I need drop nolint-comments... And this is true, because I cannot change external implementation.

But when the same golangci-lint version under Linux checks the same code with the same config but without (already dropped) nolint-directive (gitlab pipeline), I/m getting linting error:
```
copylocks: func passes lock by value: github.com/zeebe-io/zeebe/clients/go/pkg/entities.Job contains github.com/zeebe-io/zeebe/clients/go/pkg/pb.ActivatedJob contains google.golang.org/protobuf/internal/impl.MessageState contains sync.Mutex (govet)
				job entities.Job,
				    ^
```
vicious circle

ashanbrown:
Because some linters may be implementation dependent based on tags or OS, we allow excluding nolint checks.  This can be done either by adding a linter to the "ignore-unused" in settings, or by adding `nolintlint` to the `nolint` directive, such as  `//nolint:my-linter,nolintlint`.

Fixes #1542.

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -331,6 +331,8 @@ linters-settings:
     require-explanation: true
     # Enable to require nolint directives to mention the specific linter being suppressed. Default is false.
     require-specific: true
+    # Ignore nolints for specific linters
+    ignore: []
   rowserrcheck:
     packages:
       - github.com/jmoiron/sqlx
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -332,7 +332,7 @@ linters-settings:
     # Enable to require nolint directives to mention the specific linter being suppressed. Default is false.
     require-specific: true
     # Ignore nolints for specific linters
-    ignore: []
+    ignore-unused: []
   rowserrcheck:
     packages:
       - github.com/jmoiron/sqlx
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -366,6 +366,7 @@ type NoLintLintSettings struct {
 	RequireSpecific    bool     `mapstructure:"require-specific"`
 	AllowNoExplanation []string `mapstructure:"allow-no-explanation"`
 	AllowUnused        bool     `mapstructure:"allow-unused"`
+	IgnoreUnused       []string `mapstructure:"ignore-unused"`
 }
 
 type TestpackageSettings struct {
--- a/pkg/golinters/nolintlint.go
+++ b/pkg/golinters/nolintlint.go
@@ -54,6 +54,12 @@ func NewNoLintLint() *goanalysis.Linter {
 			for _, n := range pass.Files {
 				nodes = append(nodes, n)
 			}
+
+			ignoredLinters := make(map[string]struct{}, len(settings.IgnoreUnused))
+			for _, l := range settings.IgnoreUnused {
+				ignoredLinters[l] = struct{}{}
+			}
+
 			issues, err := lnt.Run(pass.Fset, nodes...)
 			if err != nil {
 				return nil, fmt.Errorf("linter failed to run: %s", err)
@@ -65,6 +71,9 @@ func NewNoLintLint() *goanalysis.Linter {
 				if ii, ok := i.(nolintlint.UnusedCandidate); ok {
 					expectedNolintLinter = ii.ExpectedLinter
 					expectNoLint = true
+					if _, ignored := ignoredLinters[expectedNolintLinter]; ignored {
+						continue
+					}
 				}
 				issue := &result.Issue{
 					FromLinter:           NolintlintName,
--- a/pkg/golinters/nolintlint/nolintlint.go
+++ b/pkg/golinters/nolintlint/nolintlint.go
@@ -188,6 +188,7 @@ func (l Linter) Run(fset *token.FileSet, nodes ...ast.Node) ([]Issue, error) {
 				}
 				lintersText, explanation := fullMatches[1], fullMatches[2]
 				var linters []string
+				ignoreUnused := false
 				if len(lintersText) > 0 {
 					lls := strings.Split(lintersText[1:], ",")
 					linters = make([]string, 0, len(lls))
@@ -195,6 +196,9 @@ func (l Linter) Run(fset *token.FileSet, nodes ...ast.Node) ([]Issue, error) {
 						ll = strings.TrimSpace(ll)
 						if ll != "" {
 							linters = append(linters, ll)
+						}
+						if ll == "nolintlint" {
+							ignoreUnused = true
 						}
 					}
 				}
@@ -205,7 +209,7 @@ func (l Linter) Run(fset *token.FileSet, nodes ...ast.Node) ([]Issue, error) {
 				}
 
 				// when detecting unused directives, we send all the directives through and filter them out in the nolint processor
-				if l.needs&NeedsUnused != 0 {
+				if l.needs&NeedsUnused != 0 && !ignoreUnused {
 					if len(linters) == 0 {
 						issues = append(issues, UnusedCandidate{BaseIssue: base})
 					} else {
--- a/test/testdata/configs/nolintlint_unused.yml
+++ b/test/testdata/configs/nolintlint_unused.yml
@@ -1,0 +1,5 @@
+linters-settings:
+  nolintlint:
+    allow-unused: false
+    ignore-unused:
+      - deadcode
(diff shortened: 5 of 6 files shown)


ldez:
To try this PR, I rebased it locally.

These cases work well:
```go
package bar
//nolint:nolintlint,nakedret,gocyclo
func foo() {}
```

```go
package bar
//nolint:nakedret,gocyclo,nolintlint
func foo() {}
```

But this case is not handled correctly:
```go
package bar

//nolint:nolintlint // inception
//nolint
func foo() {}
```

The related test cases:

<details>

```go

		{
			desc:  "when ignore multi-line",
			needs: NeedsExplanation,
			contents: `
package bar

//nolint:nolintlint // inception
//nolint
func foo() {}`,
		},
		{
			desc:     "when no explanation is needed for a specific linter",
			needs:    NeedsExplanation,
			excludes: []string{"lll"},
			contents: `
package bar

func foo() {
	thisIsAReallyLongLine() //nolint:lll
}`,
		},
		{
			desc:  "when ignore one line",
			needs: NeedsExplanation,
			contents: `
package bar

//nolint:nolintlint // inception
func foo() {}`,
		},
```

</details>


---

I tried another approach: https://github.com/golangci/golangci-lint/compare/master...ldez:fix/ignore-nolintlint
Take a look at this commit https://github.com/golangci/golangci-lint/commit/497f95457da70a1a2e9eb2a1d89f3096236fe84a

WDYT? Maybe we can mix both approaches.

---

Also, I'm not sure about the `ignore` option: the `nolint:nolintlint` is more in the `nolint` way from my point of view.

ashanbrown:
@ldez I'm happy to roll with your changes if you'd like to put up a replacement PR for this one.  It'd be easier for me to review.

On another note, I'm working on changes to automatically fix nolintlint errors (just so you have a heads up... in case you were thinking along the same lines).

ashanbrown:
@ldez I was just looking over the test cases in more detail and I've incorporated one of them into the test and included it with the upstream branch merge.  I think what's at issue in the `when ignore multi-line` test case is whether we can skip having to provide an explanation.  I would question whether this is a particularly useful case because the act of adding the additional `nolint` statement for this seems worse than the cure, which is just to add an explanation, which would require no additional lines of code.

ldez:
@ashanbrown I still not a fan of the `ignore` option.

From my point of view, the fact to skip only one rule (`allow-unused`) instead of all rules is unexpected because when the `nolint:` directive is used for the other linters all the rules are ignored.

I can understand the approach because nolintlint is a linter for the nolint directives but it's a bit weird.

ashanbrown:
@ldez I think you're saying that you'd favor having `nolint:nolintlint` cause a line to behave as if the `nolintlint` linter does not apply at all to that line, matching the behavior for other linters.  Have I got that right?  Do you also favor removing the `ignore-unused` config option (I had mislabed this as just `ignore` in the example but I've corrected that).

My concern about being loose with the requirements for `nolint` is that it would allow the user to skip providing an explanation exactly where an explanation is most necessary (if that is something we require for the project).  It would also allow the line to drift vs project standards from being machine-readably formatted.  I think those are requirements that should not never be ignored because the cost to be in compliance is trivial (vs other linters, where the cost of correcting things could be much more significant).  Sorry if I'm repeating my self here; I realize it could be that you understand my reasoning here and just favor consistency over enforcement but I just want to be sure.  I really like the fact that there is no reason we need ever allow `nolint` directives to become sloppy, where there is a reason that we might choose to ignore a `nolint` due to implementation concerns.

ldez:
About `ignore`, the name was a problem but the option is also a problem: as it's based on file and neither on lines nor symbols.
It's a bit too large for me and in this case, I think that the `issues.exclude-rules` is better to do that.

There are 4 options (`allow-leading-space`, `allow-unused`, `require-explanation`, `require-specific`) and your PR only excludes one. I can understand your point of view about the `require-explanation` rule but what about the `allow-leading-space` rule?

ldez:
In fact, we are trying to resolve some edge cases with few linters/OS without handling the OS-specific behavior so the implementation cannot be "perfect".

So I will follow you :+1:

ashanbrown:
I'm starting to think that `issues.exclude-rules` could be used to resolve any issues that come up.  It would be nice if the code itself could somehow indicate which `nolint` statements were only optionally expected to be triggered, but maybe this is more trouble than it's worth.  Unless there are objections, I think I'll just close this PR.

ldez:
I agree with you, `issues.exclude-rules` can be better in this case.

PR #1572, opened 2020-12-22
Resolve custom linters' path relative to config file directory

### LINKED ISSUE #1085: Custom plugins path should be relative to the configuration file
When I run golangci-lint from a directory which is not where the configuration file (e.g.: `.golangci.yml`) exists, then the custom plugin _path_ is resolved relative to the current working directory, and not to where the configuration file resides.

In my case the issue I'm facing is that the vim linter [Ale](https://github.com/dense-analysis/ale) doesn't know what is the root directory of the go module I'm working on, and runs the linter from the package directory with something like: `cd $(dirname $file) && golangci-lint run .`

I would like the relative paths from `.golangci.yml` to be relative to the file itself, so that the resolving behaviour would be agnostic to the cwd.

In the example below, the plugin will (with latest master) be resolved as `$(pwd)/.go/bin/foobar.so`. I think it should instead be resolved as `~/some-project/.go/.bin/foobar.so`, making this deterministic regardless of the current working directory:

```
# ~/some-project/.golangci.yml
linters-settings:
  custom:
    foobar:
      path: .go/bin/foobar.so
```

It's also possible to address this issue without breaking the current behavior by taking inspiration from Jest's [`<rootDir>`](https://jestjs.io/docs/en/configuration#rootdir-string) which is being expanded when the configuration file is loaded. So we could do something like:

```
linters-settings:
  custom:
    foobar:
      path: <rootDir>/.go/bin/foobar.so
```

siliconbrain:
Resolves #1085 by interpreting non-absolute plugin paths relative to the config file's directory (or the current working directory if no config file is used).

### DIFF
--- a/pkg/lint/lintersdb/manager.go
+++ b/pkg/lint/lintersdb/manager.go
@@ -3,8 +3,10 @@ package lintersdb
 import (
 	"fmt"
 	"os"
+	"path/filepath"
 	"plugin"
 
+	"github.com/spf13/viper"
 	"golang.org/x/tools/go/analysis"
 
 	"github.com/golangci/golangci-lint/pkg/config"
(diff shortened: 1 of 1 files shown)


ernado (approved):
LGTM

PR #1575, opened 2020-12-26
Add support for gofmts, a string formatter and list sorter

ashanbrown:
gofmts allows formatting embedded json, sql and go strings and sorting groups of nodes.  The json and SQL formatting may not be ideal for all users but the goal is to favor standardization.

I'd welcome general feedback of the value of such a linter.  The original linter is at https://github.com/ashanbrown/gofmts and the readme there includes a lot of examples.

I'm a bit concerned about the implicit understanding that sort issues replace entire lines while other issues do not so I'd welcome suggestions for an API for the gofmts library that could better describe this short of returning an entire golangci-lint style replacement.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	4d63.com/gochecknoglobals v0.0.0-20201008074935-acfc0b28355a
 	github.com/Djarvur/go-err113 v0.0.0-20200511133814-5174e21577d5
 	github.com/OpenPeeDeeP/depguard v1.0.1
+	github.com/ashanbrown/gofmts v0.1.4
 	github.com/ashanbrown/makezero v0.0.0-20201205152432-7b7cdbb3025a
 	github.com/bombsimon/wsl/v3 v3.1.0
 	github.com/daixiang0/gci v0.2.7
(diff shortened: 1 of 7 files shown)


ldez:
Due to my past experiences, linking code and linters is something fragile and can be expensive when changing tooling.
Personally, I don't like to use directives in the code just to configure a specific linter.
I don't use `//nolint` for the same reasons.

Also, I don't think that using a unique formatting style in a project is really a good thing.
A piece of JSON can be on a single line because it's a very short snippet but also on several lines because the snippet is big.

In conclusion, for me, this linter is far too specific and opinionated.

ashanbrown:
@ldez Thanks for the feedback.  The background for this is that at my company we've found that formatting of embedded json and SQL strings inside go files really is a complete free-for-all and we see a lot of value in being able to enforce a standard so developers don't have to think about this.  Writing a linter based on directives was the best way I could come up to do it.  For single-line strings, the directive doesn't need to be (and is not typically) included.  Anyhow, I wanted to share this approach to gain some ideas but I'm not particularly eager to pursue it if it's contentious.

As an alternative, I was just looking into whether I could set this up as a plugin but it looks like golangci-lint doesn't yet support `https://pkg.go.dev/golang.org/x/tools/go/analysis#SuggestedFix` (although the docs for that say "This API is experimental and may change in the future." so it might be premature to try to support it).

PR #1576, opened 2020-12-26
Allow fix tests to run concurrently

ashanbrown:
I noticed the "fix" tests were taking a while to run because they don't run concurrently.  It would take about 12 seconds before and with this change takes about 4 on my local machine.

This change does two things:
1) Ensures cleanup doesn't happen until after all tests are done.
2) Disables use of the shared file lock.

I admittedly don't know what resources the shared file lock is designed to protect, but I suspect (2) is safe.  This PR does not set `t.Parallel()` at the top-level of `TestFix()`.  It would be nice to do that but again I don't know how that would play with disabling use of the file lock.


test/fix_test.go
```diff
@@ -27,7 +27,9 @@ func TestFix(t *testing.T) {
 	if os.Getenv("GL_KEEP_TEMP_FILES") == "1" {
 		t.Logf("Temp dir for fix test: %s", tmpDir)
 	} else {
-		defer os.RemoveAll(tmpDir)
+		t.Cleanup(func() {
```
ashanbrown, line 30: `t.Cleanup` waits for all tests to run, whereas `defer` does not.

test/register_cleanup.go
```diff
@@ -0,0 +1,12 @@
+// +build go1.14
```
ashanbrown, line 1: I'm not completely sure whether it's worth using build tags just to make this work but it is annoying how long the fix tests to run when doing development, so I think it's worth considering.

ldez:
I think that supporting the build of the project with an unmaintained version of Go is useless, encourages bad practice, and creates unnecessary complexity.

IMHO we can drop the support of go1.13, go1.16 will come soon so I think it's the right moment to drop go1.13.

ashanbrown:
@ldez I'd certainly be ok with deprecating go 1.13 and I agree it would be much cleaner than continuing to try to support go1.13.  Given that it's a significant decision, I'd defer to @idenx or @ernado.

ernado:
We can drop 1.13 support if it will really help us to simplify  code.

I'd better wait until go1.16 if "hack for 1.13 can be isolated

ernado:
Oops, wrong button, sorry

ashanbrown:
I don't have a simpler hack than these build tags but I'm open to suggestions.  I'm also happy to sit in this until go 1.16 is released in February.

ernado:
Build tags are ok.

ashanbrown:
@ldez What's your take on this?  Ok to merge for now or would you rather sit on this until go1.16 is released?  I'm happy to patch it up once go1.16 is out.  I could even put up a PR immediately and we could merge it in a few months.

ernado:
Let's merge it and cleanup later. I've re-checked and 10-12 lines of version specific code is minor overhead.

ernado (approved):
LGTM

ashanbrown:
@ldez Do you have any concerns about merging this as is or would you rather we wait?  As mentioned, I can put up a PR to clean this up immediately, to be merged when we stop supporting go 1.13.

ldez (commented):
For me, we can wait, but it's just my opinion.

ldez (approved):
LGTM

PR #1587, opened 2020-12-28
Add ifshort linter

esimonov:
`ifshort` is a linter for checking that the short syntax for `if`-statements is used whenever possible. For example, it will tell that the variables `ok` and `err` are unnecessarily defined outside the scope of the `if`-statements - the only places where they are referred to:

```go
func someFunc(k string, m map[string]interface{}) {
	_, ok := m[k]
	if !ok { // could be `if _, ok := m[k]; !ok`
		return
	}

	err := otherFunc1()
	if err != nil { // could be `if err := otherFunc1(); err != nil`
		otherFunc2(err)
	}
}
```

At the same time, it won't complain about the variables that actually require to be present in a wider scope.

https://github.com/esimonov/ifshort


esimonov:
Currently trying to figure out what causes CI checks to fail... 
```
level=info msg="[lintersdb] Active 38 linters: [bodyclose deadcode depguard dogsled dupl errcheck exhaustive funlen gochecknoinits goconst gocritic gocyclo gofmt goimports golint gomnd goprintffuncname gosec gosimple govet ineffassign interfacer lll misspell nakedret noctx nolintlint rowserrcheck scopelint staticcheck structcheck stylecheck typecheck unconvert unparam unused varcheck whitespace]"
level=info msg="[runner/filename_unadjuster] Pre-built 0 adjustments in 31.479618ms"
fatal error: concurrent map read and map write
```

Do you have any ideas what may be wrong please? What worries me the most is that the error log (~3500 lines) does not mention my linter at all, except for the `go: downloading github.com/esimonov/ifshort` part on the very beginning.
I'm also quite sure that my linter does not perform any concurrent operations on maps, and `go run -race` does not report any issues.
Is that something you've seen before? Any help would be much appreciated.

ldez:
@esimonov could you rebase your PR? Thank you

esimonov:
Done; test on unix are now passed, but `check_generated` is still not. Unfortunately, its error message does not tell me much:
```
     "v1.4": {
       "Error": "golangci-lint version 'v1.4' isn't supported: we support only v1.14.0 and later versions"
make: *** [fast_check_generated] Error 1
Makefile:52: recipe for target 'fast_check_generated' failed
```

Apparently, tests on MacOS also failed (didn't fail before, and I haven't changed anything in my linter since they were passing).

SVilgelm:
@esimonov I think we need to merge the #1591 and after that you can rebase over the master branch and generated should pass. MacOS fails sometimes, just re-run it

SVilgelm:
@esimonov the PR is merger, please rebase

esimonov:
Thanks, it worked!

ldez:
Hello,

have you thought about using tag/version?

There are several advantages:
- it's more "Go module idiomatic"
- that allow the dependabot to update automatically golangci-lint when you release a new version

esimonov:
@ldez Thank you for your suggestion. I updated my PR.

ldez (approved):
LGTM

Issue #1502, opened 2020-11-10
Error: build flag -mod=readonly only valid when using modules


lukasmalkmus:
Thank you for creating the issue!

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
golangci-lint has version v1.32.2 built from (unknown, mod sum: "h1:CgIeFWTLJ3Nt1w/WU1RO351j/CjN6LIVjppbJfI9nMk=") on (unknown)
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
run:
  modules-download-mode: readonly
  skip-dirs:
    - .git
    - .github
    - .vscode
    - bin
    - dist
    - docs
    - man
    - tools

linters:
  disable-all: true
  enable:
    - bodyclose
    - deadcode
    - dogsled
    - dupl
    - errcheck
    - exhaustive
    - exportloopref
    - goconst
    - gofmt
    - goimports
    - golint
    - gosec
    - gosimple
    - govet
    - ineffassign
    - misspell
    - nolintlint
    - prealloc
    - staticcheck
    - structcheck
    - stylecheck
    - typecheck
    - unconvert
    - unparam
    - unused
    - varcheck
    - whitespace

linters-settings:
  gofmt:
    simplify: true
  goimports:
    local-prefixes: github.com/lukasmalkmus/xxx
  govet:
    check-shadowing: true
```

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
go version go1.15.4 darwin/amd64
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/lukasmalkmus/Library/Caches/go-build"
GOENV="/Users/lukasmalkmus/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/lukasmalkmus/Code/Go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/lukasmalkmus/Code/Go"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
[...]

PR #1592, opened 2020-12-29
fix(dep): Pin golang.org/x/tools dependency

sayboras:
## Description

Currently, golangci-lint is failed for all repos using vendoring.

As suggested by @lukasmalkmus, pinning golang.org/x/tools to
9cf592e881e9 to fix the issue.

Relates #1502

Note: I tried to avoid replace directive, but can't without downgrading mvdan.cc/gofumpt package.

## Testing

Before

```
$ make lint
golangci-lint run
ERRO Running error: context loading failed: failed to load packages: failed to load with go/packages: err: exit status 1: stderr: build flag -mod=readonly only valid when using modules
make: *** [lint] Error 3
```

After

```
$ make lint
golangci-lint run
```


ldez:
https://github.com/golangci/golangci-lint/issues/1502#issuecomment-752069595

ldez:
I created another fix: #1593

@sayboras WDYT?

sayboras:
@ldez thanks, it's much better than pinning version 💯 . Let me close this one in favor of #1593

Issue #1589, opened 2020-12-28
forbidigo uses makezero in the lint reports


SVilgelm:
Thank you for creating the issue!

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

I enabled `forbidigo` and got these errors:

```
internal/i18n/main.go:15:2: use of `fmt.Println` forbidden by pattern `^fmt\.Print(|f|ln)$` (makezero)
        fmt.Println(string(i18n.GenerateJSON()))
        ^
pkg/cmd/gen/i18n/i18n.go:17:2: use of `fmt.Println` forbidden by pattern `^fmt\.Print(|f|ln)$` (makezero)
        fmt.Println(string(i18n.GenerateJSON()))
        ^
pkg/cmd/gen/config/config.go:55:2: use of `fmt.Println` forbidden by pattern `^fmt\.Print(|f|ln)$` (makezero)
        fmt.Println(string(data))
        ^
```

As you can see It uses `makezero` in the report

Issue was introduced in here: https://github.com/golangci/golangci-lint/pull/1569/files#diff-f0557183cd3a50b72b0c7c3477c73b79c2e8bdd805652fe7c234cc23bf661b95R49

the `forbidigoName` should be used instead of `makezeroName`

SVilgelm:
@ashanbrown I assigned issue to you, please fix it

PR #1595, opened 2020-12-29
Verify linter name in integration tests

ashanbrown:
This PR ensures that the expected linter is triggered in the integration tests (after a case where returned the wrong linter name for a linter -- see #1589).

The PR infers the name of the linter if it is the only explicitly listed linter (enabled with `-E<linter>`).  It can also be specified by adding the argument `expected_linter: ` to the test header or by adding the linter name:

    ERROR <linter name> "string"

This PR also enforces a strict syntax for error matches in tests by requiring that a string be Go-syntax string.  Note that we still don't verify line number, although we could do that in a later test.

This PR also fixes some incorrect matches in tests where we were just matching a double-quoted subsection of an expected name.


pkg/printers/text.go
```diff
@@ -52,7 +53,7 @@ func (p *Text) Print(ctx context.Context, issues []result.Issue) error {
 }
 
 func (p Text) printIssue(i *result.Issue) {
-	text := p.SprintfColored(color.FgRed, "%s", i.Text)
+	text := p.SprintfColored(color.FgRed, "%s", strings.TrimSpace(i.Text))
```
ashanbrown, line 56: This change is actually because `paralleltest` returns issues with newlines at the end, causing us to not have the linter name at the end.  `paralleltest` does not have a customer issue reporter step, so I'm not sure it's wrong to make this adjustment generally.

test/errchk.go
```diff
@@ -184,47 +203,35 @@ func wantedErrors(file, short string) (errs []wantedError) {
...
 		}
-		for _, m := range mm {
-			replacedOnce := false
-			rx := lineRx.ReplaceAllStringFunc(m[1], func(m string) string {
-				if replacedOnce {
-					return m
-				}
-				replacedOnce = true
-				n := lineNum
-				if strings.HasPrefix(m, "LINE+") {
```
ashanbrown, line 200: as far as I could tell, nothing was using this logic that appears to have to do with matching line numbers.

test/testdata/staticcheck_in_megacheck.go
```diff
@@ -18,3 +18,9 @@ func StaticcheckNolintMegacheckInMegacheck() {
 	var x int
 	x = x //nolint:megacheck
 }
+
+func Staticcheck2() {
```
ashanbrown, line 22: moved this from stylecheck_in_megacheck since it appeared to be misplaced.

test/linters_test.go
```diff
@@ -124,7 +124,6 @@ func testOneSource(t *testing.T, sourcePath string) {
 		"--allow-parallel-runners",
 		"--disable-all",
 		"--print-issued-lines=false",
-		"--print-linter-name=false",
```
ashanbrown, line 127: We need the linter name to verify it

test/testdata/default_exclude.go
```diff
@@ -4,13 +4,13 @@
 /*Package testdata ...*/
 package testdata
 
-// InvalidFuncComment, both golint and stylecheck will complain about this, // ERROR `ST1020: comment on exported function ExportedFunc1 should be of the form "ExportedFunc1 ..."`
+// InvalidFuncComment, both golint and stylecheck will complain about this, // ERROR stylecheck `ST1020: comment on exported function ExportedFunc1 should be of the form "ExportedFunc1 ..."`
```
ashanbrown, line 7: explicitly indicating the expected linter here

test/testdata/godox.go
```diff
@@ -3,11 +3,11 @@
 package testdata
 
 func todoLeftInCode() {
-	// TODO implement me // ERROR godox.go:6: Line contains FIXME/TODO: "TODO implement me"
```
ashanbrown, line 6: I believe the only thing being matched here was the stuff between double-quotes.

test/testdata/goerr113.go
```diff
@@ -4,17 +4,17 @@ package testdata
 import "os"
 
 func SimpleEqual(e1, e2 error) bool {
-	return e1 == e2 // ERROR `err113: do not compare errors directly, use errors.Is() instead: "e1 == e2"`
+	return e1 == e2 // ERROR `err113: do not compare errors directly, use errors.Is\(\) instead: "e1 == e2"`
```
ashanbrown, line 7: This change makes this a valid regex in a go string

ldez:
I wanted to merge the PR so I rebased it, but now the CI is failing on a problem that I don't reproduce...

The CI is going to drive me crazy.

ldez (approved):
LGTM

ashanbrown:
Thanks @ldez!

PR #1641, opened 2021-01-11
Update docs for v1.35.1

SVilgelm:


ldez:
@SVilgelm I think it can be good to announce the release in the team discussion before creating a release.

v1.35.1 contains linters changes, so it cannot be a patch version from my point of view.

ldez (approved):

PR #1647, opened 2021-01-12
Documentation on contributing

butuzov:
This PR fix the reference to `govet` in contributing docs as linter that has one option. Not critical, just cleaning.

### DIFF
--- a/docs/src/docs/contributing/new-linters.mdx
+++ b/docs/src/docs/contributing/new-linters.mdx
@@ -19,8 +19,8 @@ After that:
 3. Add the new struct for the linter (which you've implemented in `pkg/golinters/{yourlintername}.go`) to the
    list of all supported linters in [`pkg/lint/lintersdb/manager.go`](https://github.com/golangci/golangci-lint/blob/master/pkg/lint/lintersdb/manager.go)
    to the function `GetAllSupportedLinterConfigs`. Enable it by default only if you are sure.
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

could you explain a bit more the change?

SVilgelm (approved):
Looks good to me. Thank you

SVilgelm:
@butuzov Yeah, could you please explain why did you change the example from `govet` to `unparam`? I really don't see a reason to make this change

butuzov:
Hi Guys.

I had run into an issue yesterday, that `govet` hasn't reported for shadoowing variables/packages. That snowballed into some investigation and this PR one of the outcomes.

---- 

This is minor change to the website docs and can be left in place, but I guess we will skip the initial frustration of the person reading at the website "For example, `govet` has only 1 option: `check-shadowing`.", then looking to the GovetSettings struct and finding that there are more. I have looked for the next linter that has **only** one setting (that can be passed via CLI flag).

What I have missed is that there is another option for `unparam` (one not passed to CLI. So here are the options we have:
1) Replace `govet` with other linter with only one setting (next in the list is `nakedret`)
2) Close this issue and left `govet` in tutorial as "linter with only one option"/s

butuzov:
Again, this quite a minor thing, and not critical.

ldez (approved):
LGTM

PR #1660, opened 2021-01-21
Add docs on using homebrew tap

danielhelfand:
While use of the tap is discouraged, there are times that homebrew core doesn't immediately update the formula to the latest version. This provides docs on using the tap to help get the latest released quicker through brew.

An example of this is recently as brew install only installs v1.33 at the moment.

### DIFF
--- a/docs/src/docs/usage/install/index.mdx
+++ b/docs/src/docs/usage/install/index.mdx
@@ -57,7 +57,7 @@ Note: Previously we used a [homebrew tap](https://github.com/golangci/homebrew-t
 isn't immediately available via homebrew core due to manual updates that need to occur from homebrew core maintainers. In this case, the tap formula, which is updated automatically,
 can be used to install the latest version of `golangci-lint`:
 
-```
+```sh
 brew tap golangci/tap
 brew install golangci/tap/golangci-lint
 ```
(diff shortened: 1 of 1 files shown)


docs/src/docs/usage/install/index.mdx
```diff
@@ -53,7 +53,13 @@ brew install golangci-lint
 brew upgrade golangci-lint
 ```
 
-Note: previously we use [tap](https://github.com/golangci/homebrew-tap), but it is obsoleted. We recommend using official formula instead of an obsolete tap.
+Note: Previously we used a [homebrew tap](https://github.com/golangci/homebrew-tap). We recommend using official formula instead of the tap, but sometimes the most recent release 
+isn't immediately available via homebrew core. In this case, the tap formula can be used to install the latest version of `golangci-lint`:
```
SVilgelm, line 57: Could you please add a bit more details, like the tap formula is updated automatically, when the core requires some manual actions and can be delivered with a delay

SVilgelm (changes requested):

docs/src/docs/usage/install/index.mdx
danielhelfand: Updated, but let me know if I can be more specific with anything.

SVilgelm (approved):
Looks good to me. Thank you

PR #1670, opened 2021-01-29
fix rowserrcheck reports false positive

### LINKED ISSUE #943: How can I report some issues about rowserrcheck linter?
Hello there,

I found a few cases where [rowserrcheck](https://github.com/jingyugao/rowserrcheck) reports false positive.
I'd like to report those issues but the `rowserrcheck` repository doesn't have enabled the Github issues.

How should I report them? Should I do it in this repository and mention the author?

jingyugao:
Fixes #943

This PR fix that rowserrcheck reports false positive when handling `phi instruction`.

For example
```go
func badcase() {
	var db *sql.DB
	var rows *sql.Rows
	if rand.Float64() < 0.5 {
		rows, _ = db.Query("select 1")
	}
	if err := rows.Err(); err != nil {
		panic(err)
	}
}
```
which ssa is as below:
```
func badcase():
0:                                                                entry P:0 S:2
	t0 = math/rand.Float64()                                        float64
	t1 = t0 < 0.5:float64                                              bool
	if t1 goto 1 else 2
1:                                                              if.then P:1 S:1
	t2 = (*database/sql.DB).Query(nil:*database/sql.DB, "select 1":string, nil:[]interface{}...) (*database/sql.Rows, error)
	t3 = extract t2 #0                                   *database/sql.Rows
	t4 = extract t2 #1                                                error
	jump 2
2:                                                              if.done P:2 S:2
	// t5'Err is checked,but t5 is not t3's referrer. t5 is phi(...,t3)'referrer.
        t5 = phi [0: nil:*database/sql.Rows, 1: t3] #rows    *database/sql.Rows
	t6 = (*database/sql.Rows).Err(t5)                                 error
	t7 = t6 != nil:error                                               bool
	if t7 goto 3 else 4
3:                                                              if.then P:1 S:0
	t8 = change interface interface{} <- error (t6)             interface{}
	panic t8
4:                                                              if.done P:1 S:0
	return
```
if we move the assignment statement out of the if block.
```go
func passcase() {
	var db *sql.DB
	var rows *sql.Rows
	if rand.Float64() < 0.5 {
		rows, _ = db.Query("select 1")
	}
	if err := rows.Err(); err != nil {
		panic(err)
	}
}
```
ths ssa is below
```
0:                                                                entry P:0 S:2
	t0 = math/rand.Float64()                                        float64
	t1 = t0 < 0.5:float64                                              bool
	t2 = (*database/sql.DB).Query(nil:*database/sql.DB, "select 1":string, nil:[]interface{}...) (*database/sql.Rows, error)
	t3 = extract t2 #0                                   *database/sql.Rows
	t4 = extract t2 #1                                                error
        // here t3 is t2'referrer's referrer
	t5 = (*database/sql.Rows).Err(t3)                                 error
	t6 = t5 != nil:error                                               bool
	if t6 goto 1 else 2
1:                                                              if.then P:1 S:0
	t7 = change interface interface{} <- error (t5)             interface{}
	panic t7
2:                                                              if.done P:1 S:0
	return
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,7 +14,6 @@ require (
 	github.com/esimonov/ifshort v1.0.0
 	github.com/fatih/color v1.10.0
 	github.com/go-critic/go-critic v0.5.3
-	github.com/go-sql-driver/mysql v1.4.0
 	github.com/go-xmlfmt/xmlfmt v0.0.0-20191208150333-d5b6f63a941b
 	github.com/gofrs/flock v0.8.0
 	github.com/golangci/check v0.0.0-20180506172741-cfe4005ccda2
@@ -33,7 +32,6 @@ require (
 	github.com/jgautheron/goconst v0.0.0-20201117150253-ccae5bf973f3
 	github.com/jingyugao/rowserrcheck v0.0.0-20210130005344-c6a0c12dd98d
 	github.com/jirfag/go-printf-func-name v0.0.0-20191110105641-45db9963cdd3
-	github.com/jmoiron/sqlx v1.2.0
 	github.com/kulti/thelper v0.2.1
 	github.com/kunwardeep/paralleltest v1.0.2
 	github.com/kyoh86/exportloopref v0.1.8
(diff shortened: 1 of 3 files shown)


ldez:
@jingyugao there is an issue with the authorship of your commits, could you recreate your commit with a valid author?

also could you merge https://github.com/jingyugao/rowserrcheck/pull/13

ldez (approved):

PR #1677, opened 2021-02-01
Expose more config options for forbidigo

ashanbrown:
The latest version of forbidigo excludes godoc examples by default.  This PR adds a config option to disable that feature if so desired.  Forbidigo now also allows us to ignore its own `//permit` nolint directive in favor of `//nolint`.  I don't think it makes sense to have multiple ways of ignore errors.


ldez:
https://github.com/golangci/golangci-lint/pull/1678

ldez (approved):
LGTM

PR #1688, opened 2021-02-04
Compile binaries with -trimpath enabled

amenzhinsky:
Hi there,

This just slightly improves builds reproducibility, the short description of the flag is bellow:
```
-trimpath
	remove all file system paths from the resulting executable.
	Instead of absolute file system paths, the recorded file names
	will begin with either "go" (for the standard library),
	or a module path@version (when using modules),
	or a plain import path (when using GOPATH).
```


ldez (approved):
LGTM

PR #1720, opened 2021-02-08
Add plugin option for Goland

xxpxxxxp:
I'm writing a plugin for Goland to support `golangci-lint` inline, also providing possible quick fixes: https://github.com/xxpxxxxp/intellij-plugin-golangci-lint
Should be stable now, I'd like to suggest it as an option for Goland users.
Thank you all authors, `golangci-lint` is widely used in my org and improved code quality significantly.

### DIFF
--- a/docs/src/docs/usage/integrations.mdx
+++ b/docs/src/docs/usage/integrations.mdx
@@ -19,6 +19,7 @@ title: Integrations
 
 2. Sublime Text - [plugin](https://github.com/alecthomas/SublimeLinter-contrib-golang-cilint) for SublimeLinter.
 3. GoLand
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

what is the interest of your plugin compared to native integration (FileWatcher)?

xxpxxxxp:
Well, FileWatcher is tough to work with, one has to save the current file then find issues in the output panel, go back to fix them.
This plugin is hinting issues **inline**, providing quick fix to some issues, see example below

image

bombsimon (approved):
I've been using this plugin with IntelliJ and I think it's a nice addition instead of using file watchers. Sounds reasonable to reference this in the docs.

PR #1729, opened 2021-02-10
Add "revive" linter

### LINKED ISSUE #238: Support revive
https://github.com/mgechev/revive

sspaink:
Closes #238

This PR adds support for the [revive linter](https://github.com/mgechev/revive). Tested by running `T=revive.go make test_linters`, added a simple example config but all the configuration settings in revive have been implemented.

I assume having this linter disabled by default is what would be wanted, not sure what it takes to be enabled by default.

Just let me know if there is anything you would like me to change, thanks!


ldez:
Hello,

could you rebase your PR? thank you

test/testdata/configs/revive.yml
```diff
@@ -0,0 +1,7 @@
+linters-settings:
+  revive:
+    ignore-generated-header: true
+    severity: warning
+    rules:
+      - name: indent-error-flow
+        severity: warning
```
ldez, line 7:
```suggestion
        severity: warning

```

pkg/golinters/revive.go
```diff
@@ -0,0 +1,187 @@
...
+					},
+					LineRange: &result.Range{
+						From: results[i].Position.Start.Line,
+						To:   results[i].Position.End.Line,
+					},
+					FromLinter: reviveName,
+				}, pass))
+			}
+			return nil, nil
+		}
```
ldez, line 128:
```suggestion
			for _, r := range results {
				issues = append(issues, goanalysis.NewIssue(&result.Issue{
					Severity: string(r.Severity),
					Text:     fmt.Sprintf("%q", r.Failure.Failure),
					Pos: token.Position{
						Filename: r.Position.Start.Filename,
						Line:     r.Position.Start.Line,
						Offset:   r.Position.Start.Offset,
						Column:   r.Position.Start.Column,
					},
					LineRange: &result.Range{
						From: r.Position.Start.Line,
						To:   r.Position.End.Line,
					},
					FromLinter: reviveName,
				}, pass))
			}
```

pkg/golinters/revive.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+
+			revive := lint.New(ioutil.ReadFile)
+
+			lintingRules, err := reviveConfig.GetLintingRules(conf)
+			if err != nil {
+				return nil, err
+			}
+
+			packages, err := dots.ResolvePackages(files, normalizeSplit([]string{}))
```
ldez, line 74: `normalizeSplit([]string{})` seems useless here, because it will always return an empty slice.

pkg/golinters/revive.go
```diff
@@ -0,0 +1,187 @@
...
+			}
+			return nil, nil
+		}
+	}).WithIssuesReporter(func(*linter.Context) []goanalysis.Issue {
+		return issues
+	}).WithLoadMode(goanalysis.LoadModeSyntax)
+}
+
+func normalizeSplit(strs []string) []string {
+	res := []string{}
```
ldez, line 135:
```suggestion
	var res []string
```

pkg/golinters/revive.go
```diff
@@ -0,0 +1,187 @@
...
+		conf.Confidence = cfg.Confidence
+	}
+
+	// By default golangci-lint ignores missing doc comments, follow same convention by removing this default rule
+	// Relevant issue: https://github.com/golangci/golangci-lint/issues/456
+	delete(conf.Rules, "exported")
+
+	if len(cfg.Rules) != 0 {
+		// Clear default rules, only use rules defined in config
+		conf.Rules = map[string]lint.RuleConfig{}
```
ldez, line 169:
```suggestion
		conf.Rules = make(map[string]lint.RuleConfig, len(cfg.Rules))
```

pkg/golinters/revive.go
```diff
@@ -0,0 +1,187 @@
...
+	for _, r := range cfg.Rules {
+		conf.Rules[r.Name] = lint.RuleConfig{Arguments: r.Arguments, Severity: lint.Severity(r.Severity)}
+	}
+
+	conf.ErrorCode = cfg.ErrorCode
+	conf.WarningCode = cfg.WarningCode
+
+	if len(cfg.Directives) != 0 {
+		// Clear default Directives, only use Directives defined in config
+		conf.Directives = map[string]lint.DirectiveConfig{}
```
ldez, line 180:
```suggestion
		conf.Directives = make(map[string]lint.DirectiveConfig, len(cfg.Directives))
```

ldez (changes requested):

pkg/golinters/revive.go
```diff
@@ -0,0 +1,187 @@
...
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
+	"github.com/golangci/golangci-lint/pkg/lint/linter"
+	"github.com/golangci/golangci-lint/pkg/result"
+
+	reviveConfig "github.com/mgechev/revive/config"
+	"github.com/mgechev/revive/lint"
+	"golang.org/x/tools/go/analysis"
+)
```
ldez, line 20:
```suggestion
import (
	"encoding/json"
	"fmt"
	"go/token"
	"io/ioutil"
	"strings"

	"github.com/mgechev/dots"
	reviveConfig "github.com/mgechev/revive/config"
	"github.com/mgechev/revive/lint"
	"golang.org/x/tools/go/analysis"

	"github.com/golangci/golangci-lint/pkg/config"
	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
	"github.com/golangci/golangci-lint/pkg/lint/linter"
	"github.com/golangci/golangci-lint/pkg/result"
)
```

ldez (changes requested):

pkg/golinters/revive.go
sspaink: I agree the suggested change looks nicer, but gocritic complains about it and suggests pointers or indexing which causes the build to fail.
```
pkg\golinters\revive.go:108:4: rangeValCopy: each iteration copies 184 bytes (consider pointers or indexing) (gocritic)
                        for _, r := range results {
                        ^
```

ldez: the other suggestion a more important, so keep your current code on that point.

pkg/golinters/revive.go
sspaink: Good catch, this string slice should have contained a list of excluded packages. I decided to remove "normalizeSplit" and just pass an empty string slice because it would make more sense to just use golanci-lint's option to skip certain directories `--skip-dirs` then add another exclude setting. Hopefully that sounds good?

ldez (approved):
LGTM

PR #1732, opened 2021-02-13
Remove comparison to gometalinter in benchmarks

pkuca:
- Remove gometalinter run from bench test
- Clone remote tool repo to local repository in bench test


ldez:
As it's a breaking change, we cannot remove the linter like that.

SVilgelm:
so, it's archived, but the lates code a completely compatible with golangci-lint, personally I never seen any suggestions from this linter, but it's enabled.
Instead of removing the linter, let's mark it as deprecated and printing a warning message.

pkuca:
"mark it as deprecated and printing a warning message" - what's the recommended way to do both of those things? I may have missed other occurrences while searching.

ernado:
I'm not sure that we done such thing before, so there can be no examples.

SVilgelm:
@pkuca @ernado please see the #1755 PR, I added a feature to mark the linters as deprecated.

SVilgelm:
@pkuca The PR #1755 is merged, so we marked the interfacer as deprecated. Please rebase this PR, but don't delete interfacer

ldez (approved):
LGTM

PR #1739, opened 2021-02-16
Use upstream gocyclo.

ldez:
Gocyclo has been updated https://github.com/fzipp/gocyclo/commits/main

we don't need  a fork anymore https://github.com/golangci/gocyclo

PR #1741, opened 2021-02-17
Use upstream gocyclo library

invidian:
I guess in the past it was not possible to use it directly, hence fork
was used.

This commit also simplifies the logic a bit of the linter.

Signed-off-by: Mateusz Gozdek <mgozdekof@gmail.com>


ldez:
It's a duplicate of #1739

invidian:
Oh, thanks for pointing out @ldez. Do you mind also pulling my part of code which simplifies some bits please? I think they are valuable.

PR #1744, opened 2021-02-17
durationcheck: False positive when multiplying with int type struct field

ldez:
Related to https://github.com/charithe/durationcheck/issues/7

### DIFF
--- a/test/testdata/durationcheck.go
+++ b/test/testdata/durationcheck.go
@@ -1,9 +1,38 @@
 //args: -Edurationcheck
 package testdata
 
-import "time"
+import (
+	"fmt"
+	"time"
+)
 
-func waitFor(someDuration time.Duration) {
-	timeToWait := someDuration * time.Second // ERROR "Multiplication of durations: `someDuration \\* time.Second` "
+type durationCheckData struct {
+	i int
+	d time.Duration
+}
+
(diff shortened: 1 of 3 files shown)


test/testdata/durationcheck.go
```diff
@@ -1,9 +1,38 @@
...
+}
+
+func durationcheckCase02() {
+	dcd := durationCheckData{d: 10 * time.Second}
+	_ = dcd.d * time.Second // ERROR "Multiplication of durations: `dcd.d \\* time.Second`"
+}
+
+func durationcheckCase03() {
+	seconds := 10
+	fmt.Print(time.Duration(seconds) * time.Second)
```
bombsimon, line 26: Why doesn't this yield an error? 🤔

bombsimon (approved):
LGTM! Added a question but since it's related to upstream I'll just approve anyway.

test/testdata/durationcheck.go
ldez: because `time.Duration` is a `int64` (`type Duration int64`) so in this case: `seconds` (`int`) is just converted to `int64` (`Duration`).

```
time.Duration(seconds) * time.Second
=> 10 * time.Second
```

PR #1749, opened 2021-02-20
fix: schema error in linter-settings

ejain:
Screen Shot 2021-02-19 at 10 03 02 PM

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -310,7 +310,7 @@ linters-settings:
       - github.com/sirupsen/logrus
     packages-with-error-message:
       # specify an error message to output when a blacklisted package is used
-      - github.com/sirupsen/logrus: "logging is allowed only by logutils.Log"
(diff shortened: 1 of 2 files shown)


ejain:
Looks like YAML doesn't support dots in keys, so this doesn't work for anything but standard packages... Not sure where the schema is defined?

ejain:
Looks like the schema is defined in https://github.com/SchemaStore/schemastore/blob/master/src/schemas/json/golangci-lint.json.

ldez:
Hello, the problem seems related to viper https://github.com/spf13/viper/issues/324

ldez:
after more investigation, the fact to use a list instead of a map in the YAML is a hack to be able to use key's name with a dot.

ldez:
So the schema is right, the configuration examples also, but both are not compatible.

The list syntax is currently the only way to define the configuration, because of a bug inside viper.

I think you can open an issue for tracking this.

I will close this PR.

PR #1754, opened 2021-02-20
Add "enable-new" command that enables any linters not explicitly disabled

ashanbrown:
This command updates the identified configuration file and adds new linters to
it, trying to preserve the original file as closely as possible

This offers an alternative to the deprecated "enable-all" setting.  It allows
projects to keep up-to-date with the latest linters without introducing new
linters when the golangci-lint version is upgraded.

Running this on golangci-lint itself, the output looks like:

```
Enabling the following new linters in ".golangci.yml":
asciicheck: Simple linter to check that your code does not contain non-ASCII identifiers [fast: true, auto-fix: false]
cyclop: checks function and package cyclomatic complexity [fast: true, auto-fix: false]
durationcheck: check for two durations multiplied together [fast: true, auto-fix: false]
errorlint: go-errorlint is a source code linter for Go software that can be used to find code that will cause problemswith the error wrapping scheme introduced in Go 1.13. [fast: true, auto-fix: false]
exhaustivestruct: Checks if all struct's fields are initialized [fast: true, auto-fix: false]
exportloopref: checks for pointers to enclosing loop variables [fast: true, auto-fix: false]
forbidigo: Forbids identifiers [fast: true, auto-fix: false]
gci: Gci control golang package import order and make it always deterministic. [fast: true, auto-fix: true]
gochecknoglobals: check that no global variables exist [fast: true, auto-fix: false]
gocognit: Computes and checks the cognitive complexity of functions [fast: true, auto-fix: false]
godot: Check if comments end in a period [fast: true, auto-fix: true]
godox: Tool for detection of FIXME, TODO and other comment keywords [fast: true, auto-fix: false]
Add "enable-new" command that enables any linters that are not explicitly disabled
goerr113: Golang linter to check the errors handling expressions [fast: true, auto-fix: false]
gofumpt: Gofumpt checks whether code was gofumpt-ed. [fast: true, auto-fix: true]
goheader: Checks is file header matches to pattern [fast: true, auto-fix: false]
gomodguard: Allow and block list linter for direct Go module dependencies. This is different from depguard where there are different block types for example version constraints and module recommendations. [fast: true, auto-fix: false]
ifshort: Checks that your code uses short syntax for if-statements whenever possible [fast: true, auto-fix: false]
makezero: Finds slice declarations with non-zero initial length [fast: true, auto-fix: false]
maligned: Tool to detect Go structs that would take less memory if their fields were sorted [fast: true, auto-fix: false]
nestif: Reports deeply nested if statements [fast: true, auto-fix: false]
nlreturn: nlreturn checks for a new line before return and branch statements to increase code clarity [fast: true, auto-fix: false]
paralleltest: paralleltest detects missing usage of t.Parallel() method in your Go test [fast: true, auto-fix: false]
prealloc: Finds slice declarations that could potentially be preallocated [fast: true, auto-fix: false]
predeclared: find code that shadows one of Go's predeclared identifiers [fast: true, auto-fix: false]
revive: Fast, configurable, extensible, flexible, and beautiful linter for Go. Drop-in replacement of golint. [fast: true, auto-fix: false]
sqlclosecheck: Checks that sql.Rows and sql.Stmt are closed. [fast: true, auto-fix: false]
testpackage: linter that makes you use a separate _test package [fast: true, auto-fix: false]
thelper: thelper detects golang test helpers without t.Helper() call and checks the consistency of test helpers [fast: true, auto-fix: false]
tparallel: tparallel detects inappropriate usage of t.Parallel() method in your Go test codes [fast: true, auto-fix: false]
wrapcheck: Checks that errors returned from external packages are wrapped [fast: true, auto-fix: false]
wsl: Whitespace Linter - Forces you to use empty lines! [fast: true, auto-fix: false]```

### CODE REVIEW

**Comment by ldez** [comment]
There several formats for the config file: YAML, TOML, JSON, ...

File: go.mod [hunk 1]
```diff
@@ -77,6 +77,7 @@ require (
 	golang.org/x/text v0.3.4 // indirect
 	golang.org/x/tools v0.1.0
 	gopkg.in/yaml.v2 v2.4.0
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
```
SVilgelm, line 80: lets use either v2 or v3

SVilgelm (changes requested):

ldez:
I'm not sure about this feature, maybe we can think of a more global interactive mode.
Maybe opening up an issue to talk about it can be a good idea.

PR #1765, opened 2021-02-22
Deprecate maligned, add govet `fieldalignment` as replacement

ldez:
Last week, [maligned has been deprecated](https://github.com/mdempsky/maligned).

The recommendation is to switch to https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment

Related to #1555

Currently, we don't support `SuggestedFixes` but it's a more global topic.

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -297,6 +297,7 @@ linters-settings:
           - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
 
     # enable or disable analyzers by name
+    # run `go tool vet help` to see all analyzers
     enable:
       - atomicalign
     enable-all: false
--- a/pkg/golinters/govet.go
+++ b/pkg/golinters/govet.go
@@ -15,6 +15,7 @@ import (
 	_ "golang.org/x/tools/go/analysis/passes/ctrlflow" // unused, internal analyzer
 	"golang.org/x/tools/go/analysis/passes/deepequalerrors"
 	"golang.org/x/tools/go/analysis/passes/errorsas"
+	"golang.org/x/tools/go/analysis/passes/fieldalignment"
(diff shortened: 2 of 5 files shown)


SVilgelm (approved):

Josca:
@ldez please, what is correct way to fix message `WARN [runner] The linter 'maligned' is deprecated due to: The repository of the linter has been archived by the owner. Use govet 'fieldalignment' instead.`? Just add `govet` and remove `maligned`?

ldez:
yes you just have to disable or remove `maligned` and add `fieldalignment` to the govet configuration:

```yml
linters:
  disable:
    - maligned

linters-settings:
  govet:
    enable:
      - fieldalignment
```

```yml
linters:
  disable:
    - maligned

linters-settings:
  govet:
    enable-all: true

```

Josca:
@ldez ok, thanks, now it works and it writes a lot of warnings, e.g.:
`fieldalignment: struct with 16 pointer bytes could be 8` for:
~~~bash
type RhAccount struct {
	ID   int
	Name string
}
~~~
What does it mean, please? I can not find any explanation anywhere.

Josca:
- oh, sorry, I found that already: https://medium.com/@felipedutratine/how-to-organize-the-go-struct-in-order-to-save-memory-c78afcf59ec2

iamtakingiteasy:
> * oh, sorry, I found that already: https://medium.com/@felipedutratine/how-to-organize-the-go-struct-in-order-to-save-memory-c78afcf59ec2

This is about "struct size" `fieldalignment` linting, not about "pointer bytes". "Ponter bytes" refers to how much of a struct GC has to scan in order to check the memory.

See explaination at https://github.com/golang/go/issues/44877#issuecomment-794565908

#1825 is about making this second type of validation non-default as per author intention indicated in comment above.

PR #1777, opened 2021-02-25
feat: add stringlencompare

johejo:
https://github.com/johejo/stringlencompare

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -37,6 +37,7 @@ require (
 	github.com/jgautheron/goconst v1.4.0
 	github.com/jingyugao/rowserrcheck v1.1.0
 	github.com/jirfag/go-printf-func-name v0.0.0-20200119135958-7558a9eaa5af
+	github.com/johejo/stringlencompare v0.0.2
 	github.com/julz/importas v0.0.0-20210419104244-841f0c0fe66d
 	github.com/kisielk/errcheck v1.6.0
 	github.com/kulti/thelper v0.4.0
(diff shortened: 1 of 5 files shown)


SVilgelm:
Maybe it's better to integrate the whole https://github.com/google/syzkaller/blob/master/tools/syz-linter/linter.go instead of `stringlen`?

johejo:
I've separated it this way because it contains rules specific to the syszkaller project.

I thought this would be a better idea, assuming golangci-lint users turns on / off in the `.golangci.yaml` config.
Do you have any cool ideas?

ldez:
https://github.com/google/syzkaller has not been made as lib so it will not be fun to use it as a dependency.
Also `syzkaller` depend on `golangci-lint`: https://github.com/google/syzkaller/blob/76f7fc952d5c6a94c61aa2628568ffddb533272a/go.mod#L14

Maybe the package `tools` can be transformed as a module.

Otherwise, I'm not sure that you can use a [MIT](https://github.com/johejo/stringlencompare/blob/main/LICENSE) license and add the code from Google with [Apache2](https://github.com/google/syzkaller/blob/master/LICENSE) license like that (even with file headers).

cc @dvyukov

dvyukov:
I am happy to re-license if you to pull in that code. I am the only one who touched tools/syz-linter/* and I am happy to offload this code.

SVilgelm:
@johejo and @dvyukov, Guys, could you please make a decision about the license. We are happy to integrate the linter

johejo:
> I am happy to re-license if you to pull in that code. I am the only one who touched tools/syz-linter/* and I am happy to offload this code.

Do I need to change (properly adjust) license of stringlencompare?
 If there is no problem, rebase the PR so that it can be merged.

johejo:
The license of stringlencompare has been changed to the same Apache License 2.0 as the syzkaller project.
https://github.com/johejo/stringlencompare/releases/tag/v0.0.2

ldez:
I found a rule in go-critic which really closes: https://go-critic.github.io/overview#emptystringtest

Maybe it can be an improvement of this rule?

WDYT?

johejo:
I didn't know that go-critic had such a rule. Thanks to tell me.

go-critic doesn't seem to be able to detect this case.
https://github.com/golangci/golangci-lint/blob/3b0455cba13a64ff18b553dda4b8a3b1d42dd35e/test/testdata/stringlencompare.go#L10

```
$ gocritic check -enable="emptyStringTest" ./test/testdata/stringlencompare.go
./test/testdata/stringlencompare.go:6:5: emptyStringTest: replace `len(str) == 0` with `str == ""`
```

johejo:
Can you tell me what additional cases stringlencompare should support?

quasilyte:
Note that go-critic, a linter that is already included in golangci-lint, has this check.
https://go-critic.com/overview.html#emptyStringTest-ref

ldez:
My suggestion is to improve the go-critic rule instead of adding a new linter.
@quasilyte seems to agree with that, then I will close this PR.

PR #1786, opened 2021-02-26
Using a version instead of commit id for goconst

SVilgelm:
Just an updating a reference in the go, mod file, the version is same.
There was a mistake in the tag names (no `v` prefix), that was fixed: https://github.com/jgautheron/goconst/issues/16


ldez (approved):
LGTM

Issue #1787, opened 2021-02-26
Add linters from https://github.com/gostaticanalysis


ldez:
I propose to introduce the following linters:

- [x] https://github.com/gostaticanalysis/nilerr: finds code that returns nil even though it checks that error is not nil.
- [x] https://github.com/gostaticanalysis/forcetypeassert: finds forced type assertions.

PR #1788, opened 2021-02-26
Add nilerr linter.

ldez:
Finds the code that returns nil even if it checks that the error is not nil.

https://github.com/gostaticanalysis/nilerr

Related to  #1787


SVilgelm (approved):
Looks good to me. Thank you

PR #1793, opened 2021-02-28
Use same default linter as go vet.

ldez:
https://github.com/golang/go/blob/5ff7ec98b7727b3641df25200345b1aa50b6ff35/src/cmd/vet/main.go#L39-L66

also add missing `framepointer`


SVilgelm (approved):
Looks good to me. Thank you

PR #1817, opened 2021-03-05
Add gomoddirectives linter.

ldez:
https://github.com/ldez/gomoddirectives

```yml
  gomoddirectives:
    # Allow local `replace` directives. Default is false.
    replace-local: false
    # List of allowed `replace` directives. Default is empty.
    replace-allow-list:
      - launchpad.net/gocheck
    # Allow to not explain why the version has been retracted in the `retract` directives. Default is false.
    retract-allow-no-explanation: false
    # Forbid the use of the `exclude` directives. Default is false.
    exclude-forbidden: false
```


.golangci.example.yml
```diff
@@ -449,6 +449,12 @@ linters-settings:
     servingv1: knative.dev/serving/pkg/apis/serving/v1
     # using `autoscalingv1alpha1` alias for `knative.dev/serving/pkg/apis/autoscaling/v1alpha1` package
     autoscalingv1alpha1: knative.dev/serving/pkg/apis/autoscaling/v1alpha1
+  gomodreplace:
```
carnott-snap, line 452: I know my linter did not, so feel free to say we should add `gomodxxx`, but any chance we can expand the scope to restrict exclude and retract statements.

`retract` is new, and does generate stdout warnings when go-getting, but the code will build and run without them.

[`exclude`](https://github.com/golang/go/wiki/Modules#gomod) seems to be an escape hatch like replace, but for resolving conflicts or removing broken modules. it has a similar virality (must be in ALL main modules) to `replace`, so my gut say disallow.

ldez: yes the scope can be expanded. I need to get more information about `exclude` directives.

about `retract` directives I don't see what rule do you want to apply, could you explain?

.golangci.example.yml
```diff
@@ -449,6 +449,12 @@ linters-settings:
...
+    # Allow local `replace` directives. Default is false.
+    local: false
```
carnott-snap, line 454: I am failing to think of a use case where I want to disallow all replace except for local. (please correct me if my understanding of how this works is wrong.) iirc we discussed the following use cases:
- disallow only local replace
- disallow all replace
- disallow all replace except for some modules

ldez: a use case: https://github.com/hashicorp/consul/blob/master/go.mod

.golangci.example.yml
ldez: the `exclude` are not transitive but I don't see how to handle that differently.

https://golang.org/ref/mod#go-mod-file-exclude

carnott-snap: for my repos, the lint for `exclude` and `replace` would be the same: disallow all. later we could add nuance to allow some `exclude` cases, but I have not seen it in the wild.

`retract` could be tricky. the ask is to "disallow any dependencies that themselves are retracted". now that I lay this out, it may be decent to do this as a separate linter, since the scope of "recurse through all dependencies and detect if their version is retracted" could get complex. your call. we may also run up against the fact that this resolution information only exists in `cmd/go/internal`.

another ask for `retract` is more a syntax thing: we should probably enforce that `modfile.File.Retract.Rationale != ""`.

.golangci.example.yml
carnott-snap: can you elaborate on how it is used or why it is required?

if this is a library it means that tests run from within and without would behave differently, something that seems bad. if this is a main module, it still seems gross to maintain, and I think using the `allow-list` syntax is cleaner.

.golangci.example.yml
```diff
@@ -449,6 +449,12 @@ linters-settings:
...
+    # List of allowed `replace` directives. Default is empty.
+    allow-list:
+      - launchpad.net/gocheck
```
carnott-snap, line 457: which of these does this allow:
- `replace launchpad.net/gocheck v0.0.0 => /path/to/gocheck`
- `replace launchpad.net/gocheck v0.0.0 => example.com/path/to/gocheck v0.0.0`
- `replace example.com/path/to/gocheck v0.0.0 => launchpad.net/gocheck v0.0.0`

ldez: currently only the last

.golangci.example.yml
ldez: I don't know exactly why they are using that, I saw that pattern several times.

I think it is a way of facilitating the development with the multi modules

carnott-snap: we can add it if people really want it, but that sounds like a things for local development, not something that should be merged or released

.golangci.example.yml
carnott-snap: that seems reasonable, as long as it is clear that `local: true` is not required to get the second

ldez: I change the behavior https://github.com/ldez/gomodreplace/blob/13fd25eefbe7e332e3d67b7497365363ae58e59f/gomodreplace_test.go#L29

.golangci.example.yml
```diff
@@ -449,6 +449,16 @@ linters-settings:
     servingv1: knative.dev/serving/pkg/apis/serving/v1
     # using `autoscalingv1alpha1` alias for `knative.dev/serving/pkg/apis/autoscaling/v1alpha1` package
     autoscalingv1alpha1: knative.dev/serving/pkg/apis/autoscaling/v1alpha1
+  gomoddirectives:
+    # Allow local `replace` directives. Default is false.
+    replace-local: false
+    # List of allowed `replace` directives. Default is empty.
+    replace-allow-list:
```
carnott-snap, line 456: would nesting not be easier to maintain:
```suggestion
    replace:
      # Allow local `replace` directives. Default is false.
      local: false
      # List of allowed `replace` directives. Default is empty.
      allow-list:
```

.golangci.example.yml
```diff
@@ -449,6 +449,16 @@ linters-settings:
...
+      - launchpad.net/gocheck
+    # Allow to not explain why the version has been retracted in the `retract` directives. Default is false.
+    retract-allow-no-explanation: false
+    # Forbid the use of the `exclude` directives. Default is false.
+    exclude-forbidden: false
```
carnott-snap, line 461: any reason to not use the same default as `replace`:
```suggestion
    # Allow the use of the `exclude` directives. Default is false.
    exclude: false
```

pkg/golinters/gomoddirectives.go
```diff
@@ -0,0 +1,64 @@
...
+func NewGoModDirectives(settings *config.GoModDirectivesSettings) *goanalysis.Linter {
+	var issues []goanalysis.Issue
+	var once sync.Once
+
+	var opts gomoddirectives.Options
+	if settings != nil {
+		opts.ReplaceAllowLocal = settings.ReplaceLocal
+		opts.ReplaceAllowList = settings.ReplaceAllowList
+		opts.RetractAllowNoExplanation = settings.RetractAllowNoExplanation
+		opts.ExcludeForbidden = settings.ExcludeForbidden
```
carnott-snap, line 27: if you use a custom type or line up the structs, you can do a direct assignment:
```suggestion
		opts = gomoddirectives.Options(settings)
```

.golangci.example.yml
ldez: I don't want to do that

.golangci.example.yml
ldez: because I don't want to exclude the use of `exclude` by default.

carnott-snap: then I would think we want to make `replace` off by default too?

pkg/golinters/gomoddirectives.go
ldez: it's not the same type, so it's not possible.

.golangci.example.yml
ldez: it's not the same goal ihmo, so I prefer is ban `replace` by default but allow `exclude` by default.

pkg/golinters/gomoddirectives.go
carnott-snap: actually, it is: https://play.golang.org/p/Gna-Uv1PAaS

.golangci.example.yml
carnott-snap: out of curiosity, why?

pkg/golinters/gomoddirectives.go
ldez:
```
# github.com/golangci/golangci-lint/pkg/golinters
pkg/golinters/gomoddirectives.go:24:33: cannot convert *settings (type "github.com/golangci/golangci-lint/pkg/config".GoModDirectivesSettings) to type gomoddirectives.Options
```

ldez: https://play.golang.org/p/0PbSOa7_aI4

carnott-snap (approved):
looks sane, meets the use cases we discussed; I have not tested this locally, but will trust the author/ci to assert its correctness. should we add some tests to verify anything, or are we trusting that is done in the upstream library?

ldez:
@carnott-snap
> should we add some tests to verify anything,

It's not possible with the current test system, but I added a lot of tests in my lib.

> or are we trusting that is done in the upstream library?

FYI I'm a maintainer of golangci-lint, so yes we trust :smile:

pkg/golinters/gomoddirectives.go
carnott-snap: you changed the name of `ReplaceAllowLocal`: https://play.golang.org/p/o-97b9emRD_X

bombsimon (approved):
I trust you @ldez! ❤️

PR #1862, opened 2021-03-21
fix: wrong linters load mode

ldez:
The `LoadMode` `LoadModeTypesInfo` in the `goanalysis.Linter` says that the linter needs to resolve all the types.

The `LoadMode` `WithLoadForGoAnalysis` in the `linter.Config` says that the linter needs more information about dependencies and sizes.

Each of these loading modes has an impact on how the linter can be executed, how the cache is used, and on the speed.

For example, if a linter uses the `LoadMode` `LoadModeTypesInfo` in the `goanalysis.Linter`, this linter will no be able to analyze a single file with elements not defined in that file.

Another example, the impact of the wrong load mode on the speed of `ineffassign`:

|        | real | user  | sys  |
|--------|------|-------|------|
| before | 2.88 | 12.71 | 1.28 |
| after  | 1.53 | 4.39  | 0.64 |

**The loading mode depends on the internal requirements of a linter, it's neither a way to optimize speed nor a way to allow a linter to be executed on a single file.**
But if the loading mode is wrong, it has an impact on the capacities and the performance of a linter.

### DIFF
--- a/pkg/golinters/forcetypeassert.go
+++ b/pkg/golinters/forcetypeassert.go
@@ -15,5 +15,5 @@ func NewForceTypeAssert() *goanalysis.Linter {
 		"finds forced type assertions",
 		[]*analysis.Analyzer{a},
 		nil,
-	).WithLoadMode(goanalysis.LoadModeTypesInfo)
+	).WithLoadMode(goanalysis.LoadModeSyntax)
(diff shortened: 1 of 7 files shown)


bombsimon (approved):
LGTM

PR #1871, opened 2021-03-25
docs: add cache directory information.

### LINKED ISSUE #734: GOLANGCI_LINT_CACHE change not documented
We run golangci-lint in a docker container. Up through version 1.18.0, we just set GOCACHE to /tmp/.cache and everything worked fine. When upgrading to 1.19.0, it failed with a message "failed to initialize build cache at /.cache/golangci-lint: mkdir /.cache: permission denied". This seems to be because golangci-lint was changed to use `GOLANGCI_LINT_CACHE` for its cache directory, or `os.UserCacheDir` if that is not defined. But this change is not mentioned in the README or the release notes.

ldez:
Fixes #734

### DIFF
--- a/docs/src/docs/usage/configuration.mdx
+++ b/docs/src/docs/usage/configuration.mdx
@@ -57,6 +57,6 @@ than the default and have more strict settings:
 
 ## Cache
 
-GolangCI-Lint stores his cache into the [default user cache directory](https://golang.org/pkg/os/#UserCacheDir).
+GolangCI-Lint stores its cache in the [default user cache directory](https://golang.org/pkg/os/#UserCacheDir).
 
-You can override the default cache directory with the environment variable `GOLANGCI_LINT_CACHE`, the path must be an absolute path.
+You can override the default cache directory with the environment variable `GOLANGCI_LINT_CACHE`; the path must be absolute.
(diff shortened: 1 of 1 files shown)


docs/src/docs/usage/configuration.mdx
```diff
@@ -54,3 +54,9 @@ than the default and have more strict settings:
 ```yaml
 { .GolangciYaml }
 ```
+
+## Cache
+
+GolangCI-Lint stores his cache into the [default user cache directory](https://golang.org/pkg/os/#UserCacheDir).
```
AlekSi, line 60:
```suggestion
GolangCI-Lint stores its cache in the [default user cache directory](https://golang.org/pkg/os/#UserCacheDir).
```

Definitely not "his", and probably not "into" (but I'm not a native speaker too).

docs/src/docs/usage/configuration.mdx
```diff
@@ -54,3 +54,9 @@ than the default and have more strict settings:
...
+
+You can override the default cache directory with the environment variable `GOLANGCI_LINT_CACHE`, the path must be an absolute path.
```
AlekSi, line 62:
```suggestion
You can override the default cache directory with the environment variable `GOLANGCI_LINT_CACHE`; the path must be absolute.
```

Just to avoid repetition.

AlekSi (changes requested):

AlekSi:
(and thank you!)

bombsimon (approved):

PR #1906, opened 2021-04-11
Add tagliatelle linter

ldez:
tagliatelle is a linter that handles struct tags.

for now, it only handle the string casing.

https://github.com/ldez/tagliatelle

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -478,6 +478,20 @@ linters-settings:
     retract-allow-no-explanation: false
     # Forbid the use of the `exclude` directives. Default is false.
     exclude-forbidden: false
+  tagliatelle:
+    # check the struck tag name case
+    case:
+      # use the struct field name to check the name of the struct tag
+      use-field-name: true
+      rules:
+        # any struct tag type can be used.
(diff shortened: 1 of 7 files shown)


.golangci.yml
```diff
@@ -128,11 +128,6 @@ issues:
       linters:
         - gomnd
 
-    # https://github.com/go-critic/go-critic/issues/926
-    - linters:
-        - gocritic
-      text: "unnecessaryDefer:"
-
```
bombsimon, line 135: nit: Would this fit better in a separate PR?

bombsimon (approved):
Nice linter, lgtm! Would be nice with examples or links to each styling in the upstream repo but that's not really a part of this PR and more a nice to have. Might not be obvious to everyone what the difference between f.ex. `camel` and `goCamel` is.

.golangci.yml
ldez: I used that to purge the cache :wink:

ldez:
I improved the readme to add illustrations of the different string casing.

https://github.com/ldez/tagliatelle

PR #1917, opened 2021-04-19
fix: comma in exclude pattern leads to unexpected results

### LINKED ISSUE #665: Comma in exclude pattern leads to unexpected results
## Bug Report

I tried to configure an exclude pattern to get rid of some goconst errors. These errors take the form of

```
string `get` has 12 occurrences, make it a constant
```

So I made an exclude filter like this:

```yaml
issues:
  exclude:
  - "string `get` has [0-9]+ occurences, make it a constant"
```

and then noticed that *all* goconst errors disappeared. After adding some debugging statements to the code, I could see that the final assembled regex is

```
(?i)(string `get` has [0-9]+ occurences| make it a constant|Error return value of .((os\.)?std(out|err)\..*|.*Close|.*Flush|os\.Remove(All)?|.*printf?|os\.(Un)?Setenv). is not checked|(comment on exported (method|function|type|const)|should have( a package)? comment|comment should be of the form)|func name will be used as test\.Test.* by other packages, and that stutters; consider calling this|(possible misuse of unsafe.Pointer|should have signature)|ineffective break statement. Did you mean to break out of the outer loop|Use of unsafe calls should be audited|Subprocess launch(ed with variable|ing should be audited)|G104|(Expect directory permissions to be 0750 or less|Expect file permissions to be 0600 or less)|Potential file inclusion via variable)
```

Notice how the comma turned the exclusion pattern into two distinct elements. The documentation in the readme just says

> List of regexps of issue texts to exclude, empty list by default.

I don't really know why this is tagged as a mapstructure in the code, but right now the behaviour or the documentation is misleading.

## Version of golangci-lint:

1.17.1 or master branch

## Config file: `cat .golangci.yml`

```
linters:
  enable:
   - goconst
  disable-all: true

issues:
  exclude:
  - iamnotarealstring, make it a constant
```

## Go environment: `go version && go env`

```
go version go1.13 linux/amd64
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/xrstf/.cache/go-build"
GOENV="/home/xrstf/.config/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GONOPROXY=""
GONOSUMDB=""
GOOS="linux"
GOPATH="/home/xrstf/gospace"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/linux_amd64"
GCCGO="gccgo"
AR="ar"
CC="gcc"
CXX="g++"
CGO_ENABLED="1"
GOMOD=""
CGO_CFLAGS="-g -O2"
CGO_CPPFLAGS=""
CGO_CXXFLAGS="-g -O2"
CGO_FFLAGS="-g -O2"
CGO_LDFLAGS="-g -O2"
PKG_CONFIG="pkg-config"
GOGCCFLAGS="-fPIC -m64 -pthread -fmessage-length=0 -fdebug-prefix-map=/tmp/go-build282403496=/tmp/go-build -gno-record-gcc-switches"
```

cc @alvaroaleman

ldez:
Fixes #665

to check the PR:

<details>
<summary>main.go</summary>

```go
package main

import (
	"fmt"
)

func main() {
	foo := "get"
	bar := "get"
	more := "get"
	again := "get"

	fmt.Println(foo+bar+more, again)

	goo()
}

func goo() {
	foo := "example"
	bar := "example"
	more := "example"
	again := "example"

	fmt.Println(foo+bar+more, again)
}
```

</details>

<details>

```yml
linters:
  disable-all: true
  enable:
    - goconst

issues:
  exclude:
    - string .get. has \d+ occurrences, make it a constant
```

</details>

before the PR:
```console
$ golangci-lint run
```

after the PR:
```console
$ ./golangci-lint run
main.go:19:9: string `example` has 4 occurrences, make it a constant (goconst)
        foo := "example"
               ^
```

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -313,24 +313,15 @@ func fixSlicesFlags(fs *pflag.FlagSet) {
 			return
 		}
 
-		// custom join to handle string with comma.
-		var g string
-		for i, v := range s {
-			if strings.Contains(v, ",") {
-				// add quotes to escape comma because spf13/pflag use a CSV parser:
-				// https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/string_slice.go#L43
-				g += `"` + v + `"`
-			} else {
-				g += v
-			}
-
-			if i < len(s)-1 {
-				g += ","
-			}
+		var safe []string
+		for _, v := range s {
+			// add quotes to escape comma because spf13/pflag use a CSV parser:
+			// https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/string_slice.go#L43
+			safe = append(safe, `"`+v+`"`)
 		}
 
 		// calling Set sets Changed to true: next Set calls will append, not overwrite
-		_ = f.Value.Set(g)
+		_ = f.Value.Set(strings.Join(safe, ","))
 	})
 }
 
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -313,8 +313,24 @@ func fixSlicesFlags(fs *pflag.FlagSet) {
 			return
 		}
 
+		// custom join to handle string with comma.
+		var g string
+		for i, v := range s {
+			if strings.Contains(v, ",") {
+				// add quotes to escape comma because spf13/pflag use a CSV parser:
+				// https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/string_slice.go#L43
+				g += `"` + v + `"`
+			} else {
+				g += v
+			}
+
+			if i < len(s)-1 {
+				g += ","
+			}
+		}
+
 		// calling Set sets Changed to true: next Set calls will append, not overwrite
-		_ = f.Value.Set(strings.Join(s, ","))
+		_ = f.Value.Set(g)
 	})
 }
 
--- a/pkg/config/reader.go
+++ b/pkg/config/reader.go
@@ -7,7 +7,7 @@ import (
 	"path/filepath"
 	"strings"
 
-	homedir "github.com/mitchellh/go-homedir"
+	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/viper"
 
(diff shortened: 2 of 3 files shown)


pkg/commands/run.go
```diff
@@ -313,8 +313,24 @@ func fixSlicesFlags(fs *pflag.FlagSet) {
...
+			}
+
+			if i < len(s)-1 {
+				g += ","
+			}
+		}
+
 		// calling Set sets Changed to true: next Set calls will append, not overwrite
-		_ = f.Value.Set(strings.Join(s, ","))
+		_ = f.Value.Set(g)
```
bombsimon, line 333: Wouldn't it be easier to read and less complex if we just added quotes around all flags so the CSV parser would just do the right thing? 

```suggestion
		// add quotes to escape comma because spf13/pflag use a CSV parser:
		// https://github.com/spf13/pflag/blob/85dd5c8bc61cfa382fecd072378089d4e856579d/string_slice.go#L43
		for i, v := range s {
			s[i] = "` + v + `"`
		}
		

		_ = f.Value.Set(strings.Join(s, ","))
```

Or if we need to keep `s`, create a new `[]string` and insert the quoted flags to those.

ldez: It's less efficient and less precise but it's possible.

bombsimon: Doesn't `strings.Contains` have to iterate over every byte in every flag just do determine the branch? Also flags are limited in amount so I can't imagine one could save more than a couple of microseconds on this difference and thus for me personally readability and maintainability is more important.

ldez: I don't know why but the suggestion buttons are disabled so I cannot accept your suggestion, so I will do it by hand.

bombsimon (approved):
LGTM!

PR #1922, opened 2021-04-21
Update wsl to 3.3.0, sort config in example config

bombsimon:
* Allow forcing short declarations to cuddle only with themselves
* Allow a blank line after each comment group at the start of a block (not just between groups)

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -518,26 +518,19 @@ linters-settings:
     multi-func: false # Enforces newlines (or comments) after every multi-line function signature
 
   wsl:
-    # If true append is only allowed to be cuddled if appending value is
-    # matching variables, fields or types on line above. Default is true.
-    strict-append: true
-    # Allow calls and assignments to be cuddled as long as the lines have any
-    # matching variables, fields or types. Default is true.
-    allow-assign-and-call: true
-    # Allow assignments to be cuddled with anything. Default is false.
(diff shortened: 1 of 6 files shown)


.golangci.example.yml
```diff
@@ -518,26 +518,19 @@ linters-settings:
...
   wsl:
-    # If true append is only allowed to be cuddled if appending value is
-    # matching variables, fields or types on line above. Default is true.
-    strict-append: true
-    # Allow calls and assignments to be cuddled as long as the lines have any
-    # matching variables, fields or types. Default is true.
-    allow-assign-and-call: true
-    # Allow assignments to be cuddled with anything. Default is false.
+    # See https://github.com/bombsimon/wsl/blob/master/doc/configuration.md for
+    # documentation of available settings. These are the defaults for
```
bombsimon, line 522: Not sure if this is the best way to link to an external document but it felt easier to not fill this file with documentation and easier to show examples of what each option does. Open to reverting this if it's a bad idea.

bombsimon:
CI seems really flaky, have to restart unit tests several times, mainly macOS failing. Also not sure why Nancy is blocking on `jwt-go` in this PR. Tried to look at the latest PRs but the jwt dependency hasn't been changed for a long time.

ldez:
The problem with `jwt-go` is not related to a change inside your PR but to the declaration of a CVE on this dependent.

We can skip because it's an indirect and unused dependency.

ldez (approved):
LGTM :+1:

PR #1925, opened 2021-04-22
update go-critic to v0.5.6

pierrre:
Fix https://github.com/golangci/golangci-lint/issues/1923

PR #1926, opened 2021-04-22
Set the minimum Go version to build the binary to go1.15

ldez:
We will try to support only the latest two versions of Go to following the [release policy of Go](https://golang.org/doc/devel/release.html#policy)

Since the recommended way to install `golangci-lint` is to download and/or install a pre-built binary we see don't see a major problem with this.

Linked to #1925

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -66,7 +66,6 @@ jobs:
     strategy:
       matrix:
         golang:
-          - 1.14
           - 1.15
           - 1.16
     steps:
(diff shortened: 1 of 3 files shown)


ldez:
@ernado could you change the [check settings](https://github.com/golangci/golangci-lint/settings/branches) to remove `tests-on-unix (1.14)` as an expected check? Thank you.

SVilgelm (approved):

bombsimon (approved):

denis-tingaikin (approved):

ernado (approved):
Repo settings were updated to reflect this change.

PR #1929, opened 2021-04-24
dev: cleanup config package

ldez:
The purpose of this cleaning is to facilitate contribution and maintenance.

The PR provides a homogeneous design:
- each configuration section have a dedicated file
- no anonymous structure
- the linters in the main `Config` structure are sorted alphabetically and the linters settings structures are also sorted alphabetically.

I created one commit by change like that it's easier to review.


bombsimon (approved):
Nice, looks much better structured and sorted! 🎉 

Not sure the last commit is really provides increased maintainability, to me it's a bit stuttering to prefix every file in a directory with the same name (`config/config_*`) but I guess that's just personal preference.

<sup>Yes, I saw that `reader.go` was left but tbh I don't really see why this one wouldn't be prefixed as well since it's reading/parsing config.</sup>

bombsimon:
I have no strong feeling hence the approve! If you think this is better than keeping the names go ahead and merge or wait for an additional review. I just appreciate your effort in maintaining this project, really appreciated!

ldez:
I think I will drop the last commit because we are, at least 2, not to be a fan of the prefix.

PR #1930, opened 2021-04-24
gosec: add configuration

### LINKED ISSUE #177: Support gosec (gas) linter options
It would be nice to support `gas` linter options like `-include` and `-exclude` as part of the `linters-settings` in the golangci-lint config files.
Thanks.

ldez:
Fixes #177


bombsimon (approved):
Nice!

Issue #1868, opened 2021-03-24
version: prints to Stderr instead of StdOut


clebs:
- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).
- [x] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

<details><summary>Commands print to Stderr instead of StdOut</summary>

Trying to fetch the version in a shell script: `CI_VERSION=$(golangci-lint version --format short)` I noticed the variable was empty and I could see the version printed in the terminal.
After investigating I found out that `golangci-lint` uses Cobra's `Println` function:
https://github.com/golangci/golangci-lint/blob/c1d7cfd721401395e6a454b6af55bfc2dc1089d1/pkg/commands/version.go#L35-L38

According to Cobra, that function prints to `Stderr` per default:
https://github.com/spf13/cobra/blob/b55fa79836c63eab81e884e6ec1d7e4ae5c7f9b5/command.go#L1201

</details>

<details><summary>Version of golangci-lint</summary>
Installed automatically with VS Code

```console
$ golangci-lint --version
golangci-lint has version v1.38.0 built from (unknown, mod sum: "h1:hgZsLRzZrjhpp44Ak+fhXNzgrbDF39ETf22a+Jd3fJQ=") on (unknown)
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
run:
  modules-download-mode: readonly
```

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
go version go1.16.2 darwin/amd64
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="<redacted>/Library/Caches/go-build"
GOENV="<redacted>/Library/Application Support/go/env"
GOEXE=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="<redacted>/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="<redacted>"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/Cellar/go/1.16.2/libexec"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
[...]

PR #1869, opened 2021-03-24
Set version command output to Stdout

clebs:
Default command output was set to Stderr, now set to Stdout.

Fixes #1868

PR #1932, opened 2021-04-25
Set stderr and stdout explicitly

kamilsk:
Related to: https://github.com/golangci/golangci-lint/pull/1869

Motivation: the patch uses stdout implicitly and affects testability.

Tested:

```bash
$ go run ./cmd/golangci-lint/ version 2>/dev/null
golangci-lint has version (devel) built from (unknown, mod sum: "") on (unknown)
$ go run ./cmd/golangci-lint/ version 1>/dev/null
```

### DIFF
--- a/pkg/commands/executor.go
+++ b/pkg/commands/executor.go
@@ -30,6 +30,11 @@ import (
 	"github.com/golangci/golangci-lint/pkg/timeutils"
 )
 
+var (
+	stderr = io.Writer(os.Stderr)
+	stdout = io.Writer(os.Stdout)
+)
+
 type Executor struct {
 	rootCmd    *cobra.Command
 	runCmd     *cobra.Command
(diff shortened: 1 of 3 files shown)


kamilsk:
I think that only `main` function should decide what input/output it uses. For example: https://github.com/octomation/go-tool/blob/271d0b52b43c560e24c2e66615a2d1c49347a1a3/main.go#L42-L43

It will require a bit more editing, but I can do it if needed.

ldez:
What is the problem that solves this PR? :thinking:

kamilsk:
It solves the original problem https://github.com/golangci/golangci-lint/issues/1868, but uses an explicit approach instead of implicitly.

kamilsk:
I suggest to restore the original code (revert https://github.com/golangci/golangci-lint/pull/1869/files) and configure stderr/stdout explicitly.

ldez:
For me, #1869 fixes #1868, what is the value to do it differently?

kamilsk:
After #1869 the only possible way to test it is integration test, because you need to check os.Stdout. The pr fixes that and you can write unit test instead of integration test to check output, just provide buffer as stdout/stderr.

kamilsk:
Feel free to close the pr but please provide an argue.

Issue #1944, opened 2021-04-29
'Windows_NT' got converted to 'windows_nt' which is not a GOOS value.


mayocream:
- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).
- [ ] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

<details><summary>Go environment</summary>

```console
❯ go version && go env
go version go1.16 windows/amd64
set GO111MODULE=on
set GOARCH=amd64
set GOBIN=
set GOCACHE=C:\Users\Mayo\AppData\Local\go-build
set GOENV=C:\Users\Mayo\AppData\Roaming\go\env
set GOEXE=.exe
set GOFLAGS=
set GOHOSTARCH=amd64
set GOHOSTOS=windows
set GOINSECURE=
set GOMODCACHE=C:\Users\Mayo\go\pkg\mod
set GONOPROXY=
set GONOSUMDB=
set GOOS=windows
set GOPATH=C:\Users\Mayo\go
set GOPRIVATE=
set GOPROXY=https://goproxy.cn,direct
set GOROOT=C:\ProgramData\scoop\apps\go\current
set GOSUMDB=sum.golang.org
set GOTMPDIR=
set GOTOOLDIR=C:\ProgramData\scoop\apps\go\current\pkg\tool\windows_amd64
set GOVCS=
set GOVERSION=go1.16
set GCCGO=gccgo
set AR=ar
set CC=gcc
set CXX=g++
set CGO_ENABLED=1
set GOMOD=NUL
set CGO_CFLAGS=-g -O2
set CGO_CPPFLAGS=
set CGO_CXXFLAGS=-g -O2
set CGO_FFLAGS=-g -O2
set CGO_LDFLAGS=-g -O2
set PKG_CONFIG=pkg-config
set GOGCCFLAGS=-m64 -mthreads -fmessage-length=0 -fdebug-prefix-map=C:\Users\Mayo\AppData\Local\Temp\go-build434575085=/tmp/go-build -gno-record-gcc-switches
```

</details>

```bash
❯ curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.39.0
[...]

PR #1945, opened 2021-04-29
fix(install): windows os type

mayocream:
ref: https://github.com/golangci/golangci-lint/issues/1944


ldez:
Let's discuss in #1944 before trying to create a fix.

PR #1947, opened 2021-04-30
Update Wrapcheck to v2, add configuration

tomarrell:
Another wrapcheck update, this time with some configuration.

Recently added support to specify custom signatures to ignore when checking for wrapped errors making it more useful to people who want to use non-stdlib error wrapping.

Have given it a whirl, all looks good.

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -551,6 +551,9 @@ linters-settings:
     multi-func: false # Enforces newlines (or comments) after every multi-line function signature
 
   wrapcheck:
+    # See https://github.com/tomarrell/wrapcheck#configuration for documentation
+    # about available settings. The set of signatures below is considered the
+    # default set if no sigs are provided.
     ignoreSigs:
       - .Errorf(
       - errors.New(
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -551,11 +551,9 @@ linters-settings:
     multi-func: false # Enforces newlines (or comments) after every multi-line function signature
 
   wrapcheck:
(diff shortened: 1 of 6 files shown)


.golangci.example.yml
```diff
@@ -550,6 +550,15 @@ linters-settings:
     multi-if: false   # Enforces newlines (or comments) after every multi-line if statement
     multi-func: false # Enforces newlines (or comments) after every multi-line function signature
 
+  wrapcheck:
+    ignoreSigs:
```
ldez, line 557:
```suggestion
    # An array of strings that specify substrings of signatures to ignore.
    # If this set, it will override the default set of ignored signatures.
    ignoreSigs:
```

pkg/golinters/wrapcheck.go
```diff
@@ -1,19 +1,29 @@
...
 
+	"github.com/golangci/golangci-lint/pkg/config"
 	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
 )
 
 const wrapcheckName = "wrapcheck"
 
-func NewWrapcheck() *goanalysis.Linter {
+func NewWrapcheck(cfg *config.WrapcheckSettings) *goanalysis.Linter {
+	c := wrapcheck.NewDefaultConfig()
```
ldez, line 14: Maybe you can use `settings` instead of `cfg` and `cfg` instead of `c`:
```go
func NewWrapcheck(settings *config.WrapcheckSettings) *goanalysis.Linter {
	cfg := wrapcheck.NewDefaultConfig()
```

ldez (changes requested):

PR #1956, opened 2021-05-05
importas: add message if settings contain no aliases

StephenBrown2:
I wanted to use the strict feature of importas, but forgot to also move my aliases under the `alias:` key, and couldn't figure out why they weren't being applied. This message would have saved me a couple hours debugging the code to find out I should have just read the docs more closely.

Initially I had it as a warning so it would show when running normally, but the tests (with --no-config) still created a default config for ImportAs, and so the output was flagged. An info message will be shown with --verbose, and should be sufficient, as that was the first thing I did to try to figure out if importas was getting filtered out for some reason.


ldez (approved):
LGTM

PR #1959, opened 2021-05-07
Flip importas configuration

cbandy:
The standalone [importas](https://github.com/julz/importas) tool takes arguments in [the form `{package}:{alias}`](https://github.com/julz/importas/blob/841f0c0fe66d8a78d66676e024272dfd69b57668/flags.go#L12) and [stores them in a map](https://github.com/julz/importas/blob/841f0c0fe66d8a78d66676e024272dfd69b57668/flags.go#L25) in which [`package` is the key](https://github.com/julz/importas/blob/841f0c0fe66d8a78d66676e024272dfd69b57668/config.go#L17). The intent is to describe the one alias that is appropriate for a unique [package path](https://golang.org/ref/mod#modules-overview).

The golangci-lint configuration does the opposite; the `alias` is the key. This effectively "reserves" an alias for one package. This difference in configuration limits the number of packages that can be configured and doesn't align with the goals and documentation of the standalone tool.

This PR aligns our configuration with the standalone tool. It is a breaking change, but another breaking change to this linter's configuration is waiting to be released in ffe80615b079a2cfe9b28e2d8323d58061b795d6.

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -410,13 +410,13 @@ linters-settings:
     # List of aliases
     alias:
       # using `servingv1` alias for `knative.dev/serving/pkg/apis/serving/v1` package
-      servingv1: knative.dev/serving/pkg/apis/serving/v1
+      knative.dev/serving/pkg/apis/serving/v1: servingv1
       # using `autoscalingv1alpha1` alias for `knative.dev/serving/pkg/apis/autoscaling/v1alpha1` package
-      autoscalingv1alpha1: knative.dev/serving/pkg/apis/autoscaling/v1alpha1
+      knative.dev/serving/pkg/apis/autoscaling/v1alpha1: autoscalingv1alpha1
       # You can specify the package path by regular expression,
       # and alias by regular expression expansion syntax like below.
       # see https://github.com/julz/importas#use-regular-expression for details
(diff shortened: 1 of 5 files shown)


ldez:
Hello,

It's not possible to flip the configuration due to the configuration parsing system.
There is a bug when the key contains dot.

cbandy:
I see. Are you open to changing it to a list of strings? These match how the standalone tool works exactly.

```yaml
  importas:
    # if set to `true`, force to use alias.
    no-unaliased: true
    # List of aliases
    aliases:
      # using `servingv1` alias for `knative.dev/serving/pkg/apis/serving/v1` package
      - knative.dev/serving/pkg/apis/serving/v1:servingv1
      # using `autoscalingv1alpha1` alias for `knative.dev/serving/pkg/apis/autoscaling/v1alpha1` package
      - knative.dev/serving/pkg/apis/autoscaling/v1alpha1:autoscalingv1alpha1
      # You can specify the package path by regular expression,
      # and alias by regular expression expansion syntax like below.
      # see https://github.com/julz/importas#use-regular-expression for details
      - knative.dev/serving/pkg/apis/(\w+)/(v[\w\d]+):$1$2
```

PR #1976, opened 2021-05-13
fix: example config yaml

butuzov:
Minor fix, remove one tab character in the `.golangci.example.yaml` which breaks yaml parsing (in the case if .golangci.example.yaml used to automate `golangci-lint` testing).


ldez (approved):
LGTM

PR #1985, opened 2021-05-15
docs: exhaustivestruct example explanation

butuzov:
Comment add context on `struct-patterns` patterns, so it's not unexpected for users that patterns behave in next way:

1. e.g. `*.Text` can match struct Test if golangci-lint launched against the file
   `golangcli-lint run example/example.go`

1. e.g. `*.Text` do not match struct Test if golangci-lint launched against package
   `golangcli-lint run example`


ldez (approved):
LGTM

PR #2000, opened 2021-05-19
docs: fix pattern of forbidigo in example config yaml 

ytakaya:
The pattern of forbidigo in `.golangci.example.yaml` was not the correct pattern so I fixed.
It has been modified with reference to the original [usage](https://github.com/ashanbrown/forbidigo#usage).

`ginkgo\\.F.*` in the pattern before modification does not behave as expected.

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -145,9 +145,9 @@ linters-settings:
   forbidigo:
     # Forbid the following identifiers
     forbid:
-      - fmt.Errorf # consider errors.Errorf in github.com/pkg/errors
-      - fmt.Print.* # too much log noise
-      - ginkgo\\.F.* # these are used just for local development
+      - ^fmt\.Errorf$ # consider errors.Errorf in github.com/pkg/errors
+      - ^fmt\.Print.*$ # too much log noise
+      - ^ginkgo\.F[A-Z].*$ # these are used just for local development
     # Exclude godoc examples from forbidigo checks.  Default is true.
     exclude_godoc_examples: false
 
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -143,9 +143,9 @@ linters-settings:
       - 'example.com/package.ExampleStruct'
 
   forbidigo:
-    # Forbid the following identifiers
+    # Forbid the following identifiers (identifiers are written using regexp):
     forbid:
-      - ^print.*$ # Identifiers are written using regexp.
+      - ^print.*$
       - 'fmt\.Print.*'
     # Exclude godoc examples from forbidigo checks.  Default is true.
     exclude_godoc_examples: false
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -145,9 +145,8 @@ linters-settings:
   forbidigo:
     # Forbid the following identifiers
     forbid:
-      - ^fmt\.Errorf$ # consider errors.Errorf in github.com/pkg/errors
-      - ^fmt\.Print.*$ # too much log noise
-      - ^ginkgo\.F[A-Z].*$ # these are used just for local development
+      - ^print.*$ # Identifiers are written using regexp.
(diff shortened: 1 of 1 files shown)


butuzov:
Can we leave in place simpler patterns (which are the working ones?)

butuzov:
Maybe it just worth mentioning that pattern should be a valid regular expression?

ytakaya:
I think that we need to encourage user to use valid regular expression because `regexp` package is used in forbidigo.
https://github.com/ashanbrown/forbidigo/blob/f30240a5170c8f6a62cf603ea541c0ff2008a6be/forbidigo/forbidigo.go#L70
But simplicity is also important, so using symbols like `^` and `$` may be excessive.

I think the biggest problem is that there are patterns in the documentation that don't work.
`ginkgo\\.F.*` does not match an identifier like `ginkgo.FContext()` or `ginkgo.FDescribe()`, so it needs to be fixed to `ginkgo\.F.*` or `"ginkgo\\.F.*"`.
By using double quotation marks, `\` is escaped and becomes the valid regular expression.

Based on the above, here's what I think is a good example.
```
- fmt\.Errorf
- fmt\.Print.*
- ginkgo\.F.*
```

butuzov:
Absolutely agree! How about:
- we list the same package/function with various interpretations 
- mention that `forbidigo` require valid regexp
- will use single quotes for YAML string (so we don't have double quoting `\\`)

```yaml
- fmt.Print
- 'fmt\.Print.*'
- '^fmt.Print.*$'
```

ldez:
@butuzov I understand the idea, but it can be confusing for users to use 3 variations of the same expression.

I would prefer:
```yml
  forbidigo:
    # Forbid the following identifiers
    forbid:
      - ^print.*$
      - 'fmt\.Print.*'

```

butuzov:
Can we leave at least one simplified pattern?

ldez:
There no real simplified pattern: a dot in a regex doesn't mean dot but every character.

ex: `fmt.Print` matches `afmtaPrint()` and `samfmt.Print()`

butuzov:
... and `fmt.Print`, and `afmtaPrint()`... and after it matches last one  - user will return to example config and read that value require to be `valid regexp` (which is missing in your example and initial author's PR ). Note about `regexp` is important, and can help users in the future.

butuzov:
Compare it to the `exhaustivestruct` struct patterns which I was initially sure are regexp (and they are not).

ldez:
I agree with adding a comment about the regex syntax requirement. :+1:

ytakaya:
I've modified it based on your opinions.
Does this make any difference to your recognition?

.golangci.example.yml
```diff
@@ -145,9 +145,8 @@ linters-settings:
   forbidigo:
     # Forbid the following identifiers
     forbid:
-      - ^fmt\.Errorf$ # consider errors.Errorf in github.com/pkg/errors
-      - ^fmt\.Print.*$ # too much log noise
-      - ^ginkgo\.F[A-Z].*$ # these are used just for local development
+      - ^print.*$ # Identifiers are written using regexp.
+      - 'fmt\.Print.*'
```
butuzov:
```suggestion
    # Forbid the following identifiers (Identifier are written using regexp):
    forbid:
      - ^print.*$ 
      - 'fmt\.Print.*'
```

butuzov: super minor nit.

ldez: @butuzov your suggestion is not based on the right (HEAD) commit

butuzov:
my suggestion deleted as pointer @ldez it wasn't based on right commit.

```diff
+ # Forbid the following identifiers (identifiers are written using regexp):
- # Forbid the following identifiers    
  forbid:
```

butuzov (approved):

ashanbrown (commented):
👍

PR #2013, opened 2021-05-24
nolintlint: fix false positive.

ldez:
the comments:

> // nolintlint generates potential issues for every nolint directive and they are filtered out here

> // nolintlint provides a linter to ensure that all //nolint directives are followed by explanations

produce false positives: `nolintlint`  is seen as `nolint` directive.


bombsimon (approved):
LGTM!

PR #2015, opened 2021-05-25
change variable declaration

sivchari:
if capacity can be secured, change to reserve capacity with make method

### DIFF
--- a/pkg/config/linters_settings_gocritic.go
+++ b/pkg/config/linters_settings_gocritic.go
@@ -72,7 +72,7 @@ func gocriticCheckerTagsDebugf() {
 
 	tagToCheckers := buildGocriticTagToCheckersMap()
 
-	var allTags []string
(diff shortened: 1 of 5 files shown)


ldez:
Hi,

`make` doesn't secure a slice, it's just a memory optimization.

The impact of this optimization in golangci-lint is about 0.

IMO it's a useless optimization.

ldez:
@sivchari if you are searching for something to work on it, I think you can try to fix the false positives of `prealloc`.

My advice: don't try to be a "human linter", spend time on creating or improving a linter.

Also If you try to change the slice initialization, there are at least 7 missing fixes.

For now, and for now only, I think we have to decline this PR: either we are able to define a rule on that for the project and use a linter to handle that, or we don't have to handle that.

sivchari:
Ok,  I've comprehended.
I close this PR once, and I improve prealloc or makezero, creating new linter.

PR #2016, opened 2021-05-26
Add varerr linter

sivchari:
varerr checks that you initialize error type.
https://github.com/sivchari/varerr

e.g.

```go
var err error

err = funcA()
if err !=  nil {
    ... 
}

err = funcB()
if err !=  nil {
    ...
}
```

output
```console
the error type `err` is initialized with var
```

above code can been fixed
```go
if err := funcA(); err != nil {
    ...
}

if err := funcB(); err != nil {
    ...
}
```

In addition,  this linter ignore a particular issue by including the directive // novarerr on that line.
```go
var err error // novarerr
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -65,6 +65,7 @@ require (
 	github.com/shazow/go-diff v0.0.0-20160112020656-b6b7b6733b8c
 	github.com/shirou/gopsutil/v3 v3.21.4
 	github.com/sirupsen/logrus v1.8.1
+	github.com/sivchari/varerr v0.1.1
 	github.com/sonatard/noctx v0.0.1
 	github.com/sourcegraph/go-diff v0.6.1
 	github.com/spf13/cobra v1.1.3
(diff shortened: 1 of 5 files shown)


ldez:
Hello,

what is the problem with the fact to define a var for an error?

sivchari:
Thanks for comment, Idez.

In the past, in my team development, I have used several methods where only the error type was the return value. And we had a mixture of both the way I pointed out in this linter and the way we handled errors in the scope of the if block.
I have encountered a situation where new Go writers were wondering which format was appropriate due to the mixed code format.

Of course, defining with var is not a bad idea. 
However, the Go language has a language specification rule of implicit and explicit blocks that allows error checking to be written in a smart way.
Therefore, except for cases where error checking becomes complicated, we believe that smart error checking is a good thing.

In light of the above, we created this linter to eliminate confusion and unify code formatting

ldez:
Sorry but I don't see the goal of this linter, the fact to declare a var for an error is a classical pattern.

example:
```go
package main

import (
	"log"
	"os"
)

func main() {
	foo := "foo"

	if os.Getenv("EXAMPLE") != "" {
		var err error
		foo, err = funcD()
		if err != nil {
			log.Fatal(err)
		}
	}

	println(foo)
}
```

Your linter will lead to 99,999% of false positives.

Sorry, but I will have to decline this PR.

Before creating a linter you have to find a real problem to detect: a linter is not a language tutorial.

I will repeat myself: if you are searching for something to work on it, I think you can try to fix the false positives of [prealloc](https://github.com/alexkohler/prealloc) or take a look to [makezero](https://github.com/ashanbrown/makezero).
Fixing the false positives of [prealloc](https://github.com/alexkohler/prealloc) is something useful for the golangci-lint.

PR #2018, opened 2021-05-27
gomodguard: fix problem where duplicate issues were being reported

ryancurrah:
Gomodguard was saving issues to a global variable and when it was the process issues functioned returned results it would end up returning duplicates. This issue solves that by the process function returning only new issues found.

Also ensure we only configure and initialize gomodguard once by removing those steps from the `Run` func.

https://github.com/ryancurrah/gomodguard/compare/v1.2.0...v1.2.1

### DIFF
--- a/pkg/golinters/gomodguard.go
+++ b/pkg/golinters/gomodguard.go
@@ -72,7 +72,7 @@ func NewGomodguard() *goanalysis.Linter {
 		}
 
 		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
-			files := []string{}
+			var files []string
 
 			for _, file := range pass.Files {
 				files = append(files, pass.Fset.PositionFor(file.Pos(), false).Filename)
--- a/pkg/golinters/gomodguard.go
+++ b/pkg/golinters/gomodguard.go
@@ -24,8 +24,8 @@ func NewGomodguard() *goanalysis.Linter {
 		issues   []goanalysis.Issue
 		mu       = sync.Mutex{}
 		analyzer = &analysis.Analyzer{
-			Name: gomodguardName,
-			Doc:  gomodguardDesc,
+			Name: goanalysis.TheOnlyAnalyzerName,
+			Doc:  goanalysis.TheOnlyanalyzerDoc,
 		}
 	)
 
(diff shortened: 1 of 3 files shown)


pkg/golinters/gomodguard.go
```diff
@@ -21,75 +24,70 @@ func NewGomodguard() *goanalysis.Linter {
 		issues   []goanalysis.Issue
 		mu       = sync.Mutex{}
 		analyzer = &analysis.Analyzer{
-			Name: goanalysis.TheOnlyAnalyzerName,
-			Doc:  goanalysis.TheOnlyanalyzerDoc,
+			Name: gomodguardName,
+			Doc:  gomodguardDesc,
```
ldez, line 28: Could revert this change?
The `TheOnlyAnalyzer` is a specific pattern.

ldez (changes requested):

pkg/golinters/gomodguard.go
ryancurrah: Yep.

ryancurrah: Both or just the name?

ldez: you can keep `goanalysis.TheOnlyanalyzerDoc` only

ryancurrah: Reverted both.

pkg/golinters/gomodguard.go
```diff
@@ -28,68 +31,63 @@ func NewGomodguard() *goanalysis.Linter {
...
+		}
 
-			processor, err := gomodguard.NewProcessor(processorCfg)
-			if err != nil {
-				lintCtx.Log.Warnf("running gomodguard failed: %s: if you are not using go modules "+
-					"it is suggested to disable this linter", err)
-				return nil, nil
-			}
+		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
+			files := []string{}
```
ldez, line 75:
```suggestion
			var files []string
```

ldez (changes requested):
nitpic

ryancurrah:
> nitpic

resolved

ldez (approved):
LGTM

PR #2069, opened 2021-06-21
gofumpt: Add lang-version option

kailun-qin:
Without the Go language version input, `gofumpt` cannot apply some rules
which are version dependent.

Defaulting `lang-version` to the Go version of the running environment
if not specified.

Signed-off-by: Kailun Qin <kailun.qin@intel.com>

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -259,6 +259,9 @@ linters-settings:
     simplify: true
 
   gofumpt:
+    # Select the Go language version to target. The default is `1.15`.
+    lang-version: "1.15"
+
     # Choose whether or not to use the extra rules that are disabled
     # by default
     extra-rules: false
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -259,7 +259,7 @@ linters-settings:
     simplify: true
 
   gofumpt:
-    # Select the Go language version to target. The default is `1.15`.
+    # Select the Go version to target. The default is `1.15`.
     lang-version: "1.15"
 
     # Choose whether or not to use the extra rules that are disabled
--- a/pkg/golinters/gofumpt.go
+++ b/pkg/golinters/gofumpt.go
@@ -34,9 +34,9 @@ func NewGofumpt() *goanalysis.Linter {
 		nil,
 	).WithContextSetter(func(lintCtx *linter.Context) {
 		settings := lintCtx.Settings().Gofumpt
-		langVersion := getLangVersion(settings)
+
 		options := format.Options{
-			LangVersion: langVersion,
+			LangVersion: getLangVersion(settings),
 			ExtraRules:  settings.ExtraRules,
 		}
 
@@ -100,12 +100,9 @@ func NewGofumpt() *goanalysis.Linter {
 }
 
 func getLangVersion(settings config.GofumptSettings) string {
-	langVersion := settings.LangVersion
-
-	if langVersion != "" {
-		return langVersion
+	if settings.LangVersion == "" {
+		// TODO: defaults to "1.15", in the future (v2) must be set by using build.Default.ReleaseTags like staticcheck.
+		return "1.15"
 	}
-
-	// TODO: defaults to "1.15", in the future (v2) must be set by using build.Default.ReleaseTags like staticcheck.
-	return "1.15"
+	return settings.LangVersion
 }
--- a/pkg/golinters/gofumpt.go
+++ b/pkg/golinters/gofumpt.go
@@ -11,6 +11,7 @@ import (
 	"golang.org/x/tools/go/analysis"
 	"mvdan.cc/gofumpt/format"
 
+	"github.com/golangci/golangci-lint/pkg/config"
 	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
 	"github.com/golangci/golangci-lint/pkg/lint/linter"
 )
@@ -32,6 +33,13 @@ func NewGofumpt() *goanalysis.Linter {
 		[]*analysis.Analyzer{analyzer},
 		nil,
 	).WithContextSetter(func(lintCtx *linter.Context) {
+		settings := lintCtx.Settings().Gofumpt
+		langVersion := getLangVersion(settings)
+		options := format.Options{
+			LangVersion: langVersion,
+			ExtraRules:  settings.ExtraRules,
+		}
+
 		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
 			var fileNames []string
 			for _, f := range pass.Files {
@@ -46,12 +54,12 @@ func NewGofumpt() *goanalysis.Linter {
 				if err != nil {
 					return nil, fmt.Errorf("unable to open file %s: %w", f, err)
 				}
-				output, err := format.Source(input, format.Options{
-					ExtraRules: lintCtx.Settings().Gofumpt.ExtraRules,
-				})
+
+				output, err := format.Source(input, options)
 				if err != nil {
 					return nil, fmt.Errorf("error while running gofumpt: %w", err)
 				}
+
 				if !bytes.Equal(input, output) {
 					out := bytes.Buffer{}
 					_, err = out.WriteString(fmt.Sprintf("--- %[1]s\n+++ %[1]s\n", f))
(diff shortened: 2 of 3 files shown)


ldez (changes requested):
All the code related to `format.Options` must be move outside the loop:

```go
			settings := lintCtx.Settings().Gofumpt

			// Get the target Go language version to be used in gofumpt.
			// Reuse logic from `https://github.com/mvdan/gofumpt/blob/ba6406b58f3630c5246474b20f6ee5c83d1cc6b9/gofmt.go#L127-L133`.
			langVersion := settings.LangVersion
			if langVersion == "" {
				outGoVersion, errGoVersion := exec.Command("go", "list", "-m", "-f", "{{.GoVersion}}").Output()
				outGoVersion = bytes.TrimSpace(outGoVersion)
				if errGoVersion == nil && len(outGoVersion) > 0 {
					langVersion = string(outGoVersion)
				}
			}

			options := format.Options{
				LangVersion: langVersion,
				ExtraRules:  settings.ExtraRules,
			}

			var issues []goanalysis.Issue

			for _, f := range fileNames {
				input, err := ioutil.ReadFile(f)
				if err != nil {
					return nil, fmt.Errorf("unable to open file %s: %w", f, err)
				}

				output, err := format.Source(input, options)
				if err != nil {
					return nil, fmt.Errorf("error while running gofumpt: %w", err)
				}
				// ...
			}
```

kailun-qin:
> All the code related to `format.Options` must be move outside the loop:
> 
> ```go
> 			settings := lintCtx.Settings().Gofumpt
> 
> 			// Get the target Go language version to be used in gofumpt.
> 			// Reuse logic from `https://github.com/mvdan/gofumpt/blob/ba6406b58f3630c5246474b20f6ee5c83d1cc6b9/gofmt.go#L127-L133`.
> 			langVersion := settings.LangVersion
> 			if langVersion == "" {
> 				outGoVersion, errGoVersion := exec.Command("go", "list", "-m", "-f", "{{.GoVersion}}").Output()
> 				outGoVersion = bytes.TrimSpace(outGoVersion)
> 				if errGoVersion == nil && len(outGoVersion) > 0 {
> 					langVersion = string(outGoVersion)
> 				}
> 			}
> 
> 			options := format.Options{
> 				LangVersion: langVersion,
> 				ExtraRules:  settings.ExtraRules,
> 			}
> 
> 			var fileNames []string
> 			for _, f := range pass.Files {
> 				pos := pass.Fset.PositionFor(f.Pos(), false)
> 				fileNames = append(fileNames, pos.Filename)
> 			}
> 
> 			var issues []goanalysis.Issue
> 
> 			for _, f := range fileNames {
> 				input, err := ioutil.ReadFile(f)
> 				if err != nil {
> 					return nil, fmt.Errorf("unable to open file %s: %w", f, err)
> 				}
> 
> 				output, err := format.Source(input, options)
> 				if err != nil {
> 					return nil, fmt.Errorf("error while running gofumpt: %w", err)
> 				}
> 				// ...
> 			}
> ```

Good catch!  Updated, thanks!

pkg/golinters/gofumpt.go
```diff
@@ -46,8 +47,21 @@ func NewGofumpt() *goanalysis.Linter {
 				if err != nil {
 					return nil, fmt.Errorf("unable to open file %s: %w", f, err)
 				}
+
+				// Get the target Go language version to be used in gofumpt.
+				// Reuse logic from `https://github.com/mvdan/gofumpt/blob/master/gofmt.go`.
```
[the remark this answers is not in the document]
kailun-qin: Updated, thanks!

pkg/golinters/gofumpt.go
```diff
@@ -33,6 +34,24 @@ func NewGofumpt() *goanalysis.Linter {
 		nil,
 	).WithContextSetter(func(lintCtx *linter.Context) {
 		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
+			settings := lintCtx.Settings().Gofumpt
+
+			// Get the target Go language version to be used in gofumpt.
+			// Reuse logic from `https://github.com/mvdan/gofumpt/blob/ba6406b58f3630c5246474b20f6ee5c83d1cc6b9/gofmt.go#L127-L133`.
+			langVersion := settings.LangVersion
+			if langVersion == "" {
+				outGoVersion, errGoVersion := exec.Command("go", "list", "-m", "-f", "{{.GoVersion}}").Output()
```
bombsimon, line 43: I might be wrong but I was under the impression that `golangci-lint` are not dependent on Go itself. F.ex. when running in CI you can provide a single binary to lint the code. If that's true, this will introduce a new dependency that might impact a lot of users since the default language version is `""`.

Not sure what the solution here would be but maybe worth considering?

bombsimon:
> Good catch! Updated, thanks!

I think you can move it even further out, outside of `analyzer.Run`, no? I don't see why we need to figure out version and set options for every pass so doing it once should be enough, right @ldez?

pkg/golinters/gofumpt.go
kailun-qin: Thanks for the comment! I think it's reasonable to not introduce a new dependency.
An alternative is to use something similar to `getGoVersion` below used in `StaticCheckSettings`, where we default the version to 1.13 and use the build ReleaseTags later. https://github.com/golangci/golangci-lint/blob/a68a88ee02344ab570cafd5f53ac8d0f5adee299/pkg/golinters/staticcheck_common.go#L17-L29

bombsimon: Yeah that sounds like a good alternative! Didn't really read up on the details about this flag but if that works I think the zero dependency part of `golangci-lint` is pretty nice.

ldez:
@kailun-qin can you address changes as commit, because when you squash it's not possible to see changes.

> I think you can move it even further out, outside of analyzer.Run, no?

@bombsimon it depends on the implementation to get the default Go version.

kailun-qin:
> @kailun-qin can you address changes as commit, because when you squash it's not possible to see changes.
> 

Sorry that I noticed this before the latest commit. Will do for the followings.

> > I think you can move it even further out, outside of analyzer.Run, no?
> 
> @bombsimon it depends on the implementation to get the default Go version.

Using a hardcoding default (to `1.15`) to remove the dependency, and IMO this can be moved outside of `analyzer.Run`. PTAL @bombsimon @ldez, thanks!

bombsimon (approved):
LGTM!

ldez:
@kailun-qin please don't squash, we will squash when we will merge.

kailun-qin:
> @kailun-qin please don't squash, we will squash when we will merge.

@ldez Sure, excuse me for not catching your comment before refreshing the last commit. I'll not squash for the following ones.
Please kindly take a look at the latest version and see if any comment. Many thanks!

pkg/golinters/gofumpt.go
```diff
@@ -90,3 +98,14 @@ func NewGofumpt() *goanalysis.Linter {
...
+
+func getLangVersion(settings config.GofumptSettings) string {
+	langVersion := settings.LangVersion
+
+	if langVersion != "" {
+		return langVersion
+	}
+
+	// TODO: defaults to "1.15", in the future (v2) must be set by using build.Default.ReleaseTags like staticcheck.
+	return "1.15"
```
ldez, line 110:
```suggestion
	if settings.LangVersion == "" {
		// TODO: defaults to "1.15", in the future (v2) must be set by using build.Default.ReleaseTags like staticcheck.
		return "1.15"
	}

	return settings.LangVersion
```

pkg/golinters/gofumpt.go
```diff
@@ -32,6 +33,13 @@ func NewGofumpt() *goanalysis.Linter {
 		[]*analysis.Analyzer{analyzer},
 		nil,
 	).WithContextSetter(func(lintCtx *linter.Context) {
+		settings := lintCtx.Settings().Gofumpt
+		langVersion := getLangVersion(settings)
+		options := format.Options{
+			LangVersion: langVersion,
+			ExtraRules:  settings.ExtraRules,
+		}
```
ldez, line 41:
```suggestion
		settings := lintCtx.Settings().Gofumpt

		options := format.Options{
			LangVersion: getLangVersion(settings),
			ExtraRules:  settings.ExtraRules,
		}
```

.golangci.example.yml
```diff
@@ -259,6 +259,9 @@ linters-settings:
     simplify: true
 
   gofumpt:
+    # Select the Go language version to target. The default is `1.15`.
```
ldez, line 262:
```suggestion
    # Select the Go version to target. The default is `1.15`.
```

ldez (changes requested):

kailun-qin:
Updated, thanks!

ldez:
@kailun-qin just FYI https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes

ldez (approved):

kailun-qin:
> @kailun-qin just FYI https://docs.github.com/en/github/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes

Thanks for the hint! I'll use that way for the future suggested changes.

PR #2078, opened 2021-06-23
dev: fix interface implement check for "SkipDirs"

shmsr:
For every processor, there's a compile-time check for processors to check if they implement the `Processor` interface:
```
type Processor interface {
	Process(issues []result.Issue) ([]result.Issue, error)
	Name() string
	Finish()
}
```

But there's a typo: `SkipFiles` is getting checked instead of `SkipDirs` 

Fix:
```
var _ Processor = (*SkipDirs)(nil)
```
(or)
```
var _ Processor = &SkipDirs
```

The former is more idiomatic; so I did that.

### DIFF
--- a/pkg/result/processors/skip_dirs.go
+++ b/pkg/result/processors/skip_dirs.go
@@ -24,7 +24,7 @@ type SkipDirs struct {
 	skippedDirsCache map[string]bool
 }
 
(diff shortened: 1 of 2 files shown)


ldez (approved):

shmsr:
@ldez Good for merge?

PR #2089, opened 2021-07-01
docs: not show the number of linters

sanposhiho:
Currently more than 48 linter can be used with golangci-lint, so, the number of linters on docs is wrong.
It would be hard to change this number every time, maybe 🤔 

This PR changed to not show the number of linters.


ldez (approved):
LGTM

PR #2110, opened 2021-07-10
errcheck: allow exclude config without extra file

rliebz:
One of the challenges in standardizing on a golangci-lint configuration is that if we want to configure `errcheck`, it requires two separate files to configure: the golangci config file, plus a separate errcheck excludes file.
In particular, copying golangci-lint configs across projects often fails, as a reference to an exclude file at a particular file will not resolve.

While this config does not add any new functionality, it allows the entirety of the errcheck configuration to be done inside the
golangci-lint config file itself.

https://github.com/kisielk/errcheck#excluding-functions

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -133,9 +133,6 @@ func initFlagSet(fs *pflag.FlagSet, cfg *config.Config, m *lintersdb.Manager, is
 	fs.StringVar(&lsc.Errcheck.Exclude, "errcheck.exclude", "",
 		"Path to a file containing a list of functions to exclude from checking")
 	hideFlag("errcheck.exclude")
-	fs.StringSliceVar(&lsc.Errcheck.ExcludeFunctions, "errcheck.exclude-functions", nil,
-		"List of functions to exclude from checking")
-	hideFlag("errcheck.exclude-functions")
 	fs.StringVar(&lsc.Errcheck.Ignore, "errcheck.ignore", "fmt:.*",
 		`Comma-separated list of pairs of the form pkg:regex. The regex is used to ignore names within pkg`)
 	hideFlag("errcheck.ignore")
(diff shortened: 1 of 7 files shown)


pkg/commands/run.go
```diff
@@ -133,6 +133,9 @@ func initFlagSet(fs *pflag.FlagSet, cfg *config.Config, m *lintersdb.Manager, is
 	fs.StringVar(&lsc.Errcheck.Exclude, "errcheck.exclude", "",
 		"Path to a file containing a list of functions to exclude from checking")
 	hideFlag("errcheck.exclude")
+	fs.StringSliceVar(&lsc.Errcheck.ExcludeFunctions, "errcheck.exclude-functions", nil,
+		"List of functions to exclude from checking")
+	hideFlag("errcheck.exclude-functions")
```
ldez, line 138: as you can see the comment on the top of this section, the flags are "deprecated", could you remove those lines?

test/testdata/errcheck_exclude_functions.go
```diff
@@ -0,0 +1,19 @@
+//args: -Eerrcheck
+//config: linters-settings.errcheck.check-blank=true
+//config: linters-settings.errcheck.exclude-functions=io/ioutil.ReadFile,io/ioutil.ReadDir
```
ldez, line 3: could you use a config file instead of the config comment system?

example: https://github.com/golangci/golangci-lint/blob/36aa12425c0aaf5697c2cd7c094f84c410ddd9d7/test/testdata/forbidigo_example.go#L2

ldez (changes requested):

rliebz:
Thanks for the quick feedback, all comments addressed!

Not sure if there's anything I need to do to get the pipeline to pass. It's failing on a `make build test`, which is passing for me locally

ldez (approved):

magnetikonline:
@rliebz great to see this just land in `1.42.0` - was hoping this issue would be addressed one day! 👍

PR #2117, opened 2021-07-18
suggest golangci-lint help linters

smantic:
`golangci-lint linters` wont run when you have an unknown linter. `golangci-lint help linters` works.

### DIFF
--- a/pkg/lint/lintersdb/validator.go
+++ b/pkg/lint/lintersdb/validator.go
@@ -30,7 +30,7 @@ func (v Validator) validateLintersNames(cfg *config.Linters) error {
 	}
 
 	if len(unknownNames) > 0 {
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

what do you call an "unknown linter"?

smantic:
@ldez a linter name string that doesnt match any existing linter names. whenever `if v.m.GetLinterConfigs(name) == nil` this unknown linters error gets thrown

ldez:
why do you want to display the linter's list if you have an invalid linter in your configuration?
From my point of view, it's better to show the error.

smantic:
@ldez Please view the diff. My suggested change is just one word in the error message that is printed when you have an invalid linter in your config.

ldez:
@smantic please answer my question, because without a strong argument I will not approve that PR.

ldez (approved):
LGTM

PR #2131, opened 2021-07-24
add nilassign linter

sivchari:
nilassign finds that assigning to invalid memory address or nil pointer dereference.
https://github.com/sivchari/nilassign


sivchari:
@ldez 
hi, thanks for the review and the fix!
what should I do after this?

ldez (approved):

ldez:
After more investigation on this linter, there are too many false positives.
I will revert this PR (as it's not a part of the release), please submit a new PR when your linter will be more reliable.

sivchari:
@ldez 
Hi, thanks for review and submit issue.
I'll refer to it and make it more reliable. After I've fixed it, I'd push new PR.

PR #2136, opened 2021-07-29
feat: Add support for remote config file

alexandrevilain:
Enforcing the same lint configuration for many projects and making updates for it can become a bit time consuming.
This is why I purpose you to add support for remote config file in golangci-lint. It allows teams to maintain only one global configuration file for all their projects.

I'll be happy to discuss with you if the implementation does not match your coding style/rules.

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -51,7 +51,10 @@ func wh(text string) string {
 	return color.GreenString(text)
 }
 
-const defaultTimeout = time.Minute
+const (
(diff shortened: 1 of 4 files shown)


ldez:
Hello,

I would like to suggest opening an issue to discuss the meaning of this feature (not the technical part).

ldez:
I will close this PR according to https://github.com/golangci/golangci-lint/issues/2137#issuecomment-891621683

PR #2163, opened 2021-08-09
Added structfield linter

uudashr:
Add support for [structfield](https://github.com/uudashr/structfield) linter

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -542,6 +542,10 @@ linters-settings:
     # https://staticcheck.io/docs/options#checks
     checks: [ "all" ]
 
+  structfield:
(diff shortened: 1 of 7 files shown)


Antonboom:
Why not add an option to [exhaustivestruct](https://github.com/mbilski/exhaustivestruct)?

leonklingele:
Can this please be reopened? Use case:

The first (unnamed) `Account` instantiation should be invalid & reported, yet the second one where not all fields are set should still be valid. This is not the case with the [exhaustruct](https://github.com/GaijinEntertainment/go-exhaustruct) linter.
https://go.dev/play/p/xs88a4ykgfy

PR #2183, opened 2021-08-17
Update Go versions used by the CI

ldez:
The binaries for the releases should be built with Go1.17.

For now, the min Go version will stay on go1.15 but we will upgrade to go1.16 in few weeks.


SVilgelm (approved):

SVilgelm:
👍 still looks good to me

PR #2198, opened 2021-08-30
build: upgrade to Go 1.17

Juneezee:
Since both the CI (PR #2183) and Docker images (PR #2188) are now using Go 1.17, the Go directive should be upgraded to 1.17 as well.


ldez:
Hello, as explain [here](https://github.com/golangci/golangci-lint/pull/2183) we will not update the go version to go1.17 for now.

PR #2216, opened 2021-09-03
feat: add contextcheck linter

sylvia7788:
https://github.com/sylvia7788/contextcheck


butuzov:
if incoming context is declared unimportant, but another context still used with in short scoped function, this will trigger an error.

```go
func IgnoreIncomingContext(_ context.Context) {
   ctx := shortscopedContext()
   select {
     case <- ctx.Done():
       return
     default: 
        ...
   }
}
// got -: Function `gnoreIncomingContext$1` should pass the context parameter
```

also:
somehow I am no getting path of where issue ocurred.

pkg/lint/lintersdb/manager.go
```diff
@@ -506,6 +506,11 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithLoadForGoAnalysis().
 			WithURL("https://github.com/Antonboom/errname").
 			WithSince("v1.42.0"),
+		linter.NewConfig(golinters.NewContextCheck()).
+			WithPresets(linter.PresetBugs).
+			WithLoadForGoAnalysis().
+			WithURL("https://github.com/sylvia7788/contextcheck").
+			WithSince("v1.0.0"),
```
ldez, line 513:
```suggestion
			WithSince("v1.43.0"),
```

ldez (changes requested):
Hi,

The license is invalid, you have to set the right information.

https://github.com/sylvia7788/contextcheck/blob/4be28c822aeb486702de79e402f323d5f44b370d/LICENSE#L189

sylvia7788:
@butuzov 
1. '$1' means the first clusure of the function IgnoreIncomingContext
2. I will add one case to ignore the ignored param😊

sylvia7788:
@ldez 
Hi,

I have already modified the Copyright, please review it, thanks.

https://github.com/sylvia7788/contextcheck/blob/20c10da9027bad9d852ae23374cb8d25d7ecb745/LICENSE#L189

pkg/lint/lintersdb/manager.go
```diff
@@ -506,6 +506,11 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithLoadForGoAnalysis().
 			WithURL("https://github.com/Antonboom/errname").
 			WithSince("v1.42.0"),
+		linter.NewConfig(golinters.NewContextCheck()).
+			WithPresets(linter.PresetBugs).
```
johejo, line 510: I don't think it's necessarily a bug not to pass the context.
Of course, I know in most cases it is better to pass the context.

ldez: This linter doesn't check if the context is passed but the heritage of a context.
In a lot of cases, it's a bug.

johejo (commented):
LGTM

ldez (approved):
LGTM

PR #2219, opened 2021-09-05
new-linter: ireturn (checks for function return type)

butuzov:
This PR introduces a new linter `ireturn` that checks for function's returned types and adds the possibility to reject or allow specific types of interfaces.

https://github.com/butuzov/ireturn

TL/DR
_Accept Interfaces, Return Concrete Types_

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -447,6 +447,27 @@ linters-settings:
       - pkg: knative.dev/serving/pkg/apis/(\w+)/(v[\w\d]+)
         alias: $1$2
 
+  ireturn:
+    # ireturn allows to use either `allow` or `reject` settings at the same time.
+    # both settings are a lists of the keywords (`empty` for interface{}`,
+    # `error` for errors, `stdlib` for standard library, `anon` for anonymouse
+    # interfaces ) and reglar expresssions matched to interfaces or packages.
+
+    allow:
+    # default settings are allow to use errors, empty interfaces, anonymouse
+    # interfaces, and interfaces provided by standard library.
+    - anon
+    - error
+    - empty
+    - stdlib
+    # You can specify idiomatic endings for interface
+    - (or|er)$
+
+
+    # or you can specify reject patterns
+    reject:
+    - github.com\/user\/package\/v4\.Type
+
   lll:
     # max line length, lines longer will be reported. Default is 120.
     # '\t' is counted as 1 character by default, and can be changed with the tab-width option
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -448,14 +448,14 @@ linters-settings:
         alias: $1$2
 
   ireturn:
-    # ireturn allows to use either `allow` or `reject` settings at the same time.
-    # both settings are a lists of the keywords (`empty` for interface{}`,
-    # `error` for errors, `stdlib` for standard library, `anon` for anonymouse
-    # interfaces ) and reglar expresssions matched to interfaces or packages.
+    # ireturn allows using either `allow` or `reject` settings at the same time.
+    # both settings are lists of the keywords (`empty` for interface{}`,
+    # `error` for errors, `stdlib` for standard library, `anon` for anonymous
+    # interfaces ) and regular expressions matched to interfaces or packages.
 
     allow:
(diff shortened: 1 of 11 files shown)


.golangci.example.yml
```diff
@@ -447,6 +447,27 @@ linters-settings:
       - pkg: knative.dev/serving/pkg/apis/(\w+)/(v[\w\d]+)
         alias: $1$2
 
+  ireturn:
+    # ireturn allows to use either `allow` or `reject` settings at the same time.
+    # both settings are a lists of the keywords (`empty` for interface{}`,
+    # `error` for errors, `stdlib` for standard library, `anon` for anonymouse
+    # interfaces ) and reglar expresssions matched to interfaces or packages.
```
ldez, line 454:
```suggestion
  ireturn:
    # ireturn allows using either `allow` or `reject` settings at the same time.
    # both settings are lists of the keywords (`empty` for interface{}`,
    # `error` for errors, `stdlib` for standard library, `anon` for anonymous
    # interfaces ) and regular expressions matched to interfaces or packages.
```

.golangci.example.yml
```diff
@@ -447,6 +447,27 @@ linters-settings:
...
+
+    allow:
+    # default settings are allow to use errors, empty interfaces, anonymouse
+    # interfaces, and interfaces provided by standard library.
```
ldez, line 458:
```suggestion
    # default settings are allowed to use errors, empty interfaces, anonymous
    # interfaces, and interfaces provided by the standard library.
```

go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/ashanbrown/makezero v0.0.0-20210520155254-b6261585ddde
 	github.com/bkielbasa/cyclop v1.2.0
 	github.com/bombsimon/wsl/v3 v3.3.0
+	github.com/butuzov/ireturn v0.0.0-20210903145654-645c1cbdbffd
```
ldez, line 16: Can you create a tag? Like that bot can handle the updates.
And it's a good practice.

pkg/lint/lintersdb/manager.go
```diff
@@ -456,6 +458,12 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithSince("v1.36.0").
 			WithPresets(linter.PresetStyle).
 			WithURL("https://github.com/esimonov/ifshort"),
+
+		linter.NewConfig(golinters.NewIreturn(ireturnCfg)).
+			WithSince("v1.43.0").
+			WithPresets(linter.PresetStyle).
+			WithURL("https://github.com/butuzov/ireturn"),
+
```
ldez, line 466: can you remove the extra empty lines and move the linter near the end of the list?

test/testdata/ireturn_allow.go
```diff
@@ -0,0 +1,13 @@
+// args: -Eireturn
+// config: linters-settings.ireturn.allow=["IreturnAllowDoer"]
```
ldez, line 2: I recommend using a config file instead of the annotation/CLI approach.

ldez (changes requested):
in the `test/testdata/` files can you try to use at least one imported element (from stdlib)?

go.mod
butuzov: Sure, just I wanted to add one more thing before going 0.1.0.

butuzov:
All issues from the previous CR are fixed, I will be glad if you can take a second look to spot any issues that left?

pkg/config/linters_settings.go
```diff
@@ -186,6 +187,11 @@ type ExhaustiveStructSettings struct {
 	StructPatterns []string `mapstructure:"struct-patterns"`
 }
 
+type IreturnSettings struct {
+	Allow  []string `mapstructure:"allow"`
+	Reject []string `mapstructure:"return"`
```
ldez, line 192:
```suggestion
	Reject []string `mapstructure:"reject"`
```

ldez (changes requested):

pkg/config/linters_settings.go
butuzov: WoW. How can I miss it? Merci

ldez (approved):

PR #2220, opened 2021-09-05
add nobreak linter

sivchari:
`nobreak` finds infinite for statement loop.
https://github.com/sivchari/nobreak


ldez:
Hello,

this linter is a bit too naive: `break` is not the only way to stop this kind of loop (ex: `return`), and in some cases (like tests) the exit of the loop is performed with non-standard functions.

ldez:
As this linter is called `nobreak` but in fact, the goal is to detect infinite loop, and `break` is not the only element to stop/exit of an infinite loop, I will close this PR.

Thanks anyway.

PR #2221, opened 2021-09-06
add tenv linter

sivchari:
tenv is analyzer that detects using os.Setenv instead of t.Setenv.
https://github.com/sivchari/tenv


butuzov:
Does it check the version from `go.mod` or compiled runtime?

ldez (changes requested):
The current implementation produces too many false-positive, the linter must be improved.

sivchari:
@butuzov 
I checked rutime.Verision()

sivchari:
@ldez 
thaks for review.
ok, I'd improve it.

may I ask you to share the false positive you found when you have a moment?

butuzov:
> @butuzov
> I checked rutime.Verision()

Maybe it's better to [parse mod file](https://godocs.io/golang.org/x/mod/modfile#File) to see if it require go.1.17 (see type `Go`)? So you can activate linter if it's forced or module's go is `1.17`

sivchari:
I've fixed

johejo (commented):
I feel that os.Setenv in places where *testing.T cannot be referenced should not be warned by default.

sivchari:
Why do you think that ? Please tell me reason.
And,  if tenv is changed as you say implementation, tenv will can't detect following case.

```go
package main_test

import (
	"os"
	"testing"
)

var env string

func setup() {
	env = os.Getenv("a")
	os.Setenv("a", "b")
}

func cleanup() {
	os.Setenv("a", env)
}

func TestF(t *testing.T) {
	setup()
	// do something ...
	t.Cleanup(cleanup)
}
```
What do you think ?

johejo:
The following is based on my experience. I don't know if it applies to all Gophers.

The refactoring is probably
1. Pass *testing.T to the helper
2. Replace os.Setenv with t.Setenv in the helper

The first step is changing the function signatures, which can be painful at times.

If the default warning is only for the case that can reference *testing.T, user probably don't need to change the function signatures so it's probably not too difficult.

sivchari:
This linter would rather have that refactoring as one of its goals.

Also, one of the motivations for using tenv is to change things that don't use testing.* 
So the pain you mentioned is the right one, and changing this function signature is probably not difficult because it only in the test file.

johejo:
By the way, tenv seems to target only *_test.go.
Is this intentional?
Helpers are often cut out into separate packages.

sivchari:
Yes, it's intentional.

It depends on the product whether you want to get the environment variables for testing in other packages.

Therefore, considering that this linter will support many methods, it's a proportionally large number of false-positive will be generated.

We can also determine by package name, but in Go, the _test suffix is just whether to include it in the build or not, so this is also not practical.

However, the file name _test.go is a naming convention, so we use this method.

ldez:
I agree with @johejo that the linter should report problems only on functions that use `b *testing.B` or `t *testing.T` or `tb testing.TB`.
The fact to check all the content of a `*_test.go` file can be an option.

sivchari:
@ldez 

thanks for comment.
I would like to change my opinion to one that is more supportive and easier to use.

@johejo 

What about the policy of checking all *_test.go with the -all command and supporting *testing.T, *testing.B, and *testing.TB by default?

sivchari:
I update check range and add all option.
Please review when you have a chance.

.golangci.example.yml
```diff
@@ -613,6 +613,13 @@ linters-settings:
       name: true
       begin: true
 
+  tenv:
+    # The following configurations enable checks prior to Go 1.17.
+    force: false
```
ldez, line 618: The `force` option should be removed because it's just a way to enable/disable.
Then is just the same thing that adds or removes the linter.

ldez (changes requested):

.golangci.example.yml
sivchari: The force option is an option that determines whether to enable it for products older than Go1.7.
If force should be removed, what would be the best approach to implement the same?

ldez: I think you don't have to handle the go1.17 constraint inside your linter.
And inside golangci-lint I don't think we have to handle that in this PR.

sivchari: Does this mean that we should run the linter for Go1.17, which does not provide t.Setenv?

Is it correct that the motivation for users of `tenv` is to replace `os.Setenv` with `t.Setenv`, so we don't need the force option and the go.mod version check?

ldez: Your linter, by itself, doesn't have to care about the go version.
Inside golangci-lint, it's another topic, because a user has to choose to enable or disable a linter.
We can help by detecting the go version of the project but for me, it's outside of the scope of your PR.

sivchari: ok, I comprehended.
I'd remove force option and go version check.
thanks

> Inside golangci-lint, it's another topic, because a user has to choose to enable or disable a linter.

I agree with your opinion.

.golangci.example.yml
```diff
@@ -622,6 +622,11 @@ linters-settings:
       name: true
       begin: true
 
+  tenv:
+    # The all option will run against all method in test file.
+    # By default, only methods that take *testing.T, *testing.B, and testing.TB as arguments are checked.
```
ldez, line 627:
```suggestion
    # The option `all` will run against whole test files (`_test.go`) regardless of method/function signatures.
    # By default, only methods that take `*testing.T`, `*testing.B`, and `testing.TB` as arguments are checked.
```

ldez (changes requested):
The report messages can be improved, instead of `func TestF is not using testing.Setenv` I suggest this:
```
# the name of the variable (`t`, `b`, `tb`) can be used in the majority of the cases

os.Setenv() can be replaced by `t.Setenv()` in <Method/Function name>

os.Setenv() can be replaced by `b.Setenv()` in <Method/Function name>

os.Setenv() can be replaced by `tb.Setenv()` in <Method/Function name>

os.Setenv() can be replaced by `testing.Setenv()` in <Method/Function name>
```

---

I created a sample (2 files) with comments to explain what is for me the expected behavior of this linter:

<details>
<summary>foo.go</summary>

```go
package sandbox

import (
	"os"
	"testing"
)

// no report
var errF = os.Setenv("Foo", "bar")

func Foo() {
	// no report
	os.Setenv("Foo", "bar")
}

func Bar(t *testing.T) {
	// report by default (signature based)
	os.Setenv("Foo", "bar")
}

func Bor(b *testing.B) {
	// report by default (signature based)
	os.Setenv("Foo", "bar")
}

func Bur(tb testing.TB) {
	// report by default (signature based)
	os.Setenv("Foo", "bar")
}
```

</details>


<details>
<summary>foo_test.go</summary>

```go
package sandbox

import (
	"os"
	"testing"
)

// no report
var fii = os.Setenv("Foo", "bar")

func foo() {
	// report only the option "all" enabled
	os.Setenv("Foo", "bar")
}

func TestMain(m *testing.M) {
	// Never report error with `TestMain(m *testing.M)` functions
	os.Setenv("Foo", "bar")
	os.Exit(m.Run())
}

func TestName(t *testing.T) {
	// report by default (signature based)
	os.Setenv("Foo", "bar")
}

func BenchmarkName(b *testing.B) {
	// report by default (signature based)
	os.Setenv("Foo", "bar")
	for i := 0; i < b.N; i++ {
		b.Name()
	}
}

```

</details>

sivchari:
```go
// never reported
var errF = os.Setenv("Foo", "bar")
```
Why you expect behavior this line isn't reported with enabling `all` option ?

sivchari:
@ldez 

```console
Error:     analysistest.go:448: a/a_test.go:23:2: diagnostic "os.Setenv() can be replaced by `t.Setenv()` in TestF" does not match pattern "os.Setenv() can be replaced by `t.Setenv()` in TestF"
Error:     analysistest.go:448: a/a_test.go:24:2: diagnostic "os.Setenv() can be replaced by `t.Setenv()` in TestF" does not match pattern "os.Setenv() can be replaced by `t.Setenv()` in TestF"
Error:     analysistest.go:448: a/a_test.go:31:2: diagnostic "os.Setenv() can be replaced by `b.Setenv()` in BenchmarkF" does not match pattern "os.Setenv() can be replaced by `b.Setenv()` in BenchmarkF"
Error:     analysistest.go:448: a/a_test.go:32:2: diagnostic "os.Setenv() can be replaced by `b.Setenv()` in BenchmarkF" does not match pattern "os.Setenv() can be replaced by `b.Setenv()` in BenchmarkF"
Error:     analysistest.go:448: a/a_test.go:38:2: diagnostic "os.Setenv() can be replaced by `tb.Setenv()` in testTB" does not match pattern "os.Setenv() can be replaced by `tb.Setenv()` in testTB"
Error:     analysistest.go:512: a/a_test.go:23: no diagnostic was reported matching "os.Setenv() can be replaced by `t.Setenv()` in TestF"
Error:     analysistest.go:512: a/a_test.go:24: no diagnostic was reported matching "os.Setenv() can be replaced by `t.Setenv()` in TestF"
Error:     analysistest.go:512: a/a_test.go:31: no diagnostic was reported matching "os.Setenv() can be replaced by `b.Setenv()` in BenchmarkF"
Error:     analysistest.go:512: a/a_test.go:32: no diagnostic was reported matching "os.Setenv() can be replaced by `b.Setenv()` in BenchmarkF"
Error:     analysistest.go:512: a/a_test.go:38: no diagnostic was reported matching "os.Setenv() can be replaced by `tb.Setenv()` in testTB"
```

detail: https://github.com/sivchari/tenv/pull/12

if you have a time, please teach me that why this test is  failed.

ldez:
Please re-read my message there are 2 files: `foo.go` and `foo_test.go`

sivchari:
sorry, I fixed it.
please review.

sivchari:
sorry, I fixed it.
please review.

pkg/lint/lintersdb/manager.go
```diff
@@ -522,6 +524,11 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithLoadForGoAnalysis().
 			WithURL("https://github.com/Antonboom/nilnil").
 			WithSince("v1.43.0"),
+		linter.NewConfig(golinters.NewTenv(tenvCfg)).
+			WithSince("v1.43.0").
+			WithPresets(linter.PresetPerformance, linter.PresetStyle).
```
johejo, line 529: Is there any performance improvement in this linter pointed out?

sivchari: thanks, this is mistook.
I fixed it.

johejo (approved):

ldez (approved):

PR #2227, opened 2021-09-06
feat: Add new linter argslen

guerinoni:
https://github.com/guerinoni/argslen

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -83,6 +83,12 @@ output:
 # all available settings of specific linters
 linters-settings:
 
+  argslen:
+  # max arguments allowed for function.
+  maxArguments: 5
+  # should the linter execute on test files as well
+  skipTests: false
+
   cyclop:
     # the maximal code complexity to report
     max-complexity: 10
--- a/go.mod
+++ b/go.mod
@@ -34,6 +34,7 @@ require (
 	github.com/gordonklaus/ineffassign v0.0.0-20210225214923-2e10b2664254
 	github.com/gostaticanalysis/forcetypeassert v0.0.0-20200621232751-01d4955beaa5
 	github.com/gostaticanalysis/nilerr v0.1.1
+	github.com/guerinoni/argslen v0.0.0-20210906213733-bd64655c3950
 	github.com/hashicorp/go-multierror v1.1.1
 	github.com/jgautheron/goconst v1.5.1
 	github.com/jingyugao/rowserrcheck v1.1.0
--- a/test/testdata/argslen.go
+++ b/test/testdata/argslen.go
@@ -1,0 +1,7 @@
+//args: -Eargslen
+//config: linters-settings.argslen.maxArguments=1
+package testdata
+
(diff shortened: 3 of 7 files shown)


.golangci.example.yml
```diff
@@ -83,6 +83,12 @@ output:
 # all available settings of specific linters
 linters-settings:
 
+  argslen:
+  # max arguments allowed for function.
+  maxArguments: 5
+  # should the linter execute on test files as well
+  skipTests: false
+
```
butuzov, line 91: block alignment seams off.

test/testdata/argslen.go
```diff
@@ -0,0 +1,7 @@
+//args: -Eargslen
+//config: linters-settings.argslen.maxArguments=1
```
butuzov, line 2: @ldez recommendation is to use a config file instead of the annotation/CLI approach.

test/testdata/argslen.go
```diff
@@ -0,0 +1,7 @@
...
+package testdata
+
+func argslenFunc(s1, s2, s3 string) { // ERROR "args number for function `argslenFunc` is too high, 3 vs 1 allowed"
```
butuzov, line 5: Can the error be reworded (just to make it more clear)?
1) args number is reached the allowed maximum of 1 argument?
2) args number for function `argslenFunc` is too high, found 3 of 1 allowed

or maybe someone can come with even more clear statement.

go.mod
```diff
@@ -34,6 +34,7 @@ require (
 	github.com/gordonklaus/ineffassign v0.0.0-20210225214923-2e10b2664254
 	github.com/gostaticanalysis/forcetypeassert v0.0.0-20200621232751-01d4955beaa5
 	github.com/gostaticanalysis/nilerr v0.1.1
+	github.com/guerinoni/argslen v0.0.0-20210906213733-bd64655c3950
```
butuzov, line 37: @ldez  pointer in my PR that it's better to use semver and create tagged releases.

butuzov (changes requested):

butuzov:
Federico, is it possible for you to remove binary from your repository (one committed in https://github.com/guerinoni/argslen/commit/c86baf59c617f14bd556595ec652e07468b3639b)?

You also can use [bfg](https://rtyley.github.io/bfg-repo-cleaner/) to remove ti from your repository entierly.

test/testdata/argslen.go
bombsimon: I agree with @butuzov about rephrasing this and I think "args number (for function)" can be hard to understand for some people. If we talk about too many arguments I think that should be the error message. Something like.

> Too many arguments for function `argslenFunc`, 1 allowed but got 3

ldez:
Hello,

there is already a rule in revive to handle the same thing as this linter: https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#argument-limit

ldez:
Then we will decline this PR, as this rule is already checked by another linter.

PR #2229, opened 2021-09-11
Add linter noioutil

le0tk0k:
Hi team.

[noioutil](https://github.com/le0tk0k/noioutil) finds "io/ioutil" package.
(https://golang.org/doc/go1.16#ioutil)

Example

```go
package main

import (
	"fmt"
	"io/ioutil"
)

func main() {
	bytes, _ := ioutil.ReadFile("foo.go")
	fmt.Println(string(bytes))
}
```

```
$ go vet -vettool=$(which noioutil) main.go
# command-line-arguments
./main.go:5:2: "io/ioutil" package is used
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -45,6 +45,7 @@ require (
 	github.com/kyoh86/exportloopref v0.1.8
 	github.com/ldez/gomoddirectives v0.2.2
 	github.com/ldez/tagliatelle v0.2.0
+	github.com/le0tk0k/noioutil v0.3.0
 	github.com/maratori/testpackage v1.0.1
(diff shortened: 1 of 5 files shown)


bombsimon:
Thanks for your contribution. This is already handled by [depguard](https://golangci-lint.run/usage/linters/#depguard).

```sh
$ cat .golangci.yml        
linters:
  disable-all: true
  enable:
    - depguard

linters-settings:
  depguard:
    list-type: blacklist
    include-go-root: true
    packages:
      - "io/ioutil"


$ golangci-lint run                                                   
main.go:5:2: `io/ioutil` is in the blacklist (depguard)
        "io/ioutil"
        ^
```

PR #2230, opened 2021-09-12
docs: Update docs of installation with go

sanposhiho:
> installing executables with 'go get' in module mode is deprecated.

So, we have to use `go install` on golang v1.16+

```
$ go get -u github.com/golangci/golangci-lint/cmd/golangci-lint@v1.42.1
go get: installing executables with 'go get' in module mode is deprecated.
        Use 'go install pkg@version' instead.
        For more information, see https://golang.org/doc/go-get-install-deprecation
        or run 'go help get' or 'go help install'.

```


butuzov (approved):
lgtm

SVilgelm (approved):
LGTM

PR #2236, opened 2021-09-16
Add nilnil linter

Antonboom:
https://github.com/Antonboom/nilnil

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -172,10 +172,6 @@ linters-settings:
     # minimal code complexity to report, 30 by default (but we recommend 10-20)
     min-complexity: 10
 
-  nestif:
-    # minimal complexity of if statements to report, 5 by default
-    min-complexity: 4
-
   goconst:
     # minimal length of string constant, 3 by default
     min-len: 3
@@ -497,6 +493,31 @@ linters-settings:
     # make an issue if func has more lines of code than this setting and it has naked returns; default is 30
     max-func-lines: 30
 
+  nestif:
+    # minimal complexity of if statements to report, 5 by default
+    min-complexity: 4
+
+  nilnil:
+    # By default, nilnil checks all returned types below.
+    checked-types:
+      - ptr
+      - func
+      - iface
+      - map
+      - chan
+
+  nolintlint:
+    # Enable to ensure that nolint directives are all used. Default is true.
+    allow-unused: false
+    # Disable to ensure that nolint directives don't have a leading space. Default is true.
+    allow-leading-space: true
+    # Exclude following linters from requiring an explanation.  Default is [].
+    allow-no-explanation: [ ]
+    # Enable to require an explanation of nonzero length after each nolint directive. Default is false.
+    require-explanation: true
+    # Enable to require nolint directives to mention the specific linter being suppressed. Default is false.
+    require-specific: true
+
   prealloc:
     # XXX: we don't recommend using this linter before doing performance profiling.
     # For most programs usage of prealloc will be a premature optimization.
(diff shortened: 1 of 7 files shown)


ldez:
Hello,

I'm not sure to agree with the rule behind this linter.

And currently, it produces too many false positives, an example:
```go
package foobar

import (
	"net/url"
)

type Foo struct {
	URL *url.URL
}

func example(val string) (*Foo, error) {
	endpoint, err := url.Parse(val)
	if err != nil {
		return nil, err
	}

	if endpoint.Host != "example.com" {
		return nil, nil
	}

	return &Foo{URL: endpoint}, nil
}

func fiibar() (string, error) {
	return "", nil
}

```
```
foo.go:18:3: return both the `nil` error and invalid value: use a sentinel error instead (nilnil)
                return nil, nil
                ^
```

bombsimon:
> And currently, it produces too many false positives, an example:

What's false positive here, you do return `nil, nil` on line 18. The author (and me as well) wants you to return `nil, errors.New("invalid domain name")`

Also I'm not sure we should reject linters because it's not useful for _everyone_. I would say that a big part of the linters are super opinionated and will never be enabled by the vast majority. That's also why we don't enable new linters by default.

ldez:
> Also I'm not sure we should reject linters because it's not useful for everyone. 

It's not my point, I tested this linter on several real codebases, and it produces, from my point of view, a lot of false positives.

The cases where we need to not return an error are commons.

bombsimon:
> It's not my point, I tested this linter on several real codebases, and it produces, from my point of view, a lot of false positives.
>
> The cases where we need to not return an error are commons.

I think this is similar to `wrapcheck` and `goerr113`, there's a lot of places where you can or want to return an inline or unwrapped error but the linters forces you to handle them in a specific way for a consistent code style. This linter seem to prefer returning and checking error instead of nilness which is a pretty common pattern and found in stdlib in a quite a few places as well. I would not call those false positives but I understand what you're saying.

> I have just "blocked" the PR and expressed my doubts, then the discussion is still opened.

Yeah I know and that's why I contributed to it with my thoughts to the discussion. :smiley: Would of course be nice if more people had opinions about this as well.

> We enable the new linters by default when a user uses enable-all or presets wink

I think the outcome of the discussion(s) to keep `enable-all` was that people using it (including me) accepted to get linters with unintentional behaviors and if so manually disable them after upgrading `golangci-lint` so that would be the case for us using `enable-all` here as well.

Antonboom:
@ldez, what exactly is false positive? And what behavior do you expect?

Please, see linter readme and examples for more details.

There is no technical way to determine when it is valid to `return nil, nil` and when it is not - it depends on the context. The linter suggests that you generally abandon this pattern in favor of errors and checks all danger types (aptr, map, interface, channel and funcs).

If you use `return nil, nil` all over the place, then just do not include the linter, as @bombsimon wrote above.

ldez:
What I consider as false-positive you consider that a feature, then there is no false-positive.
I am still not convinced by this linter because for me it's not a relevant rule, but if some people think that it's useful, I will follow.

Antonboom:
@ldez thank u!

In fact, the linter did not appear out of nowhere. Using in my working projects `return nil, nil` resulted in:
- checks like `if err != nil ... if val != nil` at multiple levels of app (and you have to go through the whole chain to figure out what's the matter)
- occasional panics
- comments like "be careful, the function can return nil, nil" (the same is in the [std](https://github.com/Antonboom/nilnil#examples))

I myself would like to get more evidence base besides personal experience and common sense and I will be grateful for it.<br>
I tried to express my position in readme:
```
I understand that each case needs to be analyzed separately,
but I hope that the linter will make you think again - 
is it necessary to use an ambiguous API or is it better to do it using a sentinel error?

In any case, you can just not enable the linter.
```

So no problem. I just awaiting for feedback from other reviewers.

@bombsimon, thank u for support :)

ldez (approved):
LGTM

bombsimon (approved):
:100:

PR #2237, opened 2021-09-16
nlreturn: add block-size option

butuzov:
Adds configuration `block-size` that defines the size of the block (including return statement that is still "OK") so no return split is required.

Example

```yaml
linters-settings:
  nlreturn:
    block-size: 2
```


butuzov:
Currently blocked as v2.2.0 has and off-by-one error that brings  `panic`

.golangci.example.yml
```diff
@@ -158,6 +158,11 @@ linters-settings:
     # Exclude godoc examples from forbidigo checks.  Default is true.
     exclude_godoc_examples: false
 
+  nlreturn:
```
ldez, line 161: can move this section in this alphabetical place?

ldez (changes requested):

.golangci.example.yml
butuzov: sure. done.

butuzov:
@ldez it's blocked because the current version that supports block-size also have off-by-one error and panics. I am waiting for @ssgreg to merge my PR

butuzov:
@ldez maybe its worth make abc order for other parts too (linters settings, linters list in the manager) as another PR? It's so easy to miss (as I did)

ssgreg (approved):

butuzov:
@ssgreg awesome! can you release `v2.2.1`?

ldez (approved):

PR #2240, opened 2021-09-18
add varnamelen linter

blizzy78:
Please see https://github.com/blizzy78/varnamelen for details about this linter.


pkg/golinters/varnamelen.go
```diff
@@ -0,0 +1,48 @@
...
+		cfg,
+	)
+}
+
+func boolString(b bool) string {
+	if b {
+		return "true"
+	}
+	return "false"
+}
```
ldez, line 48:
```suggestion
```

pkg/golinters/varnamelen.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+func NewVarnamelen(settings *config.VarnamelenSettings) *goanalysis.Linter {
+	a := varnamelen.NewAnalyzer()
+
+	cfg := map[string]map[string]interface{}{}
+	if settings != nil {
+		vnlCfg := map[string]interface{}{
+			"checkReceiver": boolString(settings.CheckReceiver),
+			"checkReturn":   boolString(settings.CheckReturn),
```
ldez, line 21:
```suggestion
			"checkReceiver": strconv.FormatBool((settings.CheckReceiver),
			"checkReturn":   strconv.FormatBool((settings.CheckReturn),
```

pkg/lint/lintersdb/manager.go
```diff
@@ -520,6 +522,11 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithLoadForGoAnalysis().
 			WithURL("https://github.com/Antonboom/nilnil").
 			WithSince("v1.43.0"),
+		linter.NewConfig(golinters.NewVarnamelen(varnamelenCfg)).
+			WithPresets(linter.PresetStyle).
+			WithLoadForGoAnalysis().
```
ldez, line 527:
```suggestion
```

pkg/golinters/varnamelen.go
```diff
@@ -0,0 +1,48 @@
...
+
+		cfg[a.Name] = vnlCfg
+	}
+
+	return goanalysis.NewLinter(
+		a.Name,
+		"checks that the length of a variable's name matches its scope",
+		[]*analysis.Analyzer{a},
+		cfg,
+	)
```
ldez, line 40:
```suggestion
	).WithLoadMode(goanalysis.LoadModeSyntax)
```

ldez (changes requested):
There are some false positives with testing variables:

```go
package example

import "testing"

func TestName(t *testing.T) {
	println("FOO")
	println("FOO")
	println("FOO")
	println("FOO")
	println("FOO")
	println("FOO")
	t.Fail()
}

func BenchmarkName(b *testing.B) {
	for i := 0; i < b.N; i++ {
		println("FOO")
		println("FOO")
		println("FOO")
		println("FOO")
		println("FOO")
		println("FOO")
		b.Name()
	}
}
```

```console
$ ./golangci-lint run
foo_test.go:5:15: parameter name 't' is too short for the scope of its usage (varnamelen)
func TestName(t *testing.T) {
              ^
foo_test.go:15:20: parameter name 'b' is too short for the scope of its usage (varnamelen)
func BenchmarkName(b *testing.B) {
                   ^
```

At least `t *testing.T`, `b *testing.B`, `tb testing.TB`, `m *testing.M` must not produce reports.

blizzy78:
> At least `t *testing.T`, `b *testing.B`, `tb testing.TB`, `tb testing.TB`, `m *testing.M` must be not produce reports.

I'm curious, is that a requirement specifically for golangci-lint?

ldez:
All these elements are related to Go conventions and their names must always be short.

blizzy78:
Would it be necessary to check, for example, if `t *testing.T` is an actual Go standard library `testing.T`? After all, `testing` could be any arbitrary import name.

Also, regarding conventions, I'd guess `ctx context.Context` would have to be on the list of exceptions as well? My linter would produce a report for that if `linters-settings.varnamelen.min-name-length >= 4`

ldez:
> Would it be necessary to check, for example, if `t *testing.T` is an actual Go standard library `testing.T`? After all, `testing` could be any arbitrary import name.

Everything is possible but in most cases `testing` is used.

> Also, regarding conventions, I'd guess `ctx context.Context` would have to be on the list of exceptions as well?

Yes `ctx context.Context` is also in the main convention.

> My linter would produce a report for that if `linters-settings.varnamelen.min-name-length >= 4`

I don't think that is the solution. Your linter claims to follow a Go convention, in this case, you have to be precise and follow also the related conventions. Known short names should be excluded from the analysis.

blizzy78:
> you can check the packages aliases (even if I think it's useless)

I think I'll pass on checking aliases for now if that's okay for you. If not, please let me know.

blizzy78:
I've changed the linter such that the following parameters will now always be ignored:

- `t *testing.T`
- `b *testing.B`
- `tb Testing.TB`
- `pb *testing.PB`
- `m *testing.M`
- `ctx context.Context`

blizzy78:
I think the tests on Mac OS are failing because of something completely unrelated, no?

blizzy78:
@ldez, is there anything else you'd like me to change?

ldez:
I just want to run a performance test before merging

Issue #2137, opened 2021-07-29
Add support for remote config file


alexandrevilain:
### your feature request related to a problem? Please describe.

When working in a multiple services environnent it can be hard to maintain and to update lint rules at the team/project level.

When a library is no more maintained or with a big security issue, adding the rule to stop using it for each project can be pretty time consuming.

### Describe the solution you'd like.

Maintaining a single repository with lint rules for the project/team.
With a CI/CD pipeline pushing the latest rules to an S3 bucket , nexus repository or something else.

Then for each project/service the Makefile would be like:
```
.PHONY: lint
lint:
   golangci-lint run --config=https://team.repository.com/lint/.golangci-lint.yaml
```

### Describe alternatives you've considered.

- Using git submodules: It become as time consuming as updating the configuration file for each services,
- Cloning the repository for each CI pipeline to retrieve the lastest configuration file:  Developpers needs to do it on their dev computer too, bad dev experience.

### Additional context

Here is the implementation pull request: https://github.com/golangci/golangci-lint/pull/2136

bombsimon:
Would it make more sense to get the config as a dependent macro? You could mark that as `.PHONY` as well if you want to download it each time.

```make

.PHONY: lint
/tmp/.golangci.yml:
  curl https://team.repository.com/lint/.golangci-lint.yaml --output /tmp/.golangci.yml

lint: /tmp/.golangci.yml
  golangci-lint run --config=/tmp/.golangci.yml
```

alexandrevilain:
Hi! Thanks, that's a good workaround! 

Wouldn't be easier to get it integrated inside golangci-lint or is it out of scope ?

ldez:
Every remote interaction is a possible security issue, it would be better to not embed this kind of remote access in the binary.

SVilgelm:
also, the reading remote file requires some additional configuration options:
- custom CA certificate
- ignoring TLS
- timeout
- custom http headers

it is really better to download the config file by curl

SVilgelm:
also it would be great to support git protocol

thepwagner:
Does https://github.com/golangci/golangci-lint/pull/2241 satisfy these use cases?

If you want to use `curl`, it should make combining the "team" settings with any local overrides easier.

[...]

Issue #1141, opened 2020-05-17
Make and promote recommended presets like in eslint, tslint


jirfag:
## Motivation
1. Some linters are duplicated partially or fully: [errcheck, goerr13, staticcheck], [modguard, depguard], [golint, stylecheck]
2. Some really cool linters aren't used because there are too many linters and only some of them are enabled by default.
3. Configuration became too large. Even maintainers don't know all configuration options and linters.

## Proposal
We need to research how eslint, tslint, etc work with that. My first guess is to introduce presets like `golangci:recommended` and add ability to share and reuse such presets like [airbnb-eslint](https://www.npmjs.com/package/eslint-config-airbnb).

thepwagner:
https://github.com/golangci/golangci-lint/issues/2137 is perhaps related, especially when this idea gets into shared presets like `airbnb-eslint`.

ringerc:
I quickly hit issues with trying to promote a preferred default set of `golangci-lint` rules for my org, while preserving individual teams' ability to customise, add to and override them on a per-component / per-repo basis. Short of hacks in build automation like pre-processing a `.golangci.yaml.d` snippets directory to generate a config as part of build automation I didn't find a sensible way to do it.

The lightest-way to satisfy this might be to support a config-directory in `golangci-lint` like a `.golangci.yaml.d`. If the configuration points to a directory, all files in that directory are read in locale "C" sort-order and merged. How well this would work will depend a bit on how the config is expressed though; to be useful it has to be possible to match and delete array entries created by earlier configs.

ldez:
Just to link the discussion topic: https://github.com/golangci/golangci-lint/discussions/3954

chaindead:
### There is my proposal for problem

What I actually want is boring: inherit a *versioned* base config and adjust it locally, without copy-paste drift and without the MVS surprise.

Borrow the model GitLab CI already proved at scale: an `include` that names one or more base configs, with your local file deep-merged on top. No new subcommand and no separate step - `golangci-lint run` keeps working as-is, so editors and CI don't need to know anything changed.

```yaml
version: "2"

include:
  # A versioned base from another repo. `ref` is pinned explicitly, so no MVS.
  - project: github.com/acme/go-linter
    ref: v1.4.0
    file: .golangci.base.yml
  # ...or a plain URL (CI / public bases).
  - remote: https://raw.githubusercontent.com/acme/go-linter/v1.4.0/.golangci.base.yml
[...]

PR #2241, opened 2021-09-18
feat: sharable config presets

thepwagner:
Add an implementation of "shareable presets", as discussed by https://github.com/golangci/golangci-lint/issues/1141 .

This is configured by a new root element, like:

```yaml
presets:
- ./.golangci.myorg.yml  # an additional configuration file on disk, maybe maintained via curl
- github.com/thepwagner/golangci-lint-yaml # a go module, which must contain a configuration file
```

That feature set is aiming to align with the opinions in https://github.com/golangci/golangci-lint/issues/2137 - minimize additional remote access by delegating to the go module system. Presets can be pinned in `go.sum`, fetched with `go mod download`, vendored, change-managed by Dependabot, etc. This mirrors alternatives like [eslint-config-airbnb](https://www.npmjs.com/package/eslint-config-airbnb). Authentication is delegated to go modules: I've tested using `ssh` and a private repo.

Presets are not loaded recursively, to limit overhead and complexity.

Preset configurations are merged rather ungracefully: viper does the heavy lifting. The exception is the `.linters.enabled` and `.linters.disabled` lists, which are hand-merged to avoid the most expected repetition. This merge operation is the biggest area for improvements.

The included test references a public repo that worksonmymachine(TM), this is just for a demo and should be replaced with something embedded or controlled by this org.
https://github.com/thepwagner/archivist/pull/130 is a standalone sample of the changes required to a project to adopt a shared preset. The ugly bit is treating the lint configuration as a Go dependency.

### DIFF
--- a/pkg/config/reader.go
+++ b/pkg/config/reader.go
@@ -5,10 +5,12 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/mitchellh/go-homedir"
 	"github.com/spf13/viper"
+	"golang.org/x/tools/go/packages"
 
 	"github.com/golangci/golangci-lint/pkg/fsutils"
 	"github.com/golangci/golangci-lint/pkg/logutils"
@@ -43,17 +45,18 @@ func (r *FileReader) Read() error {
 		return fmt.Errorf("can't parse --config option: %s", err)
 	}
 
+	v := viper.New()
 	if configFile != "" {
-		viper.SetConfigFile(configFile)
+		v.SetConfigFile(configFile)
 	} else {
-		r.setupConfigFileSearch()
+		r.setupConfigFileSearch(v)
 	}
 
-	return r.parseConfig()
+	return r.parseConfig(v)
 }
 
-func (r *FileReader) parseConfig() error {
-	if err := viper.ReadInConfig(); err != nil {
+func (r *FileReader) parseConfig(v *viper.Viper) error {
+	if err := v.ReadInConfig(); err != nil {
 		if _, ok := err.(viper.ConfigFileNotFoundError); ok {
 			return nil
 		}
@@ -61,7 +64,7 @@ func (r *FileReader) parseConfig() error {
 		return fmt.Errorf("can't read viper config: %s", err)
 	}
 
-	usedConfigFile := viper.ConfigFileUsed()
+	usedConfigFile := v.ConfigFileUsed()
 	if usedConfigFile == "" {
 		return nil
 	}
@@ -72,8 +75,12 @@ func (r *FileReader) parseConfig() error {
 	}
 	r.log.Infof("Used config file %s", usedConfigFile)
 
-	if err := viper.Unmarshal(r.cfg); err != nil {
+	if err := v.Unmarshal(r.cfg); err != nil {
 		return fmt.Errorf("can't unmarshal config by viper: %s", err)
+	}
+
+	if err := r.mergePresets(v); err != nil {
+		return fmt.Errorf("can't merge presets: %s", err)
 	}
 
 	if err := r.validateConfig(); err != nil {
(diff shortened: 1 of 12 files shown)


test/testdata/configs/preset_module.yml
```diff
@@ -0,0 +1,3 @@
+# TODO: host a preset in the golangci org?
+presets:
+ - github.com/thepwagner/golangci-lint-yaml
```
miry, line 3: Could it be located in `golangci-lint/test/testdata` ?

thepwagner: Yep, but I'm not sold on that path: the goal of this test is to exercise loading the configuration from a go module.

Using a `replace` directive, we could load from a go module that happens to be housed at `./test/testdata/...`:
```
require github.com/golangci/golangci-lint-yaml v0.0.1
replace github.com/golangci/golangci-lint-yaml => ./test/testdata/some-new-dir
```

I figured since https://github.com/golangci/golangci-lint/issues/1141 mentions "presets like `golangci:recommended`", using such a preset here would be the best solution. To shoe-horn that idea into this proposal, a module like `github.com/golangci/golangci-recommended` would be the equivalent of `golangci:recommended` preset.

--- 

Any repo with a `.golangci.yml` should work - using an existing dependency like [github.com/ldez/gomoddirectives](https://github.com/ldez/gomoddirectives/blob/v0.2.2/.golangci.yml) would have been less egregious to the dependency graph, but I thought the pattern of a module containing _only_ the configuration was more interesting to include in the proposal.

miry: I see.

In my case I would like have any github or https path (not exactly the go module or some convention of names).

```yaml
presets:
  - local/path/config.yml
  - https://raw.githubusercontent.com/golangci/golangci-lint/master/.pre-commit-hooks.yaml
  - github:golangci/golangci-lint/master/subfolder/docs/common.yaml
```

pkg/config/reader.go
```diff
@@ -128,6 +135,105 @@ func (r *FileReader) validateConfig() error {
...
+}
+
+const configName = ".golangci"
+
+func (r *FileReader) loadPresets(v *viper.Viper) ([]*viper.Viper, error) {
+	cfgBase := filepath.Dir(v.ConfigFileUsed())
+	presets := make([]*viper.Viper, 0, len(r.cfg.Presets))
+	for _, preset := range r.cfg.Presets {
+		presetV := viper.New()
+		if strings.HasPrefix(preset, ".") {
```
miry, line 199: It means the local path always should be started with `./` ?

thepwagner: Or `../`. The pattern for distinguishing modules from local paths is inspired by the [right hand side of replace directives](https://golang.org/ref/mod#go-mod-file-replace) in `go.mod`.

> If the path on the right side of the arrow is an absolute or relative path (beginning with ./ or ../), it is interpreted as the local file path to the replacement module root directory, which must contain a go.mod file.
> [...]
> If the path on the right side is not a local path, it must be a valid module path.

`/` might also be a valid prefix to trigger file loading that I ~missed~ did not think would be used.

test/testdata/configs/preset_module.yml
thepwagner: > any https path

Per https://github.com/golangci/golangci-lint/issues/2137#issuecomment-891621683 , I don't think `https://` paths would be a welcome addition. If you want an HTTPS resource, they can be `curl`-ed first and used as a file preset.

> any github path

That might be useful, e.g. to have multiple configurations in a single repository.

Would it be acceptable for that path to also be a valid go module? Again, I  wanted to avoid adding outgoing network calls, authentication etc. Only using files provided by `go mod` is an attempt to walk that line.

miry: It make sense. Thank you for the hint. 

In this case then I would propose first support only local path and think about go mod solution in separate PR.

thepwagner:
@ldez do you have any feedback on this?

Nivl:
I know this is old but I do like this idea. This solves a bunch of issues for organizations with hundreds of repos. The breaking changes in 1.53 (like https://github.com/golangci/golangci-lint/issues/3906) would have been much more manageable with something like this in place.

Not a huge fan of having to hardcode the packages in a `lint.go` file though but it's not terrible either. At the very least a V1 that supports merging local files would go really far. Users can use curl, submodules, etc. to bring in anything else in the short term.

@ldez I see that you've set yourself as reviewer. Do you have any opinions about it? I don't mind opening a PR to work/help on that depending on what @thepwagner wants to do.

ldez:
This implementation based on external files will create problems, for example with the invalidation of the cache, another is related to security with URL.

It's not the first PR on this topic and the problems are always the same.

So before working on something, I recommend making proposals inside #1141.

Issue #2261, opened 2021-09-27
Full-file version of new-from-rev?


natefinch:
### Your feature request related to a problem? Please describe.

I want to encourage people to fix linting errors in their repos without requiring them to fix every single error in the whole repo at once.  New-From-rev seems like it would be good for this, except that it's too narrowly focused. It forces you to fix linting errors on any lines you change or new code you add... but if you never touch other lines of code, it'll never bug you about fixing them.

I'd like a version of new-from-rev where it'll show all linting errors from any *file* you touch.  That way the ripple effects go a lot further, and you'll be a lot more likely to fix more things as you go along.

### Describe the solution you'd like.

I figure this could be an almost copy and paste of the new-from-rev code, except any matches in files that have been touched are shown.

### Describe alternatives you've considered.

You could have a script that makes a fake patch that includes the full code of all the lines in the PR and use new-from-patch, but that seems like a huge hack that might be prone to breaking.

### Additional context.

I am willing to code this up if people think it's a good idea.

SVilgelm:
@natefinch I'm ok with this feature, it has pros and cons. Please feel free to implement this. But think about not copy-pasting, but refactoring. I don't want to see one more flag with similar functionality, it would be better if we have one more flag that will control the functionality of `new-from-rev`

natefinch:
Yep. That's exactly how I'm doing it. Just a bool that applies to either new from rev or new from patch. I should have a PR ready soon. Writing tests now

natefinch:
I thought about a different way this could be handled, but I think it has an annoying and fatal flaw. 

There could be a "distance from diff" value that defaults to 0, i.e. issues only in exact lines you touch.  If you set it to -1, it would do the whole file, as above. Otherwise you could set it to like 10, and any issues within 10 lines of lines you've changed would "count". That sounds cool, except every time you fix a line 10 lines up, the demarcation moves up 10 lines from *that* change, potentially finding more issues, etc. etc.  Super annoying.  At least with full file, you know when you're done, and there's no chance a fix to an old error brings in a new one.

bombsimon:
Sounds like a good start to just mark the whole file, I think the use cases for the distance feature would be very few and not worth a complex implementation.

ldez:
Closed by  #2264

PR #2264, opened 2021-09-29
add support for finding issues in entire files in a diff

natefinch:
This implements #2261 

It adds a flag called "whole-files" which changes the behavior of new-from-rev and new-from-patch so they report all issues in a file that has been changed, not just on the exact lines that have been updated.

Requires https://github.com/golangci/revgrep/pull/6. (and I'll need to update that import version after that PR is merged)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	github.com/golangci/lint-1 v0.0.0-20191013205115-297bf364a8e0
 	github.com/golangci/maligned v0.0.0-20180506175553-b1d89398deca
 	github.com/golangci/misspell v0.3.5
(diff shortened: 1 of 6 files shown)


natefinch:
You may note that there are no tests here... the tests are all in revgrep.  I would be happy to write tests here as well, but I couldn't find other tests of this code, and starting from scratch is rough.  Suggestions welcome.

SVilgelm:
revgrep PR is merged, please update go.mod

bombsimon (approved):
LGTM!

Thank you very much for the implementation! :rocket:

ldez (approved):

natefinch:
Thanks!  Do I need to make a separate PR for the web docs, or does that get auto-generated?

PR #2274, opened 2021-10-04
Add govarpkg linter

alexal:
Finds variables that collide with imported package names. Inspired by built-in linter in IntelliJ.

https://github.com/alexal/govarpkg

### DIFF
--- a/pkg/lint/lintersdb/manager.go
+++ b/pkg/lint/lintersdb/manager.go
@@ -529,6 +529,10 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithPresets(linter.PresetStyle).
 			WithLoadForGoAnalysis().
 			WithURL("https://github.com/sivchari/tenv"),
+		linter.NewConfig(golinters.NewGoVarPkg()).
+			WithPresets(linter.PresetStyle).
+			WithLoadForGoAnalysis().
+			WithURL("https://github.com/alexal/govarpkg"),
 
 		// nolintlint must be last because it looks at the results of all the previous linters for unused nolint directives
 		linter.NewConfig(golinters.NewNoLintLint()).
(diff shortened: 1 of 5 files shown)


pkg/lint/lintersdb/manager.go
```diff
@@ -529,6 +529,10 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithPresets(linter.PresetStyle).
 			WithLoadForGoAnalysis().
 			WithURL("https://github.com/sivchari/tenv"),
+		linter.NewConfig(golinters.NewGoVarPkg()).
+			WithPresets(linter.PresetStyle).
+			WithLoadForGoAnalysis().
```
ldez, line 534:
```suggestion
```

test/testdata/govarpkg.go
```diff
@@ -0,0 +1,13 @@
...
+	"fmt"
+	"github.com/alexal/govarpkg/pkg/tst"
+)
+
+func test() {
+	a := "test"
+	fmt.Println(a)
+	tst := tst.Name{}
+	tst.New("Alex")
+}
```
ldez, line 13:
```suggestion
//args: -Egovarpkg
package testdata

import (
	"fmt"
	"strings"
)

func govarpkg() {
	a := []string{"a", "b", "c"}

	strings := strings.Join(a, ",") // ERROR `Variable 'strings' collides with imported package name`
	fmt.Println(strings)
}

```

ldez (changes requested):

butuzov:
Isn't this covered by `importShadow` by [importShadow_](https://github.com/go-critic/go-critic/blob/f599c814bd8d49e04e1ba5d1ced26782127e42f5/checkers/importShadow_checker.go)

ldez:
Good catch @butuzov, I have looking for this rule but I didn't find it.

https://go-critic.github.io/overview#importShadow-ref

We will close the PR as it's a duplicate.

PR #2282, opened 2021-10-08
chore: use go1.17 in go.mod

ernado:
This allows lazy loading.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/golangci/golangci-lint
 
-go 1.15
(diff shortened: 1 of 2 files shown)


ldez:
I'm not sure about this change, because it will "allow" using go1.17 elements in the code, and this can be a problem for people that use the "tools" pattern (use an import to get the tools) to get golangci-lint.

Related to #2183

ernado:
Closing for now, will not move to 1.17 until 1.18 is released.

PR #2294, opened 2021-10-15
Add goBadWords Linter

Ghvstcode:
https://github.com/Ghvstcode/goBadWord

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Antonboom/nilnil v0.1.0
 	github.com/BurntSushi/toml v0.4.1
 	github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24
-	github.com/Ghvstcode/goBadWord v0.0.0-20211016122609-f965cdf7e900
+	github.com/Ghvstcode/goBadWord v0.0.0-20211016123036-94534bffdbfe
 	github.com/OpenPeeDeeP/depguard v1.0.1
 	github.com/alexkohler/prealloc v1.0.0
 	github.com/ashanbrown/forbidigo v1.2.0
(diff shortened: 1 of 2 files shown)


ldez:
Also the linter repository contains a binary, can you remove it?

SVilgelm:
Please use the `embed` feature to include the list of bad words, otherwise you will need to supply that json with golagnci-lint

Ghvstcode:
> Please use the `embed` feature to include the list of bad words, otherwise you will need to supply that json with golagnci-lint

Hey @SVilgelm 👋🏼, What is the embed feature? & How do I add the list of bad words properly or supply the json file to golangci-lint

Ghvstcode:
> This PR rewrites a lot of unrelated functionality. Can you please reset your branch and commit only related to your linter changes?

Sorry about that, Fixed!

Ghvstcode:
> Also the linter repository contains a binary, can you remove it?
> 
> https://github.com/Ghvstcode/goBadWord/tree/main/cmd/badWord

Good catch! Fixed.

ldez:
The "embed" approach will not work well with the "tools pattern" (tools defined in a go.mod file), it's preferable to not use [embed](https://pkg.go.dev/embed) in the context of golangci-lint. It's better to use a simple hardcoded list inside Go Code for the context of golangci-lint.

butuzov:
What is the name of this tool - badwords, badword or gobadword?

Ghvstcode:
> The "embed" approach will not work well with the "tools pattern" (tools defined in a go.mod file), it's preferable to not use [embed](https://pkg.go.dev/embed) in the context of golangci-lint. It's better to use a simple hardcoded list inside Go Code for the context of golangci-lint.

Aha, I see! Fixed. It’s a long list of words tho'

butuzov:
- Help Entry is ~5 kb long. If you make it shorter you can help your own project
- Entry in help refer to keyword  is different than real keywords (-bad-words vs `-badWord`)
- How to use it (I haven't found readme or something)?
- project require LICENCE
- panics...
- it doesn't catch what it suppose to (i tested with variable names `foobar` and the same keyword).

@Ghvstcode this project to fresh even for review. Can you take a bit more work polishing it?

Ghvstcode:
> here is what I have found:
> 
> * Help Entry is ~5 kb long. Make it shorter to help the project be more user-oriented.
> * Entry in help refer to keyword  is different than real keywords (`-bad-words` vs `-badWord`).
> * How to use it (I haven't found readme or something)?
> * golangci-lint linters require LICENCE
> * It panics at least in one place...
> * It doesn't catch what it suppose to (i tested with variable names `foobar` and the same keyword).
> 
> @Ghvstcode this project is way too fresh and untested even for review. Can you take a bit more work polishing it?

Hey, I would do some work on it and come back with another PR. Thanks for your time.

PR #2304, opened 2021-10-23
Add documentation for go-critic and ruleguard settings

sebastien-rosset:
Add examples for `gocritic` and `ruleguard` settings.
This documents the settings for `go-critic` which has been upgraded to `0.6.1`.

Depends on #2308

PR #2308, opened 2021-10-25
gocritic: add support for variable substitution in ruleguard path settings

### LINKED ISSUE #1662: Path for gocritic ruleguard rules depends on the current working directory
I am using [`ruleguard`](https://github.com/quasilyte/go-ruleguard) via the `gocritic` linter. As part of using this linter, you must set the path to your rules:

https://github.com/golangci/golangci-lint/blob/750309d1574e9a4afc32346364f9e4ae36955692/.golangci.example.yml#L178-L180

Unfortunately, if you currently set a relative path, it will be relative to the directory that you run `golangci-lint` from. This means that you must always run `golangci-lint` from the same directory or use an absolute path, which may not work across different environments.

https://github.com/golangci/golangci-lint/pull/1572 dealt with a similar issue for custom plugins. It fixed the issue by adding an explicit check to see if the path was relative or absolute.

I looked into doing this for the `ruleguard` settings, but it appears that the [settings code](https://github.com/golangci/golangci-lint/blob/750309d1574e9a4afc32346364f9e4ae36955692/pkg/golinters/gocritic.go#L78-L109) is very generic and doesn't know anything in particular. Would a PR that added a special case to that code for `ruleguard` be accepted? I would understand if it would be considered in bad taste.

A more general alternative might be if `golangci-criitc` changed the working directory to the configuration dir directory before running the linters. However, I suspect this would cause other problems.

sebastien-rosset:
It makes it possible to use variable substitution in the .golangci.yaml file for the gocritic `ruleguard` checker:

```yaml
ruleguard:
        # path to a gorules file for the ruleguard checker
        rules: '${configDir}/rules/*.go'
```

In the above configuration, `${configDir}` is substituted by the directory where the .golangci.yaml config file was found.

For example, suppose the config file is at `/users/bob/.golangci.yaml` and the ruleguard files are at `/users/bob/ruleguard/*.go`. Then previously the gocritic settings had to be configured with something like this: 

```yaml
# The gocritic linter has a `ruleguard` checker 
ruleguard:
  # comma-separated list of glob paths to gorules files for the ruleguard checker
  rules: '/users/bob/ruleguard/*.go'
```

If the ruleguard paths are relative, they are relative to the working directory where the `golangci-lint` command runs, not the directory where the .golangci.yaml exists.

*Details:*
1. Add new `GetConfigDir()` function in `pkg/config/config.go`
1. In the gocritic settings, perform variable substitution before passing the parameters to the gocritic checkers. Specifically, resolve `${configDir}` to the value of `Config.GetConfigDir()`
1. Add code comments to document existing functions.
1. Fix `.golangci.example.yml` to remove warnings that can easily be prevented. Using the file as is was causing the following issues. The file (intentionally?) causes validation issues as is, but still it makes sense to reduce the number of issues.

```
golangci-lint run -c .golangci.example.yml 
ERRO Unable to load custom analyzer example:/path/to/example.so, plugin: not implemented 
WARN No need to enable check "rangeValCopy": it's already enabled 
WARN Gocritic settings were provided for not enabled check "unnamedresult" 
WARN Gocritic settings were provided for not enabled check "nestingreduce" 
WARN Gocritic settings were provided for not enabled check "truncatecmp" 
WARN Gocritic settings were provided for not enabled check "ruleguard" 
ERRO Running error: --presets is incompatible with --enable-all 
```

If this PR is accepted, then I will add more details to the documentation in #2304.

Fixes #1662
Fixes go-critic/go-critic#1143


daixiang0 (approved):

SVilgelm (changes requested):
LGTM, but it definitely needs some tests. Please add at least the functional tests

sebastien-rosset:
> LGTM, but it definitely needs some tests. Please add at least the functional tests

Good point. There didn't seem to be any functional test for `ruleguard` rules. 
1. Add ruleguard check in test/testdata/gocritic.go
2. Add ruleguard rules in test/testdata/configs/ruleguard
3. Add `debug`, `failOn` and `rules` parameters in the test gocritic functional test

go.mod
```diff
@@ -63,6 +63,7 @@ require (
 	github.com/pkg/errors v0.9.1
 	github.com/polyfloyd/go-errorlint v0.0.0-20210722154253-910bb7978349
 	github.com/prometheus/procfs v0.6.0 // indirect
+	github.com/quasilyte/go-ruleguard v0.3.13 // indirect
```
sebastien-rosset, line 66: Without this line, the `ruleguard` functional test fails. Specifically in [test/testdata/configs/ruleguard/dup.go](https://github.com/golangci/golangci-lint/pull/2308/files#diff-e4e65433aef638f729eaf0ac7108a187f389ad64d64ae7377792b50a5f735fd2R3), ruleguard is not able to import `github.com/quasilyte/go-ruleguard/dsl`

```
go get github.com/quasilyte/go-ruleguard
```

SVilgelm: this happens because you put the `ruleguard` package to `testdata` folder, go ignores it. Please move under `test` folder

sebastien-rosset: Ah thank you, that makes sense now. I am updating the PR.

test/ruleguard/dup.go
```diff
@@ -0,0 +1,22 @@
+// go:build ruleguard
+package gorules
```
SVilgelm, line 2: let's have the package's name same as folder's name, it confuses a bit

sebastien-rosset: done

SVilgelm: thanks

SVilgelm (approved):
LGTM

ldez:
There is something that I don't understand: I added myself as a reviewer, some people approved without waiting, after that I added myself as assignee, and the PR is merged.
I'm not sure if it's the right workflow...

SVilgelm:
sorry, didn't pay attention

SVilgelm:
we can revert if anything is wrong here

PR #2318, opened 2021-10-31
Replace ioutil with io and os

ahrtr:
The ioutil package has already been deprecated in golang 1.16, please see [go1.16#ioutil](https://golang.org/doc/go1.16#ioutil). So we need to replace all the ioutil with os and io.

The ioutil used in yaml or txt files are not updated in this PR.

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -370,7 +370,7 @@ func (c *Cache) putIndexEntry(id ActionID, out OutputID, size int64, allowVerify
 		// Truncate the file only *after* writing it.
 		// (This should be a no-op, but truncate just in case of previous corruption.)
 		//
-		// This differs from ioutil.WriteFile, which truncates to 0 *before* writing
+		// This differs from os.WriteFile, which truncates to 0 *before* writing
 		// via os.O_TRUNC. Truncating only after writing ensures that a second write
 		// of the same content to the same file is idempotent, and does not — even
 		// temporarily! — undo the effect of the first write.
--- a/internal/cache/cache_test.go
+++ b/internal/cache/cache_test.go
@@ -8,7 +8,6 @@ import (
 	"bytes"
 	"encoding/binary"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"path/filepath"
 	"testing"
@@ -22,7 +21,7 @@ func init() {
 func TestBasic(t *testing.T) {
 	t.Parallel()
 
-	dir, err := ioutil.TempDir("", "cachetest-")
+	dir, err := os.MkdirTemp("", "cachetest-")
 	if err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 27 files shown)


bombsimon (approved):
Thanks! LGTM

ldez (changes requested):
I just want to be precise: the `ioutil` package is not really deprecated, there is no depreciation defined in this package.

The Go team recommends using the new implementation in the new code, but they don't recommend removing it from the existing code (they say nothing about existing code).

> ... we encourage new code to use the new definitions in the io and os packages. 
https://golang.org/doc/go1.16#ioutil

---

For me, the code in the `testdata` package must keep, at least for now, the use of `ioutil`.

bombsimon:
What do you mean, the linked titel literally says "Deprecation of ioutil"? However, what deprecated means may vary, it's nothing guaranteeing deprecated tools to stop working.

Russ Cox explains this [on twitter](https://twitter.com/_rsc/status/1351676094664110082):
> Seeing discussion of io/ioutil deprecation in various places. 
>
> To be clear: in Go, “deprecated” means only “there is a better, preferred way.” It does not mean “will break in the future.”
>
> Today’s programs using ioutil.ReadFile and friends will keep working forever. 

I see no reason not to use the "better, preferred way" just because Go guarantees it won't stop working.

> For me, the code in the `testdata` package must keep, at least for now, the use of `ioutil`.

Do you mind elaborate on why this is a must for you?

ldez:
> What do you mean

I mean there are literally no depreciation comments in the package `ioutil`.

It was just to be clear about the real context of this PR.

> Do you mind elaborate on why this is a must for you?

As the package `ioutil` will still be here, I think it's a good idea to keep some tests that use this package.

FYI the CI fails because of changes in the `testdata` packages

bombsimon:
> As the package ioutil will still be here, I think it's a good idea to keep some tests that use this package.

Thanks for explaining. I was under the impression that it wouldn't still be here but maybe you mean in general in the Go ecosystem.

> FYI the CI fails because of the changes in the testdata package.

Yeah I didn't dig into the root cause but surely there must be a way to make testdata work without `iotuil` I hope.

But if this isn't something that should change in `golangci-lint` I think it makes more sense to close this PR and keep the usage in testdata. If we want to make this change in the future it's probably easier to just file a new PR. WDYT?

ldez:
> I was under the impression that it wouldn't still be here but maybe you mean in general in the Go ecosystem.

Yes, it will still be in Go code base and Go libraries.

> Yeah I didn't dig into the root cause but surely there must be a way to make testdata work without iotuil I hope.

the `testdata` package contains files used to test the linters. The goal of those files is to evaluate linters, then they always "work".

> But if this isn't something that should change in golangci-lint I think it makes more sense to close this PR and keep the usage in testdata. If we want to make this change in the future it's probably easier to just file a new PR. WDYT?

It's not what I'm trying to say.

There are a few weeks we dropped go1.15 constraints in the CI (2021-10-17) and updated the minimal requirement to go1.16 (2021-10-20 https://github.com/golangci/golangci-lint/pull/2041#issuecomment-948004450).

Yes, now we can use the `ioutil` package replacements through the `io` and `os` package.

The linters are here to evaluate all the Go code even `ioutil` related code, and some tests are here only for this package (ex: `func TestErrcheckIgnoreIoutil()`)

So my opinion is to remove `ioutil` from the golangci-lint code but not in the existing linter test files.

ahrtr:
Thanks. I just reverted all changes in testdata package.

ldez (approved):
LGTM

PR #2319, opened 2021-11-01
.golangci.example.yml: clarify nolintlint.allow-unused

extemporalgenome:
nolintlint's allow-usued option comment is misleading: use of "enable" suggests that setting the value to true will ensure all directives are used. Further, it's unclear what "used" means in this context (it could be misread that every possible linter must be specified in at least one nolint directive).


ldez (approved):
LGTM

Issue #2329, opened 2021-11-01
Add NebulousLabs/glyphcheck as linter


adamdecaf:
### Your feature request related to a problem? Please describe.

Homoglyph attacks are an increasing concern for source code and currently hard to detect. There are some tools available for detecting suspect UTF-8 runes. 

### Describe the solution you'd like.

Including support for running a homoglyph linter would be a good first step. 

### Describe alternatives you've considered.

There are few if any good homoglyph linters available for the Go language. Perhaps a new linter should be written.

https://github.com/search?l=Go&q=homoglyph&type=Repositories 

### Additional context.

https://github.com/NebulousLabs/glyphcheck

bombsimon:
A liter needs to be written using the [go/analysis](https://pkg.go.dev/golang.org/x/tools/go/analysis) package to be merged to `golangci-lint`. It doesn't look like any of the ones in the search result does this so someone would need to write a new linter for `golangci-lint` to pick it up.

ldez:
Hello,

`asciicheck` already check that cases

ldez:
@adamdecaf you can follow the PR #2330

PR #2330, opened 2021-11-01
Add github.com/breml/bidichk linter

breml:
[bidichk](https://github.com/breml/bidichk) checks for dangerous unicode character sequences

The following unicode characters are considered dangerous:

* U+202A: LEFT-TO-RIGHT-EMBEDDING
* U+202B: RIGHT-TO-LEFT-EMBEDDING
* U+202C: POP-DIRECTIONAL-FORMATTING
* U+202D: LEFT-TO-RIGHT-OVERRIDE
* U+202E: RIGHT-TO-LEFT-OVERRIDE
* U+2066: LEFT-TO-RIGHT-ISOLATE
* U+2067: RIGHT-TO-LEFT-ISOLATE
* U+2068: FIRST-STRONG-ISOLATE
* U+2069: POP-DIRECTIONAL-ISOLATE

See following links for background:

* ['Trojan Source' Bug Threatens the Security of All Code](https://krebsonsecurity.com/2021/11/trojan-source-bug-threatens-the-security-of-all-code/)
* [Trojan Source - Proofs-of-Concept](https://github.com/nickboucher/trojan-source)
* [Go proposal: disallow LTR/RTL characters in string literals?](https://github.com/golang/go/issues/20209)
* [cockroachdb/cockroach - PR: lint: add linter for unicode directional formatting characters](https://github.com/cockroachdb/cockroach/pull/72287)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/bkielbasa/cyclop v1.2.0
 	github.com/blizzy78/varnamelen v0.3.0
 	github.com/bombsimon/wsl/v3 v3.3.0
+	github.com/breml/bidichk v0.1.0
 	github.com/butuzov/ireturn v0.1.1
 	github.com/charithe/durationcheck v0.0.9
 	github.com/daixiang0/gci v0.2.9
--- a/go.mod
+++ b/go.mod
@@ -15,7 +15,7 @@ require (
 	github.com/bkielbasa/cyclop v1.2.0
 	github.com/blizzy78/varnamelen v0.3.0
 	github.com/bombsimon/wsl/v3 v3.3.0
-	github.com/breml/bidichk v0.1.0
+	github.com/breml/bidichk v0.1.1
 	github.com/butuzov/ireturn v0.1.1
 	github.com/charithe/durationcheck v0.0.9
 	github.com/daixiang0/gci v0.2.9
--- a/go.sum
+++ b/go.sum
@@ -106,6 +106,12 @@ github.com/blizzy78/varnamelen v0.3.0 h1:80mYO7Y5ppeEefg1Jzu+NBg16iwToOQVnDnNIoW
 github.com/blizzy78/varnamelen v0.3.0/go.mod h1:hbwRdBvoBqxk34XyQ6HA0UH3G0/1TKuv5AC4eaBT0Ec=
 github.com/bombsimon/wsl/v3 v3.3.0 h1:Mka/+kRLoQJq7g2rggtgQsjuI/K5Efd87WX96EWFxjM=
 github.com/bombsimon/wsl/v3 v3.3.0/go.mod h1:st10JtZYLE4D5sC7b8xV4zTKZwAQjCH/Hy2Pm1FNZIc=
+github.com/breml/bidichk v0.0.0-20211101211321-e7013e04fb35 h1:gVuw76onitoSg977sQ5CEK87U5sV17UHhtn0VmBaW3g=
+github.com/breml/bidichk v0.0.0-20211101211321-e7013e04fb35/go.mod h1:zbfeitpevDUGI7V91Uzzuwrn4Vls8MoBMrwtt78jmso=
+github.com/breml/bidichk v0.0.0-20211101215529-9b68d04c240e h1:vYW6FE3czRa1PWZHaarmH+rXqjxAxl5+4Cbda3793S4=
+github.com/breml/bidichk v0.0.0-20211101215529-9b68d04c240e/go.mod h1:zbfeitpevDUGI7V91Uzzuwrn4Vls8MoBMrwtt78jmso=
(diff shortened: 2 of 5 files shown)


SVilgelm:
Hi, I have checks the code of the linter. Here is my thoughts:
1. this code https://github.com/breml/bidichk/blob/master/pkg/bidichk/bidichk.go#L26-L30 canoe simplified:
```go
return check(f.Name(), f.Pos(0), pass) == nil
```
2. It will be great to have an ability to configure the list of forbidden codes: 
```go
func SetCodes(codes map[string]rune) &analysis.Analyzer {
    disallowedRunes = codes
    return Analyzer
}
```

SVilgelm:
3. please bump a release, in this case the Dependabot will be able to upgrade the version of the linter

ldez:
Hello,

`asciicheck` already checks that cases.

https://golangci-lint.run/usage/linters/#disabled-by-default-linters--e--enable

Quick example:

<details>

```yml
linters:
  disable-all: true
  enable:
    - asciicheck

```

</details>


<details>
<summary>example.go</summary>

```go
package sandbox

import "fmt"

func gitһub() {
	fmt.Println("Hello")
}

```

</details>

```console
$ golangci-lint run
foo.go:7:6: identifier "gitһub" contain non-ASCII character: U+04BB 'һ' (asciicheck)
func gitһub() {
     ^
```

https://github.com/golangci/golangci-lint/issues/2329#issuecomment-956731474

ldez:
I checked and it seems that asciicheck doesn't analyze comments, but I'm not sure how to exploit a comment with this kind of character.
It also doesn't check literal string, maybe asciicheck can be improved.

breml:
@ldez Have a look at https://github.com/nickboucher/trojan-source/blob/main/Go/commenting-out.go for examples how to exploit with comments. And yes, asciicheck does not check this case. bidichk does not operate on the parsed/tokenized Go code. It inspects the body of the file it self and therefore also string literals and comments. The whole attack works sort of similar to SQL injection attacks.

ldez:
I tried to run the file, but on Linux with go1.17 or go1.16, I don't reproduce.
Maybe it's only working on MacOS.

ldez:
> bidichk does not operate on the parsed/tokenized Go code. It inspects the body of the file it self and therefore also string literals and comments.

I think there will be a performance issue.

breml:
@SVilgelm created an initial release v0.1.0. I will have a look at your other suggestions as well.

breml:
> I tried to run the file, but on Linux with go1.17 or go1.16, I don't reproduce. Maybe it's only working on MacOS.

@ldez I am on Linux as well, using Go 1.16 and I can clearly see the bug.

I cloned the repo, entered the directory `Go`.

If you look at the [file](https://github.com/nickboucher/trojan-source/blob/main/Go/commenting-out.go), you see:

```Go
package main

import "fmt"

func main() {
	var isAdmin = false
	var isSuperAdmin = false
	isAdmin = isAdmin || isSuperAdmin
	/*‮ } ⁦if (isAdmin)⁩ ⁦ begin admins only */
		fmt.Println("You are an admin.")
	/* end admins only ‮ { ⁦*/
}
```

If you run the code with e.g. `go run commenting-out.go`, you get:

```shell
$ go run commenting-out.go 
You are an admin.
```

But this is not what you expect from the code, because both variables, `isAdmin` and `isSuperAdmin` are initialized to false, so the condition also resolves to false. Therefore the `if` should evaluate to `true` nor should the line `You are an admin.` be printed.

So what the compiler sees is (more like):

```Go
package main

import "fmt"

func main() {
	var isAdmin = false
	var isSuperAdmin = false
	isAdmin = isAdmin || isSuperAdmin
	/* } if (isAdmin) begin admin only */
		fmt.Println("You are an admin.")
	/* end admins only } */
}
```

ldez:
I tested the implementation and the performance seems ok.

SVilgelm:
checked the code again and found one more place for improvement:
```go
			if !bytes.ContainsRune(body[start:], r) {
				break
			}
			start += bytes.IndexRune(body[start:], r)
```
this code scans a file two times: `!bytes.ContainsRune` and `bytes.IndexRune`, I would recommend to improve it to scan only one time, it will increase performance.
```go
			idx := bytes.IndexRune(body[start:], r)
			if idx == -1 {
				break
			}
			start += idx
```

ldez:
@SVilgelm maybe you can open a PR or an issue on `bidichk` repo?
It can be easier to follow

knz:
See [this comment](https://github.com/cockroachdb/cockroach/pull/72287#discussion_r741183989): it may not be wise including the isolate characters in this list.

SVilgelm:
> @SVilgelm maybe you can open a PR or an issue on the `bidichk` repo?
> It can be easier to follow.

maybe, but It is hard to find a time for golangci-lint, I just need to add more details in the comments :)

ldez:
@SVilgelm I will help and I will open a PR :wink:

SVilgelm (approved):

ldez (approved):
LGTM

PR #2339, opened 2021-11-02
codeql: Enable javascript

SVilgelm:
Usage of matrix is not required, action can scan for all languages at same time


SVilgelm:
seems we don't need the JS analyzer, since most of JS code are some additional tools

PR #2344, opened 2021-11-04
exhaustive: upgrade to latest tag; add new flags and deprecate old ones

nishanths:
This makes the following changes to the `exhaustive` analyzer.

- Upgrade `github.com/nishanths/exhaustive` to latest tag. 
- Add support for new flags: `ignore-enum-members` and `checking-strategy`.
- Add tests for new flags.
- Deprecate flag `ignore-pattern`, which no longer has any effect (use `ignore-enum-members` instead).

As a side note, the `exhaustive` analzyer no longer provides any `[]analysis.SuggestedFix` in the diagnostics that it reports.

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -136,10 +136,16 @@ linters-settings:
   exhaustive:
     # check switch statements in generated files also
     check-generated: false
-    # indicates that switch statements are to be considered exhaustive if a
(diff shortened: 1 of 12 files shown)


nishanths:
@ldez: I've added you as a reviewer, but feel free to reassign!

SVilgelm (approved):
LGTM

PR #2345, opened 2021-11-04
Add github.com/fpuc/gostatement linter

fpuc:
[gostatement](https://github.com/fpuc/gostatement) is an analyzer checking for occurrence of `go` statements. You may want to use a custom func wrapping the statement utilizing recover, logging, metrics...


SVilgelm:
hmm, I'm not sure I will ever use it, but why not, if people need it for any reasons

ldez:
Hello, it seems to not be linter but just a way to display information, IMO it's out of the scope of golangci-lint.

fpuc:
> Hello, it seems to not be linter but just a way to display information, IMO it's out of the scope of golangci-lint.

Hello, I can completely understand your opinion. OTOH you can look at it from a different angle: The perspective of project maintainers is different and they have to decide what is for him/her/it a stylistic problematic construct - so it would be a linter then. 

(Of course this would fit quite nicely into the linter system and easy to add - thanks for all you great work done and the guidelines provided - and your quick review)

butuzov:
not a linter, but a nice thing to have. defiantly out of `golangci-lint` scope. 

closed.

PR #2349, opened 2021-11-06
feat: add errchkjson linter

breml:
Add [errchkjson](https://github.com/breml/errchkjson).

This linter checks the types, that are passed to json encoding function (e.g. `json.Marshal`) and has two main functions:

1. Reports types, that are not supported for json encoding (e.g. `complex128`, `func` or `chan`).
2. Reports when the `error` returned from the json encoding function can be ignored, because there is no way for json.Marshal to have an error.

For more details, see the [README.md](https://github.com/breml/errchkjson).

(This linter has been laying around on my harddisk for 2 years, after the successful inclusion of [bidichk](https://github.com/breml/bidichk) in golangci-lint (#2330) I felt motivated to polish this one and propose it for inclusion as well.)


test/testdata/errchkjson.go
```diff
@@ -0,0 +1,614 @@
...
+
+		C128         complex128           `json:"-"` // invalid exported but ommited
+		C128Slice    []complex128         `json:"-"` // invalid exported but ommited
+		C128Array    [10]complex128       `json:"-"` // invalid exported but ommited
+		MapBoolStr   map[bool]string      `json:"-"` // invalid exported but ommited
+		MapF64Str    map[float64]string   `json:"-"` // invalid exported but ommited
+		F            func()               `json:"-"` // invalid exported but ommited
+		Ch           chan struct{}        `json:"-"` // invalid exported but ommited
+		UnsafePtr    unsafe.Pointer       `json:"-"` // invalid exported but ommited
+		MapStructStr map[structKey]string `json:"-"` // invalid exported but ommited
```
ldez, line 382:
```suggestion
		F64                  float64                         `json:"-"` // unsafe exported but omitted
		F64Ptr               *float64                        `json:"-"` // unsafe exported but omitted
		F64Slice             []float64                       `json:"-"` // unsafe exported but omitted
		F64Array             [10]float64                     `json:"-"` // unsafe exported but omitted
		MapStrF64            map[string]float64              `json:"-"` // unsafe exported but omitted
		MapEIStr             map[interface{}]string          `json:"-"` // unsafe exported but omitted
		Number               json.Number                     `json:"-"` // unsafe exported but omitted
		NumberPtr            *json.Number                    `json:"-"` // unsafe exported but omitted
		NumberSlice          []json.Number                   `json:"-"` // unsafe exported but omitted
		MapNumberStr         map[json.Number]string          `json:"-"` // unsafe exported but omitted
		Ei                   interface{}                     `json:"-"` // unsafe exported but omitted
		Stringer             fmt.Stringer                    `json:"-"` // unsafe exported but omitted
		Mt                   marshalText                     `json:"-"` // unsafe exported but omitted
		MapMarshalTextString map[marshalText]string          `json:"-"` // unsafe exported but omitted
		ExportedStruct       ExportedUnsafeAndInvalidStruct  `json:"-"` // unsafe exported but omitted
		ExportedStructPtr    *ExportedUnsafeAndInvalidStruct `json:"-"` // unsafe exported but omitted

		C128         complex128           `json:"-"` // invalid exported but omitted
		C128Slice    []complex128         `json:"-"` // invalid exported but omitted
		C128Array    [10]complex128       `json:"-"` // invalid exported but omitted
		MapBoolStr   map[bool]string      `json:"-"` // invalid exported but omitted
		MapF64Str    map[float64]string   `json:"-"` // invalid exported but omitted
		F            func()               `json:"-"` // invalid exported but omitted
		Ch           chan struct{}        `json:"-"` // invalid exported but omitted
		UnsafePtr    unsafe.Pointer       `json:"-"` // invalid exported but omitted
		MapStructStr map[structKey]string `json:"-"` // invalid exported but omitted
```

test/testdata/errchkjson_omit_safe.go
```diff
@@ -0,0 +1,614 @@
...
+		MapBoolStr   map[bool]string
+		MapF64Str    map[float64]string
+		F            func()
+		Ch           chan struct{}
+		UnsafePtr    unsafe.Pointer
+		MapStructStr map[structKey]string
+	}
+)
+
+// JSONMarshalSaveStructWithUnexportedFields contains a struct with unexported, unsafe fields.
```
ldez, line 285:
```suggestion
// JSONMarshalSaveStructWithUnexportedFieldsWithNoSafe contains a struct with unexported, unsafe fields.
```

test/testdata/errchkjson_omit_safe.go
```diff
@@ -0,0 +1,614 @@
...
+	_ = unexportedInStruct.f
+	_ = unexportedInStruct.ch
+	_ = unexportedInStruct.unsafePtr
+	_ = unexportedInStruct.mapStructStr[structKey{1}]
+	_, _ = json.Marshal(unexportedInStruct)   // ERROR "Error return value of `encoding/json.Marshal` is not checked"
+	_, err = json.Marshal(unexportedInStruct) // struct containing unsafe but unexported fields is safe
+	_ = err
+}
+
+// JSONMarshalSaveStructWithOmittedFields contains a struct with omitted, unsafe fields.
```
ldez, line 350:
```suggestion
// JSONMarshalSaveStructWithOmittedFieldsWithNoSafe contains a struct with omitted, unsafe fields.
```

pkg/golinters/errchkjson.go
```diff
@@ -0,0 +1,31 @@
+package golinters
+
+import (
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/breml/errchkjson"
```
ldez, line 6:
```suggestion
	"github.com/breml/errchkjson"
	"golang.org/x/tools/go/analysis"
```

ldez (changes requested):

test/testdata/errchkjson.go
```diff
@@ -0,0 +1,614 @@
...
+	}
+	_ = ommitInStruct.MapStructStr[structKey{1}]
+	_, _ = json.Marshal(ommitInStruct)   // struct containing unsafe but omitted, exported fields is safe
+	_, err = json.Marshal(ommitInStruct) // ERROR "Error return value of `encoding/json.Marshal` is checked but passed argument is safe"
+	_ = err
+}
+
+// JSONMarshalUnsaveTypes contains a multitude of test cases to marshal different combinations of types to JSON,
+// that can potentially lead to json.Marshal returning an error.
+func JSONMarshalUnsaveTypes() {
```
ldez, line 392:
```suggestion
// JSONMarshalUnsafeTypes contains a multitude of test cases to marshal different combinations of types to JSON,
// that can potentially lead to json.Marshal returning an error.
func JSONMarshalUnsafeTypes() {
```

test/testdata/errchkjson.go
```diff
@@ -0,0 +1,614 @@
...
+	_, _ = json.Marshal(unexportedInStruct)   // struct containing unsafe but unexported fields is safe
+	_, err = json.Marshal(unexportedInStruct) // ERROR "Error return value of `encoding/json.Marshal` is checked but passed argument is safe"
+	_ = err
+}
+
+// JSONMarshalSaveStructWithOmittedFields contains a struct with omitted, unsafe fields.
+func JSONMarshalSaveStructWithOmittedFields() {
+	var err error
+
+	var ommitInStruct struct {
```
ldez, line 354:
```suggestion
	var omitInStruct struct {
```

test/testdata/errchkjson.go
```diff
@@ -0,0 +1,614 @@
...
+		MapBoolStr   map[bool]string      `json:"-"` // invalid exported but omitted
+		MapF64Str    map[float64]string   `json:"-"` // invalid exported but omitted
+		F            func()               `json:"-"` // invalid exported but omitted
+		Ch           chan struct{}        `json:"-"` // invalid exported but omitted
+		UnsafePtr    unsafe.Pointer       `json:"-"` // invalid exported but omitted
+		MapStructStr map[structKey]string `json:"-"` // invalid exported but omitted
+	}
+	_ = ommitInStruct.MapStructStr[structKey{1}]
+	_, _ = json.Marshal(ommitInStruct)   // struct containing unsafe but omitted, exported fields is safe
+	_, err = json.Marshal(ommitInStruct) // ERROR "Error return value of `encoding/json.Marshal` is checked but passed argument is safe"
```
ldez, line 386:
```suggestion
	_ = omitInStruct.MapStructStr[structKey{1}]
	_, _ = json.Marshal(omitInStruct)   // struct containing unsafe but omitted, exported fields is safe
	_, err = json.Marshal(omitInStruct) // ERROR "Error return value of `encoding/json.Marshal` is checked but passed argument is safe"
```

test/testdata/errchkjson.go
```diff
@@ -0,0 +1,614 @@
...
+		InnerStruct innerStruct
+	}
+	_, _ = json.Marshal(outerStruct)   // struct with only safe types
+	_, err = json.Marshal(outerStruct) // ERROR "Error return value of `encoding/json.Marshal` is checked but passed argument is safe"
+	_ = err
+}
+
+type (
+	structKey                      struct{ id int }
+	ExportedUnsafeAndInvalidStruct struct { // unsafe unexported but ommited
```
ldez, line 256:
```suggestion
	ExportedUnsafeAndInvalidStruct struct { // unsafe unexported but omitted
```

test/testdata/errchkjson.go
```diff
@@ -0,0 +1,614 @@
...
+
+func (mt marshalText) MarshalText() ([]byte, error) {
+	return []byte(`mt`), nil
+}
+
+var _ encoding.TextMarshaler = marshalText(struct{}{})
+
+// JSONMarshalSaveTypes contains a multitude of test cases to marshal different combinations of types to JSON,
+// that are save, that is, they will never return an error, if these types are marshaled to JSON.
+func JSONMarshalSaveTypes() {
```
ldez, line 23:
```suggestion
// JSONMarshalSafeTypes contains a multitude of test cases to marshal different combinations of types to JSON,
// that are safe, that is, they will never return an error, if these types are marshaled to JSON.
func JSONMarshalSafeTypes() {
```

test/testdata/errchkjson_omit_safe.go
```diff
@@ -0,0 +1,614 @@
...
+
+func (mt marshalText) MarshalText() ([]byte, error) {
+	return []byte(`mt`), nil
+}
+
+var _ encoding.TextMarshaler = marshalText(struct{}{})
+
+// JSONMarshalSaveTypesWithNoSafe contains a multitude of test cases to marshal different combinations of types to JSON,
+// that are save, that is, they will never return an error, if these types are marshaled to JSON.
+func JSONMarshalSaveTypesWithNoSafe() {
```
ldez, line 23:
```suggestion
// JSONMarshalSafeTypesWithNoSafe contains a multitude of test cases to marshal different combinations of types to JSON,
// that are safe, that is, they will never return an error, if these types are marshaled to JSON.
func JSONMarshalSafeTypesWithNoSafe() {
```

test/testdata/errchkjson_omit_safe.go
```diff
@@ -0,0 +1,614 @@
...
+		InnerStruct innerStruct
+	}
+	_, _ = json.Marshal(outerStruct)   // ERROR "Error return value of `encoding/json.Marshal` is not checked"
+	_, err = json.Marshal(outerStruct) // struct with only safe types is safe, but omit-safe is set
+	_ = err
+}
+
+type (
+	structKey                      struct{ id int }
+	ExportedUnsafeAndInvalidStruct struct { // unsafe unexported but ommited
```
ldez, line 256:
```suggestion
	ExportedUnsafeAndInvalidStruct struct { // unsafe unexported but omitted
```

test/testdata/errchkjson_omit_safe.go
```diff
@@ -0,0 +1,614 @@
...
+
+		C128         complex128           `json:"-"` // invalid exported but ommited
+		C128Slice    []complex128         `json:"-"` // invalid exported but ommited
+		C128Array    [10]complex128       `json:"-"` // invalid exported but ommited
+		MapBoolStr   map[bool]string      `json:"-"` // invalid exported but ommited
+		MapF64Str    map[float64]string   `json:"-"` // invalid exported but ommited
+		F            func()               `json:"-"` // invalid exported but ommited
+		Ch           chan struct{}        `json:"-"` // invalid exported but ommited
+		UnsafePtr    unsafe.Pointer       `json:"-"` // invalid exported but ommited
+		MapStructStr map[structKey]string `json:"-"` // invalid exported but ommited
```
ldez, line 382:
```suggestion
		F64                  float64                         `json:"-"` // unsafe exported but omitted
		F64Ptr               *float64                        `json:"-"` // unsafe exported but omitted
		F64Slice             []float64                       `json:"-"` // unsafe exported but omitted
		F64Array             [10]float64                     `json:"-"` // unsafe exported but omitted
		MapStrF64            map[string]float64              `json:"-"` // unsafe exported but omitted
		MapEIStr             map[interface{}]string          `json:"-"` // unsafe exported but omitted
		Number               json.Number                     `json:"-"` // unsafe exported but omitted
		NumberPtr            *json.Number                    `json:"-"` // unsafe exported but omitted
		NumberSlice          []json.Number                   `json:"-"` // unsafe exported but omitted
		MapNumberStr         map[json.Number]string          `json:"-"` // unsafe exported but omitted
		Ei                   interface{}                     `json:"-"` // unsafe exported but omitted
		Stringer             fmt.Stringer                    `json:"-"` // unsafe exported but omitted
		Mt                   marshalText                     `json:"-"` // unsafe exported but omitted
		MapMarshalTextString map[marshalText]string          `json:"-"` // unsafe exported but omitted
		ExportedStruct       ExportedUnsafeAndInvalidStruct  `json:"-"` // unsafe exported but omitted
		ExportedStructPtr    *ExportedUnsafeAndInvalidStruct `json:"-"` // unsafe exported but omitted

		C128         complex128           `json:"-"` // invalid exported but omitted
		C128Slice    []complex128         `json:"-"` // invalid exported but omitted
		C128Array    [10]complex128       `json:"-"` // invalid exported but omitted
		MapBoolStr   map[bool]string      `json:"-"` // invalid exported but omitted
		MapF64Str    map[float64]string   `json:"-"` // invalid exported but omitted
		F            func()               `json:"-"` // invalid exported but omitted
		Ch           chan struct{}        `json:"-"` // invalid exported but omitted
		UnsafePtr    unsafe.Pointer       `json:"-"` // invalid exported but omitted
		MapStructStr map[structKey]string `json:"-"` // invalid exported but omitted
```

test/testdata/errchkjson_omit_safe.go
```diff
@@ -0,0 +1,614 @@
...
+	}
+	_ = ommitInStruct.MapStructStr[structKey{1}]
+	_, _ = json.Marshal(ommitInStruct)   // ERROR "Error return value of `encoding/json.Marshal` is not checked"
+	_, err = json.Marshal(ommitInStruct) // struct containing unsafe but omitted, exported fields is safe, but omit-safe is set
+	_ = err
+}
+
+// JSONMarshalUnsaveTypes contains a multitude of test cases to marshal different combinations of types to JSON,
+// that can potentially lead to json.Marshal returning an error.
+func JSONMarshalUnsaveTypes() {
```
ldez, line 392:
```suggestion
// JSONMarshalUnsafeTypes contains a multitude of test cases to marshal different combinations of types to JSON,
// that can potentially lead to json.Marshal returning an error.
func JSONMarshalUnsafeTypes() {
```

ldez (changes requested):

breml:
@ldez Thanks for finding all the typos. I will rebase all the changes as soon as I am back at my keyboard.

breml:
@ldez rebased all the typo changes and updated to errchkjson v0.1.1, which adds a check for structs without any exported fields (as requested on [reddit](https://www.reddit.com/r/golang/comments/qo0x7g/go_linter_that_checks_types_that_are_json_encoded/hjo0a4n/))

breml:
@ldez What do you think about this [comment on reddit](https://www.reddit.com/r/golang/comments/qo0x7g/go_linter_that_checks_types_that_are_json_encoded/hjk8q4m/)?

> Interesting, but the basic errcheck linter would probably conflict with the decisions of this one, no?. It gets mad whenever there is a possible error value that you don't check, even if it's "unreachable".

My tests suggest, that there is no problem, if the err is omitted with `_`. But there might be a problem, if the the error is omitted entirely, for example with:

```Go
enc := json.NewEncoder(os.Stdout)
enc.Encode(xyz) // error is omitted
```

Is there a way to throw away a report from an other linter (in this case errcheck), because errchkjson is more specific?

ldez:
> updated to errchkjson v0.1.1, which adds a check for structs without any exported fields (as requested on reddit)

This check is already done by another linter.

ldez:
> Is there a way to throw away a report from an other linter (in this case errcheck), because errchkjson is more specific?

A linter cannot interact with another linter.

ldez:
IMO, for the golangci-lint context, your linter must not handle the error and let the other linters, that are specific to this, handle the error.

breml:
> IMO, for the golangci-lint context, your linter must not handle the error and let the other linters, that are specific to this, handle the error.

I do not agree with this, because `errcheck` is generic in the sense that it reports every occasion, where an arbitrary function returns an error and this error is omitted. `errchkjson` is more specific in the sense, that it is limited to the json encoding functions but in contrast to `errcheck` it does understand the different cases, when the json encoding functions in fact may return an error and in which cases it is safe to ignore the error.

Therefore I tried to find a way, how these two linters can co-exist and both perform their job to the best extend possible. The latest commit is the current solution I found so far.

breml:
Is there anything missing on this PR? What is the best way to move forward with this?

pkg/golinters/errchkjson.go
```diff
@@ -0,0 +1,80 @@
...
+		if errcheckCfg == nil {
+			errcheckCfg = &config.ErrcheckSettings{}
+		}
+		errcheckCfg.ExcludeFunctions = append(errcheckCfg.ExcludeFunctions,
+			"encoding/json.Marshal",
+			"encoding/json.MarshalIndent",
+			"(*encoding/json.Encoder).Encode",
+		)
+	}
+
```
ldez, line 36: As I already said, a linter cannot interact with another linter.

The fact to create links between 2 linters means that 1 linter can be a part of the other linter.

ldez (changes requested):

pkg/golinters/errchkjson.go
breml: @ldez Thanks for having a look at my PR again, I really appreciate it.

> As I already said, a linter cannot interact with another linter.

I do not exactly understand what you mean by "cannot". My understanding of "cannot" is, that it is technically not possible. But I don't think this is true, because the implementation of the PR does work as intended.
Or do you mean "cannot" in the sense of "MUST NOT" (or "SHALL NOT") as defined in https://datatracker.ietf.org/doc/html/rfc2119? That is: it is prohibited by guideline (even though, it could technically be possible).
If it is the later, I would like to know, if this guideline is written down somewhere and I would like to better understand the reasoning behind it. I understand, that in some cases the exclude patterns have been used to handle exactly such situations, e.g. for duplicates between gosec and errcheck (see `EXC0008`).

> The fact to create links between 2 linters means that 1 linter can be a part of the other linter.
So what is your recommendation? I can see the following options (in no particular order):

1. Add the functionality of errchkjson to errcheck. I do not know, if this special handling is in the scope of errcheck, because it defines its scope pretty general with "... checking for unchecked errors ..." and as far as I can tell, there is currently no special handling based on the input arguments in errcheck. Additionally, errchkjson also performs checks on the input types for the json encoding functions (can a type be encoded) and this is cleary out of scope for errcheck and would have to stay in this linter.
2. Extend errchkjson to also include everything that is done by errcheck so the user is free to choose one linter or the other depending of the fact, if he wants to have the special handling of the json encoding functions. In my opinion, this does not make any sense.
3. We add the necessary exceptions to the ExcludePatterns in `pkg/config/issues.go`, but I am not sure how exactly this would work given the fact, that the exclude pattern should only be added under special circumstances (both linters enabled and omitSafe not true).
4. Add the errchkjson linter (without adding config settings to errcheck) and mention in the documentation / .golangci-lint.yaml, how the user can add the necessary excludes. This would work, but it is not really user friendly.
5. Keep the current implementation.

PR #2358, opened 2021-11-14
Ensure that the Issues key in JSON format is a list

rafikdraoui:
This makes the data more consistent and easier to use with other tools. For example, a third-party tool can iterate over `Issues` without having to make a preliminary null check.

Before:
```console
$ golangci-lint run --out-format=json | jq
{
  "Issues": null,
  "Report": {
    "Linters": [...]
  }
}
```

After:
```console
$ golangci-lint run --out-format=json | jq
{
  "Issues": [],
  "Report": {
    "Linters": [...]
  }
}
```

### DIFF
--- a/pkg/printers/json.go
+++ b/pkg/printers/json.go
@@ -30,6 +30,9 @@ func (p JSON) Print(ctx context.Context, issues []result.Issue) error {
 		Issues: issues,
 		Report: p.rd,
 	}
+	if res.Issues == nil {
+		res.Issues = []result.Issue{}
+	}
 
 	outputJSON, err := json.Marshal(res)
 	if err != nil {


SVilgelm:
hmm, cannot reproduce the `nolintnolint` issue on local

pkg/commands/run.go
```diff
@@ -479,9 +479,9 @@ func (e *Executor) executeRun(_ *cobra.Command, args []string) {
 
 // to be removed when deadline is finally decommissioned
 func (e *Executor) setTimeoutToDeadlineIfOnlyDeadlineIsSet() {
-	// nolint:staticcheck
-	deadlineValue := e.cfg.Run.Deadline
+	deadlineValue := e.cfg.Run.Deadline // nolint
```
SVilgelm, line 482: this is weird, I'm not sure why `nolintnolint` fails on this.
@ldez do you have any ideas?

ldez: It's a flaky behavior, increased/produced by the cache used by the golangci-lint GH action.
It's a real pain, I often have to relaunch the CI several times before I can merge a PR.

ldez:
another way to set struct tags:
```go
type JSONResult struct {
	Issues []result.Issue `json:",omitempty"`
	Report *report.Data   `json:",omitempty"`
}
```

```console
$ golangci-lint run --out-format=json | jq
{
  "Report": {
    "Linters": [...]
  }
}
```

pkg/commands/run.go
ldez: pb partially fixed in #2359

SVilgelm (approved):

ldez:
@SVilgelm have you seen my suggestion?

SVilgelm:
yeah, I saw that, but I think it is better to explicitly print an empty array instead of omitting empty values. In this case the tools that should parse an output can avoid the checks for null or if the value does not exist

ldez:
I think it's better to use the same rule for all the fields.
`*report.Data` ->
https://github.com/golangci/golangci-lint/blob/aa3550ab837926f70329a449adb4d826517016b1/pkg/report/data.go#L3-L18

rafikdraoui:
I agree with @Svilgelm that ensuring that there is always in the output a key named `Issues` whose value is an array would make it easier for parsing. I suspect that the list of issues is what external tools mostly care about, rather than the internal configuration of golangci-lint or its logging output in `Report`.

That said, if the decision is to not go forward with this, it does feel slightly better to have the key omitted rather than having it set to nil. But it doesn't change the fact that external tools will have to be aware of the quirk and work around it (which is fine, if the tool integrates with golangci-lint, then it already is committed to be "hard wired" to it :smiley:)

SVilgelm:
We can improve the `Report`'s part In a separate PR, it requires some additional discussion

SVilgelm:
I'm merging this one

ldez:
@SVilgelm when we don't agree, maybe we can discuss a bit, and if we don't find an agreement maybe we invite another maintainer in the topic instead of just merging.

PR #2359, opened 2021-11-14
fix: reduce flaky in the CI and clean the .golangci.yml of the project

ldez:
- transforms one`//nolint` comment to an `exclude-rules` (less flaky)
- clean `.golangci.yml` (remove obsolete stuff, and change `gomnd` config)
- minor changes (slice init and godoc comment)

Note: I think that `nolintlint` have a concurrency/cache problem, it needs to be fixed in another PR.


SVilgelm (approved):
LGTM

PR #2360, opened 2021-11-15
chore: fix wrong default value in comment

mookjp:
* The default value for `allow-unused` is false
  * https://github.com/golangci/golangci-lint/blob/ecbb9c475ffb689c0d7d35ed84e385c5d9905f5f/pkg/config/linters_settings.go#L49


ashanbrown (approved):
👍

PR #2363, opened 2021-11-17
Automatically detect local prefixes

### LINKED ISSUE #931: Add `prefix-current-module` to goimports
I'd like to add a bool option `prefix-current-module` to the goimports config.

It would basically check if the linted package is a module, and slap the module path (based on `go.mod` if exists) to the list of local prefixes. It would make the configuration more portable.

WDYT?

sagikazarmark:
This PR is an attempt to automatically detect local prefixes from module paths.
The rationale behind this feature is making config files less verbose/redundant and a little bit more portable.

It is not perfect, but it's working.

More specifically, I'm not sure if loading module information for every linter is a good idea, so module information loading should be incorporated into the LoadMode, but I'm not sure how since it's a string.

I'm also not sure if the gci -> goimports fallback is necessary.

Last, but not least the config option name feels a bit weird, but I couldn't come up with a better one.

Any feedback would be highly appreciated.

Fixes #931

### DIFF
--- a/pkg/golinters/gci.go
+++ b/pkg/golinters/gci.go
@@ -3,6 +3,7 @@ package golinters
 import (
 	"bytes"
 	"fmt"
+	"strings"
 	"sync"
 
 	"github.com/daixiang0/gci/pkg/gci"
@@ -32,9 +33,30 @@ func NewGci() *goanalysis.Linter {
 		nil,
 	).WithContextSetter(func(lintCtx *linter.Context) {
 		localFlag := lintCtx.Settings().Gci.LocalPrefixes
+
+		if localFlag == "" && lintCtx.Settings().Gci.LocalPrefixModule {
+			var modules []string
+
+			for _, module := range lintCtx.Modules {
+				modules = append(modules, module.Path)
+			}
+
+			localFlag = strings.Join(modules, ",")
+		}
+
 		goimportsFlag := lintCtx.Settings().Goimports.LocalPrefixes
 		if localFlag == "" && goimportsFlag != "" {
 			localFlag = goimportsFlag
+		}
+
+		if localFlag == "" && lintCtx.Settings().Goimports.LocalPrefixModule {
+			var modules []string
+
+			for _, module := range lintCtx.Modules {
+				modules = append(modules, module.Path)
+			}
+
+			localFlag = strings.Join(modules, ",")
 		}
 
 		analyzer.Run = func(pass *analysis.Pass) (interface{}, error) {
--- a/pkg/config/linters_settings.go
+++ b/pkg/config/linters_settings.go
@@ -226,7 +226,8 @@ type FunlenSettings struct {
 }
 
 type GciSettings struct {
-	LocalPrefixes string `mapstructure:"local-prefixes"`
+	LocalPrefixes     string `mapstructure:"local-prefixes"`
+	LocalPrefixModule bool   `mapstructure:"local-prefix-module"`
 }
 
 type GocognitSettings struct {
--- a/pkg/config/linters_settings.go
+++ b/pkg/config/linters_settings.go
@@ -277,7 +277,8 @@ type GoHeaderSettings struct {
 }
 
 type GoImportsSettings struct {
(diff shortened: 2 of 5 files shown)


pkg/golinters/gci.go
```diff
@@ -32,11 +33,32 @@ func NewGci() *goanalysis.Linter {
...
+
+		if localFlag == "" && lintCtx.Settings().Gci.LocalPrefixModule {
+			var modules []string
+
+			for _, module := range lintCtx.Modules {
+				modules = append(modules, module.Path)
+			}
+
+			localFlag = strings.Join(modules, ",")
+		}
```
SVilgelm, line 45: this block looks redundant, the fallback to goimports and next same block will never be executed

SVilgelm (changes requested):
it is better to grab the local prefix from go.mod file if it exists during the initialization of linter itself
and please add the tests

sagikazarmark:
@SVilgelm thanks for the feedback!

Is there already an example somewhere that reads the go.mod file or uses the module name? Speaking from experience, it can get tricky.

SVilgelm:
Here is an example of parsing go.mod file
```go
import (
	"golang.org/x/mod/modfile"
)

func getModulePath() string {
	mfile, _ := modfile.ParseLax("go.mod", []byte{}, nil)
	return mfile.Module.Mod.Path
}
```

ernado:
Great idea in general, just please make sure to make it opt-in so we don't break backward compatibility (looks like it is opt in via `local-prefix-module: true` which is good)

pkg/golinters/gci.go
```diff
@@ -32,11 +33,32 @@ func NewGci() *goanalysis.Linter {
...
+		if localFlag == "" && lintCtx.Settings().Goimports.LocalPrefixModule {
+			var modules []string
+
+			for _, module := range lintCtx.Modules {
+				modules = append(modules, module.Path)
+			}
+
+			localFlag = strings.Join(modules, ",")
+		}
+
```
bombsimon, line 61: These looks exactly the same with some duplicated code, maybe would be nice to create a function?

```go
localFlagFn := func(modules) string {
	var modules []string

	for _, module := range lintCtx.Modules {
		modules = append(modules, module.Path)
	}

	return strings.Join(modules, ",")
}
```

Then just call it twice:

```go
if localFlag == "" && lintCtx.Settings().Gci.LocalPrefixModule {
	localFlag = localFlagFn(lintCtx.Settings().Gci.LocalPrefixModule)
}

goimportsFlag := ...


if localFlag == "" && lintCtx.Settings().Goimports.LocalPrefixModule {
	localFlag = localFlagFn(lintCtx.Settings().Goimports.LocalPrefixModule)
}
```

sagikazarmark: Absolutely!

sagikazarmark:
@SVilgelm I've spent some time trying to figure out how to get the most accurate module information and it's a little bit complicated than just calling that modfile package.

First, we need to find the go.mod file. The easiest (and probably the most accurate) way to do it is running `go env GOMOD` (that requires the Go tool to be in the path though). Then we can parse the go.mod file.

It might also make sense to just defer this task to a library that already does this well: https://github.com/gobuffalo/here

Last, but not least: I don't know if it's possible to analyse multiple packages from different modules at the same time, but if it is, then the current solution will provide the most accurate result. (It doesn't necessarily have to live in the context though, can be moved to the linter implementations as well.

What do you think?

ldez:
you can take a look at what I did here: https://github.com/ldez/gomoddirectives/blob/82673faed925c794e4a179ae238dd0b10f260732/module.go#L22-L24

Personally, I don't think that autodetecting is a good idea because it will not work well in some cases (like golangci-lint by file inside an IDE)

sagikazarmark:
Can we get to an agreement for whether this change would be merged or not before I invest any more time in it?

@ldez that's a fair point and I think the current solution would be better from that perspective as well (if no module info is available, no prefix will be applied, whereas manually looking for the go.mod file might be problematic).

ldez:
Fixed by #4522

PR #2367, opened 2021-11-18
Handle exit codes in the version output

alexander-melentyev:


pkg/commands/version.go
```diff
@@ -36,7 +38,8 @@ func (e *Executor) initVersion() {
 		RunE: func(cmd *cobra.Command, _ []string) error {
 			switch strings.ToLower(e.cfg.Version.Format) {
 			case "short":
-				fmt.Println(e.version)
+				fmt.Fprintln(os.Stdout, e.version)
+				os.Exit(exitcodes.Success)
```
ldez, line 42: This line is a bit useless, because exit code `0` is the default

pkg/commands/version.go
```diff
@@ -47,9 +50,11 @@ func (e *Executor) initVersion() {
 				if err != nil {
 					return err
 				}
-				fmt.Println(string(data))
+				fmt.Fprintln(os.Stdout, string(data))
+				os.Exit(exitcodes.Success)
```
ldez, line 54: This line is a bit useless, because exit code `0` is the default

pkg/commands/version.go
```diff
@@ -47,9 +50,11 @@ func (e *Executor) initVersion() {
...
 			default:
-				fmt.Printf("golangci-lint has version %s built from %s on %s\n", e.version, e.commit, e.date)
+				fmt.Fprintf(os.Stdout, "golangci-lint has version %s built from %s on %s\n", e.version, e.commit, e.date)
+				os.Exit(exitcodes.Success)
```
ldez, line 57: This line is a bit useless, because exit code `0` is the default

ldez (changes requested):

bombsimon (changes requested):
Even though 0 is the default exit code if we want to be explicit you don't have to add it for _every_ case, add it in the end of the function.

But then again, this isn't really how cobra works, cobras `RunE` will get a nil error return and exit with a successful exit code.

ldez:
@bombsimon your argument is better than mine.
And WDYT about `fmt.Println("")` vs `fmt.Fprintf(os.Stdout,"")`?

bombsimon:
> @bombsimon your argument is better than mine. And WDYT about `fmt.Println("")` vs `fmt.Fprintf(os.Stdout,"")`?

I think it makes more sense to use `fmt.Println` since the output to use with `fmt.Fprintf` doesn't seem to be configurable for this part so it doesn't make too much sense to me.

PR #2379, opened 2021-11-24
Update wrapcheck configuration to include ignoreSignRegexps

timkral:
This provides an update to the `wrapcheck` linter configuration to account for a change introduced here - https://github.com/tomarrell/wrapcheck/pull/21

cc: @tomarrell


ldez (approved):
LGTM

PR #2380, opened 2021-11-25
revive: fix enableAllRules

mdelah:
Was broken by commit [e5450981](https://github.com/mgechev/revive/commit/e5450981) not being reflected in golangci-lint's copy of normalizeConfig().


nashikb:
Perhaps I'm missing something, but shouldn't golangci-lint support the enableAllRules option of revive so that golangci-lint doesn't have to hard-code the list of revive rules? This seems to be setting up for subtle malbehavior down the road.

ldez:
The problem is the rules and the related functions are not exported.

ldez (approved):
LGTM

PR #2382, opened 2021-11-28
add containedctx linter

sivchari:
containedctx is a linter that detects struct contained context.Context field
https://github.com/sivchari/containedctx

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -73,7 +73,7 @@ require (
 	github.com/shazow/go-diff v0.0.0-20160112020656-b6b7b6733b8c
 	github.com/shirou/gopsutil/v3 v3.21.12
 	github.com/sirupsen/logrus v1.8.1
-	github.com/sivchari/containedctx v1.0.0
+	github.com/sivchari/containedctx v1.0.1
 	github.com/sivchari/tenv v1.4.7
 	github.com/sonatard/noctx v0.0.1
 	github.com/sourcegraph/go-diff v0.6.1
--- a/go.mod
+++ b/go.mod
@@ -73,6 +73,7 @@ require (
 	github.com/shazow/go-diff v0.0.0-20160112020656-b6b7b6733b8c
 	github.com/shirou/gopsutil/v3 v3.21.12
 	github.com/sirupsen/logrus v1.8.1
+	github.com/sivchari/containedctx v1.0.0
 	github.com/sivchari/tenv v1.4.7
 	github.com/sonatard/noctx v0.0.1
 	github.com/sourcegraph/go-diff v0.6.1
--- a/go.sum
+++ b/go.sum
@@ -701,8 +701,8 @@ github.com/sirupsen/logrus v1.6.0/go.mod h1:7uNnSEd1DgxDLC74fIahvMZmmYsHGZGEOFrf
 github.com/sirupsen/logrus v1.7.0/go.mod h1:yWOB1SBYBC5VeMP7gHvWumXLIWorT60ONWic61uBYv0=
 github.com/sirupsen/logrus v1.8.1 h1:dJKuHgqk1NNQlqoA6BTlM1Wf9DOH3NBjQyu0h9+AZZE=
 github.com/sirupsen/logrus v1.8.1/go.mod h1:yWOB1SBYBC5VeMP7gHvWumXLIWorT60ONWic61uBYv0=
-github.com/sivchari/containedctx v1.0.0 h1:xQM7v922emhkPuslBeTH/sMBFyeacXX0MvYoXoTOu/w=
-github.com/sivchari/containedctx v1.0.0/go.mod h1:PwZOeqm4/DLoJOqMSIJs3aKqXRX4YO+uXww087KZ7Bw=
+github.com/sivchari/containedctx v1.0.1 h1:fJq44cX+tD+uT5xGrsg25GwiaY61NGybQk9WWKij3Uo=
+github.com/sivchari/containedctx v1.0.1/go.mod h1:PwZOeqm4/DLoJOqMSIJs3aKqXRX4YO+uXww087KZ7Bw=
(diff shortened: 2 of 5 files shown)


Antonboom:
@sivchari, FYI – it's OK pass `Context` in the struct for tracing purpose.

For example:
```go
result := client.ExternalServiceCall()
span, ctx := tracer.ExtractSpan(result.Ctx)
// Deal with external service span.
```

sivchari:
[context](https://pkg.go.dev/context)

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx

Your example code depends on package design of 3rd library.
And, this linter don't detect external code.
Additional, if you wanna to contain context.Context in struct, you can disable this linter.

Antonboom:
@sivchari I'm familiar with context doc :)
Yes, you can blame the third-party library, but it's hard to argue with it when it is the de facto standard.

I have nothing against linter. It was just notification for u, that not everything is so simple.

sivchari:
I'm not sure which library you mean by de facto standard, but Many third-library declared interface to pass context to first parameter.
The third-library for using data tracing (e.g. dogstatsd, prometheus API ) shod be pass the first parameter too.
However, it's also true, as you say, that some third libraries contain context.Context in struct.
But Go team recommends not to contain context.Context in struct, so this linter is meant to keep it that way.
Thanks for comment.

bombsimon:
> But Go team recommends not to contain context.Context in struct, so this linter is meant to keep it that way.

Do you have a reference for that? I think it's pretty common when you implement servers or clients to have some base context, contexts with timeouts or contexts with cancellation on a struct that implements either of these.

I generally don't think `golangci-lint` should be opinionated and decide what's useful or not, it's just new to me that this would be a bad pattern.

ldez:
@bombsimon https://pkg.go.dev/context

> think it's pretty common when you implement servers or clients to have some base context, contexts with timeouts or contexts with cancellation on a struct that implements either of these.

this is a bad practice :scream_cat:

bombsimon:
> @bombsimon https://pkg.go.dev/context
> 
> > think it's pretty common when you implement servers or clients to have some base context, contexts with timeouts or contexts with cancellation on a struct that implements either of these.
> 
> this is a bad practice 🙀

Thanks! I just think I've never interpreted the text and example from the docs as spanning over _any_ struct. The example in the docs that says

> Do not store Contexts inside a struct type; instead, pass a Context explicitly to each function that needs it. The Context should be the first parameter, typically named ctx:

I think is referring to the receiving value in `DoSomething` which I totally get is bad practice. What I was more thinking about would be an implementation of some kind of protocol that passes a `ctx` to each handler, just like [the context that is stored on each `http.Request`](https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/http/request.go;l=319-323;drc=refs%2Ftags%2Fgo1.17.3) in stdlib.

But I guess there's no way to tell how the struct is used and since a lot of code bases doesn't implement this kind of behavior I can see the use case for the linter.

ldez:
`http.Request` is the exception that confirms the rule :wink:

sivchari:
> Do you have a reference for that? 

Sorry, this is an interpretation with my own opinions. However, as the Go team commented as `The Context should be the first parameter, typically named ctx` comments, this pattern is bad.

FYI
https://go.dev/blog/context-and-structs

> Exception to the rule: preserving backwards compatibility

https://github.com/golang/go/blob/ecf6b52b7f4ba6e8c98f25adf9e83773fe908829/src/net/http/request.go#L319-L323

```go
	// ctx is either the client or server context. It should only
	// be modified via copying the whole Request using WithContext.
	// It is unexported to prevent people from using Context wrong
	// and mutating the contexts held by callers of the same request.
	ctx context.Context
```

The reason it is included in net/http is for backward compatibility reasons.

bombsimon:
> `http.Request` is the exception that confirms the rule 😉
> 
> https://go.dev/blog/context-and-structs
>
> >  When designing an API with context, remember the advice: pass context.Context in as an argument; don’t store it in structs.

Heh, yeah usually that's the case with stdlib. 😄 But the gRPC example is a better example regarding what I meant with server implementation. It passes `context.Context` as an argument to each handler implementing a server but for internals it uses the struct to store it.

Another example where I don't see a problem storing it on structs would be in table driven testing.

```go
cases := []struct{
    description string
    path string
    ctx context.Context
}{
    {
      description: "Test foo",
      path:        "/foo/bar",
      ctx:          context.WithValue(context.Background(), someVal, "some-value")),
    },
}
```

Not saying this linter isn't good and obviously you provided links with more in depth information about this, thanks! I didn't know this was so generic.

So to clarify, I'm not talking about how contexts are passed to different methods or packages, I'm talking about contexts used to keep track of state in internal implementations.

sivchari:
Back to the linter. I'm improving that.
Because, in the current function, the linter detects test case (e.g. table-driven-test).
table-driven-test is general and context is sometime contained in struct too.
So I'm improving to exclude test file.

bombsimon:
> Back to the linter. I'm improving that. Because, in the current function, the linter detects test case (e.g. table-driven-test). table-driven-test is general and context is sometime contained in struct too. So I'm improving to exclude test file.

You don't have to do that in your linter for `golangci-lint` though, generally skipping tests in `golangci-lint` is handled with the implementation of that feature in `golangci-lint` itself. I'm also not sure skipping whole test files is the right way to go, I can see use cases of wanting to use this linter on test files as well.

sivchari:
> You don't have to do that in your linter for golangci-lint though, generally skipping tests in golangci-lint is handled with the implementation of that feature in golangci-lint itself. I'm also not sure skipping whole test files is the right way to go, I can see use cases of wanting to use this linter on test files as well.

Oh, I've mistook. 
Thanks !

ldez (approved):
LGTM

PR #2386, opened 2021-11-30
Add support for multiple outputs

### LINKED ISSUE #481: Support to output result in multiple formats
It would be great if it would be possible to produce results in multiple formats of a run.
This can be useful if you e.g. want to run golangci-lint in CI and send a human-readable output (line-number) in a plain-text report to the user and also have a checkstyle result to process it and show it in the CI tool.

Syntax could be like:
```
golangcilint run --out-format <FORMAT>:(<FILENAME>|stdout|stderr)[,<FORMAT>:(<FILENAME>|stdout|stderr)]..
```

lafriks:
Outputs can be specified using `--out-format` by separating them by comma.

Support providing file path by separating format name and path by colon symbol `:`.

Fixes  #481

### DIFF
--- a/test/linters_test.go
+++ b/test/linters_test.go
@@ -2,6 +2,8 @@ package test
 
 import (
 	"bufio"
+	"fmt"
+	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -100,7 +102,7 @@ func TestGciLocal(t *testing.T) {
 func TestMultipleOutputs(t *testing.T) {
 	sourcePath := filepath.Join(testdataDir, "gci", "gci.go")
 	args := []string{
-		"--disable-all", "--print-issued-lines=false", "--print-linter-name=false", "--out-format=line-number,json",
+		"--disable-all", "--print-issued-lines=false", "--print-linter-name=false", "--out-format=line-number,json:stdout",
 		sourcePath,
 	}
 	rc := extractRunContextFromComments(t, sourcePath)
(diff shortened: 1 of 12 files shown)


SVilgelm (changes requested):
Thank you very much, it really logs good, but definitely requires some additional tests to cover the files and stderr|stdout as well

lafriks:
stdout should be already covered as default but can be easily added to have for one format to be specified.

I can easily add file output(s) test case.

For stderr I would need some guidance as I'm not very familiar with testing helpers used in this project.

lafriks:
@SVilgelm I added test cases for stderr and file outputs. As testing helper command execution are combined with stdout It's not cleanly to test this. If stdout and stderr are read into separate variables and that would break quite some other tests as they expect combined output.

test/linters_test.go
```diff
@@ -97,6 +99,69 @@ func TestGciLocal(t *testing.T) {
...
+		ExpectOutputContains(`"Issues":[`)
+}
+
+func TestFileOutput(t *testing.T) {
+	f, err := os.CreateTemp("", "golangci_lint_test_result")
+	require.NoError(t, err)
+	f.Close()
+
+	resultPath := f.Name()
+	defer os.Remove(resultPath)
```
SVilgelm, line 142: Please `use t.TempDir`, it does not require removing and closing the file
```suggestion
	resultPath = t.TempDir() + "golangci_lint_test_result"
```

SVilgelm (changes requested):

test/linters_test.go
lafriks: changed, fixed also lint error

SVilgelm (approved):
LGTM

BradErz:
Is there any ETA on this one being merged?

lafriks:
@ldez I have added you write permissions to forked repository. I use such approach to somehow separate my own developed projects and OSS projects I contribute to only time to time 😄

ldez:
I know very well this pattern, but still, I recommend not using it.
If you want to separate fork and sources, it's better to create a user instead of an organization.

Note: I don't understand why do you want to separate forks and sources.

lafriks:
**Off-topic**: I just started to do that some time ago because I kept having too many repositories that can't really be deleted because of open PR and it can take quite some time to get merged, sometimes months

ldez (approved):
LGTM

fho:
@lafriks  thanks a lot for implementing it :-)

PR #2387, opened 2021-11-30
Normalize exclude-rules paths for Windows

### LINKED ISSUE #1398: Normalize path in exclude rules (OS path separator)
When specifying path's w/in exclude rules, the path regex requires special attention for cross-platform path separators.  Currently, you have to specify paths using syntax such as `folder[/\\]file.go` in order to gain cross-platform freedom.  Some settings within the configuration appear to support automatic replacement of `/` with the operating system's default (e.g. on windows it becomes `\`).

Identify all configuration's that involve a path (such as `exclude-rules`), and extend the functionality that currently replaces the path separator when executed on a windows platform while utilizing `skip-files` and `skip-dir`.  Currently, I'm 

No other alternatives considered.  Honestly unsure of the implications of this request.  Just throwing it out there for discussion :)

arjenvanderende:
Minor change to normalize the paths in `exclude-rules`. This re-uses the same logic that is already in place for `skip-dirs`.

Fixes  #1398

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -2080,6 +2080,7 @@ issues:
 
     # Exclude known linters from partially hard-vendored code,
     # which is impossible to exclude via `nolint` comments.
+    # `/` will be replaced by current OS file path separator to properly work on Windows.
     - path: internal/hmac/
       text: "weak cryptographic primitive"
       linters:
--- a/pkg/result/processors/exclude_rules.go
+++ b/pkg/result/processors/exclude_rules.go
@@ -44,7 +44,8 @@ func createRules(rules []ExcludeRule, prefix string) []excludeRule {
 			parsedRule.source = regexp.MustCompile(prefix + rule.Source)
 		}
 		if rule.Path != "" {
-			parsedRule.path = regexp.MustCompile(rule.Path)
+			path := normalizePathInRegex(rule.Path)
+			parsedRule.path = regexp.MustCompile(path)
 		}
 		parsedRules = append(parsedRules, parsedRule)
 	}


ldez:
I blocked this PR because I think it will create regressions and side effects.
I'm not using Windows so I'm quickly able to validate the behavior.

kohenkatz:
@ldez Can you elaborate on what you think this will break? As a Windows user, it looks correct to me, but I'd like to see if there's anything specific that you think should be tested. This code is already used for `skip-files` and `skip-dirs`, so why would it not work for `exclude-rules.path`?

kamilsk (dismissed):

kamilsk:
> When a PR is blocked, put an approved without any comment is weird.

OK, got it. `normalizePathInRegex` is already used by SkipDirs and SkipFiles, and

> This re-uses the same logic that is already in place for skip-dirs

sounds safe and valid.

Also, `separatorToReplace = regexp.QuoteMeta(string(filepath.Separator))` looks OK for Windows


```go
package main

import (
	"regexp"
	"strings"
)

func main() {
	path := '\\'
	separatorToReplace := regexp.QuoteMeta(string(path))

	file := "some/.*/file.go"
	file = strings.ReplaceAll(file, "/", separatorToReplace)
	re := regexp.MustCompile(file)

	if !re.MatchString("some\\path\\to\\file.go") {
		panic("wrong")
	}
}
```

on Windows I see paths in format `pkg\lint\lintersdb\manager.go`, see https://github.com/golangci/golangci-lint/runs/7469347948?check_suite_focus=true

So, I think there is no risks, but it will be great if some windows-specific tests will be added here.

ldez (approved):
LGTM

PR #2453, opened 2022-01-03
Add decorder linter

bosix:
This pr adds decorder - a declaration order linter for go source files (https://gitlab.com/bosi/decorder)

Please don't hesitate to ask any questions you may have :)


ldez:
Hello, could you give the sources (go team wiki, or pages that define and explain) of your rules?

bosix:
Hi @ldez,
I'm quite sure what do you want. The source code is open source and can be found [here](https://gitlab.com/bosi/decorder). I can expand the readme file and add more details about what exactly the linter checks if you want.

ldez:
I know where the code is :wink: , I'm asking for references/sources/documentation for your rules.

The goal is to define the correct default configuration and to be able to justify the rules.

butuzov:
just alternative tool: https://github.com/MarioCarrion/nit

bosix:
@ldez I've extended the readme file with a more detailed explanation what rules the linter apply. You may find the implementation of the rules directly inside the repo.

@butuzov thank you for your suggestion. The linter you suggest has no configuration options so you have to take all listed rules "as it". Since I don't want to enforce everything (e.g. always use parenthesis for declaration of variables), the linter is no option for me personally.

bombsimon:
To me it's clear that the linter just enforces a specific coding style and aiming for consistency and no matter _what_ you want to be consistent about it's a common practice to do so. It feels like a good tool to have at companies with multiple developers when you try to enforce code styles, I think it's easier to adapt and implement if they can be managed with static analysis.

I think people interested in this can configure their preferred order and then we could just use the upstream defaults (`type,const,var,func`) here. If I understand it correctly it's also possible to just specify some of them which would mean the others could be in any order, f.ex `const,var` which would require `const` before `var` but `type` can be either before, between or after, right?

ldez:
Sorry but you don't understand my point: I want something that explains why the default rules are the default rules.

If there is no explanation, the rules must be disabled by default.

bosix:
@idenx thank you for the clarification. Now I get your point. No there is no "explanation" why go code should be styled like suggested in the defaults and I agree that It's better to just suggest a linter by default when there is a good reason for it (better than enforcing consistency across multiple projects like @bombsimon mentioned). Can my linter be disabled by default completely by golangci-lint anywhere or do you think it's better to disable all rules so even if a user enabled the linter he/she has to explicitly enable each rule?

@bombsimon yes you're right. You can only define the order of a subset (e.g. var and const).

ldez:
> Can my linter be disabled by default completely by golangci-lint anywhere or do you think it's better to disable all rules so even if a user enabled the linter he/she has to explicitly enable each rule?

Your linter will be disabled by default, but many users use `enable-all` option by default, so in this case your linter will be "enable by default". Thus, the default behavior of your linter should be as opinionated as possible.

ldez:
@bosix the users will not apply the configuration from the example, but the example must show the default values (in comment).

The default behavior must be the default behavior without any configuration.

bosix:
@idenx alright. I'll adjust the config example. 

The rules are disabled by the default configuration in `pkg/config/linters_settings.go`. Shouldn't this aply when a user does not have anythink configured for the new linter?

bosix:
@ldez fixed. Please have a look at it

bombsimon (approved):
LGTM

pkg/lint/lintersdb/manager.go
```diff
@@ -189,6 +191,12 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithPresets(linter.PresetComplexity).
 			WithURL("https://github.com/bkielbasa/cyclop"),
 
+		linter.NewConfig(golinters.NewDecorder(decorderCfg)).
+			WithSince("v1.44.0").
+			WithLoadForGoAnalysis().
```
ldez, line 196:
```suggestion
```

test/testdata/decorder.go
```diff
@@ -0,0 +1,19 @@
+// args: -Edecorder
+//config_path: testdata/configs/decorder.yml
+package testdata
+
+const (
+	decoc = 1
```
ldez, line 6:
```suggestion
	decoc = math.MaxInt
```

add the import:
```
import "math"
```

We add one import of stdlib by test because some side effects can be hidden without that.

ldez (changes requested):

pkg/lint/lintersdb/manager.go
bosix: Suggestion is applied

test/testdata/decorder.go
bosix: Suggestion is applied

ldez (approved):
LGTM

PR #2459, opened 2022-01-04
docs: more revive examples

butuzov:
adjustments to #2458

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -747,6 +747,9 @@ linters-settings:
       - name: exported
         severity: warning
         disable: false
+        arguments:
+        - "checkPrivateReceivers"
+        - "sayRepetitiveInsteadOfStutters"
       # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#file-header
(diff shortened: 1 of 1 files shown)


.golangci.example.yml
```diff
@@ -878,8 +882,8 @@ linters-settings:
         severity: warning
         disable: false
         arguments:
-          - - ID
-          - - VM
+        - [ "ID" ] # Whitelist
+        - [ "VM" ] # Blacklist
```
ldez, line 886: my syntax works :wink: , but if you prefer `- ["xx"]` is also possible.
```suggestion
        - [ "ID" ] # Allowlist
        - [ "VM" ] # Denylist
```

ldez (changes requested):

.golangci.example.yml
butuzov: I didn't said it not works, but whitelist and blacklists was discovery for me (and I spend some time checking what is what in revive). So this is important thing to mention, imo.

ldez (approved):
LGTM

PR #2463, opened 2022-01-06
pkg/result/processors: compile nolint regexp only once

quasilyte:
`regexp.MatchString` compiles regexp for every invocation,
there is no pattern caching there.

This change introduces a precompiled regexp to save some time
while checking comments for nolint directives.

Found using go-perfguard with cpu profile collected while
running golangci-lint on itself.

### DIFF
--- a/pkg/result/processors/nolint.go
+++ b/pkg/result/processors/nolint.go
@@ -17,6 +17,7 @@ import (
 )
 
 var nolintDebugf = logutils.Debug("nolint")
+var nolintRe = regexp.MustCompile(`^nolint( |:|$)`)
 
 type ignoredRange struct {
 	linters                []string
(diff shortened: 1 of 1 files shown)


ldez:
@quasilyte could you sign the CLA?

quasilyte:
@ldez, done.

pkg/result/processors/nolint.go
```diff
@@ -232,9 +232,11 @@ func (p *Nolint) extractFileCommentsInlineRanges(fset *token.FileSet, comments .
 	return ret
 }
 
+var nolintRe = regexp.MustCompile(`^nolint( |:|$)`)
```
ldez, line 235: can you move this variable to the top of the file?

ldez (changes requested):

pkg/result/processors/nolint.go
quasilyte: Sure.

ernado (approved):
LGTM

ldez (approved):

PR #2467, opened 2022-01-09
depguard: updates configuration

timkral:
Hi,

This change updates the [Depguard linter to v1.1.0](https://github.com/OpenPeeDeeP/depguard/releases/tag/v1.1.0) and updates `golangci-lint` to include the new configuration available for it. Highlights:

1. Include the new `ignore-file-rules` configuration to the `depguard` settings. This allows the linter to ignore certain portions of a code base. Read more in the [README](https://github.com/OpenPeeDeeP/depguard#ignore-file-rules) and here's the [original PR](https://github.com/OpenPeeDeeP/depguard/pull/21).
1. Allow multiple guards in the `depguard` settings. This becomes important because of the new `ignore-file-rules` configuration. If we allow the `depguard` linter to ignore part of a code base, a developer may want to execute different guards on different parts of their code. A classic example of this would be a guard for tests and a guard for non-tests.
1. Change the settings to use updated language. Based on the PR [here](https://github.com/OpenPeeDeeP/depguard/pull/20).

cc: @dixonwille

Please let me know if there are any changes required. Thanks for the consideration.
-Tim


.golangci.yml
```diff
@@ -1,6 +1,6 @@
 linters-settings:
   depguard:
-    list-type: blacklist
+    list-type: denylist
```
timkral, line 3: My IDE was showing an error for this change based on the [JSON schema here](https://json.schemastore.org/golangci-lint.json). So not sure if this is OK to change or not.

pkg/golinters/depguard.go
```diff
@@ -63,47 +129,31 @@ func NewDepguard() *goanalysis.Linter {
...
 				Cwd:   "",  // fallbacked to os.Getcwd
 				Build: nil, // fallbacked to build.Default
 			}
-			issues, err := dg.Run(loadConfig, prog)
-			if err != nil {
-				return nil, err
-			}
-			if len(issues) == 0 {
-				return nil, nil
-			}
```
timkral, line 86: Note that this is taken out so that we don't accidentally short-circuit additional guards.

pkg/commands/run.go
```diff
@@ -193,6 +193,10 @@ func initFlagSet(fs *pflag.FlagSet, cfg *config.Config, m *lintersdb.Manager, is
 		"Depguard: check list against standard lib")
 	hideFlag("depguard.include-go-root")
 
+	fs.StringSliceVar(&lsc.Depguard.IgnoreFileRules, "depguard.ignore-file-rules", nil,
+		"Depguard: rules to ignore certain files for consideration")
+	hideFlag("depguard.ignore-file-rules")
+
```
ldez, line 199: Can you remove this? We don't add new flags.

https://github.com/golangci/golangci-lint/blob/29eedbf6e56508cf1ca52ad0eeed067b1c3a3066/pkg/commands/run.go#L128

pkg/config/linters_settings.go
```diff
@@ -178,6 +178,14 @@ type DepGuardSettings struct {
 	Packages                 []string
 	IncludeGoRoot            bool              `mapstructure:"include-go-root"`
 	PackagesWithErrorMessage map[string]string `mapstructure:"packages-with-error-message"`
+	IgnoreFileRules          []string          `mapstructure:"ignore-file-rules"`
+	AdditionalGuards         []struct {
```
ldez, line 182: can you create a named struct?

pkg/golinters/depguard.go
```diff
@@ -28,7 +90,10 @@ func setDepguardListType(dg *depguard.Depguard, lintCtx *linter.Context) error {
 	return nil
 }
 
-func setupDepguardPackages(dg *depguard.Depguard, lintCtx *linter.Context) {
+func setupDepguardPackages(
+	dg *depguard.Depguard,
+	packagesWithErrorMessage map[string]string,
+) {
```
ldez, line 96:
```suggestion
func setupDepguardPackages(dg *depguard.Depguard, packagesWithErrorMessage map[string]string) {
```

ldez (changes requested):

ldez:
I took advantage of the PR to do some cleaning.

ldez (approved):
LGTM

PR #2475, opened 2022-01-11
revive: adding missing lint rules

aniforprez:
This PR adds a couple of rules that are currently not present on the revive linter. These rules are available as of the current revive version

### DIFF
--- a/pkg/golinters/revive.go
+++ b/pkg/golinters/revive.go
@@ -305,6 +305,8 @@ var allRules = append([]lint.Rule{
 	&rule.NestedStructs{},
 	&rule.IfReturnRule{},
 	&rule.UselessBreak{},
+	&rule.OptimizeOperandsOrderRule{},
+	&rule.TimeEqualRule{},
 }, defaultRules...)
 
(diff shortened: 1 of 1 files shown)


pkg/golinters/revive.go
```diff
@@ -305,6 +305,8 @@ var allRules = append([]lint.Rule{
 	&rule.NestedStructs{},
 	&rule.IfReturnRule{},
 	&rule.UselessBreak{},
+	&rule.OptimizeOperandsOrderRule{},
+	&rule.TimeEqualRule{},
```
butuzov, line 309: Can you also add a sample configuration to [`.golangci.example.yml`](https://github.com/golangci/golangci-lint/blob/master/.golangci.example.yml)?

ldez, line 309:
```suggestion
	&rule.TimeEqualRule{},
	&rule.BannedCharsRule{},
	&rule.OptimizeOperandsOrderRule{},
```

ldez (changes requested):

PR #2476, opened 2022-01-11
docs: updates usage settings for `go-mnd`

butuzov:


.golangci.example.yml
```diff
@@ -389,10 +389,17 @@ linters-settings:
...
-        # ignored-numbers: 1000
-        # ignored-files: magic_.*.go
-        # ignored-functions: math.*
+        checks:
+        - argument
+        - case
+        - condition
+        - operation
+        - return
+        - assign
```
butuzov, line 398: its also possible to supply it as comma-separated string, however having an array as example better from a documentation perspective (easier to add validation checks for `json schema`)

ldez:
The linter requires a "comma separated list of checks", it's not a slice.

I need to check but I think that if you set a slice only the last value will be used.

butuzov:
Save your time, I am cleaning up json schema and already have tested this, this is the result of my checks https://github.com/tommy-muehle/go-mnd/pull/30 , this configuration is correctly recognized by `go-mnd`

ldez:
I will check in all cases because it's what I do for all PRs :smile_cat: :wink:

ldez (approved):
LGTM

PR #2482, opened 2022-01-14
exhaustive: drop deprecated/unused settings

butuzov:
Both settings are no more used by `exhaustive` linter:
- `ignore-pattern`
- `checking-strategy`

See https://github.com/nishanths/exhaustive/blob/v0.7.11/exhaustive.go#L203-L205


ldez (approved):
LGTM

PR #2483, opened 2022-01-14
godot: add period option

butuzov:
+ added ability to customize `period` setting
+ added default value for period (`true`)
+ update `.golangci.example.yml`

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -325,7 +325,7 @@ linters-settings:
     # exclude todo and fixme comments
     - "^fixme:"
     - "^todo:"
-    # don't check that each sentence ends with a period (default: true)
+    # check that each sentence ends with a period (default: true)
     period: false
     # check that each sentence starts with a capital letter (default: false)
     capital: true
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -322,10 +322,13 @@ linters-settings:
     scope: declarations
     # list of regexps for excluding particular comment lines from check
     exclude:
-      # example: exclude comments which contain numbers
-      # - '[0-9]+'
+    # exclude todo and fixme comments
+    - "^fixme:"
+    - "^todo:"
+    # check that each sentence ends with a dot
+    period: false
     # check that each sentence starts with a capital letter
-    capital: false
+    capital: true
 
   godox:
     # report any comments starting with keywords, this is useful for TODO or FIXME comments that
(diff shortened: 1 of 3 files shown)


pkg/config/linters_settings.go
```diff
@@ -29,6 +29,10 @@ var defaultLintersSettings = LintersSettings{
 	Godox: GodoxSettings{
 		Keywords: []string{},
 	},
+	Godot: GodotSettings{
+		Scope:  "declarations",
+		Period: true,
```
bombsimon, line 34: In `.golangci.example.yml` `capital` is set to true but since it's lacking in these settings it will bey default be set to `false`. Wouldn't it make more sense to have the example configuration represent the same as the default configuration? At least that's how it's usually configured I think?

ldez: yes, the values in `.golangci.example.yml` must not be the default values. The default values are in the comments.

butuzov: Got it. Will change once get to laptop.

butuzov: @ldez @bombsimon changes applied, will be glad if you can check it again.

.golangci.example.yml
```diff
@@ -318,14 +318,17 @@ linters-settings:
...
     # list of regexps for excluding particular comment lines from check
     exclude:
-      # example: exclude comments which contain numbers
-      # - '[0-9]+'
-    # check that each sentence starts with a capital letter
-    capital: false
+    # exclude todo and fixme comments
+    - "^fixme:"
+    - "^todo:"
+    # don't check that each sentence ends with a period (default: true)
```
ldez, line 328:
```suggestion
    # check that each sentence ends with a period (default: true)
```

I think it's better to explain the meaning of the option instead of the meaning of value in the example.

ldez (changes requested):

.golangci.example.yml
```diff
@@ -318,14 +318,17 @@ linters-settings:
...
+    scope: toplevel
     # list of regexps for excluding particular comment lines from check
     exclude:
-      # example: exclude comments which contain numbers
-      # - '[0-9]+'
-    # check that each sentence starts with a capital letter
-    capital: false
+    # exclude todo and fixme comments
+    - "^fixme:"
+    - "^todo:"
```
ldez, line 327:
```suggestion
      # exclude todo and fixme comments
      - "^fixme:"
      - "^todo:"
```

ldez (changes requested):

ldez (approved):
LGTM

butuzov:
Thank you @bombsimon and @ldez

PR #2484, opened 2022-01-15
docs: split configuration page into multiple sections

ldez:
Split the integration of `.golangci.example.yml` into multiple sections.
Switch the CLI section and the file section: put the file section first because it's the main use case.

<details>
<summary>before the PR</summary>

There is a small bug at the end of the capture (the black background disappears), but it's just a capture artifact. (the real current render [configuration](https://golangci-lint.run/usage/configuration/))

Screenshot 2022-01-15 at 12-01-14 Configuration golangci-lint

</details>

<details>
<summary>with the PR</summary>

Screenshot 2022-01-19 at 13-00-19 Configuration golangci-lint

</details>

### DIFF
--- a/scripts/expand_website_templates/main.go
+++ b/scripts/expand_website_templates/main.go
@@ -347,7 +347,7 @@ func extractExampleSnippets(example []byte) (*SettingSnippets, error) {
 
 	snippets := SettingSnippets{}
 
-	buffer := bytes.NewBufferString("")
+	builder := strings.Builder{}
 
 	for j, node := range root.Content {
 		switch node.Value {
@@ -385,7 +385,7 @@ func extractExampleSnippets(example []byte) (*SettingSnippets, error) {
 				return nil, err
 			}
 
-			_, _ = buffer.WriteString(
+			_, _ = builder.WriteString(
 				fmt.Sprintf(
 					"### `%s` configuration\n\nSee the dedicated [linters-settings](/usage/linters) documentation section.\n\n",
 					node.Value,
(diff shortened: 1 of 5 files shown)


scripts/expand_website_templates/main.go
```diff
@@ -317,58 +317,168 @@ func getThanksList() string {
...
+				},
+			},
 		}
+
+		_, _ = fmt.Fprintf(builder, "### %s\n\n", nextNode.Content[i].Value)
+		_, _ = fmt.Fprintln(builder, "```yaml")
+
+		const ident = 2
+		encoder := yaml.NewEncoder(builder)
+		encoder.SetIndent(ident)
```
bombsimon, line 449: Feels like a weird `const` when it's now defined two times in separate blocks and also when you bumped min occurrences to 3. I would either just do `encoder.SetIndent(2)` or move the `const` to the global scop at the top of the file.

scripts/expand_website_templates/main.go
```diff
@@ -317,58 +317,168 @@ func getThanksList() string {
...
+		HeadComment: root.HeadComment,
+		LineComment: root.LineComment,
+		FootComment: root.FootComment,
+		Line:        root.Line,
+		Column:      root.Column,
+	}
+
+	snippets := SettingSnippets{}
+
+	buffer := bytes.NewBufferString("")
```
bombsimon, line 350: The old code and now two new functions use `strings.Builder` combined with `fmt.Fprint`. Any reason to not use the same here for consistency?

bombsimon (approved):
Sorry for the slow review.

Looks great, much easier to read when split up!

PR #2496, opened 2022-01-18
varnamelen: explicit default values

butuzov:
adds explicit default values 

(defined in varnamelen linter)
- `MaxDistance` (int:5)
- `MinNameLength` (int:3)


ldez:
I will push force because a CI check (CLA) did not start (and I don't know why).

ldez (approved):

PR #2497, opened 2022-01-19
Add "grouper" linter

leonklingele:
https://github.com/leonklingele/grouper

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -807,18 +807,33 @@ linters-settings:
       - unusedwrite
 
   grouper:
-    # const
-    const-require-single-const: false
-    const-require-grouping: false
-    # import
-    import-require-single-import: false
-    import-require-grouping: false
-    # type
-    type-require-single-type: false
-    type-require-grouping: false
-    # var
-    var-require-single-var: false
-    var-require-grouping: false
+    # Require the use of a single global 'const' declaration only.
+    # Default: false
+    const-require-single-const: true
+    # Require the use of grouped global 'const' declarations.
+    # Default: false
+    const-require-grouping: true
+
+    # Require the use of a single 'import' declaration only.
+    # Default: false
+    import-require-single-import: true
+    # Require the use of grouped 'import' declarations.
+    # Default: false
+    import-require-grouping: true
+
+    # Require the use of a single global 'type' declaration only.
+    # Default: false
+    type-require-single-type: true
+    # Require the use of grouped global 'type' declarations.
+    # Default: false
+    type-require-grouping: true
+
+    # Require the use of a single global 'var' declaration only.
+    # Default: false
+    var-require-single-var: true
+    # Require the use of grouped global 'var' declarations.
+    # Default: false
+    var-require-grouping: true
 
   ifshort:
     # Maximum length of variable declaration measured in number of lines, after which linter won't suggest using short syntax.
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -806,6 +806,20 @@ linters-settings:
       - unusedresult
       - unusedwrite
 
+  grouper:
+    # const
+    const-require-single-const: false
+    const-require-grouping: false
+    # import
+    import-require-single-import: false
+    import-require-grouping: false
+    # type
+    type-require-single-type: false
+    type-require-grouping: false
+    # var
+    var-require-single-var: false
+    var-require-grouping: false
+
   ifshort:
     # Maximum length of variable declaration measured in number of lines, after which linter won't suggest using short syntax.
     # Has higher priority than max-decl-chars.
--- a/pkg/config/linters_settings.go
+++ b/pkg/config/linters_settings.go
@@ -131,6 +131,7 @@ type LintersSettings struct {
 	Gosec            GoSecSettings
 	Gosimple         StaticCheckSettings
 	Govet            GovetSettings
+	Grouper          GrouperSettings
 	Ifshort          IfshortSettings
 	ImportAs         ImportAsSettings
 	Ireturn          IreturnSettings
@@ -374,6 +375,21 @@ func (cfg GovetSettings) Validate() error {
 		return errors.New("disable-all and disable can't be combined")
 	}
 	return nil
+}
+
+type GrouperSettings struct {
+	// const
+	ConstRequireSingleConst bool `mapstructure:"const-require-single-const"`
+	ConstRequireGrouping    bool `mapstructure:"const-require-grouping"`
+	// import
+	ImportRequireSingleImport bool `mapstructure:"import-require-single-import"`
+	ImportRequireGrouping     bool `mapstructure:"import-require-grouping"`
+	// type
+	TypeRequireSingleType bool `mapstructure:"type-require-single-type"`
+	TypeRequireGrouping   bool `mapstructure:"type-require-grouping"`
+	// var
+	VarRequireSingleVar bool `mapstructure:"var-require-single-var"`
+	VarRequireGrouping  bool `mapstructure:"var-require-grouping"`
 }
 
 type IfshortSettings struct {
--- a/pkg/config/linters_settings.go
+++ b/pkg/config/linters_settings.go
@@ -378,17 +378,14 @@ func (cfg GovetSettings) Validate() error {
 }
 
 type GrouperSettings struct {
-	ConstRequireSingleConst bool `mapstructure:"const-require-single-const"`
-	ConstRequireGrouping    bool `mapstructure:"const-require-grouping"`
-
+	ConstRequireSingleConst   bool `mapstructure:"const-require-single-const"`
+	ConstRequireGrouping      bool `mapstructure:"const-require-grouping"`
 	ImportRequireSingleImport bool `mapstructure:"import-require-single-import"`
 	ImportRequireGrouping     bool `mapstructure:"import-require-grouping"`
-
-	TypeRequireSingleType bool `mapstructure:"type-require-single-type"`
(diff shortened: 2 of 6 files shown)


pkg/lint/lintersdb/manager.go
```diff
@@ -408,6 +408,11 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithAlternativeNames("vet", "vetshadow").
 			WithURL("https://golang.org/cmd/vet/"),
 
+		linter.NewConfig(golinters.NewGrouper(m.cfg)).
+			WithSince("v1.22.0"). // TODO
```
leonklingele, line 412: What do use here?

pkg/config/linters_settings.go
```diff
@@ -368,6 +369,10 @@ func (cfg GovetSettings) Validate() error {
 	return nil
 }
 
+type GrouperSettings struct {
+	Config map[string]map[string]interface{} `mapstructure:"config"`
```
leonklingele, line 373: `config` or `settings`? What's the preferred name?

pkg/lint/lintersdb/manager.go
bombsimon: The current version is `1.43.0` so this will be a part of the next minor release, it should be mentioned [under section 3](https://golangci-lint.run/contributing/new-linters/) in the docs.

```suggestion
			WithSince("v1.44.0").
```

ldez: https://github.com/golangci/golangci-lint/blob/e3d0247ee24aed2fc231b21d3eaa04f0fd7eca47/pkg/lint/lintersdb/manager.go#L165-L166

.golangci.yml
```diff
@@ -83,6 +89,7 @@ linters:
     - gosec
     - gosimple
     - govet
+    - grouper
```
ldez, line 92:
```suggestion
```

pkg/config/linters_settings.go
ldez: The configuration must be explicit:
```go
type GrouperSettings struct {
	ImportRequireSingleImport bool `mapstructure:"import-require-single-import"`
	ImportRequireGrouping     bool `mapstructure:"import-require-grouping"`
}
```

.golangci.example.yml
```diff
@@ -516,6 +516,12 @@ linters-settings:
       - shadow
     disable-all: false
 
+  grouper:
+    config:
+      grouper:
+        import-require-single-import: false
+        import-require-grouping: false
+
```
ldez, line 524:
```suggestion
  grouper:
    import-require-single-import: false
    import-require-grouping: false
```

pkg/golinters/grouper.go
```diff
@@ -0,0 +1,27 @@
...
+
+	var linterCfg map[string]map[string]interface{}
+	if cfg != nil {
+		linterCfg = cfg.LintersSettings.Grouper.Config
+	}
+
+	return goanalysis.NewLinter(
+		"grouper",
+		"An analyzer to analyze expression groups.",
+		analyzers,
```
ldez, line 24:
```suggestion
		[]*analysis.Analyzer{grouper.New()},
```

pkg/golinters/grouper.go
```diff
@@ -0,0 +1,27 @@
...
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
+)
+
+func NewGrouper(cfg *config.Config) *goanalysis.Linter {
+	analyzers := []*analysis.Analyzer{
+		grouper.New(),
+	}
```
ldez, line 14:
```suggestion
```

pkg/golinters/grouper.go
```diff
@@ -0,0 +1,27 @@
...
+
+	var linterCfg map[string]map[string]interface{}
+	if cfg != nil {
+		linterCfg = cfg.LintersSettings.Grouper.Config
+	}
```
ldez, line 19:
```suggestion
	linterCfg := map[string]map[string]interface{}{}
	if settings != nil {
		linterCfg["grouper"] = map[string]interface{}{
			"import-require-single-import": settings.ImportRequireSingleImport,
			"import-require-grouping":      settings.ImportRequireGrouping,
		}
	}
```

pkg/lint/lintersdb/manager.go
```diff
@@ -408,6 +408,11 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithAlternativeNames("vet", "vetshadow").
 			WithURL("https://golang.org/cmd/vet/"),
 
+		linter.NewConfig(golinters.NewGrouper(m.cfg)).
```
ldez, line 411:
```suggestion
		linter.NewConfig(golinters.NewGrouper(grouperCfg)).
```

and:

```go
func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
	var bidichkCfg *config.BiDiChkSettings
	// ...
	var grouperCfg *config.GrouperSettings
	// ...
	if m.cfg != nil {
		bidichkCfg = &m.cfg.LintersSettings.BiDiChk
		// ...
		grouperCfg = &m.cfg.LintersSettings.Grouper
		// ...
	}

	// ...
}
```

ldez (changes requested):

pkg/config/linters_settings.go
```diff
@@ -368,6 +369,10 @@ func (cfg GovetSettings) Validate() error {
...
+}
```
ldez, line 374:
```suggestion
type GrouperSettings struct {
	ImportRequireSingleImport bool `mapstructure:"import-require-single-import"`
	ImportRequireGrouping     bool `mapstructure:"import-require-grouping"`
}
```

ldez (changes requested):

leonklingele:
Thanks for the thorough review! I basically reused the linter definition of `gomnd`: https://github.com/golangci/golangci-lint/blob/e3d0247ee24aed2fc231b21d3eaa04f0fd7eca47/pkg/golinters/gomnd.go And, to be honest, find it quite nice not having to explicitly list all flags here again. What's the benefit of an explicit configuration?

ldez:
> Thanks for the thorough review! I basically reused the linter definition of gomnd: https://github.com/golangci/golangci-lint/blob/e3d0247ee24aed2fc231b21d3eaa04f0fd7eca47/pkg/golinters/gomnd.go 

As you can see, now, go-mnd uses the same pattern as the other linters.
https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/gomnd.go

> And, to be honest, find it quite nice not having to explicitly list all flags here again. What's the benefit of an explicit configuration?

- the 2 extra levels `grouper.config.grouper` are useless.
- we use Go, we have types, so we use types.
- It's better to get a clear overview of the configuration.
- Allow checking the configuration options.

leonklingele:
@ldez ptal :)

.golangci.example.yml
```diff
@@ -559,6 +559,20 @@ linters-settings:
...
+    const-require-grouping: false
+    # import
+    import-require-single-import: false
+    import-require-grouping: false
+    # type
+    type-require-single-type: false
+    type-require-grouping: false
+    # var
+    var-require-single-var: false
+    var-require-grouping: false
```
ldez, line 574: Can you improve the documentation?

- add an explanation for each option
- add the default value in comments for each option
- add a value for each option that is not the default value

ldez (changes requested):

pkg/golinters/grouper.go
```diff
@@ -0,0 +1,37 @@
...
+			"const-require-grouping":     settings.ConstRequireGrouping,
+			// import
+			"import-require-single-import": settings.ImportRequireSingleImport,
+			"import-require-grouping":      settings.ImportRequireGrouping,
+			// type
+			"type-require-single-type": settings.TypeRequireSingleType,
+			"type-require-grouping":    settings.TypeRequireGrouping,
+			// var
+			"var-require-single-var": settings.VarRequireSingleVar,
+			"var-require-grouping":   settings.VarRequireGrouping,
```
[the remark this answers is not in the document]
leonklingele: Actually, I prefer to leave it as-is. The comments serve as a spacer between the different analyzer options. Could also add a suffix of ` analyzer`.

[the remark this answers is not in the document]
ldez: The names of the fields already contain the information.
Comments should be added when the code is not explicit but if the code is clear the comments add noise.

leonklingele:
oki, ptal :)

pkg/golinters/grouper.go
```diff
@@ -0,0 +1,36 @@
+package golinters
+
+import (
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
+
+	grouper "github.com/leonklingele/grouper/pkg/analyzer"
```
ldez, line 9:
```suggestion
	grouper "github.com/leonklingele/grouper/pkg/analyzer"
	"golang.org/x/tools/go/analysis"

	"github.com/golangci/golangci-lint/pkg/config"
	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
```

ldez (changes requested):
I still think that the "groups" are not needed.

leonklingele:
I hope that's it for now :)

ldez (approved):
LGTM

PR #2498, opened 2022-01-19
gomnd: new configuration

ldez:
I rewrote the `gomnd` configuration to be like the other linters configuration.
I added a compatibility section, so no breaking changes.

The old configuration and the new configuration cannot be used at the same time: if the old configuration is defined, only this configuration will be used.

### DIFF
--- a/pkg/golinters/gomnd.go
+++ b/pkg/golinters/gomnd.go
@@ -8,20 +8,29 @@ import (
 	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
 )
 
-func NewGoMND(cfg *config.Config) *goanalysis.Linter {
-	analyzers := []*analysis.Analyzer{
-		mnd.Analyzer,
-	}
-
+func NewGoMND(settings *config.GoMndSettings) *goanalysis.Linter {
 	var linterCfg map[string]map[string]interface{}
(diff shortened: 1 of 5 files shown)


pkg/golinters/gomnd.go
```diff
@@ -8,20 +8,29 @@ import (
...
-	if cfg != nil {
-		linterCfg = cfg.LintersSettings.Gomnd.Settings
+
+	if settings != nil {
+		// TODO(ldez) For compatibility only, must be drop in v2.
+		if len(settings.Settings) > 0 {
+			linterCfg = settings.Settings
+		} else {
+			linterCfg = map[string]map[string]interface{}{
+				"mnd": {
```
butuzov, line 20: did you mean `gomnd` ?

ldez: The analyzer name is `mnd`, so it's the right value.

https://github.com/tommy-muehle/go-mnd/blob/c5e4f67f0b1b0a8114d42c8b95709f4e6dd30b37/analyzer.go#L19

butuzov: I see error too. You are right.

butuzov (approved):

butuzov:
It looks better in your PR.

PR #2362, opened 2021-11-16
Add errchkjson linter

breml:
After the discussion on #2349, in particular in comment https://github.com/golangci/golangci-lint/pull/2349#discussion_r748742253, I decided to submit an alternative PR for the inclusion of _errchkjson_.

In this implementation there is by default no overlap with other linters (in particular _errcheck_ and _staticcheck_). By setting the respective configuration for _errchkjson_, the user can opt-in for the check for "error free json encoding". If the user decides to do so, it is recommended to exclude some functions from being checked by _errcheck_. The necessary instructions are provided to the user as comment in `.golangci.example.yml`.

closes #2349

PR #2499, opened 2022-01-19
docs: improve linters settings section

butuzov:
- removed `unused` settings (non-existing)
- fixed `promlinter` linter name (UnitAbbreviations)
- small refactoring on `default` values
- reordering
- added `varcheck`'s setting `exported-fields`

tl/dr: mostly cosmetic change.

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -580,8 +580,8 @@ linters-settings:
     # You can specify idiomatic endings for interface
     - (or|er)$
 
-    # reject:
-    # - github.com\/user\/package\/v4\.Type
+    reject:
+    - github.com\/user\/package\/v4\.Type
 
   lll:
     # max line length, lines longer will be reported. Default is 120.
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -128,6 +128,26 @@ linters-settings:
     # default: true (disabled)
     disable-dec-num-check: false
 
+  depguard:
+    list-type: denylist
+    include-go-root: false
+    packages:
+      - github.com/sirupsen/logrus
+    packages-with-error-message:
+      # specify an error message to output when a denied package is used
+      - github.com/sirupsen/logrus: "logging is allowed only by logutils.Log"
+    # create additional guards that follow the same configuration pattern
+    # results from all guards are aggregated together
+    additional-guards:
+      - list-type: denylist
+        include-go-root: false
+        packages:
+          - github.com/stretchr/testify
+        # specify rules by which the linter ignores certain files for consideration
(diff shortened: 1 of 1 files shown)


.golangci.example.yml
```diff
@@ -579,9 +580,8 @@ linters-settings:
     # You can specify idiomatic endings for interface
     - (or|er)$
 
-    # Reject patterns
-    reject:
-    - github.com\/user\/package\/v4\.Type
+    # reject:
+    # - github.com\/user\/package\/v4\.Type
```
ldez, line 584: can you restore the previous section? and uncomment the `keyword` section.

ldez:
> - commenting out `ireturn`'s `reject` values (only one value allowed)

I don't understand this sentence, can you explain a bit more?

ldez (changes requested):

.golangci.example.yml
```diff
@@ -163,21 +183,12 @@ linters-settings:
...
-    #
-    # if check-error-free-encoding is set to true and errcheck linter is enabled,
-    # it is recommended to add the following exceptions to prevent from false positives:
-    #
-    #     linters-settings:
-    #       errcheck:
-    #         exclude-functions:
-    #           - encoding/json.Marshal
-    #           - encoding/json.MarshalIndent
-    #           - (*encoding/json.Encoder).Encode
```
breml, line 178: This has been the result of a lengthy discussion in #2349 and #2362. Me personally would not recommend to remove the section about `linters-settings`, because as mentioned in the above comment, one will have false positives from `errcheck`, if this setting is set to true without the exceptions.

.golangci.example.yml
butuzov: done

.golangci.example.yml
ldez: @butuzov can you restore this section

butuzov:
> I don't understand this sentence, can you explain a bit more?

My fault. What I meant to say is `reject` or `allow`, not the one value for the setting.

.golangci.example.yml
butuzov: Clear. thank you, this wasn't clear while cleaning examples.

butuzov: @ldez will do that tomorrow ( not allowed to use git while reading fairy tales for my son)

ldez: I edited your PR because I want to create another one base on yours.

ldez (approved):
LGTM

.golangci.example.yml
butuzov: got it.

PR #2503, opened 2022-01-20
importas: fix no-extra-aliases

ldez:


butuzov (approved):

butuzov:
Thank you for noticing this.

PR #2525, opened 2022-01-27
contextcheck: bump to v1.0.5 && add noCache flag

sylvia7788:
The pkg that the current buildssa.SSA.Pkg depends on cannot be built because its info is nil. 
https://github.com/golang/tools/blob/0f0bbfd77bf2de5c3a4bc5cc6126522f42ef47c7/go/analysis/passes/buildssa/buildssa.go#L55-L68
If you want to get a function with Block, you can only get it from the analysis results of the previous dependent packages, but the cache will cause the analysis of the dependent packages to be skipped, so nocache is added.
https://github.com/sylvia7788/contextcheck/compare/v1.0.4...v1.0.5

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -83,7 +83,7 @@ require (
 	github.com/spf13/viper v1.10.1
 	github.com/ssgreg/nlreturn/v2 v2.2.1
 	github.com/stretchr/testify v1.7.0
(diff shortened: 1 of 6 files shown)


ldez:
The `noCache` option disables the cache for all the linters, so we cannot follow this way.
So you will have to find another solution/way.

PR #2526, opened 2022-01-27
Add "funcresult" linter

leonklingele:
https://github.com/leonklingele/funcresult


ldez:
Hello,

we already have a linter that handle those cases: https://go-critic.com/overview#unnamedresult

leonklingele:
@ldez it's not the same. The `funcresult` linter unconditionally _enforces_ the use of either named or unnamed result parameters, while `unnamedResult` only _conditionally_ decides whether result params should be _named_.

ldez:
For me, it's something that can be added to the existing linter.

leonklingele:
I've opened https://github.com/go-critic/go-critic/issues/1201. Let's see if I can integrate it to `go-critic`.

leonklingele:
seeing that this was only closed due to the use of the `require-named` option, I've removed support for it from this PR.
@ldez can you please reopen?

ldez:
> seeing that this was only closed due to the use of the require-named option,

It's not what I said: https://github.com/golangci/golangci-lint/pull/2701#issuecomment-1086678811

PR #2701 already provides the feature, so I will not reopen this one.

OSS is about sharing, neither authorship nor ownership, I recommend contributing on the linter added in #2701.

PR #2550, opened 2022-02-05
Updating tagliatelle to 0.3.1

pkprzekwas:
There is a bug making tags like `json:",inline"` failing. It was fixed in `0.3.1` release of tagliatelle (https://github.com/ldez/tagliatelle/commit/c13286cdb4c6fd98c732a65bb0505247ab9a8363), hence PR with version update.

Example:
Screenshot 2022-02-05 at 09 41 26

Note: It was working just fine before upgraded to `golangci-lint` 1.44.0


butuzov:
Thank you, but we are using @dependabot to update our dependencies.

PR #2560, opened 2022-02-08
docs(decorder): fix `disable-init-func-first-check: false` elaboration

scop:


scop:
The same issue is at https://golangci-lint.run/usage/linters/#decorder, I'm not aware where that should be fixed.

ldez (approved):

Issue #1981, opened 2021-05-14
golangci-lint cli returning non-zero value for issues raised at a severity level not of error


jufemaiz:
- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [x] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/) (not applicable)

<details><summary>Running golangci-lint CLI with severity set to `info` for a given linter returns a non-0 value</summary>

</details>

<details><summary>v1.39.0, v1.40.0</summary></details>

<details><summary>Config file</summary>

```console
---
linters:
  # please, do not use `enable-all`: it's deprecated and will be removed soon.
  # inverted configuration with `enable-all` and `disable` is not scalable during updates of golangci-lint
  disable-all: true
  enable:
    - dupl
    - godox

issues:
  # List of regexps of issue texts to exclude, empty list by default.
  # But independently from this option we use default exclude patterns,
  # it can be disabled by `exclude-use-default: false`. To list all
  # excluded by default patterns execute `golangci-lint run --help`
  exclude:
  - 'declaration of "(err|ctx)" shadows declaration at'

  # Excluding configuration per-path, per-linter, per-text and per-source
  exclude-rules:
    - path: _test\.go
      linters:
        - dupl

run:
  timeout: 5m
  skip-dirs:
    - test/testdata_etc
    - internal/cache
    - internal/renameio
    - internal/robustio
    - tmp/

# golangci.com configuration
# https://github.com/golangci/golangci/wiki/Configuration
service:
  golangci-lint-version: 1.40.x
  prepare:
    - echo "here I can run custom commands, but no preparation needed for this repo"

severity:
  # Default value is empty string.
  # Set the default severity for issues. If severity rules are defined and the issues
  # do not match or no severity is provided to the rule this will be the default
  # severity applied. Severities should match the supported severity names of the
  # selected out format.
  # - Code climate: https://docs.codeclimate.com/docs/issues#issue-severity
  # -   Checkstyle: https://checkstyle.sourceforge.io/property_types.html#severity
  # -       Github: https://help.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-error-message
  default-severity: error

  # The default value is false.
  # If set to true severity-rules regular expressions become case sensitive.
  case-sensitive: false

[...]

Issue #1839, opened 2021-03-13
Coloring Options for Severity


butuzov:
**Is your feature request related to a problem? Please describe.**
As a new dev in a team, I want to be able to distinguish issues by their severity. But Severity is an opinionated thing and it really depends, on what you "mean" and how you treat error. In `golangci-lint` 
- We have a severity concept for issues introduced quite along time ago.
- We also have coloring for issues.

Why not combine those two?
 
**Describe the solution you'd like**
Custom colors per severity?

**Describe alternatives you've considered**
Filtering out non-errors, or non-critical things.

**Additional context**
Here is demo of implementation on how it  difference:
![golangci-lint-colors](https://user-images.githubusercontent.com/651824/111017184-7ed55d00-83ba-11eb-8b5d-28da9d97a7d8.jpg)
![golangci-lint](https://user-images.githubusercontent.com/651824/111017186-81d04d80-83ba-11eb-9f8c-e6140c5caf5f.jpg)

Upvote or Downvote?

wandergeek:
Yes please!

PR #2595, opened 2022-02-17
feat(severity): color warning severity, non-zero exit when issue != default severity

jaredallard:
**What this PR does**: This PR adds a new `print-severity` field that toggles printing severity information on the `text` printer. By default, we now change the color of the error if the severity is set to `warning`. Also changed is when issues are discovered, we only return a non-zero exit code if it's equal to the default severity (e.g. `error`). If it's a non-default severity, e.g. `warning`, we return a zero exit code.

**Imporant Note**: I know this is super opinionated, so let me know if there's a different direction we want to take this. I'm not attached to this implementation at all! Also happy to split this into multiple PRs 😄

### DIFF
--- a/pkg/config/output.go
+++ b/pkg/config/output.go
@@ -29,6 +29,7 @@ type Output struct {
 	Color               string
 	PrintIssuedLine     bool   `mapstructure:"print-issued-lines"`
 	PrintLinterName     bool   `mapstructure:"print-linter-name"`
+	PrintSeverity       bool   `mapstructure:"print-severity"`
 	UniqByLine          bool   `mapstructure:"uniq-by-line"`
 	SortResults         bool   `mapstructure:"sort-results"`
 	PrintWelcomeMessage bool   `mapstructure:"print-welcome"`
--- a/pkg/printers/text.go
+++ b/pkg/printers/text.go
@@ -14,6 +14,7 @@ import (
 
 type Text struct {
 	printIssuedLine bool
+	printSeverity   bool
 	useColors       bool
 	printLinterName bool
 
@@ -21,9 +22,10 @@ type Text struct {
 	w   io.Writer
 }
 
-func NewText(printIssuedLine, useColors, printLinterName bool, log logutils.Log, w io.Writer) *Text {
+func NewText(printIssuedLine, useColors, printLinterName, printSeverity bool, log logutils.Log, w io.Writer) *Text {
 	return &Text{
 		printIssuedLine: printIssuedLine,
+		printSeverity:   printSeverity,
 		useColors:       useColors,
 		printLinterName: printLinterName,
 		log:             log,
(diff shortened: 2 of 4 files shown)


pkg/config/output.go
```diff
@@ -29,6 +29,7 @@ type Output struct {
 	Color               string
 	PrintIssuedLine     bool   `mapstructure:"print-issued-lines"`
 	PrintLinterName     bool   `mapstructure:"print-linter-name"`
+	PrintSeverity       bool   `mapstructure:"print-severity"`
```
jaredallard, line 32: I don't really like this `print-severity` field in general, but I want to try to maintain older behaviour if possible. I was thinking of adding something like `exit-non-zero-severity` to replace the default severity field. Still need to look at this codebase more to figure out how that's implemented.

pkg/printers/text.go
```diff
@@ -56,10 +58,18 @@ func (p *Text) Print(ctx context.Context, issues []result.Issue) error {
...
+	issueColor := color.FgRed
+	if i.Severity == "warning" {
+		issueColor = color.FgYellow
+	}
+
+	text := p.SprintfColored(issueColor, "%s", strings.TrimSpace(i.Text))
 	if p.printLinterName {
 		text += fmt.Sprintf(" (%s)", i.FromLinter)
 	}
+	if p.printSeverity {
```
jaredallard, line 70: The only time I really found this useful is if coloring is turned off, so I'm happy to remove it / restructure this.

ldez:
I have a problem with your PR because it contains 2 topics:
- color on the severity
- the exit code depends on the severity

It's 2 different topics, so it means 2 different PRs/issues.

Regarding the exit code, as far as I remember, discussions have already taken place on this subject and overall the decision was made not to correlate severity and exit code.
So I recommend looking for those discussions, and if you don't them, you can open an issue to handle this topic. 

Related to:
- #1839
- #1981

butuzov:
Hi, I did some related work in the past, so cannot really review this code as I am biased. I have tested both (actually one) features on myself and the company I was working for at that moment. I am also not aware of the discussion on exit codes and did my own research. However, I can give a small intro on the topic:

Both `severity` and `coloring` are connected, but you better approach it from the point of the severity of exit codes first.
 - You either fail execution or not (exit with `1` or `0`). 
 - `golangci-lint` option `issues-exit-code` needs to have precedence, to have backward compatibility.
 - You can start incorporating severity from the point it's implemented in `golangci-lint` see `severity` configuration.
 
What I found after using this on my own flow:
- No need for extra colors, `red` for errors, and a custom one for `warnings` is pretty much top of what people use.
- No need for extra severity types (like `info` or `debug`). `warnings` and `errors` are what may be the top of what is needed.

Problems:
- Should custom severity hold extra exit code? complicated questions - as what should we do if there are three different types of issues?
- Customization: currently it's supported only for the linters, and you can define whatever severity you want. You can't set up `severity` per issue (like for example for `exclude-rules`). Some of the linters provide their own severity (like `revive`) - what should we do in this case ? Define rule twice? Follow original rules and break backward compatibility? 

IMO, this feature is a bit controversial. I did implement that for myself, I didn't like that it gives a bit more space than needed.  And tbh I didn't look for the answers from the community.  The support for the idea itself was low at(hat)m, I am not even talking about implementation and support.

PR #2596, opened 2022-02-18
bump github.com/daixiang0/gci to v0.3.1

ldez:
https://github.com/daixiang0/gci/compare/76d765e3ab48...v0.3.1

**This update is done by hand because dependabot is not able to update non-semver dependency.**


bombsimon (approved):
😅

PR #2599, opened 2022-02-18
forbidigo: add example that uses comments

craigfurman:
Add an example of a recently-added forbidigo feature, comments (https://github.com/ashanbrown/forbidigo/pull/11).

Is the docs site (https://golangci-lint.run/usage/linters/#forbidigo) generated from this file?

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -280,6 +280,9 @@ linters-settings:
     forbid:
       - ^print.*$
       - 'fmt\.Print.*'
+      # Optionally put comments at the end of the regex, surrounded by `(# )?`
+      # Escape any special characters.
+      - 'fmt\.Print.*(# Do not commit print statements\.)?'
     # Exclude godoc examples from forbidigo checks.
     # Default: true
     exclude_godoc_examples: false


butuzov:
Can message be a bit more explanatory? ( in the example ). Cool feature, however, sad it removed proper error message.

ldez:
> Is the docs site (https://golangci-lint.run/usage/linters/#forbidigo) generated from this file?

Yes the linters page is generated from the `.golangci.example.yml` file.

craigfurman:
> Cool feature, however, sad it removed proper error message

I'm not sure I understand, what proper error message does this feature remove?

> Can message be a bit more explanatory? ( in the example ).

Could do, although given that this is a contrived example, we'd have to invent some context. Maybe "don't use fmt, use some logger instead", but it feels to name any arbitrary logger in a golangci example. Please leave a suggestion if you have something specific in mind though 👍

butuzov:
Never mind about the error, i didn't notice you move regexp compilation into `parse`. And yeah, more actionable examples 
like 
```yaml
       - '\.Debug\(\)\.(# Do not commit Debug code)?' # db Debug()
```

> http_test.go:64:16: use of `suite.DB.Exec(dropSQLQuery).Debug().Error` forbidden because "Do not commit Debug code" (forbidigo)


btw, maybe note in comments that

craigfurman:
Done

butuzov (approved):

butuzov:
Will merge one get to the laptop. Thank you!

PR #2608, opened 2022-02-21
Add 'floatcompare' linter

mweb:
https://github.com/mweb/floatcompare


.golangci.example.yml
```diff
@@ -280,6 +280,11 @@ linters-settings:
       - '*.Test'
       - 'example.com/package.ExampleStruct'
 
+  floatcomparison:
```
ldez, line 283:
```suggestion
  floatcompare:
```

pkg/lint/lintersdb/manager.go
```diff
@@ -279,6 +281,12 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithLoadForGoAnalysis().
 			WithURL("https://github.com/kyoh86/exportloopref"),
 
+		linter.NewConfig(golinters.NewFloatCompare(floatCompareStructCfg)).
+			WithSince("v.1.45.0").
```
ldez, line 285:
```suggestion
			WithSince("v1.46.0").
```

ldez (changes requested):
COuld you rebase your PR? Thanks.

ldez:

### Pull Request Description

### Linter

- [ ] They must work with `T=<name of the linter test file>.go make test_linters`.

### `.golangci.example.yml`

### Others Requirements

### Recommendations

- [ ] The linter repository should have a CI, tests, a readme and linting.

---

pkg/golinters/floatcompare.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/mweb/floatcompare"
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
+)
+
+const floatcompareName = "floatcompare"
```
ldez, line 11:
```suggestion
```

ldez (changes requested):

mweb:
Thanks for the review, I think I have fixed most of the open points. In my view, there are three open points:

- [ ]  It must not be a duplicate of another linter or a rule of a linter. (the team will help to verify that)

With the first two, I have no idea how I should be doing this, and it is written that you might help. The last one is recommended, this means it is optional? I will have a look how I could provide such a release with my linter.

mweb:
I added binaries to the release of the linter, and I believe I added a description about the linter to the pull request or do I have to add the description to the commit message?

ldez:
> do I have to add the description to the commit message?

You don't have to squash:
- It's better during a review to address changes as commits, it's easier to follow the changes.
- We squash the commits of the PR during the merge.

So the commit message doesn't need to be updated.

ldez:
I have several problems with this linter.

It is closed to https://go-critic.com/overview.html#truncatecmp

go-critic already tried to implement [this rule](https://github.com/go-critic/go-critic/pull/509) but the rule was removed and moved [outside](https://github.com/go-critic/checkers-contrib/blob/master/floatCompare_checker.go) go-critic.

It's unclear why the rule has been pushed outside go-critic but I think it's because of the number of [false-positives](https://github.com/go-critic/go-critic/issues/511).

mweb:
> I have some problems with this linter.
> 
> go-critic already tried to implement [this rule](https://github.com/go-critic/go-critic/pull/509) but the rule was removed and moved [outside](https://github.com/go-critic/checkers-contrib/blob/master/floatCompare_checker.go) go-critic.
> 
> It's unclear why the rule has been pushed outside go-critic but I think it's because of the number of [false-positives](https://github.com/go-critic/go-critic/issues/511).

This is interesting since the linter implements some mentioned problems in a later [pull request](https://github.com/go-critic/go-critic/pull/561).
There are some valid points that the current floatcompare linter doesn't implement, for example x != x for nan checks and the x == x. The other raised point for comparison against numbers, I might consider this to be a configurable parameter since some people might be ok with this in their code.

Would it help if the linter was improved this way, or do you prefer not to have such a linter integrated?

> 
> Otherwise, it is closed to https://go-critic.com/overview.html#truncatecmp

On this point, I disagree since these two rules are very different since this checks for truncate problems whereas float comparison is a different problem as you might see in the description of the [floatcompare linter](https://github.com/mweb/floatcompare) and the referenced [link](https://floating-point-gui.de/).

Issue #1490, opened 2020-11-03
gci+goimports = duplicated imports


howardjohn:
Config:
```yaml
linters:
  disable-all: true
  enable:
  - goimports
  - gci
  fast: false

linters-settings:
  goimports:
    local-prefixes: istio.io/
```

Input:
```go
package main

import (
	"fmt"
	"istio.io/istio/pkg/test/framework/resource"
	"os"

	"istio.io/istio/pkg/test/framework/resource"
)

var (
	_ = resource.Cluster
	_ = fmt.Println
	_ = os.Stdout
)
```

Output:
```go
package main

import (
	"fmt"
	"os"

	"istio.io/istio/pkg/test/framework/resource"
	"istio.io/istio/pkg/test/framework/resource"

	"istio.io/istio/pkg/test/framework/resource"
)

var (
	_ = resource.Cluster
	_ = fmt.Println
	_ = os.Stdout
)
```

Output seems nondeterministic

cc @daixiang0

howardjohn:
Input:

```go
import (
	"fmt"
	"os"

	"istio.io/istio/pkg/test/framework/resource"
	"istio.io/istio/pkg/test/framework/resource"
	"istio.io/istio/pkg/test/framework/components/prometheus"
)
```

Output:
```
import (
	"fmt"
	"os"

	"istio.io/istio/pkg/test/framework/components/prometheus"
)
```

howardjohn:
Also can only reproduce through golangci-lint, not direct calls on `goimports -local istio.io -w main.go; gci -local istio.io -w main.go`

ldez:
I try with different cases:

```go
import (
	"io/ioutil"
	"github.com/golangci/sandbox/pkg/render"
	"github.com/ghodss/yaml"
	"log"
	"encoding/json"

	"github.com/golangci/sandbox/pkg/router"
)
```

```go
import (
	"github.com/golangci/sandbox/pkg/render"
	"github.com/golangci/sandbox/pkg/router"
	"encoding/json"
	"io/ioutil"
	"log"
	"github.com/ghodss/yaml"
)
```

and all the linters that handle the imports (`gofmt`, `goimports`, `gci`, `gofumpt`).

Results:

| linters                           | results |
|-----------------------------------|---------|
| gofmt + goimports                 | FAIL    |
| gofmt + gci                       | FAIL    |
| goimports + gci                   | FAIL    |
| goimports + gofumpt               | FAIL    |
[...]

PR #2609, opened 2022-02-22
Support SuggestedFixes

SVilgelm:
Conver the SuggesterFixes to Replacement by:
1. applying the fixes to a file in memory
2. creating a diff by comparing with original file
3. using `ExtractIssuesFromPatch` function to create the issues

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -63,6 +63,7 @@ require (
 	github.com/nishanths/exhaustive v0.7.11
 	github.com/nishanths/predeclared v0.2.1
 	github.com/pkg/errors v0.9.1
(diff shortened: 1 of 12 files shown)


SVilgelm:
@ldez Ok, the PR is ready. I understand that the solution is not the best possible, but I still don't understand how to manually create a Replacement object for a TextEdit. This is kind of workaround 

PS: `gci` does not provide SuggestedFixes :(

nightlyone:
@SVilgelm 
> PS: `gci` does not provide SuggestedFixes :(

Is that fixed in PR #1490 ? At least a replacement is provided there, so maybe it can be improved from there on after merging?

PR #2618, opened 2022-02-24
Default to YAML when config file has no extension

donotnoot:
With this, it is possible to use process substitution so that there's no need to frequently download new rules.

```bash
golangci-lint run -c <(curl https://some-site.com/golangci-rules.yaml)
```


SVilgelm (approved):
LGTM

PR #2644, opened 2022-03-11
Add module-path setting to gofumpt

jpreese:
`gofumpt` v0.3.0 changed how imports are arranged, and added a `ModulePath` configuration option to control that behavior. Imports with the defined module path are no longer grouped with the standard library (https://github.com/mvdan/gofumpt/commit/487b40bce0289dc24fad95233b57828a4cd56943)

When using the `gofumpt` binary, it will actually exec out to the `go` binary in an attempt to get the module path from the `go.mod` file: https://github.com/mvdan/gofumpt/blob/master/gofmt.go#L289

This causes a conflict between the two formatters as the `gofumpt` binary will be clever and find the module path, but `golangci-lint` (currently) has no way of specifying this value.

### DIFF
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -511,6 +511,10 @@ linters-settings:
     # Default: 1.15
     lang-version: "1.17"
 
+    # Module path which contains the source code being formatted.
+    # Default: empty string
+    module-path: "github.com/golangci/golangci-lint"
+
     # Choose whether to use the extra rules.
     # Default: false
     extra-rules: true
--- a/.golangci.example.yml
+++ b/.golangci.example.yml
@@ -513,7 +513,7 @@ linters-settings:
 
     # Module path which contains the source code being formatted.
     # Default: empty string
-    module-path: "github.com/golangci/golangci-lint"
+    module-path: github.com/org/project
 
     # Choose whether to use the extra rules.
     # Default: false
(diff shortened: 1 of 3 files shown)


bombsimon (approved):
Thanks, LGTM

.golangci.example.yml
```diff
@@ -511,6 +511,10 @@ linters-settings:
     # Default: 1.15
     lang-version: "1.17"
 
+    # Module path which contains the source code being formatted.
+    # Default: empty string
+    module-path: "github.com/golangci/golangci-lint"
```
ldez, line 516:
```suggestion
    module-path: github.com/org/project
```

pkg/config/linters_settings.go
```diff
@@ -49,6 +49,7 @@ var defaultLintersSettings = LintersSettings{
 	},
 	Gofumpt: GofumptSettings{
 		LangVersion: "",
+		ModulePath:  "",
```
ldez, line 52:
```suggestion
```

ldez (changes requested):

pkg/config/linters_settings.go
bombsimon: I think we should be consistent, either we add defaults like `LangVersion` or we don't.

bombsimon: Why is `LangVersion` specified with the default value but we shouldn't specify `ModulePath`?

jpreese: I would have the same question. I'm not strongly opinionated on supplying the blank string or omitting the property entirely as long as it's consistent.

ldez: Sorry I was on my phone and the diff was a bit weird, you can keep it.

ldez (approved):

PR #2647, opened 2022-03-13
add GitLab CI example

madflow:


ldez:
Hello,

We don't plan to add a doc for all the existing CI, we documented the GitHub action because we handle the [action](https://github.com/golangci/golangci-lint-action)

https://golangci-lint.run/usage/install/#other-ci

PR #2651, opened 2022-03-16
Adding 'nopanic' linter

vladopajic:
Adding `nopanic` linter.
https://github.com/vladopajic/nopanic


ldez:
Hello,

I think that the scope of this linter is already handle by [forbidigo](https://golangci-lint.run/usage/linters/#forbidigo)

vladopajic:
Hello @ldez

The scope of this `nopanic` is to identify unexpected program termination found outside of `main` package. `forbidigo` can have arbitrary pattern specified, which can identify same issues reported by `nopanic`, but beauty of `nopanic` is that 1) it will not report issues in `main` package, 2) it is already configured to specific lint case.

sivchari:
Hello, I agree with @ldez . If panic is specified by forbidigo, it means the project wanna forbid panic syntax regardless of the outside main package.
I don't think this linter isn't good, but  I think this linter's usecase isn't common, so if someone wanna use this linter, it's better to download directly without golangci-lint.

vladopajic:
From the standpoint of `nopanic` linter, devs should always propagate error back to the caller - and this should be common usecase for all devs that are using `golangci-lint` to achieve A+ code grade. 

`nopanic` is little bit relaxed (by default, but configurable) where it assumes that forcibly termination of program is fine in main package because this is package where all errors are going to be propagated to.

PR #2658, opened 2022-03-19
docs: add installation via Cask section

axetroy:
https://github.com/axetroy/cask.rs

NOTE:

Cask does not need to do any additional work, such as create a PR when releasing a new version https://github.com/Homebrew/homebrew-core/pull/97241

Just need the `Cask.toml` file in the root of the repo. that's all.

### DIFF
--- a/Cask.toml
+++ b/Cask.toml
@@ -1,0 +1,19 @@
+[package]
+name = "github.com/golangci/golangci-lint"
+bin = "golangci-lint"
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

Your project seems to be a bit young, and there is already a lot of existing installation systems.

I'm not sure that we want to use your project for now.

I will decline, but I recommend opening an issue on this topic.

PR #2667, opened 2022-03-20
feat: add `exhaustruct` linter

xobotyi:
[This linter](https://github.com/GaijinEntertainment/go-exhaustruct) can be called a successor of [exhaustivestruct](https://github.com/mbilski/exhaustivestruct), and:

- it is at least **2.5+ times faster**, due to better algorithm;
- can receive `include` and/or `exclude` patterns;
- expects received patterns to be RegExp, therefore this package is not api-compatible with `exhaustivestruct`.

Also: deprecate `exhaustivestruct` linter

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -106,6 +106,8 @@ require (
 	mvdan.cc/unparam v0.0.0-20211214103731-d0ef000c54e5
 )
 
+require github.com/GaijinEntertainment/go-exhaustruct v1.0.0
+
 require (
 	github.com/Masterminds/semver v1.5.0 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Antonboom/nilnil v0.1.1
 	github.com/BurntSushi/toml v1.1.0
 	github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24
-	github.com/GaijinEntertainment/go-exhaustruct/v2 v2.0.2-0.20220410203528-eb118e90b8df
+	github.com/GaijinEntertainment/go-exhaustruct/v2 v2.1.0
 	github.com/OpenPeeDeeP/depguard v1.1.0
 	github.com/alexkohler/prealloc v1.0.0
 	github.com/ashanbrown/forbidigo v1.3.0
--- a/go.mod
+++ b/go.mod
@@ -8,6 +8,7 @@ require (
 	github.com/Antonboom/nilnil v0.1.1
 	github.com/BurntSushi/toml v1.1.0
 	github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24
+	github.com/GaijinEntertainment/go-exhaustruct v1.0.0
 	github.com/OpenPeeDeeP/depguard v1.1.0
 	github.com/alexkohler/prealloc v1.0.0
 	github.com/ashanbrown/forbidigo v1.3.0
@@ -105,8 +106,6 @@ require (
 	mvdan.cc/interfacer v0.0.0-20180901003855-c20040233aed
 	mvdan.cc/unparam v0.0.0-20211214103731-d0ef000c54e5
 )
-
-require github.com/GaijinEntertainment/go-exhaustruct v1.0.0
 
 require (
 	github.com/Masterminds/semver v1.5.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Antonboom/nilnil v0.1.1
 	github.com/BurntSushi/toml v1.1.0
 	github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24
-	github.com/GaijinEntertainment/go-exhaustruct v1.0.0
+	github.com/GaijinEntertainment/go-exhaustruct/v2 v2.0.1
 	github.com/OpenPeeDeeP/depguard v1.1.0
 	github.com/alexkohler/prealloc v1.0.0
 	github.com/ashanbrown/forbidigo v1.3.0
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/Antonboom/nilnil v0.1.1
 	github.com/BurntSushi/toml v1.1.0
 	github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24
-	github.com/GaijinEntertainment/go-exhaustruct/v2 v2.0.1
+	github.com/GaijinEntertainment/go-exhaustruct/v2 v2.0.2-0.20220410203528-eb118e90b8df
 	github.com/OpenPeeDeeP/depguard v1.1.0
 	github.com/alexkohler/prealloc v1.0.0
 	github.com/ashanbrown/forbidigo v1.3.0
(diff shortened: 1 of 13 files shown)


ldez:
Hello @xobotyi,

why didn't you do a PR on [exhaustivestruct](https://github.com/mbilski/exhaustivestruct) instead of creating a copy? 

I think that you will have a license issue.

xobotyi:
@idez pr is there for like.. half a year now.

Linter name and codebases are different (codebase, actually, is complete rewrite) why there should be a licensing issues?

ldez:
> Not even saying that exhaustivestruct is published under MIT licence, that does not forbid that.

The MIT license is clear:

> The above copyright notice and this permission notice shall be included in all
copies or substantial portions of the Software.

xobotyi:
@idez as said - the code is completely mine and have nothing in common (except purpose and usage of analyzer) with other linter.

ldez:
My login is LDEZ

xobotyi:
@ldez oh that's why siggestion didn't worked out😂
Visually, on mobile app it looks identical😅

ldez:
also, you don't need to ping me on every message.

ldez:
Can you rebase your PR? Thanks.

xobotyi:
Done

pkg/golinters/exhaustruct.go
```diff
@@ -0,0 +1,33 @@
+package golinters
+
+import (
+	"strings"
+
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/GaijinEntertainment/go-exhaustruct/pkg/analyzer"
```
ldez, line 8:
```suggestion
	"strings"
	
	"github.com/GaijinEntertainment/go-exhaustruct/pkg/analyzer"
	"golang.org/x/tools/go/analysis"
```

pkg/lint/lintersdb/manager.go
```diff
@@ -281,7 +283,14 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithSince("v1.32.0").
 			WithPresets(linter.PresetStyle, linter.PresetTest).
 			WithLoadForGoAnalysis().
-			WithURL("https://github.com/mbilski/exhaustivestruct"),
+			WithURL("https://github.com/mbilski/exhaustivestruct").
+			Deprecated("Owner seems to abandon linter.", "v1.45.0", "exhaustruct"),
+
+		linter.NewConfig(golinters.NewExhaustruct(exhaustructCfg)).
+			WithSince("v1.45.0").
```
ldez, line 290:
```suggestion
			WithSince("v1.46.0").
```

ldez (changes requested):

go.mod
```diff
@@ -24,6 +24,7 @@ require (
 	github.com/esimonov/ifshort v1.0.4
 	github.com/fatih/color v1.13.0
 	github.com/fzipp/gocyclo v0.5.0
+	github.com/GaijinEntertainment/go-exhaustruct v0.0.0-20220320194800-c73fb34375ae
```
ldez, line 27: Can you create a tag?
We use dependabot to update linters, so tags are required.

ldez (changes requested):

go.mod
xobotyi: done

ldez: you also have to update the PR :wink:

xobotyi: Yet not updatable since `go mod` says that there is no such version, so waiting for now🤷‍♂️

ldez: I updated the go.mod

xobotyi: Just updated.
should be fine now

ldez:
You don't need to squash, the PR is squashed during the merge.
It's better to keep the history of the PR changes.

xobotyi:
I dont squash, it is ammend (habit from gerrit - review tool that im using at work).

go.mod
ldez: FYI the module is available immediately you don't need to wait.
The fact to call the module adds the module to the Go proxy.

xobotyi: I had like `http 423: gone` during `go mod` execution, unable to recall exactly

ldez: have create several time the same tag?

pkg/golinters/exhaustruct.go
```diff
@@ -0,0 +1,32 @@
...
+
+func NewExhaustruct(settings *config.ExhaustructSettings) *goanalysis.Linter {
+	a := analyzer.Analyzer
+
+	var cfg map[string]map[string]interface{}
+	if settings != nil {
+		cfg = map[string]map[string]interface{}{
+			a.Name: {
+				"include": strings.Join(settings.Include, ","),
+				"exclude": strings.Join(settings.Exclude, ","),
```
ldez, line 21: I think there are problems here: `,` is something that can be used in a regular expression (ex: `\d{1,3}`).
So the parsing by the linter will produce unexpected behavior in this case.

<details><summary>parsing problem example</summary>

```yml
linters:
  disable-all: true
  enable:
    - exhaustruct

linters-settings:
  exhaustruct:
    include: ".*\\.ExhaustructCustom2{1,}$"
```

```console
$ ./golangci-lint run
$ 
```

```go

type ExhaustructCustom22 struct {
	A string
	B int
	c bool
	D float64
	E time.Time
}

func exhaustructCustom2() {
	_ = ExhaustructCustom22{
		A: "a",
		B: 0,
		c: false,
		D: 1.0,
		E: time.Now(),
	}

	_ = ExhaustructCustom22{
		A: "a",
		c: false,
		D: 1.0,
		E: time.Now(),
	}

	_ = ExhaustructCustom22{
		A: "a",
		c: false,
		E: time.Now(),
	}

	_ = ExhaustructCustom22{
		A: "a",
		B: 0,
		D: 1.0,
		E: time.Now(),
	}
}
```


</details>

ldez (changes requested):

pkg/golinters/exhaustruct.go
xobotyi: Oof. Haven't thought about quantifiers🤔
Maybe replace delimiter with semicolon, it is not used in regexp syntax, nor in golang (at least Im unable to recall)

ldez: I think it's better to not play with separators if you can.
You can forward the configuration as slice or struct.

xobotyi: Anyways I'll make a change later today

xobotyi: Updated linter and adopted to changes here.
Now regexps passed as-is, without concatenation.

ldez: Can you remove the panic?

xobotyi: Actually, I've already made such change, but yet away from wifi, I'll push it in 4-5 hours.
Yet only I haven't figured how to return an error dirung linter initialization (for the case provided strings are not valid regexp) 🤔

xobotyi: @ldez advices are appreciated =)

ldez: Sorry for the very late answer, time is always a problem.

I added a commit to handle the error, can you create a new tag for your linter?

xobotyi: Sure.

xobotyi: v2.1.0 is there

ldez: thank you!

ldez (approved):
LGTM

PR #2678, opened 2022-03-23
feat/upgrade-wrapcheck-config

guillaumeio:
This provides an update to the wrapcheck linter configuration to account for a change introduced here - https://github.com/tomarrell/wrapcheck/pull/25


ldez:
Hello @guillaumeio, we use dependenbot to update linters, and we apply changes during those updates.

I will decline the PR and let's the bot handle that.

guillaumeio:
Noted @ldez !

PR #2681, opened 2022-03-24
fix: help command

ldez:
The release process is blocked because of a bug with the help command.

Related to https://github.com/golangci/golangci-lint/runs/5674231408?check_suite_focus=true

```console
$ ./golangci-lint run -h                                                        
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x90 pc=0x127fbab]

goroutine 1 [running]:
github.com/golangci/golangci-lint/pkg/commands.NewExecutor({0x17101af, 0x7}, {0xc000162360, 0x16}, {0x14a3916, 0x9})
        github.com/golangci/golangci-lint/pkg/commands/executor.go:113 +0x68b
main.main()
        github.com/golangci/golangci-lint/cmd/golangci-lint/main.go:19 +0x47
```


ldez:
Without the missing step of the release, it's not possible to merge a PR because `check_generated` needs the generated files of the release.

So I will force the merge.

Issue #2649, opened 2022-03-16
Support of generics


ldez:
**Since the version v1.45.0 golangci-lint supports go1.18.**

But some linters don't work with generics, golangci-lint disable those linters automatically with go1.18.

Since v1.45.1, golangci-lint can detect the Go version used by a project.

To get the Go version golangci-lint will use, in order:
- the Go version defined by the CLI flag, ex: `--go=1.18`
- the Go version defined in the configuration file, ex:
  ```yml
  run:
    go: '1.18'
  ```
- the Go version defined in the `go.mod`, ex;
  ```go
  module github.com/org/my-project

  go 1.18
  ```
- Go version defined by the env var `GOVERSION`
- fallback on go1.17.

Notes:
- some rules in go-critic (`externalErrorReassign`) don't work with generics and must be disabled by hand
- if you are not using generics (and your dependencies) you can use the following configuration:
  ```yml
  run:
    go: '1.17'
  ```

---

The keyword `any` works well and produces no error (it's expected because it's just a type alias)

---

The problem with [SSA](https://github.com/golang/tools/tree/master/go/ssa) is now fixed but some linters need to update their code base to handle generics.

About the linters:
- for `go-critic`, there is an existing issue https://github.com/go-critic/go-critic/issues/1193

---

Linter issues: (checked if the problems are solved)
- [x] [govet(nilness,unusedwrite)](https://github.com/golang/go/issues/48525)
- [x] [staticcheck, stylecheck, gosimple, unused](https://github.com/dominikh/go-tools/releases/tag/2022.1)
- [x] [bodyclose](https://github.com/timakin/bodyclose/issues/38)
- [x] [contextcheck](https://github.com/sylvia7788/contextcheck/issues/4)
- [x] [nilerr](https://github.com/gostaticanalysis/nilerr/issues/11)
- [x] [noctx](https://github.com/sonatard/noctx/issues/11)
- [x] [rowserrcheck](https://github.com/jingyugao/rowserrcheck/issues/24)
- [x] [sqlclosecheck](https://github.com/ryanrolds/sqlclosecheck/issues/18)
- [x] [tparallel](https://github.com/moricho/tparallel/issues/13)
- [x] [wastedassign](https://github.com/sanposhiho/wastedassign/issues/41)
- [x] [unparam](https://github.com/mvdan/unparam)
- [x] ~~struccheck~~: Deprecated
[...]

PR #2705, opened 2022-03-30
docs: update FAQ page.

ldez:
Updates supported Go versions section.

### DIFF
--- a/docs/src/docs/usage/faq.mdx
+++ b/docs/src/docs/usage/faq.mdx
@@ -22,7 +22,7 @@ See [how to properly install `golangci-lint` in CI](/usage/install#ci-installati
 
 The same as the Go team (the 2 last minor versions)
 
-go1.18+ is officially supported since golangci-lint v1.45.0.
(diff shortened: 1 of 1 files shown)


docs/src/docs/usage/faq.mdx
```diff
@@ -18,17 +18,11 @@ Run `golangci-lint` in CI and check the exit code. If it's non-zero - fail the b
...
+The same as the Go team (the 2 last minor versions)
 
-Long answer:
-
-1. go < 1.9 isn't supported
-2. go1.9 is officially supported by golangci-lint <= v1.10.2
-3. go1.10 is officially supported by golangci-lint <= 1.15.0.
-4. go1.11 is officially supported by golangci-lint <= 1.17.1.
-5. go1.12+ are officially supported by the latest version of golangci-lint (>= 1.18.0).
+go1.18+ is officially supported since golangci-lint v1.45.0.
```
bombsimon, line 25: I think it's worth mentioning that even though Go 1.18 is supported from `v1.45.0` not every linter in `golangci-lint` will work. Just like we did with the error message I think it's a good idea to link to https://github.com/golangci/golangci-lint/issues/2649 as long as the issue is still open.

bombsimon (approved):
LGTM!

PR #2743, opened 2022-04-03
dev: remove some TODO

ldez:
- use go1.18 to test the previous version of golangci-lint
- update gomnd configuration

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -33,13 +33,19 @@ linters-settings:
   goimports:
     local-prefixes: github.com/golangci/golangci-lint
   gomnd:
-    # TODO(ldez) must be rewritten after the v1.44.0 release.
-    settings:
-      mnd:
-        # don't include the "operation" and "assign"
-        checks: argument,case,condition,return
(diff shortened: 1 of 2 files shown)


.golangci.yml
```diff
@@ -33,13 +33,19 @@ linters-settings:
...
+    checks:
+      - argument
+      - case
+      - condition
+      - return
+    ignored-numbers:
+      - '0'
+      - '1'
+      - '2'
+      - '3'
```
bombsimon, line 46: Must say a bit confusing that a list of numbers are treated as strings. (Don't mean this PR, saw this is how it's handled upstream but didn't know).

bombsimon (approved):
LGTM

.golangci.yml
ldez: Yes I know, I documented that https://golangci-lint.run/usage/linters/#gomnd.

ldez: FYI it's for the compatibility with the old configuration.

PR #2744, opened 2022-04-03
dev: support build tags about go version in linter tests

ldez:
Skip linter tests about go1.18 when using go1.17.

Allows using go1.17 and go1.18 in CI.

### DIFF
--- a/test/fix_test.go
+++ b/test/fix_test.go
@@ -49,6 +49,11 @@ func TestFix(t *testing.T) {
 				input,
 			}
 			rc := extractRunContextFromComments(t, input)
+			if rc == nil {
+				t.Logf("Skipped: %s", input)
+				return
+			}
+
 			args = append(args, rc.args...)
 
 			cfg, err := yaml.Marshal(rc.config)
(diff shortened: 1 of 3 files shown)


bombsimon (commented):
Nice!

Just to clarify, the idea is that the upstream dependencies that require Go 1.18 should add this build tag so we can bump their dependencies and still build in CI when we check compatibility with Go 1.17?

Do we need a way to communicate this? It feels very tribal knowledge if it's not documented, maybe it can be added under the [How to update existing linter](https://golangci-lint.run/contributing/faq/#how-to-update-existing-linter) section?

ldez:
> Just to clarify, the idea is that the upstream dependencies that require Go 1.18 should add this build tag so we can bump their dependencies and still build in CI when we check compatibility with Go 1.17?

No, it's just a way to write tests with go1.18 syntax, it's doesn't change linter constraints.

bombsimon:
> No, it's just a way to write tests inside golangci-lint with go1.18 syntax, it doesn't change linter constraints.

Oh, my bad, didn't read properly. 🤦  That makes more sense!

bombsimon (approved):

Issue #2784, opened 2022-04-26
go1.18.1 + some generics code + gocritic = panic (even if the directory/file is excluded)


lopezator:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [X] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

### Description of the problem

With go1.18.1, some generics code and gocritic enabled, golanci-lint panics like this, even if that directory/file is excluded via `--skip-dirs` and/or `--skip-files`: 

```
ERRO [runner] Panic: gocritic: package "slice" (isInitialPkg: true, needAnalyzeSource: true): assertion failed: goroutine 527 [running]:
runtime/debug.Stack()
....
ERRO Running error: 1 error occurred:
        * can't run linter gocritic: gocritic: gocritic: package "slice" (isInitialPkg: true, needAnalyzeSource: true): assertion failed
```

Repro repo here:

https://github.com/lopezator/gocritic-118

Related with:

https://github.com/golangci/golangci-lint/issues/2649

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version v1.45.2 built from (unknown, mod sum: "h1:9I3PzkvscJkFAQpTQi5Ga0V4qWdJERajX1UZ7QqkW+I=") on (unknown)
```

</details>

### Configuration file

<details>

```console
$ cat .golangci.yml
linters:
  enable:
    - gocritic
  disable-all: true

linters-settings:
  gocritic:
    enabled-tags:
      - diagnostic
      - experimental
      - opinionated
      - performance
      - style
    disabled-checks:
      - dupImport # https://github.com/go-critic/go-critic/issues/845
      - ifElseChain
      - octalLiteral
      - whyNoLint
      - wrapperFunc
```

</details>

### Go environment

<details>

```console
$ go version && go env
go version go1.18.1 linux/amd64
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/myhome/.cache/go-build"
GOENV="/home/myhome/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
[...]

PR #2785, opened 2022-04-26
pkg/lint: added cleanSkipped func

lopezator:
The cleanSkiped func cleans the files and directories passed
via `skipDirs` and `skipFiles` for the pkgs slice. This way,
we could skip analyzing them. This way, it's easy for the user
to avoid panics/errs when analyzing excluded files and/or
folders.

Closes https://github.com/golangci/golangci-lint/issues/2784

### DIFF
--- a/pkg/lint/runner.go
+++ b/pkg/lint/runner.go
@@ -30,7 +30,7 @@ type Runner struct {
 
 func NewRunner(cfg *config.Config, log logutils.Log, goenv *goutil.Env, es *lintersdb.EnabledSet,
 	lineCache *fsutils.LineCache, dbManager *lintersdb.Manager, pkgs []*gopackages.Package) (*Runner, error) {
-	skipFilesProcessor, err := processors.NewSkipFiles(cfg.Run.SkipFiles)
+	skipFilesProcessor, err := processors.NewSkipFiles(cfg.Run.SkipFiles, pkgs)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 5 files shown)


pkg/lint/load.go
```diff
@@ -283,6 +283,35 @@ func (cl *ContextLoader) filterDuplicatePackages(pkgs []*packages.Package) []*pa
...
+				pkgs = append(pkgs[:i], pkgs[i+1:]...)
+			}
+		}
+	}
+
+	// Clean skipFiles from pkgs.
+	filePath, err := os.Getwd()
+	if err != nil {
+		return nil, errors.New("failed to get wd")
+	}
```
lopezator, line 300: I suspect that this code block isn't compatible with `windows`. 

Because of this:

`skip-dirs` and `skip-files` work well with linux.
`skip-dirs` works will windows but `skip-files` doesn't.

I will research and find a full cross-os solution.

CC/ @ldez

lopezator:
My main doubts are:

* Do we need to analyze excluded files and folders for some reason?
* Could this skip improve the overall performance with skipping large files/directories?

ldez:
> Do we need to analyze excluded files and folders for some reason?

Yes, because the code must compile, so we must have all the files.

> Could this fix improve the overall performance when skipping large files/directories

No, for the same reason as the previous question.

lopezator:
> Yes, because the code must compile, so we must have all the files.

Maybe I am missing something, but this shouldn't this be `go build`'s job? 

I mean, you shouldn't be relying on `golangci-lint` to check if your skipped folders code compiles or not. 

I have seen that you added the `enhancement` label. 

Would you be open to include this change?

lopezator:
I am on the `golanci-lint` maintainers group, feel free to reach me via `gophers slack` if you want to discuss this further.

Thank you for the quick feedback @ldez !

ldez:
This PR cannot work as I explained https://github.com/golangci/golangci-lint/pull/2785#issuecomment-1109998418

To be analyzed the code need to compile, by "compile" I mean all the symbols need to be present.
There 2 types of analysis "fast" and "slow", the "slow" requires getting all symbols.
80% of the linters are "slow".

I will declined this PR.

PR #2790, opened 2022-05-01
Add typenaming linter

armantarkhanian:
typenaming linter will prevent gophers from using the ugly "Type" suffix in type names.

_Bad_
```golang
type UserType struct {}
```

_Good_
```golang
type User struct {}
```
Github repo: https://github.com/typenaming/typenaming

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -92,6 +92,7 @@ require (
 	github.com/timakin/bodyclose v0.0.0-20210704033933-f49887972144
 	github.com/tomarrell/wrapcheck/v2 v2.6.0
 	github.com/tommy-muehle/go-mnd/v2 v2.5.0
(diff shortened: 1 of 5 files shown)


ldez:
Hello,

sorry but I don't understand this linter: you can have a `User` and a `UserType` it's 2 different notions.

I haven't seen projects that suffix struct name by `Type`.

I recommend using your linter, which seems to be a personal use case, as a [plugin](https://golangci-lint.run/usage/linters/#custom)

You can also use [forbidigo](https://golangci-lint.run/usage/linters/#forbidigo)

```yml
linters-settings:

  forbidigo:
    forbid:
      - '.+Type'
```

So I will decline your PR.

Issue #1792, opened 2021-02-27
staticcheck randomly fails on CI


SVilgelm:
- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).
- [ ] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

<summary>Verbose output of running</summary>

```console
=== CONT  TestSourcesFromTestdataWithIssuesDir/staticcheck.go
    linters_test.go:157: [run --allow-parallel-runners --disable-all --print-issued-lines=false --out-format=line-number --max-same-issues=100 -Estaticcheck --no-config testdata/staticcheck.go]
    linters_test.go:37: 
        	Error Trace:	linters_test.go:37
        	            				linters_test.go:158
        	            				linters_test.go:57
        	Error:      	Received unexpected error:
        	            	
        	            	unmatched errors
        	            	level=warning msg="[linters context/goanalysis] Failed to get persisted facts: failed to get data from low-level cache by key fact_deprecated/facts for package fmtsort: read 0/175 bytes from /home/runner/.cache/golangci-lint/ed/ed3eaffb7621d0470f01c721ca356e360b48c69fbe6c0adf4fedb7c1b1c4f83f-a with error EOF"
        	Test:       	TestSourcesFromTestdataWithIssuesDir/staticcheck.go

...

--- FAIL: TestSourcesFromTestdataWithIssuesDir (0.00s)
             --- FAIL: TestSourcesFromTestdataWithIssuesDir/staticcheck.go (3.47s)
```

ldez:
Another case:

```
=== CONT  TestSourcesFromTestdataWithIssuesDir/staticcheck.go
    linters_test.go:36: 
        	Error Trace:	linters_test.go:36
        	            				linters_test.go:157
        	            				linters_test.go:56
        	Error:      	Received unexpected error:
        	            	staticcheck.go:26: missing error "SA1019: runtime.CPUProfile has been deprecated .*"
        	Test:       	TestSourcesFromTestdataWithIssuesDir/staticcheck.go
```

kamilsk:
most recent

- https://github.com/golangci/golangci-lint/runs/6253254323?check_suite_focus=true
- https://github.com/golangci/golangci-lint/runs/6253260233?check_suite_focus=true

```
[...]

PR #2823, opened 2022-05-04
exclude docs from unrelated go tests on ci

kamilsk:
running tests for such changes is redundant

examples to exclude
- https://github.com/golangci/golangci-lint/pull/2803
- https://github.com/golangci/golangci-lint/pull/2802
- https://github.com/golangci/golangci-lint/pull/2798

### DIFF
--- a/.github/workflows/pr-extra.yml
+++ b/.github/workflows/pr-extra.yml
@@ -4,6 +4,8 @@ on:
     branches:
       - master
   pull_request:
+    paths:
+      - '!docs/*'
 
 jobs:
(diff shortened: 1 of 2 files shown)


kamilsk:
related comment https://github.com/golangci/golangci-lint/issues/1792#issuecomment-1114559970

ldez:
> why do you need to run unrelated checks?

Because technically it's not possible to define conditional requirements/checks.
In Github, you can just add or remove a check.
All checks are required on basic PRs.

PR #2829, opened 2022-05-06
docs: update "Shell Completion" instruction on Linux

JohnTitor:
When I tried to enable shell autocompletion reading the website, noticed the content on kubernetes.io had been moved. After some playing, I also noticed the `golangci-lint completion $SHELL --help` command exists and it has a straightforward instruction.
This replaces that link with the command above. I'm open to update the link to https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/#enable-shell-autocompletion, which is the new location of kubectl's instruction, if you'd like.

(I haven't removed the macOS's instruction in case it's still useful.)

### DIFF
--- a/docs/src/docs/usage/integrations.mdx
+++ b/docs/src/docs/usage/integrations.mdx
@@ -33,7 +33,7 @@ title: Integrations
 
 ## Shell Completion
 
-`golangci-lint` can generate bash completion file.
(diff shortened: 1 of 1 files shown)


JohnTitor:
(The failure seems unrelated.)

ldez (approved):
LGTM

PR #2857, opened 2022-05-15
docs: rename .golangci.example.yml to .golangci.reference.yml

ldez:
The name of this file leads many people to think that this file is a configuration example.

This file is a configuration reference: it contains the exhaustive configuration with explanations of the options.

This is not a configuration example.


bombsimon (approved):
💯

PR #2875, opened 2022-05-22
bump golang.org/x/tools to HEAD

ldez:
https://github.com/golang/tools/compare/77aa08bb151a...904e24e9fcf9e04175d563e5112b8264262acbb0

The analyzer validation has been changed, the `Run` method is mandatory.

I created a `DummyRun` function in order to pass the validation.

For some linters, the `Run` method was defined in `WithContextSetter` (after the validation).
So I, when possible, move the creation of the `Run` method inside the creation of the analyzer.

I also did work homogenizing the linters.


kamilsk:
massive refactoring 👏 challenging to review it, I might have missed something important

PR #2884, opened 2022-05-27
docs(depguard): add `ignore-file-rules`

maratori:
Add missing `ignore-file-rules` config for `depguard` to `.golangci.reference.yml`.

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -182,6 +182,7 @@ linters-settings:
     # This allows for more precise control, but it is only available for glob patterns.
     # Default: []
     ignore-file-rules:
+      - "ignore/**/*.go"
       - "!**/*_test.go"
 
     # Create additional guards that follow the same configuration pattern.
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -163,15 +163,26 @@ linters-settings:
     include-go-root: true
 
     # A list of packages for the list type specified.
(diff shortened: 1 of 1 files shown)


.golangci.reference.yml
```diff
@@ -163,16 +163,27 @@ linters-settings:
...
       - github.com/sirupsen/logrus: 'logging is allowed only by logutils.Log'
 
+    # Specify rules by which the linter ignores certain files for consideration.
+    # Can accept both string prefixes and string glob patterns.
+    # The ! character in front of the rule is a special character
+    # which signals that the linter should negate the rule.
+    # This allows for more precise control, but it is only available for glob patterns.
+    # Default: []
+    ignore-file-rules:
+      - "!**/*_test.go"
```
ldez, line 185: can you add another example without `!`

ldez (changes requested):

.golangci.reference.yml
maratori: Done

maratori: @ldez Are you going to use Squash Merge? If not, I can squash two commits to have clear history.

I've made changes in a separate commit to make review easier.

ldez: We will squash, you don't need to handle that.

ldez (approved):
LGTM

PR #2887, opened 2022-05-28
docs(thelper): add `fuzz` config and description

maratori:
Add missing `fuzz` config of `thelper` to `.golangci.reference.yml`

See:
https://github.com/golangci/golangci-lint/blob/d92f144dffd4007253b4bbce4483da6178dd39db/pkg/config/linters_settings.go#L554


ldez (approved):
LGTM

PR #2915, opened 2022-06-12
nonamedreturns: add setting allow-error-in-defer

maratori:
TODO: change commit hash to version after new release

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -1152,6 +1152,11 @@ linters-settings:
     # Default: false
     require-specific: true
 
+  nonamedreturns:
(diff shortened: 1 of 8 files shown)


ldez:
The updates of linter are done by dependabot, we add the configuration inside the PR of the bot.

maratori:
@ldez may I ask you to manually trigger dependabot? New version is released https://github.com/firefart/nonamedreturns/releases/tag/v1.0.2

ldez:
I can't trigger that: the dependabot updates the Go dependencies every Sunday.
I don't think we will create a release this week, there is no emergency.

PR #2921, opened 2022-06-14
depguard: adjust phrasing

cory-miller:
Replaces `whitelist` and `blacklist` with `allowlist` and `blacklist` for depguard


ldez (approved):
LGTM

PR #2964, opened 2022-07-08
Disable vcs stamping in go1.18

malteehrlen:
This new go feature breaks `go list ./...` type commands that run in e.g. monorepos with several go modules but just one git repository. The fix is to run the command explicitly disabling vcs stamping with the flag `-buildvcs=false`. Since this wont have any effect on the output of the command it is fine to do it indiscriminately. 

Alternatively, a way of providing `go list` flags via the config yaml could be provided, but I dont think that is necessary.

Cheers!

### DIFF
--- a/pkg/lint/load.go
+++ b/pkg/lint/load.go
@@ -75,6 +75,11 @@ func (cl *ContextLoader) findLoadMode(linters []*linter.Config) packages.LoadMod
 
 func (cl *ContextLoader) buildArgs() []string {
 	args := cl.cfg.Run.Args
+	// disable vcs stamping to avoid git exit status 128 errors
+	if cl.cfg.Run.Go == "1.18" && len(args) == 0 {
+		return []string{"-buildvcs=false", "./..."}
+	}
+
 	if len(args) == 0 {
 		return []string{"./..."}
 	}


pkg/lint/load.go
```diff
@@ -75,6 +75,11 @@ func (cl *ContextLoader) findLoadMode(linters []*linter.Config) packages.LoadMod
 
 func (cl *ContextLoader) buildArgs() []string {
 	args := cl.cfg.Run.Args
+	// disable vcs stamping to avoid git exit status 128 errors
+	if cl.cfg.Run.Go == "1.18" && len(args) == 0 {
```
kamilsk, line 79: what about 1.19, 1.20, etc?

malte-ehrlen: good question, maybe it makes sense to take into account unreleased versions, but how far?

malte-ehrlen:
I just realized after playing around a bit that having a properly defined `go.work` file will also solve this issue, and that is probably the better solution. The `go.work` file is also something new with go 1.18 so I will test it a bit more but if it works well I will close this pr

pkg/lint/load.go
kamilsk: No, I just highlighted the fragility of the changes. Is it possible to check "feature" in runtime?

> This new go feature breaks `go list ./...`

Could someone provide details? Maybe a bug in [golang/go](https://github.com/golang/go) or a reference to the docs?

kamilsk (changes requested):
current changes are fragile, let's wait for https://github.com/golangci/golangci-lint/pull/2964#issuecomment-1179019308

PR #2970, opened 2022-07-10
feat: change err to nil

sashamelentyev:


ldez:
Without any description, it's not possible to understand.

But the error of the walk needs to be returned.

sashamelentyev:
@ldez, all code looks like this:
```go
func dirSizeBytes(path string) (int64, error) {
	var size int64
	err := filepath.Walk(path, func(_ string, info os.FileInfo, err error) error {
		if err != nil {
			return err
		}
		if !info.IsDir() {
			size += info.Size()
		}
		return nil
	})
	return size, err
}
```
we check err in first step, don't need return err in last line in filepath.Walk

kamilsk:
`feat:` is incorrect here. I suggest avoiding a prefix or using`fix:`/`refactor:`

https://www.conventionalcommits.org/en/v1.0.0/

kamilsk:
```go
		if err != nil {
			return err
		}
		if !info.IsDir() {
			size += info.Size()
		}
		return err
```

and

```go
		if err != nil {
			return err
		}
		if !info.IsDir() {
			size += info.Size()
		}
		return nil
```

are both unclear, but the second is just "optimization" of the first, nothing else

```go
		if err == nil && !info.IsDir() {
			size += info.Size()
		}
		return err
```

looks the same but holds

> the error of the walk needs to be returned

and reduces code lines

kamilsk:
https://github.com/golangci/golangci-lint/pull/2970#issuecomment-1179761123

I'm talking about the commit message

image

sashamelentyev:
@kamilsk i rename commit, but pr is closed

PR #2972, opened 2022-07-11
Update source code link

cnmade:
For new user, who want quickly check the source code


bombsimon:
There's already a link in the sidebar under `Product` that links to GitHub, you don't think this is enough? 
image

cnmade:
image
The important area , why not put source repository link here.

To developers, source code important than twitter, so what you think about it.

ldez:
The menu is enough for me.

ldez:
We agree, so we will decline the PR.

PR #2993, opened 2022-07-19
feat: sort the issue before print

golangaccount:
i use golangci to check the project code, in some file, more then 2 linter report issue so before print the issues, sort the issues make sure same file issues in same area


ldez:
Hello,

the option `--sort-results` already exists.

So I will decline your PR.

Issue #2997, opened 2022-07-19
v1.47.1 much slower


fredbi:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [X] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

### Description of the problem

Today, our CI pipelines moved to v1.47.1 and we immediately felt the pain:
linting was suddenly taking about 3x longer, exhausting the resources of the CI VM (2 vCPU, 4GB RAM), even with a very small amount of changes and the "--new-from-rev master" flag. This behavior was consistent and observed on a dozen PRs today, even those with no change in the go code base.
run is slower when compared to v1.46.2, which was our previous version.

Running v1.47.1 from docker image: golangci/golangci-lint:latest-alpine
Extra installations required for alpine:
apk add git openssh-client bash ncurses gcc musl-dev

CLI for linting:
 golangci-lint run --timeout 10m --new-from-rev master --build-tags integration

### Version of golangci-lint

golangci-lint version running on the CI container.

<details>

```console
$ golangci-lint --version
golangci-lint has version v1.47.1 built from ebd6dcbf on 2022-07-19T11:57:02Z
```

</details>

### Configuration file

<details>

```console
$ cat .golangci.yml
linters-settings:
  govet:
    check-shadowing: true
  gocyclo:
    min-complexity: 25
  maligned:
    suggest-new: true
  dupl:
    threshold: 150
  goconst:
    min-len: 3
    min-occurrences: 2
  forbidigo:
    forbid:
      - ^print.*$
      - 'fmt\.Print.*'
  funlen:
    lines: 100
    statements: 50
  gocognit:
    min-complexity: 80 # This is super high value. We should gradually lower it to 30-40.
run:
  skip-dirs:
    - integrationtest

linters:
  enable:
    - gofmt
    - goimports
    - bodyclose
    - dupl
    - forbidigo
    - gocognit
    - gocritic
    - goimports
    - gosec
    - nakedret
    - nolintlint
    - revive
    - stylecheck
    - unconvert
    - unparam
  disable:
    - maligned
    - lll
    - gochecknoinits
    - gochecknoglobals

```

</details>

### Go environment

go version running on the CI container.

<details>

```console
$ go version && go env
go version go1.18.4 linux/amd64
[...]

PR #2999, opened 2022-07-20
revive: ignore slow rules

ldez:
Related to #2997

https://github.com/golangci/golangci-lint/issues/2997#issuecomment-1190681977

### DIFF
--- a/pkg/golinters/revive.go
+++ b/pkg/golinters/revive.go
@@ -7,6 +7,7 @@ import (
 	"go/token"
 	"os"
 	"reflect"
+	"strings"
 	"sync"
 
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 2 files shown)


pkg/golinters/revive.go
```diff
@@ -388,3 +389,23 @@ func defaultConfig() *lint.Config {
...
+		"time-naming",
+		"unexported-return",
+		"unhandled-error",
+		"var-declaration",
+	}
+
+	for _, s := range f {
+		delete(conf.Rules, s)
+	}
+}
```
bombsimon, line 411: Maybe worth adding a log if the rule is present in the config so the user gets notified it's being disabled instead of just silently dropping it?

bombsimon (approved):
LGTM, added a suggestion on notifying the user. I would probably have grouped all the exluded rules to give a better overview and not have to repeat the same commet so much but that's personal preference and nitpicking.

pkg/golinters/revive.go
ldez: PTAL

bombsimon (approved):
💯

KarthikNayak:
Thanks for this, is there a release scheduled for this?

PR #3002, opened 2022-07-21
nolint: drop allow-leading-space option and add "nolint:all"

ldez:
It's a part of the road to go1.19.

https://github.com/golangci/golangci-lint/issues/1658#issuecomment-1183148066

Drop `allow-leading-space` option and add the support of `nolint:all`.

This PR doesn't break the previous syntaxes.

It will not be the real fix for issue #1658 but a fix for go1.19 gofmt.

In parallel, I started to work on a real fix for the issue with a new syntax.

### DIFF
--- a/pkg/result/processors/nolint.go
+++ b/pkg/result/processors/nolint.go
@@ -252,7 +252,7 @@ func (p *Nolint) extractInlineRangeFromComment(text string, g ast.Node, fset *to
 		}
 	}
 
-	if !strings.HasPrefix(text, "nolint:") {
+	if strings.HasPrefix(text, "nolint:all") || !strings.HasPrefix(text, "nolint:") {
 		return buildRange(nil) // ignore all linters
 	}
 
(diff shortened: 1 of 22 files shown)


pkg/result/processors/nolint.go
```diff
@@ -252,16 +252,19 @@ func (p *Nolint) extractInlineRangeFromComment(text string, g ast.Node, fset *to
...
 	// ignore specific linters
 	var linters []string
 	text = strings.Split(text, "//")[0] // allow another comment after this comment
 	linterItems := strings.Split(strings.TrimPrefix(text, "nolint:"), ",")
-	for _, linter := range linterItems {
-		linterName := strings.ToLower(strings.TrimSpace(linter))
+	for _, item := range linterItems {
+		linterName := strings.ToLower(strings.TrimSpace(item))
+		if linterName == "all" {
+			panic("OOPS")
```
kamilsk, line 266: Maybe a more explicit message would be better? Is OOPS an acronym, or is an exclamation made in response to a minor mistake?

kamilsk (approved):

pkg/result/processors/nolint.go
kamilsk: for example

> incorrect usage of "all" keyword, please use it only with "nolint:all"

ldez: It's just a test residue

ldez:
As we created a release recently, I will wait a bit before merging this PR.

PR #3006, opened 2022-07-22
fix(cli): do not offer completions for commands not taking args

scop:


ldez:
can you add a description?

ldez:
I will cherry-pick this PR into the other one.

PR #3007, opened 2022-07-22
dev: clean commands

scop:
For commands consistency.

EDIT(ldez): I rewrote some commands to be more idiomatic and consistent.

### DIFF
--- a/pkg/commands/version.go
+++ b/pkg/commands/version.go
@@ -33,7 +33,10 @@ func (e *Executor) initVersion() {
 	versionCmd := &cobra.Command{
 		Use:   "version",
 		Short: "Version",
-		RunE: func(cmd *cobra.Command, _ []string) error {
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				e.log.Fatalf("Usage: golangci-lint version")
+			}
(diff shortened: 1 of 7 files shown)


pkg/commands/version.go
```diff
@@ -33,7 +33,10 @@ func (e *Executor) initVersion() {
 	versionCmd := &cobra.Command{
 		Use:   "version",
 		Short: "Version",
-		RunE: func(cmd *cobra.Command, _ []string) error {
+		RunE: func(cmd *cobra.Command, args []string) error {
+			if len(args) != 0 {
+				e.log.Fatalf("Usage: golangci-lint version")
```
kamilsk, line 38: strange pattern https://github.com/golangci/golangci-lint/search?q=len%28args%29+%21%3D+0. why not `Args: cobra.NoArgs,` ?

kamilsk (approved):

pkg/commands/version.go
ldez: You are right, I rewrote all the commands.

ldez (approved):
LGTM.
I took the opportunity to rewrite a bit the commands

pkg/commands/version.go
kamilsk: awesome!!!

PR #3027, opened 2022-07-30
feat: add noloopclosure

fatanugraha:
noloopclosure is a linter that disallow reference capture of loop variable inside of a closure.

https://github.com/fatanugraha/noloopclosure


ldez:

### Pull Request Description

### Linter

### `.golangci.reference.yml`

### Others Requirements

### Recommendations

---

ldez:
Hello,

I feel like a duplicate of [exportloopref](https://golangci-lint.run/usage/linters/) or [looppointer](https://github.com/golangci/golangci-lint/pull/1924)

fatanugraha:
Hi @ldez I checked both linters already and it seems that those linters only check whether the loop variable is captured using `&` operator or not.

This new linter checks whether the loop variable is captured using closure or not (not necessarily a pointer) -- to avoid accidental reference because of closures.

For example, this linter will complain that `i` is a loop variable in L3

```
for i := 0; i < 5; i++ {
  go (func() { fmt.Println(i) })() // L3
}
```

I tried both linters that you mentioned for code snippet above and it passes.

ldez:
We refused looppointer because of the number of false-positives, you will have to prove that don't have the problem as looppointer

fatanugraha:
@ldez this linter is way simpler than `looppointer`.

What it does is basically:
1. check all loops
2. check whether the loop has `*ast.FuncLit` on it
3. check the object (`types.Object`) of all `*ast.Ident` on the body of that `*ast.FuncLit` and see whether it's the same as the enclosing loop variable object.

The implementation is basically the striped-down version of `go vet`'s `loopclosure`.

fatanugraha:
> We refused looppointer because of the number of false-positives, you will have to prove that don't have the same problem as looppointer

Hi @ldez what kind of proof do you require for this one?

fatanugraha:
After reading this similar [issue](https://github.com/golangci/golangci-lint/issues/1404#issuecomment-806786984) more thoroughly, I think this linter will have the same issue that `looppointer` have.

These two linters will reject a valid Go code that **might or might not cause a bug**. The idea of these two linter is to disallow a certain practices that might introduce [the bug](https://go.dev/doc/faq#closures_and_goroutines) in the first place. So I wouldn't call them a false-positive (from the linter's perspective) since that's the whole point of these two linters.

Personally I think this linter will complement `looppointer` nicely to make sure that  user's code is safe when they're migrating to Go 2 (if the Go team decided to go with this [proposal](https://github.com/golang/go/discussions/56010)) so I think these two linters can still provide some value to golangci-lint users (even though they're very opinionated).

@hawkingrei Having said all of that, I don't think this linter will be merged to master anytime soon since it cause issues to users that enables all the linter via `enable-all` flag. The best that we can do is to run them without `golangci-lint`.

@ldez Feel free to close this issue if you think we shouldn't merge this to master. Thanks!

ldez:
> @ldez Feel free to close this issue if you think we shouldn't merge this to master. Thanks!

Closed for now

PR #3039, opened 2022-08-01
docs(install): remove outdated info on source install

scop:
v1.47.3 no longer builds with 1.15 or 1.16, so remove related install instructions.


ldez (approved):
LGTM

PR #3049, opened 2022-08-03
feat: add `explicitreturn` linter

maxsond:
Source: https://tildegit.org/indigo/explicitreturn

This adds a linter that requires functions to have explicit return statements. It's broadly similar to `nakedret`, but with a slightly different philosophy. 

For one thing, the size of the code doesn't matter. This will catch inline functions which I missed when I used `nakedret` (e.g. `var x` in my test). 

As well, I would not consider https://github.com/alexkohler/nakedret/issues/18 to be a bug because the function in that issue returns an error and a nil error is still a nil value so I would want to see that stated explicitly. The author of `nakedret` doesn't seem to take that view which is fine, but was part of what pushed me to writing this separately rather than working directly on `nakedret` as well as a PR that's been open for review on that project for quite a while now.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -107,6 +107,7 @@ require (
 	mvdan.cc/gofumpt v0.3.1
 	mvdan.cc/interfacer v0.0.0-20180901003855-c20040233aed
 	mvdan.cc/unparam v0.0.0-20220706161116-678bad134442
+	tildegit.org/indigo/explicitreturn v0.0.0-20220802220915-02f24a7ea1f1
 )
 
 require (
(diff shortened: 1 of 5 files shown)


ldez:

### Pull Request Description

### Linter

### `.golangci.reference.yml`

### Others Requirements

### Recommendations

---

maratori:
@maxsond Why the linter doesn't report about the first return?
```go
func MultipleReturnPaths() (x int, y string) {
	if true {
		return // ok ?
	}
	return // want `implicit return found in function MultipleReturnPaths`
}
```

maxsond:
> Why the linter doesn't report about the first return?

This is a bug. The current implementation only looks at statements directly inside the function statement, not at statements inside other statements (in this case an if statement). Will fix this.

ldez:
Feel like a duplicate of [nonamedreturns](https://golangci-lint.run/usage/linters/#nonamedreturns) :thinking:

maxsond:
It was born out of `nonamedreturns` and I worked mainly from `nonamedreturns` when making it. My team was originally discussing that one, but we agreed that we did want named returns (we find them nice for documenting what the result is beyond just the type) so `nonamedreturns` was not what we were looking for, but we did not want named returns to then permit naked/implicit return statements because that's a confusing convention for programmers from most other languages (in our collective opinion, at least, and there's some lively discussion on this in the context of Go2 as well here: https://github.com/golang/go/issues/21291).

That's why I say it's closest to `nakedret` which we may actually use instead of this because the differences between the two which I noted above are minor enough they shouldn't impact us.

ldez:
@maxsond you have to sign the CLA.

rittneje:
@maxsond Please note that https://github.com/alexkohler/nakedret/issues/18 has now been closed as "as designed". (It was just some confusion due to a misunderstanding.)

maxsond:
Ah, cool. Then I think this linter is redundant. I'll go ahead and close it.

PR #3051, opened 2022-08-03
gci: fix `custom-order` setting and upgrade to 0.6.1

Dreamacro:
Fix `custom-order` setting not work.

And upgrade to 0.6.1

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/breml/errchkjson v0.3.0
 	github.com/butuzov/ireturn v0.1.1
 	github.com/charithe/durationcheck v0.0.9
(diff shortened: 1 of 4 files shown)


ldez:
Hello,

We have a bot that manages updates of the linters.

thanks anyway

Dreamacro:
@ldez but this PR contains another fix, not related to upgrade

ldez:
That's not how it works.

The bot will create a PR and inside this PR we (the golangci-lint team) will update the code inside the PR of the bot.

Dreamacro:
@ldez I could push force the commit that only contains the fix. Or do you think you need to open a new PR?

ldez:
you can open a PR with only the fix.

PR #3055, opened 2022-08-03
dev: remove stable from actions/setup-go

kolyshkin:
Since actions/setup-go v3 `stable` parameter is no longer required
(see https://github.com/actions/setup-go/releases/tag/v3.0.0).


ldez (approved):
Thanks, I missed that info.

I updated your PR to keep an example of the syntax as it doesn't follow the Go versioning syntax.

PR #3064, opened 2022-08-05
feat: add reassign linter

chokoswitch:
reassign is a linter to check when global variables in other packages are reassigned, for example `io.EOF = nil`
https://github.com/curioswitch/go-reassign

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -1226,10 +1226,11 @@ linters-settings:
      - UnitAbbreviations
 
   reassign:
-    # Pattern for variables that are checked for reassignment.
+    # Patterns for variables that are checked for reassignment.
     # See https://github.com/curioswitch/go-reassign#usage
-    # Defaults to "^(EOF|Err.*)$"
-    pattern: ".*"
+    # Defaults to [EOF, Err.*]
+    patterns:
+     - ".*"
 
   revive:
     # Maximum number of open files at the same time.
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -1225,6 +1225,12 @@ linters-settings:
      - CamelCase
      - UnitAbbreviations
 
+  reassign:
+    # Pattern for variables that are checked for reassignment.
+    # See https://github.com/curioswitch/go-reassign#usage
+    # Defaults to "^(EOF|Err.*)$"
+    pattern: ".*"
+
   revive:
     # Maximum number of open files at the same time.
     # See https://github.com/mgechev/revive#command-line-flags
@@ -1951,6 +1957,7 @@ linters:
     - prealloc
     - predeclared
     - promlinter
+    - reassign
     - revive
     - rowserrcheck
     - scopelint
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -1226,9 +1226,9 @@ linters-settings:
      - UnitAbbreviations
 
   reassign:
-    # Patterns for variables that are checked for reassignment.
+    # Patterns for global variable names that are checked for reassignment.
     # See https://github.com/curioswitch/go-reassign#usage
-    # Defaults to [EOF, Err.*]
+    # Default: ["EOF", "Err.*"]
     patterns:
      - ".*"
 
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -2063,6 +2063,7 @@ linters:
     - prealloc
     - predeclared
     - promlinter
+    - reassign
     - revive
     - rowserrcheck
     - scopelint
--- a/go.mod
+++ b/go.mod
@@ -21,7 +21,7 @@ require (
 	github.com/breml/errchkjson v0.3.0
 	github.com/butuzov/ireturn v0.1.1
 	github.com/charithe/durationcheck v0.0.9
-	github.com/curioswitch/go-reassign v0.1.0
+	github.com/curioswitch/go-reassign v0.1.2
 	github.com/daixiang0/gci v0.6.3
 	github.com/denis-tingaikin/go-header v0.4.3
 	github.com/esimonov/ifshort v1.0.4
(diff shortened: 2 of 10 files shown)


ldez:

### Pull Request Description

### Linter

### `.golangci.reference.yml`

### Others Requirements

### Recommendations

---

chokoswitch:
Hi @ldez - it's been more than a week so just checking in, what sort of schedule can we expect on this sort of code review? Happy to iterate on the code as reviews come in. Thanks!

ldez:
I have some doubt about this linter because by default it only detected `io.EOF` reassignment.

chokoswitch:
@ldez The default is to detect `io.EOF` and any variable prefixed with `Err`. I think this provides the least chance of false positives and can catch most patterns of reassigning errors which is common use of package variables, but do you have an idea for a better default?

ldez:
Does your linter only detect reassignment of global variables?

For you, in which cases the reassignment is a problem?

chokoswitch:
Yes the job of the linter is to detect reassignment of global variables. The background for the linter is here

https://github.com/curioswitch/go-reassign#background

Some people prefer constant errors pattern to `errors.New` out of principle because of the potential of reassignment causing problems, so a linter helps make them both robust.

ldez:
> Yes the job of the linter is to detect reassignment of global variables. 

so the description of your linter must be explicit on that.

I think you have to improve the default detection and maybe you have to provide the slice instead of one string for the pattern. 

Also, I'm not sure about regexp, why did you use it instead of simple string?

chokoswitch:
> so the description of your linter must be explicit on that.

Do you mean PR description? It says "public variables in other packages" which I think is fairly explicit

> I think you have to improve the default detection and maybe you have to provide a slice instead of one string for the pattern.

Could you describe what sort of improvement you're thinking of? Should the pattern accept any variable to detect them all?

I can change from string to slice

> Also, I'm not sure about regexp, why did you use it instead of simple string?

It's to match against unknown variable names, often defined by a user. When matching against something arbitrary like that, I think regex flexibility can be needed

ldez:
> It says "public variables in other packages" which I think is fairly explicit

"public variables" or "top-level variables" are not really explicit.
"public" is not really a  Go vocabulary, we use mainly "exposed".
"top-level" is just related to a position inside something.

> Could you describe what sort of improvement you're thinking of? Should the pattern accept any variable to detect them all?

I already said that the current default is a bit weak.
In which you will use this linter? what will be your rule?
I will not provide the list of the global vars of the std lib.

> I think regex flexibility can be needed

Regexps are CPU intensives, so you have to a clear need to use them.
regexps are powerful but they must be used carefully.

The weak default and the use of regex, give me the impression that you don't really know the concrete use cases of your linter, that's why I ask you questions.

chokoswitch:
> The weak default and the use of regex, give me the impression that you don't really know the concrete use cases of your linter, that's why I ask you questions.

Did you read the background I linked? I'm not sure how much more to add about the concrete use case which is to safely provide exposed `var ErrFoo = errors.New()` type of pattern. 

It can also be used outside the case of errors to detect more broadly, in which case the regex would likely be set more permissive but that may increase chance of false positives.

chokoswitch:
Because golangci has its own suppression mechanisms, perhaps the pattern isn't as useful when invoked via golangci (it'd still be needed for CLI invocation in the binary itself). Would it be better to instead default to checking all reassignments without the pattern option, assuming that false positives could be fixed or suppressed with golangci?

ldez:
> Would it be better to instead default to checking all reassignments without the pattern option, assuming that false positives could be fixed or suppressed with golangci?

It's not a good idea: generating false positives is not a good thing at all.

As your linter is not specific to error as global variables, I expected something different as default but I understand that you are focused on error, so we will go with the current default.

Can you add a slice for the patterns?

chokoswitch:
Thanks @ldez! I've changed the config to a slice

ldez:
There is something that is unexpected for me.


<details>
<summary>tree</summary>

```
.
├── config
│   └── config.go
├── .golangci.yml
├── go.mod
└── main.go
```

</details>

<details>
<summary>config/config.go</summary>

```go
package config

import "net/http"

var DefaultClient = &http.Client{}

```

</details>

<details>
<summary>main.go</summary>

```go
package main

import (
	"io"
	"net/http"

	"github.com/golangci/sandbox/config"
)

var DefaultClient = &http.Client{}

func reassignPattern() {
	io.EOF = nil

	config.DefaultClient = nil // <- fail
	DefaultClient = nil        // <- fail

	http.DefaultClient = nil
	http.DefaultTransport = nil
}
```

</details>


<details>

```yaml
linters:
  disable-all: true
  enable:
    - reassign
    - typecheck

linters-settings:
  reassign:
    patterns:
      - "DefaultClient"
      - "DefaultTransport"

```

</details>

chokoswitch:
Thanks @ldez sorry for the trouble. I had forgotten to rename the yaml field name and had a [bug](https://github.com/curioswitch/go-reassign/pull/4/files#diff-b126d19c61b7249f14710696139d71e966aa643a9a363ad19e639e14cef6e41eR52) with multiple slashes. Added tests for latter upstream and yaml here

ldez:
I think that the reassignment should be handled even inside the same package.

```go
package a

import (
	"errors"
)

var MyError = errors.New("My error)

func reassignPattern() {
	MyError = nil

}
```

```go
package a

import (
	"net/http"
)

var DefaultClient = &http.Client{}

func reassignPattern() {
	DefaultClient = nil
}
```

It can be seen as an improvement.

ldez (approved):
LGTM

chokoswitch:
@ldez Thanks for all the help on this! 🙏 

> I think that the reassignment should be handled even inside the same package.

Yeah this is something that came to mind. I didn't implement it yet mostly because within-package, the author has control of the code so can worry less, having their variables modified outside the package from unknown codebase can cause undefined behavior though. But I think it could be an improvement, possibly opt-in

PR #3065, opened 2022-08-05
nosnakecase: deprecate linter

sivchari:
I proposal to disable nosnakecase linter. This rule is almost duplicated with revive(var-naming). In addition, the linter detects std libs or auto generated file (e.g. mockgen, protocol buffer etc...). These results is right, but it's not helpful.
Sorry, it's my fault.

FYI
[Option to disable for imported libraries](https://github.com/sivchari/nosnakecase/issues/16)
[Difference/relation to Revive's var-naming check? ](https://github.com/sivchari/nosnakecase/issues/10)


ldez:
Hello,
You cannot remove a linter: the linter must be deprecated before

ldez (approved):
LGTM

Issue #3081, opened 2022-08-08
bug: couldn't distinguish 32-bit or 64-bit under arch x86_x64 when execute install.sh


jnan806:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [X] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

### Description of the problem

in the function uname_arch() of the install.sh file, it couldn't distinguish 32-bit or 64-bit under arch x86_x64 
<img width="370" alt="image" src="https://user-images.githubusercontent.com/43714056/183460137-32e12c04-c85e-4f29-a200-a602fa9a7a55.png">

<img width="1159" alt="image" src="https://user-images.githubusercontent.com/43714056/183463098-9e1d68a2-889a-4ecf-9dab-6db80efea0c9.png">

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
# Paste output here
```
<img width="476" alt="image" src="https://user-images.githubusercontent.com/43714056/183463203-d9e723b3-16ca-4cfa-9809-b5c8276d63f6.png">

</details>

### Configuration file

<details>

```console
$ cat .golangci.yml
# paste output here
```

</details>

### Go environment

<details>

```console
$ go version && go env
# paste output here
```

</details>

### Verbose output of running

<details>

```console
$ golangci-lint cache clean
$ golangci-lint run -v
# paste output here
```

</details>

### Code example or link to a public repository

<details>

```go
// add your code here
```

</details>

ldez:
Hello,

as far I know `x86_64` is 64bits, it's the meaning of 64, so seems more a problem with your OS than with our script.

jnan806:
i has tried several ways to get the platform of OS, use docker image i386/ubuntu:latest

jnan806:
> Hello,
> 
> as far as I know, `x86_64` is 64bits, it's the meaning of 64, so it seems more a problem with your OS than with our script.
> 
> https://en.wikipedia.org/wiki/X86-64

Yes , i know too. 😄  :) 

But i has tried several commands to get the platform of OS, and get different results:

use docker image ` i386/ubuntu:latest`
[...]

PR #3082, opened 2022-08-08
distinguish 32-bit or 64-bit under arch x86_x64

jnan806:
fix: #3081 distinguish 32-bit or 64-bit under arch x86_x64


ldez:
We will discuss in the issue #3081

Issue #3080, opened 2022-08-08
Using enable-all. Custom linter does not work.


yuqengo:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [X] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

### Description of the problem

example linter does not work with this .golangci.yml .
```
linters-settings:
  custom:
    example:
      path: ./example.so

linters:
  enable-all: true
```

```
$ golangci-lint run -v testdata/src/testlintdata/todo/todo.go                                                                                                                                                                                                                                                                                                                                                                                                                                                                                  
INFO [config_reader] Config search paths: [./ ${home}/go/src/github.com/example-plugin-linter/testdata/src/testlintdata/todo /${home}/go/src/github.com/example-plugin-linter/testdata/src/testlintdata ${home}/go/src/github.com/example-plugin-linter/testdata/src ${home}/go/src/github.com/example-plugin-linter/testdata ${home}/go/src/github.com/example-plugin-linter${home}/go/src/github.com ${home}/go/src ${home}/go ${home} /Users /] 
INFO [config_reader] Used config file .golangci.yml 
INFO Loaded ./example.so: example                 
INFO [lintersdb] Active 95 linters: [asasalint asciicheck bidichk bodyclose containedctx contextcheck cyclop deadcode decorder depguard dogsled dupl durationcheck errcheck errchkjson errname errorlint execinquery exhaustive exhaustivestruct exhaustruct exportloopref forbidigo forcetypeassert funlen gci gochecknoglobals gochecknoinits gocognit goconst gocritic gocyclo godot godox goerr113 gofmt gofumpt goheader goimports golint gomnd gomoddirectives gomodguard goprintffuncname gosec gosimple govet grouper ifshort importas ineffassign interfacer ireturn lll maintidx makezero maligned misspell nakedret nestif nilerr nilnil nlreturn noctx nolintlint nonamedreturns nosnakecase nosprintfhostport paralleltest prealloc predeclared promlinter revive rowserrcheck scopelint sqlclosecheck staticcheck structcheck stylecheck tagliatelle tenv testpackage thelper tparallel typecheck unconvert unparam unused usestdlibvars varcheck varnamelen wastedassign whitespace wrapcheck wsl] 
```

[...]

PR #3085, opened 2022-08-09
feat: Include custom linters in `enable-all`

yuqengo:
Related to https://github.com/golangci/golangci-lint/issues/3080

.golangci.yml
```
linters-settings:
  custom:
    example:
      path: ./example.so

linters:
  enable-all: true
```
Custom linters work for `enable-all`.

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -1692,7 +1692,7 @@ linters-settings:
     # Suggest the use of crypto.Hash
     # Default: false
     crypto-hash: true
-    # Suggest the use of pc.DefaultXXPath
+    # Suggest the use of rpc.DefaultXXPath
     # Default: false
     default-rpc-path: true
 


ldez:
Hello,

for security reasons, I think it's a bad idea to do that.

yuqengo:
Hello, Thanks for review!

What security reasons do you have?

ldez:
I thought about the subject and I think there is a bug because the activation of plugins should be done by default. https://github.com/golangci/golangci-lint/blob/011252127f53cf6a0f2b28dc97fed55508575b09/pkg/lint/lintersdb/manager.go#L983

The current implementation inside golangci-lint has some design problems.

I think I will fix the design problem and the behavior of the plugin in another PR.

yuqengo:
@ldez 
hello idez !

> The current implementation inside golangci-lint has some design problems.
> I think I will fix design problems and the behavior of the plugin in another PR.

If you respond with another PR, Should this PR be closed ?

PR #3099, opened 2022-08-15
dev: replace golangcitest:config by golangcitest:config_path

ldez:
Replacement of the directive `golangcitest:config` by `golangcitest:config_path`.

`golangcitest:config_path` already existed but we had a mix between the 2 directives.

The objective is to simplify maintenance and the contribution by having an identical approach for all tests.

This also simplifies the maintenance of test files which are corrected automatically by the IDEs as soon as the file is edited when you have to update the configuration.

### DIFF
--- a/test/testdata/configs/wsl.yml
+++ b/test/testdata/configs/wsl.yml
@@ -1,0 +1,3 @@
+linters-settings:
+  wsl:
+    tests: 1
--- a/test/testdata/configs/wsl.yml
+++ b/test/testdata/configs/wsl.yml
@@ -1,3 +1,3 @@
 linters-settings:
   wsl:
-    tests: 1
+    allow-cuddle-declarations: false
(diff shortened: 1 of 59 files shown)


test/testdata/configs/wsl.yml
```diff
@@ -0,0 +1,3 @@
+linters-settings:
+  wsl:
+    tests: 1
```
bombsimon, line 3: This seems to be an old setting to include tests or not which is now handled by `golangci-lint`. Let's pick something from the [reference YAML file](https://github.com/golangci/golangci-lint/blob/9da04f5070e67741bb9c98d77c2ee95f13e241df/.golangci.reference.yml#L1796-L1848)

```suggestion
linters-settings:
  wsl:
    allow-cuddle-declarations: false
```

bombsimon (approved):
Nice, this will make it easier to understand how to change settings for your tests and consistent with regular usage! 👍

PR #3100, opened 2022-08-15
dev: rewrite the tests framework

ldez:
The new "framework":
- is more simple
- is more robust
- is more readable
- provide more precise information when a test fails
- has tests


bombsimon (approved):
Very nice, easy to read, easy to follow - thanks!❤️

ldez:
There is an interesting side effect: I was able to fix the flaky tests as the tests are more readable.
So I hope this is the end of the multiple and manual re-run on the CI.

PR #3104, opened 2022-08-17
dev: follow standards by using 'want' instead of 'ERROR' for tests

### LINKED ISSUE #1222: Supports analysis package error format in golangci-lint test
In order to add a linter to golangci-lint, we need to add  tests. However, the error format is different between the analysistest package and golangci-lint, so we have to change it. If same error format, we can reuse the test cases by simply copying test files.

### golangci-lint test error format

> http.Get(url) // ERROR "net/http\.Get must not be called"
> https://github.com/golangci/golangci-lint/blob/4fb9fe59f9126844cbd42ce2365b9a585d51dc6c/test/testdata/noctx.go#L16

### analysistest package error format

> http.Get(url) // want `net/http\.Get must not be called`
> https://github.com/sonatard/noctx/blob/cca218322853bcbd277bf1ff37ec43b473a885dc/testdata/src/a/a.go#L15

https://godoc.org/golang.org/x/tools/go/analysis/analysistest#Run

Is there any reason for the different error formats?

ldez:
This is a complete rewrite of the test system.

Inspired by https://pkg.go.dev/golang.org/x/tools/go/analysis/analysistest#Run

extra: I also add a new directives: `golangcitest:expected_exitcode directive`. Still the same goal: robustness and high-quality test outputs.

Fixes #1222


test/run_test.go
```diff
@@ -659,9 +659,9 @@ func TestDisallowedOptionsInConfig(t *testing.T) {
 
 		testshared.NewRunnerBuilder(t).
 			WithConfig(c.cfg).
-			WithArgs(getCommonRunArgs()...).
+			WithArgs("--skip-dirs", "testdata_etc/,pkg/golinters/goanalysis/(checker|passes)").
```
bombsimon, line 662: I kinda like that this was a shared method before, duplicating like this risks adding maintenance if changed. Maybe move the actual valua to a constant add keep the `--skip-dirs` for readability.

```go
const skipDirsCommand = "testdata_etc/,pkg/golinters/goanalysis/(checker|passes)"

// ...
    WithArgs("--skip-dirs", skipDirsCommand")
```

And while at it maybe add a comment about why we skip those dirs because it's not clear by just reading the line to me.

test/testshared/analysis_test.go
```diff
@@ -0,0 +1,138 @@
+package testshared
```
bombsimon, line 1: Really nice tests to ensure parsing is working and document expected behavior! 🥇

test/testshared/directives.go
```diff
@@ -82,6 +88,14 @@ func ParseTestDirectives(tb testing.TB, sourcePath string) *RunContext {
 			rc.ExpectedLinter = after
 			continue
 
+		case "//golangcitest:expected_exitcode":
+			require.NotEmpty(tb, after)
+			val, err := strconv.Atoi(after)
+			require.NoError(tb, err)
+
+			rc.ExitCode = val
+			continue
```
bombsimon, line 97: Are you foreseeing any other exit codes than 0 from this directive? Right now it seems like we only set this to 0 but I wonder if it would make more sense to have a setting for tests without assertions and marking them as successful. I'm thinking something like `//golangcitest:has_no_issues` and if so set `rc.ExitCode = exitcodes.Success`?

bombsimon (commented):
Sorry for the long wait, really appreciate the work on this PR. Will make it easy to port tests but also just great improvement to the runner!

test/output_test.go
```diff
@@ -0,0 +1,77 @@
...
+			"--print-linter-name=false",
+			"--out-format=line-number,json:stderr",
+		).
+		WithDirectives(sourcePath).
+		WithTargetPath(sourcePath).
+		Runner().
+		Install().
+		Run().
+		ExpectHasIssue("testdata/gci.go:8: File is not `gci`-ed").
+		ExpectOutputContains(`"Issues":[`)
```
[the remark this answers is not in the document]
ldez: the test already existed but I can improve it.

test/testshared/directives.go
ldez: In fact, I think I will add tests about the `run.exitcode` but in another PR with other tests.
If you agree I would prefer to keep the explicit value.

test/output_test.go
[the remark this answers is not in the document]
bombsimon: Yeah I actually saw the other test after writing this which tbh actually made me more confused since I guess they could share some code and checks. If you think it's easy to have a shared way of doing it go for it but I don't have a super strong opinion on this

test/testshared/directives.go
bombsimon: Sure, let's keep it! I just don't see an obvious case when a test from `testdata` would exist either successfully or with issues found. Testing other ways of exiting is a great plus but I don't think the easiest way to do that is to use this directive with another value.

bombsimon (approved):
LGTM! 🎉

test/output_test.go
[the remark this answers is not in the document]
ldez: I updated the tests

ldez:
@bombsimon thank you for the review :+1:

Issue #2239, opened 2021-09-17
Remove --exclude-use-default, default to false, or trim it WAY down


natefinch:
### Your feature request related to a problem? Please describe.

It took me way too long to figure out why golangci-lint wasn't telling me about so many linting errors that I could see with my own eyes.

This flag, which defaults to on, silently suppresses many important linting checks, and overrides the whole rest of your linting configuration in a very non-intuitive way.

The  errors this setting suppress are, in many cases, extremely important to code quality.  

Warning against using unsafe? That's important.

Warning against creating files and directories that are too open? That's important.

Warning about missing comments? That helps keep our codebase quality up, and is even mentioned in Effective Go. The comment warnings are especially egregious, in my opinion. The flag says "the rare codebase has such comments".  That's an opinion, and I can't tell if it's saying that it's rare for people to write comments in the officially way (i.e. match the first word to the thing you're documenting), or if it's saying it's rare for people to *not* do that. But either way, it's certainly *not* rare.

This doesn't feel like a list of actual edge cases and false positives (for the most part), this feels like a list of one person's opinions about what kinds of linting errors matter.  But that's *what the configuration file is **for***.

Please reconsider this flag.  I have seen what happens when you don't have sufficient linting combined with inexperienced devs... and the fact that the most popular linting program in the Go community is suppressing so many linting errors that are **not** edge cases, is extremely distressing. 

```
      --exclude-use-default            Use or not use default excludes:
 # EXC0001 errcheck: Almost all programs ignore errors on these functions and in most cases it's ok
 - Error return value of .((os\.)?std(out|err)\..*|.*Close|.*Flush|os\.Remove(All)?|.*print(f|ln)?|os\.(Un)?Setenv). is not checked

 # EXC0002 golint: Annoying issue about not having a comment. The rare codebase has such comments
 - (comment on exported (method|function|type|const)|should have( a package)? comment|comment should be of the form)

 # EXC0003 golint: False positive when tests are defined in package 'test'
 - func name will be used as test\.Test.* by other packages, and that stutters; consider calling this

 # EXC0004 govet: Common false positives
 - (possible misuse of unsafe.Pointer|should have signature)

 # EXC0005 staticcheck: Developers tend to write in C-style with an explicit 'break' in a 'switch', so it's ok to ignore
 - ineffective break statement. Did you mean to break out of the outer loop

 # EXC0006 gosec: Too many false-positives on 'unsafe' usage
 - Use of unsafe calls should be audited

[...]

PR #3117, opened 2022-08-20
docs: add a section about default exclusions

ldez:
- add a section about default exclusions
- add the list of exclusion codes to the reference 

<details>
<summary>preview</summary>

Screenshot 2022-08-20 at 16-10-31 False Positives golangci-lint

</details>

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -2097,7 +2097,7 @@ issues:
   # it can be disabled by `exclude-use-default: false`.
   # To list all excluded by default patterns execute `golangci-lint run --help`
   #
-  # Default: []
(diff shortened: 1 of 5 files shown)


bombsimon (approved):
LGTM

I know there's been quite some issues on this topic and different opinions (f.ex. #456 and #2239). There have been comments in such issues that this cannot be changed in a v1 due to breaking changes but that there are plans to change this in v2. Do you think it's worth mentioning any of that to avoid additional issues or will it just risk getting forgotten and get outdated?

ldez:
> Do you think it's worth mentioning any of that to avoid additional issues or will it just risk getting forgotten and get outdated?

For now, it's not really useful. I already started to think about the v2 and I think we will keep "suggested" exclusions but not by default.

PR #3118, opened 2022-08-20
dev: fix output tests

ldez:
When a new linter is added, the output changes, so we cannot match the full report.
We have to focus on the `Issues` section.


bombsimon (approved):
LGTM.

What I was thinking with my comment regarding matching more than the prefix was to actually unmarshal and check for the `Issues` key but since that's done in a different test this seems reasonable

PR #3093, opened 2022-08-12
feat: add logrlint

timonwong:
https://github.com/timonwong/logrlint

logrlint is a linter for [go-logr/logr](https://github.com/go-logr/logr), which lint the number of key value pairs (should be even).

Though I'm not sure whether it's good to add a very specific linter here :(

PR #3124, opened 2022-08-21
dev: improve runner to run dir with go.mod

ldez:
Related to #3093 and https://github.com/golangci/golangci-lint/discussions/2969

Currently, we don't have any linter that required this approach, so I cannot add a test.

The test will look like that:

```go
func TestSourcesFromTestdataSubDir(t *testing.T) {
	subDirs := []string{
		"logrlint",
		"ginkgolinter",
	}

	for _, dir := range subDirs {
		t.Run(dir, func(t *testing.T) {
			testSourcesFromDir(t, filepath.Join(testdataDir, dir))
		})
	}
}
```

### DIFF
--- a/test/linters_test.go
+++ b/test/linters_test.go
@@ -1,13 +1,17 @@
 package test
 
 import (
+	"os"
 	"os/exec"
 	"path/filepath"
+	"strings"
 	"testing"
+	"time"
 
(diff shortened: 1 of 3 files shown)


bombsimon (approved):
🎉

PR #3128, opened 2022-08-22
dev: fix: linter tests

ldez:
I made a mistake here: https://github.com/golangci/golangci-lint/pull/3124/files#diff-2d27f13815d27a52a927014f02a8d7545106cf3db781eb5de69e1f89ec689376R52

I just forget to remove my dev constraint.

only one linter test was running instead of all linter tests.


bombsimon (approved):
Oops, sorry for not spotting this! 🙈 LGTM

Issue #1382, opened 2020-09-21
CI is failing (silently) on Windows


Nivl:
I'm not sure how related to the other Windows issues this is, but the Windows CI seems to be consistently failing (the errors are being ignored so it always shows the "passing" status). You can see the error here: https://github.com/golangci/golangci-lint/runs/1141826954 after I removed the option to stop ignoring errors.

To repro it's simple:

1. Remove `continue-on-error: true` from the `tests-on-windows` job in `.github/workflow/pr.yml`
1. Create a PR to run the CI.

The error seems to be related to gofmt and might be related to https://github.com/golangci/golangci-lint/issues/580. I don't have a Windows machine, so I'd rather let this to Windows people to figure it out.

PR #2825, opened 2022-05-05
ci: ensure windows pipeline

ernado:
Make windows pipeline red on error

PR #1056, opened 2020-05-01
Replacing Travis with Github actions

iwankgb:
Important notes:
* I had to update `go.mod` and `go.sum` in `tools/` - I wasn't able to run `make generate` otherwise
* goreleaser does not support `gomips` key in the configuration
* I got rid of `time` in `Makefile` to support builds on Windows.

I decided to extend builds to Windows and MacOS but:
* tests on Windows fail miserably
* tests on MacOS timeout on Go 1.14 (sometimes)

Release build was tested for [Docker](https://hub.docker.com/r/iwankgb/golangci-lint/tags) and [Github releases](https://github.com/iwankgb/golangci-lint/releases/tag/v2.6). I was not able to test Homebrew. Secrets must be set for hub.docker.com (`GOLANGCI_LINT_DOCKER_TOKEN`) and Github (`GOLANGCI_LINT_TOKEN`) authentication

Fixes #1039, supersedes #1054

PR #3134, opened 2022-08-24
dev: fix CI workflow for Windows

ldez:
During the migration from Travis (Linux and macOS only) to GitHub Actions a job for Windows has been introduced.

`continue-on-error: true` has been added in the [CI migration PR](https://github.com/golangci/golangci-lint/pull/1056).

So our CI workflow has never worked with Windows.

After a lot of runs on a CI (because I use Linux), and thanks to my previous test framework rewrite, I finally got this job to work.

There are still a few tests that are skipped but these are just a few specific tests.

The side effect: a job on Linux or Darwin takes 4m but 10m on Windows...

Fixes https://github.com/golangci/golangci-lint/issues/1382

Previous tries:
- https://github.com/golangci/golangci-lint/pull/1395
- https://github.com/golangci/golangci-lint/pull/2825

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,3 @@
 go.sum linguist-generated
+*       text=auto eol=lf
+*.ps1   text eol=crlf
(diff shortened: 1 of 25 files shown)


Nivl (approved):
😍

PR #3140, opened 2022-08-26
build(deps): bump github.com/timonwong/logrlint from 0.1.0 to 0.4.0

timonwong:
https://github.com/timonwong/logrlint/compare/v0.1.0...v0.3.0

Notable changes:
- fix a fales positive for dotdotdot args
- add support to [klog](https://github.com/kubernetes/klog)
- add config flags `disableall`, `disable` and `enable`

PR #3144, opened 2022-08-26
feat: rename logrlint to loggercheck

timonwong:
Since `logrlint` reaches v0.1.0, more loggers are supported (`klog`, `zap`). 

So as suggested in #3140, `logrlint` is now deprecated, but since it's just a repo rename, so logrlint is renamed to [loggercheck](https://github.com/timonwong/loggercheck) using alternate names.

New features:
- checks for klog and zap
- allow custom rules
- allow disabling checks individually
- allow check for logging key (require string literal/constant, ascii string)
- allow check for printf-like specifiers (%s, %d, etc)


ldez:
you cannot rename a linter.

https://github.com/golangci/golangci-lint/pull/3140#issuecomment-1228343049

ldez:
I will apply some changes to your PR.

.golangci.reference.yml
```diff
@@ -1111,6 +1111,11 @@ linters-settings:
     # Default: 1
     tab-width: 1
 
+  loggercheck:
+    logr: true
+    klog: true
+    zap: true
+
```
ldez, line 1118: If the linter has a configuration, the exhaustive configuration of the linter must be added (alphabetical case-insensitive order)

ldez:
Please wait, I haven't seen that you have used the alternative names.

So you can "rename".

ldez:
@timonwong you can do your rename, read my previous message

timonwong:
> Please wait, I haven't seen that you have used the alternative names.
> 
> So you can "rename".

@ldez Ok, I'll do it right now

ldez (changes requested):

ldez:
just a quick note about klog and kubernetes https://github.com/kubernetes-sigs/logtools/tree/main/logcheck

.golangci.reference.yml
```diff
@@ -1111,6 +1111,22 @@ linters-settings:
...
+    # Default: true
+    klog: false
+    # Allow check for the github.com/go-logr/logr library.
+    # Default: true
+    logr: false
+    # Allow check for the "sugar logger" from go.uber.org/zap library.
+    # Default: false
+    zap: true
+    # List of rules to check against, where each rule is a single logger pattern.
+    rules:
```
ldez, line 1125: what is the default value? Please add the value to the file.

ldez (changes requested):

.golangci.reference.yml
timonwong: By default the (custom) rules is empty

timonwong:
> just a quick note about klog and kubernetes https://github.com/kubernetes-sigs/logtools/tree/main/logcheck

Hmmm, if knowing this before i may not create another linter myself :joy:

ldez (approved):
LGTM

PR #3173, opened 2022-08-31
fix: remove reference to non-working service

benpbrown:
golangci.com has been shut down since April 2020, so can no longer be
used to run golangci-lint in the cloud.

See https://medium.com/golangci/golangci-com-is-closing-d1fc1bd30e0e


ldez (approved):
LGTM

Issue #2357, opened 2021-11-13
Run returns 0 on build failure


antoineco:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [X] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

### Description of the problem

I noticed while looking at the CI logs of the following PR that `golangci-lint run` was returning with an exit code of `0`, although the code doesn't compile: https://github.com/triggermesh/triggermesh/pull/263

I can see the following warning in the logs:

```
WARN [runner] Can't run linter goanalysis_metalinter: buildir: failed to load package cortex: could not load export data: no export data for "go.opentelemetry.io/contrib/exporters/metric/cortex"
```

The dependency that fails to build is indeed the one I expected, but the outcome and the message are surprising to me.
My reasoning is that:

1. If such failure occurs, it's unlikely that the code can be considered acceptable.
1. If I see this message in isolation from the change, I'm unlikely to be able to determine where the linter failed, and why.

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version v1.43.0 built from (unknown, mod sum: "h1:SLwZFEmDgopqZpfP495zCtV9REUf551JJlJ51Ql7NZA=") on (unknown)
```

</details>

### Configuration file

<details>

```yaml
linters:
  enable:
  - gofmt
  - stylecheck

run:
  timeout: 10m
```

</details>

### Go environment

<details>

```console
$ go version && go env
go version go1.17.2 linux/amd64
GO111MODULE=""
GOARCH="amd64"
GOBIN=""
GOCACHE="/home/acotten/.cache/go-build"
GOENV="/home/acotten/.config/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="linux"
GOINSECURE=""
GOMODCACHE="/home/acotten/go/pkg/mod"
GONOPROXY="github.com/triggermesh/*"
GONOSUMDB="github.com/triggermesh/*"
GOOS="linux"
GOPATH="/home/acotten/go"
GOPRIVATE="github.com/triggermesh/*"
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
[...]

PR #3184, opened 2022-09-02
PoC: Add Severity Levels, use it to return zero exit code

kavu:
Hi! 

This is my Proof-of-Concept of how it could be possible to approach problem with non-zero exit codes on issues with non-error level.

I've tested it on very simple severity config:

```yaml
severity:
  default-severity: error
  case-sensitive: true
  rules:
    - linters:
      - paralleltest
      - tparallel
      severity: warning
```  

Running with `golangci-lint run ./... && echo $?` in my project with several missing `t.Parallel()`.

I don't insist it is final or ideal, just my view on how it could be done.

What do you think, guys? 

Solves #1981

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -380,7 +380,20 @@ func (e *Executor) setOutputToDevNull() (savedStdout, savedStderr *os.File) {
 }
 
 func (e *Executor) setExitCodeIfIssuesFound(issues []result.Issue) {
-	if len(issues) != 0 {
+	if len(issues) == 0 {
+		return
+	}
+
+	// Let's count how many Errors do we have overall?
+	errorsCount := 0
+	for _, i := range issues {
+		if i.Severity == config.SeverityErrorLevel {
+			errorsCount++
+		}
+	}
+
+	// Okay, we've got some errors, so we need to return some non-zero exit-code
+	if errorsCount > 0 {
 		e.exitCode = e.cfg.Run.ExitCodeIfIssuesFound
 	}
 }
--- a/pkg/config/severity.go
+++ b/pkg/config/severity.go
@@ -2,15 +2,24 @@ package config
 
 const severityRuleMinConditionsCount = 1
 
+type SeverityLevel string
+
+const (
+	SeverityDebugLevel   SeverityLevel = "debug"
+	SeverityInfoLevel    SeverityLevel = "info"
+	SeverityWarningLevel SeverityLevel = "warning"
+	SeverityErrorLevel   SeverityLevel = "error"
+)
+
 type Severity struct {
-	Default       string         `mapstructure:"default-severity"`
+	Default       SeverityLevel  `mapstructure:"default-severity"`
 	CaseSensitive bool           `mapstructure:"case-sensitive"`
 	Rules         []SeverityRule `mapstructure:"rules"`
 }
 
 type SeverityRule struct {
 	BaseRule `mapstructure:",squash"`
-	Severity string
+	Severity SeverityLevel
 }
 
 func (s *SeverityRule) Validate() error {
--- a/pkg/golinters/revive.go
+++ b/pkg/golinters/revive.go
@@ -142,7 +142,7 @@ func reviveToIssue(pass *analysis.Pass, object *jsonObject) goanalysis.Issue {
 	}
 
 	return goanalysis.NewIssue(&result.Issue{
-		Severity: string(object.Severity),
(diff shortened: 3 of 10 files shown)


ldez:
https://github.com/golangci/golangci-lint/pull/2595#issuecomment-1043739800

kavu:
@ldez I've seen these discussions, but issues are still, so I decided to take a chance and purpose my solution.

So, do I understand correctly that

> Regarding the exit code, as far as I remember, discussions have already taken place on this subject and overall the decision was made not to correlate severity and exit code.

means that there will be no any work done on this matter any time soon and we should try to find alternative solutions if we want to ignore issues that should not result with non-zero code?

ldez:
As I said, the decision was made not to correlate severity and exit code, for me nothing changed.

For me, it's useless to create a lint report just to ignore it, the correlation between severity and exit code is a bad historical way to manage linting.
For me, it's like you write tests but you want to exit 0 even if the tests are failing.

kavu:
@ldez 
Okay, got your point. Thank you! 

But maybe you can give and advice, how we can approach out problem?

We want to add some linter and we want to give some time to developers to fix it, having some clear indication that this problem exits. But as we don't want to enforce it right away, we don't want to it to break pipelines on non-zero code. That's the reasoning behind my intentions hack this part.

What will be the best approach from your perspective? Thank you!

bombsimon:
I'm torn but I think I mostly agree with @ldez. Now that we have severity, I can kinda see why you would want to filter them out. Also judging by https://github.com/golangci/golangci-lint/issues/1981#issuecomment-841045700 `golangci-lint` would not be the first linter to allow reports combined with a successful exit code.

Personally I also think that it's a somewhat odd requirement to ask for the reports and then ignoring them, it's like running `golangci-lint run || true`. To me it's not clear what that information would serve. However now we're more talking about workflows and philosophical questions. The same discussion happens all the time with log severity, f.ex. "Are you supposed to get paged when a log with severity `error` is written?".

I don't really remember the historical reasons to support severity and even less memory about discussions where we decided to not add support to match it with exit codes. Given that we actually did a change regarding exit codes pretty recently in #2357 and that this seems to be a request from several people I would be open to reconsider this and add support for this.

ldez:
> We want to add some linter and we want to give some time to developers to fix it, ...
> But as we don't want to enforce it right away, we don't want to it to break pipelines on non-zero code ...

Why would you want to add a linter but wait to fix its reports?

kavu:
> Why would you want to add a linter but wait to fix its reports?

Dummy simple reason — we want to "flash in face" the soon-to-be-errors in the pipeline logs, so it would be visible in case someone will check it or will try to read other linter problems, but not failing the whole.

Yes, totally agree, @bombsimon approach sounds really good!

We also considering creating separate task in the pipeline with the future version of lint config, that won't be blocking anything and later will just swap it with the main task. 

Thank you for your time and help, guys!

jufemaiz:
> Why would you want to add a linter but wait to fix its reports?

Ref: https://github.com/golangci/golangci-lint/issues/1981#issuecomment-841045700 for a full outline of the reasoning with citations of examples used elsewhere (thus an expectation of the ability in the software engineering space more broadly).

ImprintNav:
@ldez What's the reasoning for having severity levels if the tool provides no means of treating them differently in terms of CI success or run-time filtering? If everything is blocked at the CI level, you might as well make everything an error. We don't have "warnings" or "info" for unit tests either precisely because we don't want to filter out unit test failures by severity level.

ldez:
You can configure `godox` to allow `TODO` but forbid `FIXME`.

```yml
# ...

linters-settings:
  godox:
    keywords:
      - FIXME
# ...
```

You will have 2 levels:
- one, non-blocking, related to tasks you will do in the future.
- one, blocking, related to tasks you should do before your commit/PR/merge.

abemedia:
I'd love to see something like this go in. The reason for this is that I want godox to leave annotations for `TODO` comments, however I don't want them to cause a CI failure.
Otherwise I find that TODO comments are easily forgotten but this gives them visibility, without enforcing anything too strict.

A good alternative might however be to default to non-zero exit codes for everything and allow users to explicitly set it i.e.
```yaml
severity:
  default-severity: error
  rules:
    - linters:
        - godox
      severity: warning
      exit-code: 0

ldez:
This PR is old and my first comment is still accurate for me: https://github.com/golangci/golangci-lint/pull/3184#issuecomment-1235438429

But the discussion can continue inside issue #1981.

PR #3196, opened 2022-09-06
dev: reference all the debug keys and env vars

ldez:
It was really painful to find the debug keys and some hidden env vars.

So I created constants to document those keys and env vars.

It can be improved in the future by adding a comment on each debug key.


bombsimon (approved):
💯

PR #3202, opened 2022-09-09
wsl: support all configs and update docs

xobotyi:
Now options docs aligned with wsl's and settings supports all of wsl configs.

### DIFF
--- a/pkg/golinters/wsl.go
+++ b/pkg/golinters/wsl.go
@@ -19,30 +19,26 @@ func NewWSL(settings *config.WSLSettings) *goanalysis.Linter {
 	var mu sync.Mutex
 	var resIssues []goanalysis.Issue
 
-	conf := &wsl.Configuration{
-		AllowCuddleWithCalls: []string{"Lock", "RLock"},
-		AllowCuddleWithRHS:   []string{"Unlock", "RUnlock"},
-		ErrorVariableNames:   []string{"err"},
-	}
+	conf := wsl.DefaultConfig()
 
 	if settings != nil {
 		conf.StrictAppend = settings.StrictAppend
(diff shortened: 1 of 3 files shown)


pkg/golinters/wsl.go
```diff
@@ -19,30 +19,26 @@ func NewWSL(settings *config.WSLSettings) *goanalysis.Linter {
 	var mu sync.Mutex
 	var resIssues []goanalysis.Issue
 
-	conf := &wsl.Configuration{
-		AllowCuddleWithCalls: []string{"Lock", "RLock"},
-		AllowCuddleWithRHS:   []string{"Unlock", "RUnlock"},
-		ErrorVariableNames:   []string{"err"},
-	}
+	conf := wsl.DefaultConfig()
```
bombsimon, line 22: Since we have default settings in [linters_settings.go](https://github.com/golangci/golangci-lint/pull/3202/files#diff-cb1fbdbfccd3fd0d8fe9d9006138fbcd1afcc3a701b37180c6fe50ed0c5c390f) which does not necessarily align with `wsl.DefaultConfig()`, should we keep them separate? This was briefly mentioned in #794 but that's from 2019 when settings looked a bit different.

Just to not risk changes in future bumps, including new major versions, maybe we can continue not rely on the upstream defaults?

```suggestion
	conf := &wsl.Configuration{}
```

bombsimon: Actually this doesn't matter, we will always have default settings in `golangci-lint` and we will unconditionally overwrite every setting - at least as long as they're known to `golangci-lint`.

bombsimon (approved):
LGTM

PR #3204, opened 2022-09-11
dev: add doc about internal package extracted from Go

ldez:
- add some readme
- update robustio
- add comments on skip-dirs
- fix flaky tests on Windows and Darwin

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -138,9 +138,8 @@ issues:
 
 run:
   timeout: 5m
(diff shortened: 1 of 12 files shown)


bombsimon (approved):
LGTM, nice readmes with the history. Although we might risk updating this in future changes it's easy enough to do that retroactively at any time like you did now. 👍

ldez:
The update of some packages will be really complex (I tried) because we don't have a clear starting point.
At some point, we will have to re-think those packages.

PR #2017, opened 2021-05-26
staticcheck: configuration for `staticcheck`, `gosimple`, `stylecheck`

ldez:
Fixes #1275

Staticcheck is split into 4 linters in golangci-lint `staticcheck`, `gosimple`, `stylecheck` and `unused`.
(more information: https://github.com/golangci/golangci-lint/discussions/1989#discussioncomment-745004)

|                                       | `staticcheck` | `stylecheck` | `gosimple` | `unused` |
|---------------------------------------|---|---|---|---|
| `go`                                  | x | x | x | x |
| `checks` (*)                          | x | x | x | - |
| `dot-import-whitelist`  (ST1003)      | - | x | - | - |
| `initialisms` (ST1001)                | - | x | - | - |
| `http-status-code-whitelist` (ST1013) | - | x | - | - |

_(*)_: each `checks` section can only filter the rules related to this section:
- `checks` on `staticcheck` only filter `SAxxxx` rules
- `checks` on `stylecheck` only filter `STxxxx` rules
- `checks` on `gosimple` only filter `Sxxxx` rules

example:

```yml
  stylecheck:
    go: "1.15"
    checks: [ "all", "-ST1000", "-ST1003", "-ST1016", "-ST1020", "-ST1021", "-ST1022" ]
    dot-import-whitelist:
      - fmt
    initialisms: [ "ACL", "API", "ASCII", "CPU", "CSS", "DNS", "EOF", "GUID", "HTML", "HTTP", "HTTPS", "ID", "IP", "JSON", "QPS", "RAM", "RPC", "SLA", "SMTP", "SQL", "SSH", "TCP", "TLS", "TTL", "UDP", "UI", "GID", "UID", "UUID", "URI", "URL", "UTF8", "VM", "XML", "XMPP", "XSRF", "XSS" ]
    http-status-code-whitelist: [ "200", "400", "404", "500" ]
```

Staticcheck available options: https://staticcheck.io/docs/options

PR #3215, opened 2022-09-13
docs: specify what code checks are related to staticcheck sub-linter

davidhsingyuchen:
## Summary

PR extracts the info in the description of #2017 into the reference YAML config file. I didn't create an issue first because this PR is simple, and it's just moving existing info written by a maintainer.

## Why

I was migrating from `staticcheck` to `golangci-lint`, so I naturally tried to configure the `checks` of the `staticcheck` linter, and it took me some time to find #2017 to understand why it didn't work. This PR can be helpful for this kind of cases.


ldez (approved):
LGTM

PR #3224, opened 2022-09-16
Instructions to installing in Fedora via Copr

enmanuelmoreira:
Hi, I created a rpm package to install golangci-lint in Fedora systems via Copr. I modified the docs which has the instructions to install it.


ldez:
Hello,

It's a problem to add (and in the first place) an installation method that the team doesn't handle.

ldez:
For now, I will decline but I think we may use a Wiki to reference this kind of thing.
I need to think about it.

PR #3225, opened 2022-09-17
feat: add linter pairedbrackets

maratori:
`pairedbrackets` is a linter that checks formatting of paired brackets.
https://github.com/maratori/pairedbrackets

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -60,6 +60,7 @@ require (
 	github.com/ldez/tagliatelle v0.3.1
 	github.com/leonklingele/grouper v1.1.0
 	github.com/lufeee/execinquery v1.2.1
+	github.com/maratori/pairedbrackets v1.0.0
 	github.com/maratori/testableexamples v1.0.0
 	github.com/maratori/testpackage v1.1.0
 	github.com/matoous/godox v0.0.0-20210227103229-6504466cf951 // v1.0
@@ -179,7 +180,7 @@ require (
 	golang.org/x/exp/typeparams v0.0.0-20220827204233-334a2380cb91 // indirect
 	golang.org/x/mod v0.6.0-dev.0.20220419223038-86c51ed26bb4 // indirect
 	golang.org/x/sync v0.0.0-20220722155255-886fb9371eb4 // indirect
-	golang.org/x/sys v0.0.0-20220829200755-d48e67d00261 // indirect
+	golang.org/x/sys v0.0.0-20220909162455-aba9fc2a8ff2 // indirect
 	golang.org/x/text v0.3.7 // indirect
 	google.golang.org/protobuf v1.28.0 // indirect
 	gopkg.in/ini.v1 v1.67.0 // indirect
--- a/go.sum
+++ b/go.sum
@@ -345,6 +345,8 @@ github.com/lufia/plan9stats v0.0.0-20211012122336-39d0f177ccd0 h1:6E+4a0GO5zZEnZ
 github.com/lufia/plan9stats v0.0.0-20211012122336-39d0f177ccd0/go.mod h1:zJYVVT2jmtg6P3p1VtQj7WsuWi/y4VnjVBn7F8KPB3I=
 github.com/magiconair/properties v1.8.6 h1:5ibWZ6iY0NctNGWo87LalDlEZ6R41TqbbDamhfG/Qzo=
 github.com/magiconair/properties v1.8.6/go.mod h1:y3VJvCyxH9uVvJTWEGAELF3aiYNyPKd5NZ3oSwXrF60=
+github.com/maratori/pairedbrackets v1.0.0 h1:kTwEZfEJbJ/0aZ+YHHs5OkviIg2SlQqqQ+Q7DvQA+1Q=
+github.com/maratori/pairedbrackets v1.0.0/go.mod h1:OQnKQBTMTq+E2PCrQfhOkhMmbovXjOGd5xzUpawsbog=
 github.com/maratori/testableexamples v1.0.0 h1:dU5alXRrD8WKSjOUnmJZuzdxWOEQ57+7s93SLMxb2vI=
 github.com/maratori/testableexamples v1.0.0/go.mod h1:4rhjL1n20TUTT4vdh3RDqSizKLyXp7K2u6HgraZCGzE=
 github.com/maratori/testpackage v1.1.0 h1:GJY4wlzQhuBusMF1oahQCBtUV/AQ/k69IZ68vxaac2Q=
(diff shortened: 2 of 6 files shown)


ldez:
This kind of linter already exists inside golangci-lint.

maratori:
What linter does the same? I haven't seen it.

ldez:
At least: gofumpt
from my memories, there is another one.

maratori:
As I noted in [readme](https://github.com/maratori/pairedbrackets#gofumpt), `gofumpt` doesn't cover all cases.

For example it will not complain about following
<table>
<thead><tr><th>Bad</th><th>Good</th></tr></thead>
<tbody>
<tr><td>

```go
fmt.Printf("%s, %s!",
	"Hello", "world")
```
```go
fmt.Printf("%s, %s!",
	"Hello", "world",
)
```
```go
fmt.Printf(
	"%s, %s!",
	"Hello", "world")
```
```go
fmt.Printf("%s %s", "Last", `item
is multiline`,
)
```

</td><td>

```go
fmt.Printf("%s, %s!", "Hello", "world")
```
```go
fmt.Printf(
	"%s, %s!", "Hello", "world",
)
```
```go
fmt.Printf(
	"%s, %s!",
	"Hello", "world",
)
```
```go
fmt.Printf("%s %s", "Last", `item
is multiline`)
```

</td></tr>
</tbody></table>

ldez:
:thinking: 

Your example:
```
fmt.Printf("%s, %s!",
	"Hello", "world")
```
is not bad from my point, it's common when the pattern is long or to improve readability.

Otherwise, I'm checking the other linters.

maratori:
> is not bad from my point of view

I understand you, and I'm sure you are not alone.
But I'm not alone either. Before implementation, I've validated an idea with colleagues.

There are a lot of opinionated linters in golangci-lint. `pairedbrackets` is one of them.
That's why most of the linters are disabled by default.

ldez:
> There are a lot of opinionated linters in golangci-lint. pairedbrackets is one of them.
> That's why most of the linters are disabled by default.

yes there are opinionated linters but here I see more false-positive than opinion.

A lot of people (I think the majority) enable all linters and try to follow them.

So the default behavior of linter must be something that we can recommend.

ldez:
Just to be clear, I'm just chatting.

ldez:
@bombsimon what is your opinion?

maratori:
> Just to be clear, I'm just chatting

👌 

> I see more false-positive than opinion

I've run the linter on our code base and found false positives only in tests: `assert.Equal(...)`, `require.Equal(...)` and so on.
All of them are ignored by default, see [config `ignore-func-calls`](https://github.com/maratori/pairedbrackets#config).
Other issues are not false positives from my point of view.

> A lot of people (I think the majority) enable all linters and try to follow them.
> So the default behavior of linter must be something that we can recommend.

I can add a config to ignore all function calls by default.
What do you think about other examples? Like function declaration, or composite literals.
Are they also too strict?

ldez:
I think the "ignore function option calls" is not a good option, because I don't want to ignore rules with some functions I just want to ignore some recommendations.
For example, I can agree with examples 3 and 4 but not on 1 and 2.

Also, I don't understand the default exclusion of the testify functions.

ldez:
There are also some problems with the other elements, for example with Composite literal: if I want to have a slice with values organized by pair (like with the `strings.NewReplacer()`).

maratori:
I didn't get your point about composite literals.

Following is a good format according to `pairedbrackets`:
```go
strings.NewReplacer([]string{
	"a", "A",
	"b", "B",
}...)
```
It doesn't expect items to be formatted in one column. It just checks opening and closing brackets.

bombsimon:
I think this linter makes sense and as someone that in general appreciate consistent styling I would probably even use this myself (although that doesn't really matter). I think the reasoning around the linter, the link to the article and the clear examples is more than enough to argue for its use case. It's absolutely opinionated but like you both stated that goes for a lot of the linters in `golanci-lint`.

I agree `ignore-func-calls` is weird, I don't see why this shouldn't apply for all functions. I would rather see option(s) to allow some styles although I fully understand that's more work. Curious about the false positives, what were they?

I think I agree with all cases the linter enforces except the one that @ldez pointed out. To me, this is "balanced" and paired (left parenthesis and first argument on same line, last argument and parenthesis on same line). Maybe an option to allow this?

```go
fmt.Sprintf("This is too long for one line but ok for two %s: %d",
    someString, someInt)
```

Regarding `enable-all`; I actually don't think a majority use it. And for those who does, I think it's more common to discover new linters but if they're not relevant for the user they'll be disabled while still using the `enable-all` setting. At least that's what I do.

ldez:
> Regarding enable-all; I actually don't think a majority use it.

I think we need data about that because we are just expressing opinions :smile_cat:.
Maybe at some point, it can be a good idea to try to get information about the usage of golangci-lint.
But it's a bit off-topic.

> And for those who does, I think it's more common to discover new linters but if they're not relevant for the user they'll be disabled while still using the enable-all setting. 

I do that too, but I have seen different behavior with the handle of a new linter:
- if a linter doesn't have the "right" default behavior they use ignore it.
- some people just follow the rules because "linters say the truth".

bombsimon:
> I think we need data about that because we are just expressing opinions 😸.

You are right! 👍 

> This is why for me the default behavior is really important.

Makes sense and I agree. With that in mind maybe a default "relaxed" mode would be nice with the option to be more strict if desired.

From my point of view the only needed change would be to allow the above example by default and optionally disallow it for those who don't agree and care to read the docs. All the others, including the ones in the readme, makes sense and is easy to understand why they're there in my opinion. 

I guess such a relax mode would do the same for functions and composite literals as well although that makes way less sense to me personally. 😄 

```go
func Foo(a int,
	b string, c bool) {
	...
}

func Foo() (int,
	error) {
	...
}

foo := []int{1,
	2, 3}
```

maratori:
Google published their style guide recently. Looks like they use "relaxed" version of linter's rules ([link1](https://google.github.io/styleguide/go/decisions#matching-braces), [link2](https://google.github.io/styleguide/go/decisions#function-formatting)):

> The closing half of a brace pair should always appear on a line with the same amount of indentation as the opening brace.

Will you accept the linter to be a part of golangci-lint, if I change it?
1. By default the linter will follow the quote above
2. `ignore-func-calls` option will be removed
3. `strict` option will be added to follow the original rules

ldez:
Yes if your linter handles the boxes we talked about.
I am unsure about the term `strict` or the option itself but we will see during the review.

Antonboom:
What are the next steps here?

ldez:
I will close this PR because:
- No feedback for a long time from the PR author.
- The code of golangci-lint has changed since 2022.

Feel free to open a new PR, when the linter and the implementation inside golangci-lint will be updated.

PR #3227, opened 2022-09-20
feat: add linter todolint

timonwong:
todolint checks for possible TODOs in comments, and required TODO in the following form, for example:

- `// TODO(timonwong): This is ok`
- `// TODO: This is not ok, missing author`
- `// Todo: This is not ok, not caps`

Which makes it easier to find out who the actual author of the comment was (git blame is not enough after refactoring by other users).

https://github.com/timonwong/todolint

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -2047,6 +2047,7 @@ linters:
     - testableexamples
     - testpackage
     - thelper
+    - logrlint
     - tparallel
     - typecheck
     - unconvert
@@ -2154,6 +2155,7 @@ linters:
     - testableexamples
     - testpackage
     - thelper
+    - logrlint
     - tparallel
     - typecheck
     - unconvert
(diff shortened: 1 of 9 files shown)


ldez:
Hello,

I think that can be an improvement of [godox](https://github.com/matoous/godox)

ldez:
Hello,

I think that can be an improvement of [godox](https://github.com/matoous/godox)

timonwong:
> I think that can be an improvement of [godox](https://github.com/matoous/godox)

Hmmm yes, I'll try to fire some PRs to that repo instead :)

> When you work on a personal project you don't need the author.

Yes for personal projects there is no need to use this linter. So a better option is to add some config flags (by default disable) to "improved" godox.

ldez:
> It's ok to hold this PR

So I close this PR for now.

PR #3268, opened 2022-10-02
feat: add allfields linter

subtle-byte:
`allfields` is a linter that checks that all fields in the struct literal are set. It is useful when we need to fill the struct with data, if we add the field to the struct the `allfields` help us not forget to update struct literals.

https://github.com/subtle-byte/allfields

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -1951,6 +1951,7 @@ linters:
   # Enable specific linter
   # https://golangci-lint.run/usage/linters/#enabled-by-default
   enable:
+    - allfields
     - asasalint
     - asciicheck
     - bidichk
@@ -2058,6 +2059,7 @@ linters:
   # Disable specific linter
   # https://golangci-lint.run/usage/linters/#disabled-by-default
   disable:
+    - allfields
     - asasalint
     - asciicheck
     - bidichk
--- a/go.mod
+++ b/go.mod
@@ -94,6 +94,7 @@ require (
 	github.com/ssgreg/nlreturn/v2 v2.2.1
 	github.com/stbenjam/no-sprintf-host-port v0.1.1
 	github.com/stretchr/testify v1.8.0
+	github.com/subtle-byte/allfields v0.1.1
 	github.com/tdakkota/asciicheck v0.1.1
 	github.com/tetafro/godot v1.4.11
 	github.com/timakin/bodyclose v0.0.0-20210704033933-f49887972144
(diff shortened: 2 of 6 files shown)


ldez:

### Pull Request Description

### Linter

### `.golangci.reference.yml`

### Others Requirements

### Recommendations

---

ldez:
Hello,

> golangci-lint already has linter [exhaustruct](https://github.com/GaijinEntertainment/go-exhaustruct) which works with struct types, but allfields works with struct literals,

So it feels like a kind of duplicate of exhaustruct, I recommend adding this feature inside exhaustruct.

subtle-byte:
> Hello,
> 
> > golangci-lint already has linter [exhaustruct](https://github.com/GaijinEntertainment/go-exhaustruct) which works with struct types, but allfields works with struct literals,
> 
> So it feels like a kind of duplicate of exhaustruct, I recommend adding this feature inside exhaustruct.

I think these two linters resolve the same task, but in different ways (from different ends) and have quite few in common. And I think it's better to have 2 separate simple linters instead of building one more complex.

subtle-byte:
I think `allfields` implements the linter described in https://github.com/golang/go/issues/42940#issuecomment-737325869.

dozn:
Your lint require annotations if you _want_ the literal to be checked for exhaustiveness. This is error-prone compared to both `exhaustivestruct` and `exhaustruct` which test for exhaustiveness by default, and instead use directives to _disable_ the check.

Someone should probably look into why we have two lints which do the same thing in the first place, as I have a decent amount of `//nolint:exhaustivestruct,exhaustruct` comments strewn about.

subtle-byte:
> Your lint requires annotations if you _want_ the literal to be checked for exhaustiveness. This is error-prone compared to `exhaustruct` which test for exhaustiveness by default, and instead use directives to _disable_ the check.
> 
> If you don't want files generated via protobuf to be checked, you can disable `exhaustruct` from running on those files using `golangci-lint`'s `exclude-rules`.

Yeah, your words make sense. My idea was that you will use allfields quite rarely - in bug-prone places. And usage of exhaustruct is quite a pain because all libs are developed knowing that in Go we can omit fields during struct creation, and libs use it, e.g. struct for storing configs, or http.Client where you might want to set only Transport and don't care about the rest.

dozn:
> And usage of exhaustruct is quite a pain because all libs are developed knowing that in Go we can omit fields during struct creation, and libs use it, e.g. struct for storing configs, or http.Client where you might want to set only Transport and don't care about the rest.

You can ignore those structs using `linters-settings` like so:

```yaml
linters-settings:
  exhaustruct:
    exclude: '.*net/http\.Client.*'
```

ldez:
For me, it's a duplicate of exhaustruct, the approach is different but this can be handled by exhaustruct with a kind of "mode".

I will decline the PR for now.

PR #3308, opened 2022-10-21
Retain exitcode after fixing issues with the -fix args

Abirdcfly:
After using the `-fix` parameter to turn on automatic issue repair, if all the issues found are repaired, the `exitcode` will be `0`. 

This behavior is not written in the documentation.

There are two directions to modify this issue:
- if we think this behavior is correct, then we need to state it in the documentation
- if this behavior is not correct, we need to modify the code.


ldez:
Hello, it's the expected behavior, you can improve the doc if you want.

PR #3348, opened 2022-11-06
feat: add uncalled linter

stevenh:
Add [uncalled linter](https://github.com/stevenh/go-uncalled) which checks for missing calls based on configured rules and supports generics.


stevenh:
While this implements the same functionality as [rowserrcheck](https://github.com/jingyugao/rowserrcheck) that linter uses [buildssa](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/buildssa) which is currently incompatible with [go generics](https://go.dev/doc/tutorial/generics) as [buildssa.SSA.SrcFuncs](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/buildssa#SSA) doesn't currently include [Instantiations](https://github.com/golang/go/issues/48525#issuecomment-1128385150).

This is a from scratch implementation that uses [inspect.Analyzer](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/inspect) which doesn't have such problems. It also doesn't suffer from the false positives when the code has multiple reassigned `Row` variables that don't result in a real check. 

Credit to [rowserrcheck](https://github.com/jingyugao/rowserrcheck) for inspiration.

ldez:

### Pull Request Description

### Linter

### `.golangci.reference.yml`

### Others Requirements

### Recommendations

---

stevenh:
I've actually refactored the original implementation which was limited to checking for missing [database/sql](https://go.dev/pkg/database/sql/) [Rows.Err()](https://go.dev/pkg/database/sql/#Rows.Err) to use a configuration that will allow it to check for any number of missing calls.

Working my way though validating your check list.

stevenh:
@ldez thanks for the checklist, very helpful. I did notice it has more details than [new linters guide](https://golangci-lint.run/contributing/new-linters/) so I'll try to find some time to update that if it would help.

I believe all the items on your list are good, but look forward to your feedback.

stevenh:
@ldez any feedback I need to address?

stevenh:
I see this is now tagged at blocked @ldez anything that I need to address?

I just updated to the latest version and rebased so it's in line with current master.

ldez:
I blocked the PR because I think this linter is a kind of duplicate but I need to investigate.

stevenh:
It does duplicate the functionality of [rowserrcheck](https://github.com/jingyugao/rowserrcheck) as mentioned in previous comment, but it uses [buildssa](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/buildssa) which is currently incompatible with [go generics](https://go.dev/doc/tutorial/generics) as [buildssa.SSA.SrcFuncs](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/buildssa#SSA) so users currently have no way to check for that issue if they are using go 1.18+

This is a from scratch implementation that uses [inspect.Analyzer](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/inspect) which doesn't have such problems. It also doesn't suffer from the false positives when the code has multiple reassigned Row variables that don't result in a real check.

Finally this uses an extensible model based on configuration, providing three separate checks currently:
* sql-rows-err - Check for missing sql Rows.Err() calls.
* http-response-body-close - Check for missing http Reponse.Body.Close() calls.
* context-cancel - Check for missing context CancelFunc() calls.

ldez:
SSA supports generics for about 1 year https://github.com/golangci/golangci-lint/issues/2649#issue-1170906525

And since go1.20, go1.18 is unmaintained.

Maybe I miss something here :thinking:

stevenh:
> SSA supports generics for about 1 year [#2649 (comment)](https://github.com/golangci/golangci-lint/issues/2649#issue-1170906525)
> 
> And since go1.20, go1.18 is unmaintained.
> 
> Maybe I miss something here 🤔

Thanks for checking back in on this.

When I wrote this SSA didn't have full generics support specifically it didn't support [instantiations](https://github.com/golang/go/issues/48525#issuecomment-1128385150) causing [rowserrcheck](https://github.com/jingyugao/rowserrcheck) to be disabled in go 1.18+ which have generics. I only mentioned 1.18 as that's when generics was introduced.

I just did a check with golangci-lint v1.51.2 and [rowserrcheck](https://github.com/jingyugao/rowserrcheck) is still disabled when running with go 1.20:
> golangci-lint run
WARN [linters_context] rowserrcheck is disabled because of generics. You can track the evolution of the generics support by following the https://github.com/golangci/golangci-lint/issues/2649.

Looking at [rowserrcheck](https://github.com/jingyugao/rowserrcheck) it hasn't been updated since 2021, so unfortunately still isn't working with generics, although should be easier to fix now.

uncalled, also has the ability to be extended with additional user checks via extra configuration so I think it still has value to the community, thoughts?

stevenh:
I believe the only concern is that it duplicates the check in `rowserrcheck` however as explained above that's currently unavailable if you're using golang v1.18 or above due to generics. 

uncalled is already fully integrated with `golangci-lint` the PR just needs rebase to resolve the conflicts, which I'm happy to do if it would be accepted.

ccoVeille:
Thanks for the explanation.

From my understanding, it doesn't duplicate it. It supersedes it.

I would be able not only to restore the feature of rowserrcheck on version above go1. 18

It will allow to provide way to define custom rules. 

rowserrcheck is only one possible things it can do.

ldez:
FYI rowserrcheck will be re-enabled https://github.com/golangci/golangci-lint/pull/3691

ccoVeille:
Thanks for the information.

Does it reduce the chance to get unused integrated into golangci-lint ?

ldez:
> Does it reduce the chance to get `uncalled` integrated into golangci-lint?

Yes because duplicated linters are not accepted, but I have to evaluate pro/cons to replace `rowserrcheck` with `uncalled`.

Antonboom:
@stevenh, hi!
For me `uncalled` seems very cool generic linter, thank you for job.

@ldez, looks like `uncalled` could make deprecated:

1. https://github.com/timakin/bodyclose
2. https://github.com/jingyugao/rowserrcheck
3. https://github.com/ryanrolds/sqlclosecheck
4. https://github.com/ykadowak/zerologlint (probably)

The killer feature is support not only std types 😎

Antonboom:
But I confused from `github.com/rs/zerolog` dependency in the linter 😔

stevenh:
> But I confused from `github.com/rs/zerolog` dependency in the linter 😔

It's there to add structured logging with control over the level when running checks. Now std lib has [slog](https://pkg.go.dev/log/slog) I'll likely look to migrate to that.

stevenh:
> @stevenh, hi! For me `uncalled` seems very cool generic linter, thank you for job.

Thanks :) 
 
> @ldez, looks like `uncalled` could make deprecated:
> 
> 1. https://github.com/timakin/bodyclose
> 2. https://github.com/jingyugao/rowserrcheck
> 3. https://github.com/ryanrolds/sqlclosecheck
> 4. [xmlencoderclose: linter that checks xml.Encoder is closed #3895](https://github.com/golangci/golangci-lint/pull/3895)
> 5. https://github.com/ykadowak/zerologlint (probably)
> 
> The killer feature is support not only std types 😎
> 
> P.S. @stevenh could you provide an example of `uncalled` config to prove my list above?

bodyclose and rowserrcheck are already present in the [config here](https://github.com/stevenh/go-uncalled/blob/main/pkg/uncalled/.uncalled.yaml)

`sqlclose` would looks something like:
```yaml
  # Check for missing sql Rows.Close() calls.
  - name: sql-rows-close
    disabled: false
    category: sql
    packages:
      - database/sql
      - github.com/jmoiron/sqlx
    methods: []
    results:
      - type: .Rows
        pointer: true
        expect:
          call: .Close
          args: []
      - type: error
        pointer: false
  # Check for missing sql Stmt.Close() calls.
  - name: sql-stmt-close
    disabled: false
    category: sql
    packages:
      - database/sql
      - github.com/jmoiron/sqlx
    methods: []
    results:
      - type: .Stmt
        pointer: true
        expect:
          call: .Close
          args: []
      - type: error
        pointer: false
```

`xmlencoderclose` would looks something like:
```yaml
  # Check for missing xml Encoder.Close() calls.
  - name: xml-encoder-close
    disabled: false
    category: xml
    packages:
      - encoding/xml
    methods: []
    results:
      - type: .Encoder
        pointer: true
        expect:
          call: .Close
          args: []
      - type: error
        pointer: false
```

`zerologlint` looks a little more complex as there are multiple ways to trigger the dispatch, but this should also be possible.

If this is of interest, I can look at creating tests to confirm the above?

stevenh:
Looks like [Rows.Close](https://pkg.go.dev/database/sql#Rows.Close) is more complex as if Next is called and returns false then Rows is closed implicitly, so will need some more thought.

stevenh:
Looking at it https://github.com/ryanrolds/sqlclosecheck suffers from the false positive when you have checked `rows.Next()` and `rows.Err()` for example:
```go
func Called(db *sql.DB) {
	rows, _ := db.Query("select id from tb")
	for rows.Next() {
		// Handle row.
	}
	if rows.Err() != nil {
		// Handle error.
		fmt.Fprintln(ioutil.Discard, "error")
	}
}
```

PR #3353, opened 2022-11-10
Add timeformat to govet analysers

johejo:
https://github.com/golang/go/issues/48801

Like go vet it is not enabled by default.


johejo:
It is disabled by default in go vet but seems to be enabled by default in gopls.

Issue #3354, opened 2022-11-10
The "unused" linting rule has misleading behaviour when interacting with generated code


adamroyjones:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [X] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

### Description of the problem

One of the default linters, unused, is described on [the golangci-lint site](https://golangci-lint.run/usage/linters/) as something that

> [c]hecks Go code for unused constants, variables, functions and types

but if a constant, variable, function, or type is only referred to in a generated file, then golangci-lint will act in a misleading way.

I _think_ this should be categorised as a bug, but it may be that I've missed a piece of configuration that I can set.

### Version of golangci-lint

```console
$ golangci-lint --version
golangci-lint has version 1.50.1 built from 8926a95f on 2022-10-22T10:50:47Z
```

### Configuration file

```console
$ cat .golangci.yml
---
run:
  timeout: 20m

linters:
  disable-all: true
  enable:
    - unused
```

### Go environment

The reproducing example at the end uses a Dockerfile.

### Verbose output of running

The reproducing example at the end uses a Dockerfile.

### Code example or link to a public repository

The "details" block below describes the following directory structure (which I've also attached as a tarball [here](https://github.com/golangci/golangci-lint/files/9982882/golangci-lint-reprex.tar.gz)):

```
.
├── bar.go
├── Dockerfile
├── foo.go
├── .golangci.yml
├── go.mod
└── rc.sh
```

First, recreate this directory structure and set `rc.sh` to be executable.

1. If you run `./rc.sh`, then golangci-lint will report that the function `foo` is unused.

2. If you remove the first line from foo/bar.go (that says the file is generated) and then run `./rc.sh`, then golangci-lint will now report that both `foo` and `bar` are unused.

3. If you replace `bar` with `Bar`, then, irrespective of the line at the top about the file being generated being present, everything is (correctly) declared as being used.

In my view, the messages in 1 and 2 are misleading. In both cases, the issue is that `bar` isn't public (and therefore isn't used); `foo` is only unused as a consequence of `bar` not being used. This is especially awkward in 1, because there's no way to resolve the problem from the error message alone.

<details>

```go
// bar.go
// Code generated by hand. DO NOT EDIT.
package foo

[...]

PR #3355, opened 2022-11-10
chore: Add a clarifying note to the bug report template

adamroyjones:
As I wasted @ldez's time in issue #3354, the least I can do is add a note to the issue template that should disambiguate things a touch and prevent this from reoccurring.


ldez (approved):
LGTM

PR #3377, opened 2022-11-26
doc: cache preserving and colored output on docker runs

nawa:


ldez:
@alexandear you don't need to update the PRs. And if it's needed can you use rebase instead of merge? thank you.

codyaray:
Nice, thanks for the pro-tips on using docker here -- exactly what I was looking for!

ldez (approved):
LGTM

PR #3423, opened 2022-12-18
doc: usage typo in tagliatelle

freitzzz:
I believe *struck* is a typo of **struct**. Sorry if spammy PR, but that word is tickling my brain :)


ldez (approved):
LGTM

PR #3466, opened 2023-01-09
feat: add constantcheck linter

AntonBraer:
https://github.com/AntonBraer/constantcheck
constantcheck is a golang analyzer for checking the possibility of replacing the literals used in the code with constants from already imported packages.


Antonboom:
Covered by https://github.com/sashamelentyev/usestdlibvars (@sashamelentyev)?

AntonBraer:
My linter finds not only standard library constants, but also constants from any imported package

Antonboom:
Hm.. to be honest – covered, but not complete, std only.

But I cannot imagine case like [this](https://github.com/AntonBraer/constantcheck#examples) in the real life 🤷.

AntonBraer:
I can give an example when you have a large project in which you use the USD currency. 
Somewhere you compare the currency with a constant that you have declared in one of the packages, somewhere directly with the “USD” literal.
To avoid this just the same, you can use my linter
I can say from my work that such cases popped up in our projects, which is why we needed to write such a linter
Unfortunately, I can't show you the production code.

Antonboom:
@AntonBraer thank u for example 👍 
Did you try to use https://github.com/jgautheron/goconst linter? :)

Antonboom:
Sorry, it was misclick ^

Antonboom:
Thank u for example!
Maybe [goconst](https://github.com/jgautheron/goconst) could help u.

Anyway, no more questions from my side 👍

AntonBraer:
I try goconst, but as far as I know it doesn't cover all cases. I don't know why, but it's true.
Just compared the outputs of both linters

ldez:
Hello,

Your linter and [`usestdlibvars`](https://github.com/sashamelentyev/usestdlibvars) share the exact same goal: detecting strings that can be replaced by an existing constant.

[`goconst`](https://github.com/jgautheron/goconst) will detect if a string is used multiple times and suggest creating a constant.

The 3 linters are on the same global topic: constants and their use.

Your linter can be seen as an enhancement of `usestdlibvars` or `goconst`.

For the moment, I am torn between suggesting a PR on one of the linter or adding your linter.

But I am sure about the fact that your linter must not do the same things as the other linters or it must handle all the cases:
- detection of std constants (`usestdlibvars`)
- detection of the repetition of strings (`goconst`)

AntonBraer:
Hey, I think that taking the functionality of two linters inside mine is not entirely fair to the creators of these linters. So my pr can be closed, and the functionality of my linter can be inserted into any of the two linters.

PR #3493, opened 2023-01-20
Update LICENSE to Apache 2.0

oksana-grishchenko:
Update LICENSE to Apache 2.0 

The golangci-lint is a great tool we used to use actively at Percona, however the GPL license requires to provide all the source code under the GPL license too, which violates our legal requirements. Our open source tools https://github.com/orgs/percona/repositories are distributed under Apache 2.0, so with this PR I'd like to suggest to use this license too. I really hope we could continue using this amazing tool!

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,674 +1,203 @@
-                    GNU GENERAL PUBLIC LICENSE
-                       Version 3, 29 June 2007
-
- Copyright (C) 2007 Free Software Foundation, Inc. <https://fsf.org/>
- Everyone is permitted to copy and distribute verbatim copies
- of this license document, but changing it is not allowed.
-
-                            Preamble
-
-  The GNU General Public License is a free, copyleft license for
-software and other kinds of works.
-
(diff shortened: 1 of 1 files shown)


ldez:
Hello,
We will not change the license.

GPL is a good license.

ldez:
Hello,

We will not change the license:
- GPL is a good license
- legally we cannot change le license

GPL doesn't require to provide of all the source code under the GPL license.
GPL says that if you made changes on the golangci-lint code you have to open source your changes.

oksana-grishchenko:
Hi @ldez! Thanks for the clarifications, actually it's great to know that. I'll discuss it again with our legal team.

ldez:
Just in case, golangci-lint depends on libraries using GPL and you too :wink: 

Example:
https://deps.dev/go/github.com%2Fpercona%2Fmongodb_exporter

oksana-grishchenko:
Exactly. That's why now I think to revise the policy is the way to resolve the problem. Thanks for your help!

PR #3513, opened 2023-01-28
chore: fix latest install to cater for existing users who installed using '-s latest'

### LINKED ISSUE #3512: Way to install latest release from install.sh has changed

Hi we installed the latest version golangci using the following command `wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s latest`, with the latest commit from https://github.com/golangci/golangci-lint/commit/d57156ec228b712ccd429367482771179e3fa050, this won't work anymore, could line 320 `if [ -z "$version" ] then` be changed to  `if [ -z "$version" ] || [ "$version" = "latest" ]; then` to allow the previous command to work as it should, I suspect this might be affecting many users?

Instead of doing `if [ -z "$version" ] then` on line 320, could you do `if [ -z "$version" ] || [ "$version" = "latest" ]; then`?

Alternative here would to be to update our pipelines to the new way of installing latest - `wget -O- -nv https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh` but again I'm assuming this must be affecting many users.

### Additional context.

Many people are using this in their pipelines and it could be affecting them negatively.

herman26:
This is affecting existing users.

Fixes #3512

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -317,7 +317,7 @@ http_copy() {
 github_release() {
   owner_repo=$1
   version=$2
-  if [ -z "$version" ]; then
+  if [ -z "$version" ] || [ "$version" = "latest" ]; then
       giturl="https://api.github.com/repos/${owner_repo}/releases/latest"
   else
       giturl="https://api.github.com/repos/${owner_repo}/releases/tags/${version}"


ldez (approved):
LGTM

Issue #3511, opened 2023-01-28
install.sh script broken in main


armsnyder:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc.).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

To reproduce, run:

```shell
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- v1.50.1
```

Result:

```
golangci/golangci-lint info checking GitHub for tag 'v1.50.1'
golangci/golangci-lint info found version:  <!DOCTYPE html> <html lang= for  <!DOCTYPE html> <html lang=/darwin/arm64
```

(exits with code 3)

### Version of golangci-lint

n/a

### Configuration file

n/a

### Go environment

n/a

### Verbose output of running

n/a

### Code example or link to a public repository

n/a

armsnyder:
And it's back to working again. Maybe https://github.com/golangci/golangci-lint/commit/d57156ec228b712ccd429367482771179e3fa050 resolved it? I'll close the issue.

ldez:
duplicate of #3509

PR #3510, opened 2023-01-28
chore: fix install.sh to point at releases API

mfine:
Uses the releases API to validate a release.

Fixes #3509

Issue #3509, opened 2023-01-27
Golangci-lint downloader fails to download binary


ssko1:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc.).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

The downloader decscribed in the installation guide https://golangci-lint.run/usage/install/#other-ci returns with an error. It does not install the golangci-lint binary into my go path.

Perhaps the Github API changed?

### Version of golangci-lint

N/A

### Configuration file

N/A

### Go environment

N/A

### Verbose output of running
```
computer@computer directory % curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.50.1

golangci/golangci-lint info checking GitHub for tag 'v1.50.1'
golangci/golangci-lint info found version:  <!DOCTYPE html> <html lang= for  <!DOCTYPE html> <html lang=/darwin/amd64
computer@computer directory %
```
### Code example or link to a public repository

N/A

mdestagnol:
Same problem here, since this morning. The script mentioned in the [install](https://golangci-lint.run/usage/install/) page is broken. It seems that it's fetching from a URL which is not returning what it used to

ssko1:
The script performs a check against github to obtain a tag... something like:
```
curl https://github.com/golangci/golangci-lint/releases/v1.49.0 -H "Accept: application/json"
```

This now returns a webpage instead of JSON like it was before?

mdestagnol:
could that be that the github download format has changed ? 
from https://github.com/golangci/golangci-lint/releases/download/v1.50.1/golangci-lint-v1.50.1-darwin-arm64.tar.gz
to https://github.com/golangci/golangci-lint/releases/download/v1.50.1/golangci-lint-1.50.1-darwin-arm64.tar.gz

(note the v in the file name)

mfine:
Maybe they turned off that access path in favor of using the [API](https://docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28#get-a-release-by-tag-name)?

```sh
[...]

PR #3514, opened 2023-01-28
chore: fix install script (revert)

ldez:
GitHub engineers have pushed a change that has reverted the bug, so we can revert the changes.

https://github.com/orgs/community/discussions/45590#discussioncomment-4802887

The problem with the changes is the use of the API, this requires a GitHub token to avoid rate limits.

https://github.com/orgs/community/discussions/45590#discussioncomment-4802681

related to #3509, #3510, #3511, #3512, #3513.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -317,14 +317,11 @@ http_copy() {
 github_release() {
   owner_repo=$1
   version=$2
-  if [ -z "$version" ]; then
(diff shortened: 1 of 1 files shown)


ldez:
I consider this as high priority fix, and as this is just a revert, I will merge without review.
This is exceptional given the situation.

PR #3516, opened 2023-01-28
docs: update output example to use valid checkstyle example; add json example

nickfiggins:
the current example listed for outputs leads to checkstyle XML being saved to a .json file, which is a bit confusing since the output is xml. this PR updates the example to use xml for the checkstyle report, and send json to stdout to still show those capabilities


ldez (approved):
LGTM

PR #3414, opened 2022-12-12
go1.20 support

ldez:
This PR adds the support of go1.20.

Note for GitHub Action users: the Go version must be quoted (`'1.20'`) otherwise it is interpreted as a float (`1.2`).

~~This PR is to evaluate and prepare golangci-lint to [go1.20](https://tip.golang.org/doc/go1.20).~~

~~This PR will evolve during the beta and RC phases of [go1.20](https://tip.golang.org/doc/go1.20).~~

Fixes #3420

PR #3534, opened 2023-02-02
feat: support Go 1.20

sivchari:
This PR is to evaluate and prepare golangci-lint to [go1.20](https://tip.golang.org/doc/go1.20).


ldez:
https://github.com/golangci/golangci-lint/pull/3414

PR #3540, opened 2023-02-02
dev: use Go1.20 to run tests with the latest golangci-lint version

cia-rana:
temporarily used go1.19 explicitly in 5afb30ed542f37b8099d2ec4c5aeefdb2835e7d4


ldez:
It's TODO for me.

So I will do it.

PR #3541, opened 2023-02-02
docs: revert url to gochecknoinits

maratori:
This PR reverts URL to `gochecknoinits` removed by mistake in #3414.


ldez:
I was not removed by mistake: we don't use the lib.

PR #3543, opened 2023-02-02
Upgrade deps

sylr:
Includes a dev version of honnef.co/go/tools which made a big refactoring of the unused linter.


ldez:
We use a bot to handle dependencies.
we will wait for a release before updating staticcheck.

PR #3571, opened 2023-02-06
fix: consider path prefix when matching path patterns

pohly:
### Problem

When someone invokes golangci-lint in a sub-directory, they can set the path prefix to make the output look like the invocation had been done in the root.
But this prefix was ignored when checking path patterns of exclude, severity, and skip files/dirs entries, so those with matching for directories only worked when golangci-lint was always invoked in the same directory.

### Solution
    
The underlying problem is that the rules from the configuration get checked before updating the path associated with the issues in the path fixer.
To make the prefix work, it now gets passed down into processors and added there to the issue path before checking against the path pattern.
    
For exclude and severity rules, this could have been done through a separate parameter, but bundling it in a new `fsutils` helper struct made the change a bit smaller.

### DIFF
--- a/docs/src/docs/usage/false-positives.mdx
+++ b/docs/src/docs/usage/false-positives.mdx
@@ -66,6 +66,11 @@ issues:
 
 Exclude issues in path by `run.skip-dirs`, `run.skip-files` or `issues.exclude-rules` config options.
 
+Beware that the paths that get matched here are relative to the current working
+directory. When the configuration contains path patterns that check for specific
+directories, the `--path-prefix` parameter can be used to extend the paths
+before matching.
+
 In the following example, all the reports from the linters (`linters`) that concerns the path (`path`) are excluded:
 
 ```yml
--- a/docs/src/docs/usage/false-positives.mdx
+++ b/docs/src/docs/usage/false-positives.mdx
@@ -66,10 +66,9 @@ issues:
 
 Exclude issues in path by `run.skip-dirs`, `run.skip-files` or `issues.exclude-rules` config options.
 
-Beware that the paths that get matched here are relative to the current working
-directory. When the configuration contains path patterns that check for specific
-directories, the `--path-prefix` parameter can be used to extend the paths
-before matching.
+Beware that the paths that get matched here are relative to the current working directory.
+When the configuration contains path patterns that check for specific directories,
+the `--path-prefix` parameter can be used to extend the paths before matching.
 
 In the following example, all the reports from the linters (`linters`) that concerns the path (`path`) are excluded:
 
--- a/pkg/fsutils/files.go
+++ b/pkg/fsutils/files.go
@@ -1,0 +1,36 @@
+package fsutils
+
+import "path/filepath"
+
+// Files combines different operations related to handling file paths and
+// content.
+type Files struct {
+	*LineCache
+	pathPrefix string
+}
+
+func NewFiles(lc *LineCache, pathPrefix string) *Files {
+	return &Files{
+		LineCache:  lc,
+		pathPrefix: pathPrefix,
+	}
+}
+
+// WithPathPrefix takes a path that is relative to the current directory (as
+// used in in issues) and adds the configured path prefix, if there is one. The
+// resulting path then can be shown to the user or compared against paths
+// specified in the configuration.
+func (f *Files) WithPathPrefix(relativePath string) string {
+	return WithPathPrefix(f.pathPrefix, relativePath)
+}
+
+// WithPathPrefix takes a path that is relative to the current directory (as
+// used in in issues) and adds the configured path prefix, if there is one. The
+// resulting path then can be shown to the user or compared against paths
+// specified in the configuration.
+func WithPathPrefix(pathPrefix, relativePath string) string {
+	if pathPrefix == "" {
+		return relativePath
+	}
+	return filepath.Join(pathPrefix, relativePath)
+}
--- a/pkg/fsutils/files.go
+++ b/pkg/fsutils/files.go
@@ -2,8 +2,7 @@ package fsutils
 
 import "path/filepath"
 
-// Files combines different operations related to handling file paths and
-// content.
+// Files combines different operations related to handling file paths and content.
 type Files struct {
 	*LineCache
 	pathPrefix string
@@ -16,18 +15,16 @@ func NewFiles(lc *LineCache, pathPrefix string) *Files {
 	}
 }
 
-// WithPathPrefix takes a path that is relative to the current directory (as
-// used in in issues) and adds the configured path prefix, if there is one. The
-// resulting path then can be shown to the user or compared against paths
-// specified in the configuration.
+// WithPathPrefix takes a path that is relative to the current directory (as used in issues)
+// and adds the configured path prefix, if there is one.
+// The resulting path then can be shown to the user or compared against paths specified in the configuration.
 func (f *Files) WithPathPrefix(relativePath string) string {
 	return WithPathPrefix(f.pathPrefix, relativePath)
 }
 
-// WithPathPrefix takes a path that is relative to the current directory (as
-// used in in issues) and adds the configured path prefix, if there is one. The
-// resulting path then can be shown to the user or compared against paths
-// specified in the configuration.
+// WithPathPrefix takes a path that is relative to the current directory (as used in issues)
+// and adds the configured path prefix, if there is one.
+// The resulting path then can be shown to the user or compared against paths specified in the configuration.
 func WithPathPrefix(pathPrefix, relativePath string) string {
 	if pathPrefix == "" {
 		return relativePath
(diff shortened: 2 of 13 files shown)


pkg/lint/runner.go
```diff
@@ -83,7 +83,7 @@ func NewRunner(cfg *config.Config, log logutils.Log, goenv *goutil.Env, es *lint
 			processors.NewIdentifierMarker(),
 
 			getExcludeProcessor(&cfg.Issues),
-			getExcludeRulesProcessor(&cfg.Issues, log, lineCache),
+			getExcludeRulesProcessor(&cfg.Issues, log, cfg.Output.PathPrefix, lineCache),
```
pohly, line 86: It looks like path prefix and line cache get passed through different functions together. To minimize changes I had originally made "path prefix" a field in the line cache (or rather, the file cache, with an accessor in the line cache).

That didn't work because the caches got created before parsing the configuration. I would have had to add a `SetPathPrefix` to add that later, which didn't feel right.

Do you have any preference?

pkg/result/processors/base_rule.go
```diff
@@ -26,14 +27,14 @@ func (r *baseRule) isEmpty() bool {
...
-func (r *baseRule) match(issue *result.Issue, lineCache *fsutils.LineCache, log logutils.Log) bool {
+func (r *baseRule) match(issue *result.Issue, pathPrefix string, lineCache *fsutils.LineCache, log logutils.Log) bool {
 	if r.isEmpty() {
 		return false
 	}
 	if r.text != nil && !r.text.MatchString(issue.Text) {
 		return false
 	}
-	if r.path != nil && !r.path.MatchString(issue.FilePath()) {
+	if r.path != nil && !r.path.MatchString(path.Join(pathPrefix, issue.FilePath())) {
```
pohly, line 37: If `pathPrefix` is empty, `pathPrefix` will ignore it. But the result is "cleaned" - I am not sure whether that is potentially harmful (performance, Windows separator?).

If "path prefix" was an attribute in `lineCache`, then a `LineCache.WithPathPrefix(p string)` method could be added to do the extension of the path in a consistent way.

pkg/result/processors/exclude_rules_test.go
```diff
@@ -37,7 +37,7 @@ func TestExcludeRulesMultiple(t *testing.T) {
 				Linters: []string{"lll"},
 			},
 		},
-	}, lineCache, nil)
+	}, "" /* path prefix */, lineCache, nil)
```
pohly, line 40: I'll add unit tests once I know how this is supposed to be handled.

pkg/lint/runner.go
pohly: In the end I went for an approach where line cache and path prefix get combined. This keeps the number of parameters that need to be passed through the same.

pkg/result/processors/base_rule.go
pohly: `path.Join` is correct. The path prefixer does the same. To handle Windows, the path regex gets mangled to match the separator used by the platform.

pkg/result/processors/exclude_rules_test.go
pohly: I've added them.

pkg/lint/runner.go
pohly: Here's a diff against the earlier approach with two parameters: https://github.com/golangci/golangci-lint/compare/9664d89f477ef60739fe6ec54792b50d43099e71..d7853eaed135202c0aff13ccd6db22abe8e27d56

pohly:
@ldez: I also looked at `skip-files` and `skip-dirs`. Both had the same problem, so I extended the PR to also fix those.

I think this is ready for review now.

pohly:
In case someone wonders: the reason why the directory has to be changed in the Kubernetes repo is that `staging/src/k8s.io/*` are all separate modules, so `golangci-lint ./staging/src/k8s.io/...` wouldn't work. But we want to maintain a single golangci.yaml for all modules with path matches for certain directories, hence this PR...

pohly:
The path prefixer test failed on Windows because I was using "path.Join", not "path/filepath.Join" as before. This change should fix that:
```patch
diff --git a/pkg/fsutils/files.go b/pkg/fsutils/files.go
index 542af531..3d369a23 100644
--- a/pkg/fsutils/files.go
+++ b/pkg/fsutils/files.go
@@ -1,6 +1,6 @@
 package fsutils
 
-import "path"
+import "path/filepath"
 
 // Files combines different operations related to handling file paths and
 // content.
@@ -32,5 +32,5 @@ func WithPathPrefix(pathPrefix, relativePath string) string {
        if pathPrefix == "" {
                return relativePath
        }
-       return path.Join(pathPrefix, relativePath)
+       return filepath.Join(pathPrefix, relativePath)
 }
```

docs/src/docs/usage/false-positives.mdx
```diff
@@ -66,6 +66,11 @@ issues:
 
 Exclude issues in path by `run.skip-dirs`, `run.skip-files` or `issues.exclude-rules` config options.
 
+Beware that the paths that get matched here are relative to the current working
+directory. When the configuration contains path patterns that check for specific
+directories, the `--path-prefix` parameter can be used to extend the paths
+before matching.
+
```
ldez, line 73:
```suggestion
Beware that the paths that get matched here are relative to the current working directory.
When the configuration contains path patterns that check for specific directories,
the `--path-prefix` parameter can be used to extend the paths before matching.

```

pkg/fsutils/files.go
```diff
@@ -0,0 +1,36 @@
+package fsutils
+
+import "path/filepath"
+
+// Files combines different operations related to handling file paths and
+// content.
```
ldez, line 6:
```suggestion
// Files combines different operations related to handling file paths and content.
```

pkg/fsutils/files.go
```diff
@@ -0,0 +1,36 @@
...
+	return &Files{
+		LineCache:  lc,
+		pathPrefix: pathPrefix,
+	}
+}
+
+// WithPathPrefix takes a path that is relative to the current directory (as
+// used in in issues) and adds the configured path prefix, if there is one. The
+// resulting path then can be shown to the user or compared against paths
+// specified in the configuration.
```
ldez, line 22:
```suggestion
// WithPathPrefix takes a path that is relative to the current directory (as used in issues)
// and adds the configured path prefix, if there is one.
// The resulting path then can be shown to the user or compared against paths specified in the configuration.
```

pkg/fsutils/files.go
```diff
@@ -0,0 +1,36 @@
...
+func (f *Files) WithPathPrefix(relativePath string) string {
+	return WithPathPrefix(f.pathPrefix, relativePath)
+}
+
+// WithPathPrefix takes a path that is relative to the current directory (as
+// used in in issues) and adds the configured path prefix, if there is one. The
+// resulting path then can be shown to the user or compared against paths
+// specified in the configuration.
```
ldez, line 30:
```suggestion
// WithPathPrefix takes a path that is relative to the current directory (as used in issues)
// and adds the configured path prefix, if there is one.
// The resulting path then can be shown to the user or compared against paths specified in the configuration.
```

pkg/lint/runner.go
```diff
@@ -29,7 +29,12 @@ type Runner struct {
 
 func NewRunner(cfg *config.Config, log logutils.Log, goenv *goutil.Env, es *lintersdb.EnabledSet,
 	lineCache *fsutils.LineCache, dbManager *lintersdb.Manager, pkgs []*gopackages.Package) (*Runner, error) {
-	skipFilesProcessor, err := processors.NewSkipFiles(cfg.Run.SkipFiles)
+	// Beware that some processors need to add the path prefix when working
+	// with paths because they get invoked before the path prefixer
+	// (exclude and severity rules) or process other paths (skip files).
```
ldez, line 34:
```suggestion
	// Beware that some processors need to add the path prefix when working with paths
	// because they get invoked before the path prefixer (exclude and severity rules)
	// or process other paths (skip files).
```

ldez (changes requested):
I think your approach can be extended to the Fixer processor, can try to fix the Fixer?

pohly:
> I think your approach can be extended to the Fixer processor, can try to fix the Fixer?

The problem with the fixer processor seems to be that it runs *after* the path prefixer (https://github.com/golangci/golangci-lint/pull/3626).

That leads to the inverse problem of what I am solving here: the fixer sees the path with prefix which doesn't match the correct path relative to the current working directory, whereas severity and exclude processors (before this PR) see the path without the prefix when they need it with prefix.

The fix for the fixer in the PR above is to strip the prefix again. That doesn't seem ideal. Perhaps a simpler solution is to move the [path prefixing and sorting](https://github.com/golangci/golangci-lint/blob/59f1c6e455a6b288d859662a54707e523a05989d/pkg/lint/runner.go#L96-L97) out of the normal set of processors and do that [after fixing](https://github.com/golangci/golangci-lint/blob/59f1c6e455a6b288d859662a54707e523a05989d/pkg/commands/run.go#L373)?

Either way, such a fix doesn't need to be in this PR.

Thanks for the comment fixes, I included all of them in this commit - I prefer to keep the commit history clean. Let me know if you prefer it otherwise and I can restore them as a separate commit.

ldez:
> Thanks for the comment fixes, I included all of them in this commit - I prefer to keep the commit history clean. Let me know if you prefer it otherwise and I can restore them as a separate commit.

I prefer to fixes as commits, it's easier to follow the review.

pohly:
Okay, I added the second commit with the comment changes.

BTW, should I also rebase each time I do a push to avoid having to merge with master?

ldez (approved):
LTGM

PR #3572, opened 2023-02-07
lll: skip go command

black-06:
Skip go command because we can't add `//nolint` to the go command line.


pkg/golinters/lll.go
```diff
@@ -19,6 +19,11 @@ import (
 
 const lllName = "lll"
 
+const (
+	generate = "//go:generate"
+	linkname = "//go:linkname"
+)
```
ldez, line 25:
```suggestion
const goCommentDirectivePrefix = "//go:"
```

pkg/golinters/lll.go
```diff
@@ -94,6 +99,10 @@ func getLLLIssuesForFile(filename string, maxLineLen int, tabSpaces string) ([]r
 		line := scanner.Text()
 		line = strings.ReplaceAll(line, "\t", tabSpaces)
 
+		if strings.HasPrefix(line, generate) || strings.HasPrefix(line, linkname) {
```
ldez, line 102:
```suggestion
		if strings.HasPrefix(line, goCommentDirectivePrefix) {
```

ldez (changes requested):
I think we can just exclude all comments with the go directive prefix.

ldez (approved):
LGTM

PR #3602, opened 2023-02-17
govet: enable timeformat by default

johejo:
Updates #3353 

It seems my comment was wrong in #3353. It was enabled by default in go vet.

https://github.com/golang/go/blob/release-branch.go1.20/src/cmd/vet/main.go


ldez (approved):
LGTM

PR #3607, opened 2023-02-17
feat: add nostdglobals linter

katsadim:
Add `nostdglobals` linter that checks whether some security unsafe standard library functions are used by the code.

Take for example `http.DefaultTransport` struct, which can be hijacked by a 3rd party library and intercept your traffic.

Therefore all this linter does is to check for occurrences of `http.DefaultTransport` and `http.DefaultClient`,and notify the user.

Please do not hesitate to ask questions and also propose changes to the linter :)

https://github.com/katsadim/nostdglobals


ldez:
:thinking: feel like a duplicate of some other linters like [forbidigo](https://github.com/ashanbrown/forbidigo)

katsadim:
true, same goal may be accomplished by [forbidigo](https://github.com/ashanbrown/forbidigo). Only difference is that `nostdglobals` is opinionated and does not permit certain global variables to be used.

ldez:
So it's just a kind of subset of forbidigo, then it's a duplicate.
We will have to decline your PR.
Thank you anyway.

PR #2971, opened 2022-07-10
dev: change err to nil

sashamelentyev:
Added condition when err is nil and and file not dir instead of err handling, cuz err variable returned anyway. 
Changed operator for size variable. Because need assign only.

PR #3611, opened 2023-02-18
fix: cache status size calculation

johejo:
related #2971 

`golangci-lint cache status` always shows 10B if cache directory exists.
That's the size of trim.txt.


ldez (approved):
LGTM

PR #3613, opened 2023-02-18
.golangci.reference.yml validation

pohly:
I recently (see https://github.com/golangci/golangci-lint/pull/3612) worked on configuration changes that touched the reader. I found no unit test for it, so I added one with the `.golangci.reference.yml` file as input.

The positive side effect is that putting something invalid there now gets flagged during testing. The negative effect is that invalid configuration entries must be commented out, which makes it a bit less readable.

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -1018,48 +1018,51 @@ linters-settings:
       - unusedwrite
 
     # Enable all analyzers.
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

sorry but the reference can be handled like that.

I don't think we need to validate the reference, and the reference by the exhaustivity is invalid by nature.

Issue #2440, opened 2021-12-30
proposal: add a way to exclude rules only for non-test code


ashanbrown:
### Your feature request related to a problem? Please describe.

I'd like to include some linter rules for only test code but I don't believe is possible to write a golang regexp that excludes lint errors for files that _don't_ end in `_test.go` (see https://stackoverflow.com/questions/42515407/is-there-a-way-to-match-everything-except-a-constant-string-using-go-regexp, explaining that in order to meet the `O(n)` golang regexps don't support lookahead).  In my particular case, I'm looking to use `forbidigo` to exclude calls to `time.Sleep` to keep people from adding sleep to tests.

### Describe the solution you'd like.

I'd suggest we add a `includeTests` field to `BaseRule`.  Then an exclude rule would like:

```
  exclude-rules:
    - linters:
        - forbidigo
      source: time\.Sleep
      includeTests: false

```

The default value for `includeTests` would be true.

### Describe alternatives you've considered.

I could build selective configuration per pattern for tests into `forbidigo` but that would require significantly changing the interface and also imagine other linters may have a similar issue.  Alternatively, a `notpath` field that specified files to skip might work, allowing me to exclude test files.  That approach might also have other applications as well.

ldez:
You can do seomthing like that:
```go
func TestName(t *testing.T) {
	compile := regexp.MustCompile(`^.*[^_][^t][^e][^s][^t]\.go$`)

	fmt.Println(compile.MatchString("foobar_test.go"))
	fmt.Println(compile.MatchString("foobar.go"))
}
```
The limitation: it doesn't handle file with less than 5 characters.

ashanbrown:
Thanks for the suggestion.  I agree that would work but it doesn't express the intent very obviously and doesn't lend itself to discovery.  Might we consider something more direct?

mihaitodor:
I bumped into the same issue when I tried to forbid [`os.MkdirTemp`](https://pkg.go.dev/os#MkdirTemp) in tests, so people use [`T.TempDir`](https://pkg.go.dev/testing#T.TempDir) instead, but I'm reluctant to go for the RegEx pattern you suggested. Would be great to have a `includeTests` feature. Happy to explore adding it if you're open to having this addition.

pohly:
I've proposed an implementation based on `not-path` instead of `notpath` (hyphen for readability) in https://github.com/golangci/golangci-lint/pull/3617.

PR #3617, opened 2023-02-19
rules: support inverted path match

pohly:
It is not possible to use `exclude-rules: path` to exclude issues for non-test files because it is impossible to write a regexp that reliably matches those (https://github.com/golangci/golangci-lint/issues/2440). The new `not-path` check solves that by inverting the meaning: it excludes if the pattern does not match.

Besides test files, this is also useful for large code bases where some specific code paths have additional constraints. For example, in Kubernetes the usage of certain functions is forbidden in test/e2e/framework but okay elsewhere. Without `not-path`, a regular expression that carefully matches all other paths has to be used, which is very cumbersome:

    - path: ^(api|cluster|cmd|hack|pkg|plugin|staging|test/(cmd|conformance|e2e/(apimachinery|apps|architecture|auth|autoscaling|chaosmonkey|cloud|common|dra|instrumentation|kubectl|lifecycle|network|node|perftype|reporters|scheduling|storage|testing-manifests|upgrades|windows)|e2e_kubeadm|e2e_node|fixtures|fuzz|images|instrumentation|integration|kubemark|list|soak|typecheck|utils)|third_party|vendor)/
      msg: "E2E framework:"
      linters:
      - forbidigo

With not-path, this becomes much simpler:

    - not-path: ^test/e2e/framework/
      msg: "E2E framework:"
      linters:
      - forbidigo
      
Fixes: #2440


ldez:
Hello,

`not-path` seems really ambiguous as a name.

pohly:
> `not-path` seems really ambiguous as a name.

Let's brainstorm, perhaps we can come up with something better.

- `issues: exclude-rules: not-path` = "exclude, but not in path ..."
- `issues: exclude-rules: except-for-path` = "exclude, except when the path matches"
- `issues: exclude-rules: allow-issues-for-path` = "exclude issues, allow them for certain paths"

cc @ashanbrown

ashanbrown:
`xpath`?  `inverse-path`?

ashanbrown:
Thinking a bit more about this, `path-except` seems like it might be clearer name.  Furthermore, adding both `paths` (plural) as an inclusive list as well as `paths-except` as an exclusive list (`applied after `path` and `paths`), might be ideal.

pohly:
I'm fine with `path-except`. I'll defer to @ldez here.

Regarding `paths` and `paths-except`, do you want those to be lists of regexps instead of the single regexp for `path` and (soon?) `path-except`? That doesn't add any new capabilities (a regexp can already match multiple different paths), so I would prefer to not add those.

pohly:
Semantically, all of the conditions are AND and there's no implied ordering. That more complex conditions get checked after others is merely an implementation detail.

ashanbrown:
My only reason for suggesting `paths` and `paths-except` is that they could make it easier to express intent and keep completely unrelated cases separately (and potentially comment them in yaml as well).  I agree it isn't technically necessary to implement the desired behavior.

pohly:
I suspect that it is already possible to keep such separate cases also separated in the YAML with different `exclude-rules`: they might share other common attributes (linter, message), but if the intent is different, then perhaps those are also different?

spatecon (approved):

ldez:
`path-except` seems right.
Keep it as a string.

pohly:
Renamed to `path-except`.

ldez:
Sorry for the delay, I will handle the PR very soon.

ldez (approved):

PR #3641, opened 2023-02-27
dev: git ignore golangci-lint.exe on Windows

alexandear:
When running integration tests on Windows the file `golangci-lint.exe` is created.


ldez (approved):
LGTM

PR #3700, opened 2023-03-16
fix: combination of --fix and --path-prefix

pohly:
This combination used to fail because the fixer was executed after the path prefixed.
Then the file paths that it processed didn't match the current working directory, leading to:

```
ERRO Failed to fix issues in file test/fix_sample/main.go: failed to get file bytes for test/fix_sample/main.go: can't read file test/fix_sample/main.go: open test/fix_sample/main.go: no such file or directory
```
 
Making the fixer a normal processor and moving it before the path prefixed avoids this problem.

Related to #2293 and #3626

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -359,7 +359,7 @@ func (e *Executor) runAnalysis(ctx context.Context, args []string) ([]result.Iss
 	lintCtx.Log = e.log.Child(logutils.DebugKeyLintersContext)
 
 	runner, err := lint.NewRunner(e.cfg, e.log.Child(logutils.DebugKeyRunner),
-		e.goenv, e.EnabledLintersSet, e.lineCache, e.DBManager, lintCtx.Packages)
+		e.goenv, e.EnabledLintersSet, e.lineCache, e.fileCache, e.DBManager, lintCtx.Packages)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


test/fix_sample/main.go
```diff
@@ -0,0 +1,6 @@
+package main
+
+func main() { 
+	println(  
+"hello")
```
pohly, line 5: It's not totally unexpected that committing intentionally broken Go code into the repo will trigger the linter :sweat_smile: 

I couldn't find any tests for fixing. Do they really not exist or have I just missed them?

I had started writing a unit test, but that wouldn't have been able to reproduce the failure because the failure only occurs in combination with other processors.

I can remove the extra commit with this file.

ldez: the tests are here https://github.com/golangci/golangci-lint/tree/master/test/testdata/fix

pohly: Thanks. I've ripped out the manual test case and modified the testing to run one test case (arbitrarily picking the first one, based on alphabetical sorting) with `--path-prefix`. I checked that this breaks with master and works in this PR.

pohly:
Rebased after PR #3571 was merged.

ldez (approved):
LGTM

PR #3701, opened 2023-03-17
bump gocritic

cristaloleg:


cristaloleg:
> Error: An error occurred: [503 Service Unavailable] error accessing OSS Index

I think I cannot do much here. 

BTW, to get rid of 3rd party deps I've switched to the official Go team tool `govulncheck`. Example CI https://github.com/cristalhq/.github/blob/main/.github/workflows/vuln.yml

ldez:
I handle the problem with the vulnerability check

pkg/golinters/gocritic.go
```diff
@@ -13,7 +13,7 @@ import (
 	"sync"
 
 	"github.com/go-critic/go-critic/checkers"
-	gocriticlinter "github.com/go-critic/go-critic/framework/linter"
```
cristaloleg, line 16: Whole `framework/linter` is deprecated and moved to `linter`. It still works just as a fallback. Will be removed in the future versions.

ldez (approved):
LGTM

Issue #3555, opened 2023-02-03
docs: unused configuration broken link


oscr:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc.).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

On the page https://golangci-lint.run/usage/linters

In the "Enabled by default" section in the end there is: "unused ⚙️" 

If you click on the ⚙️ for configuration it's a broken link: https://golangci-lint.run/usage/linters/#unused

### Version of golangci-lint

N/A

### Configuration file

N/A

### Go environment

N/A

### Verbose output of running

N/A

### Code example or link to a public repository

N/A

ldez:
Hello,

there is no configuration for `unused`, so the gear must be removed.

I will fix that.

oscr:
Awesome! Thank you **l**dez! :grin:

PR #3703, opened 2023-03-18
Staticcheck linters configs refactoring

Antonboom:
Closes https://github.com/golangci/golangci-lint/issues/3555

### DIFF
--- a/pkg/golinters/staticcheck_common.go
+++ b/pkg/golinters/staticcheck_common.go
@@ -14,16 +14,10 @@ import (
 
 var debugf = logutils.Debug(logutils.DebugKeyMegacheck)
 
-func getGoVersion(settings *config.StaticCheckSettings) string {
-	var goVersion string
-	if settings != nil {
-		goVersion = settings.GoVersion
-	}
-
-	if goVersion != "" {
+func getGoVersion(versionable interface{ GoVersion() string }) string {
+	if goVersion := versionable.GoVersion(); goVersion != "" {
 		return goVersion
 	}
-
 	return "1.17"
 }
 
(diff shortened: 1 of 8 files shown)


pkg/lint/lintersdb/manager.go
```diff
@@ -134,7 +134,7 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 		goModDirectivesCfg  *config.GoModDirectivesSettings
 		gomodguardCfg       *config.GoModGuardSettings
 		gosecCfg            *config.GoSecSettings
-		gosimpleCfg         *config.StaticCheckSettings
+		gosimpleCfg         config.StaticCheckGoSimpleSettings
```
Antonboom, line 137: I think it doesn't make sense to use pointers here (for other settings too).
Not big memory win, but code is complicated a lot

pkg/golinters/staticcheck_common.go
```diff
@@ -54,23 +48,15 @@ func setAnalyzerGoVersion(a *analysis.Analyzer, goVersion string) {
...
-	var cfg *scconfig.Config
-
-	if settings == nil || !settings.HasConfiguration() {
-		return &scconfig.Config{
-			Checks:                  []string{"*"}, // override for compatibility reason. Must drop in the next major version.
-			Initialisms:             scconfig.DefaultConfig.Initialisms,
-			DotImportWhitelist:      scconfig.DefaultConfig.DotImportWhitelist,
-			HTTPStatusCodeWhitelist: scconfig.DefaultConfig.HTTPStatusCodeWhitelist,
-		}
+func staticCheckConfig(settings interface{ ChecksList() []string }) *scconfig.Config {
```
Antonboom, line 51: Also I tried use generics, but the solution turned out to be less elegant

ldez:
#3555 is about a problem during the generation of the doc, this PR doesn't fix the problem.

the deprecated configuration must be kept.

PR #3708, opened 2023-03-19
feat: add linter to check number of params and return values of func

hitzhangjie:
When function declaration contains too many parameters and return values, it will decrease the code readability.

And it is likely to improve the code by some refactoring.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -1,4 +1,7 @@
 linters-settings:
+  toomanyparams:
+    max_returns: 3
+    max_params: 5
   depguard:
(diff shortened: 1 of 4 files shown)


ldez:
Hello,

you have to create your linter outside golangci-lint and after you will be able to create a PR with the dedicated repository of your linter.

hitzhangjie:
I check the list of linters, there's a 'tooManyResults', which may check the number of return values.
I don't see a linter listed that will check the number of input parameters.

If I miss something, please let me know.

PR #3729, opened 2023-03-25
output: add colored-tab

### LINKED ISSUE #3728: Tab output format without a color

I started writing a blog post regarding my `golangci-lint` setup and decided to try the `tab` format, which looks cool and I like it.

I tried the following config and was upset because my `lint.txt` is full of cryptic symbols related to the colouring (see screenshot)

```yaml
output:
  format: tab:lint.txt
```

image

Add a new format which outputs results in the `tab` format but without colours, so no cryptic symbols appear in the output file.

Keep using `line-number` or using a personal fork without colours. This can be also `sed`-ed but I don't like this fix.

### Additional context.

I'm happy to make a PR, I'm just not sure which fix would be better: add a new `tab-no-colour` format or add a separate config field `no-color: <boolean>` or just join https://no-color.org/ movement.

ldez:
This PR introduces 2 changes:
- remove the color of the `tab` output
- add a new output `colored-tab`

I introduce a new `colored-tab` instead of `no-color-tab` to be homogeneous with `line-number` and `colored-line-number`.

I think that the removal of the color is a minor change because I don't see in which context it can be breaking.

Fixes #3728

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -10,7 +10,7 @@ linters-settings:
   dupl:
     threshold: 100
   funlen:
-    lines: 100
+    lines: -1 # the number of lines (code + empty lines) is not a right metric and leads to code without empty line or one-liner.
     statements: 50
   goconst:
     min-len: 2
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -76,7 +76,7 @@ run:
 
 # output configuration options
 output:
-  # Format: colored-line-number|line-number|json|tab|checkstyle|code-climate|junit-xml|github-actions|teamcity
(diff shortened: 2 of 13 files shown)


cristaloleg (approved):

cristaloleg:
Works as expected, thanks for the quick PR !

cristaloleg:
I assume it's ok to slightly break compatibility. JFI

.golangci.yml
```diff
@@ -10,7 +10,7 @@ linters-settings:
   dupl:
     threshold: 100
   funlen:
-    lines: 100
+    lines: -1 # the number of lines (code + empty lines) is not a right metric and leads to code without empty line or one-liner.
```
bombsimon, line 13: Isn't it better to disable this linter?

bombsimon (approved):
LGTM, nice tests

PR #3738, opened 2023-03-27
feat: handle timeout for linters

cia-rana:
The timeout option works until the package import is complete, but if the timeout occurs while the linters are running, they run to the end without exiting. This is especially inconvenient when running golangci-lint with [golangci-lint-action](https://github.com/golangci/golangci-lint-action).
Suppose we set a timeout option for golangci-lint and the timeout expires, but the linters keep running. If the timeout we set for the job expires in the meantime, the linter's logs are not output, only the job's logs are simply output as shown below.

```
Error: The action has timed out.
```

Therefore, we have implemented a process for handling timeouts.


ldez:
Hello,
I understand the goal but the fact to stop a linter without waiting for the end of the process can corrupt the cache.

I thought of a better solution, but for now, I don't have one.

PR #3740, opened 2023-03-27
feat: remove warning when the config file is explicitly stdin

### LINKED ISSUE #3739: Support reading configuration from stdin

We would like to read configuration from stdin. The reason is we want to deep merge a few configuration files with `yq` and pipe them in. This sort of works, but gives an error.

```
$ echo '' | golangci-lint run -c /dev/stdin
WARN [config_reader] Can't pretty print config file path: can't eval symlinks for path /dev/stdin: lstat /proc/688987/fd/pipe:[50935730]: no such file or directory
```

(in real world our command is not, echo, of course)

Support config from stdin without warning/error

For our specific issue, passing multiple config files and having them deep merged would also be suitable. However, this may be complex as different folks have different meanings of "merge"

### Additional context.

_No response_

ldez:
Fixes #3739

### DIFF
--- a/pkg/config/reader.go
+++ b/pkg/config/reader.go
@@ -74,15 +74,16 @@ func (r *FileReader) parseConfig() error {
 
 	if usedConfigFile == os.Stdin.Name() {
 		usedConfigFile = ""
+		r.log.Infof("Reading config file stdin")
 	} else {
 		var err error
 		usedConfigFile, err = fsutils.ShortestRelPath(usedConfigFile, "")
 		if err != nil {
 			r.log.Warnf("Can't pretty print config file path: %v", err)
 		}
-	}
 
-	r.log.Infof("Used config file %s", usedConfigFile)
+		r.log.Infof("Used config file %s", usedConfigFile)
+	}
 
 	usedConfigDir, err := filepath.Abs(filepath.Dir(usedConfigFile))
 	if err != nil {
--- a/pkg/config/reader.go
+++ b/pkg/config/reader.go
@@ -72,13 +72,20 @@ func (r *FileReader) parseConfig() error {
 		return nil
 	}
 
-	usedConfigFile, err := fsutils.ShortestRelPath(usedConfigFile, "")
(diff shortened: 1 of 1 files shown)


pkg/config/reader.go
```diff
@@ -72,13 +72,20 @@ func (r *FileReader) parseConfig() error {
 		return nil
 	}
 
-	usedConfigFile, err := fsutils.ShortestRelPath(usedConfigFile, "")
-	if err != nil {
-		r.log.Warnf("Can't pretty print config file path: %s", err)
+	if usedConfigFile == os.Stdin.Name() {
+		usedConfigFile = ""
```
bombsimon, line 76: Since this will be used in the log `Used config file`, would it make sense to either do conditional logging (not a file if stdin) or just set this to `stdin` so the log says `Reading config file stdin`?

bombsimon (approved):
I know this change only removes the warning but if we want to support this, would it make sense to detect content on stdin and automatically use that instead of having to pipe stdin to `-c`/`--config`? I was thinking something like

```go
stat, _ := os.Stdin.Stat()
if (stat.Mode() & os.ModeCharDevice) == 0 {
	// Got data on stdin
} else {
	// Use flag or default config
}
```

Approving this either way since a change like that probably deserves it's own PR.

ldez:
> I know this change only removes the warning but if we want to support this, would it make sense to detect content on stdin and automatically use that instead of having to pipe stdin to `-c`/`--config`? I was thinking something like
> 
> ```go
> stat, _ := os.Stdin.Stat()
> if (stat.Mode() & os.ModeCharDevice) == 0 {
> 	// Got data on stdin
> } else {
> 	// Use flag or default config
> }
> ```

I already think about that but this only detects if the command is piped and not if the configuration comes from stdin.
A pipe can be used in different contexts and not only for the configuration file.

pkg/config/reader.go
ldez: I added a log.

Issue #3682, opened 2023-03-12
FAIL: TestSourcesFromTestdata/go-header_bad.go


Antonboom:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc.).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

Tests in master fail because of no expected warning from `go-header`.

The standalone run is silence too:
```console
$ go install github.com/denis-tingaikin/go-header/cmd/go-header@latest
$ go-header version   
v0.4.3

$ cat .go-header.yml
template: MY {{title}}
values:
  const:
    title: TITLE.

$ go-header /Users/anthony/golangci-lint/test/testdata/go-header_bad.go
$ echo "$?"                  
0
```

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version (devel) built with go1.20.1 from (unknown, mod sum: "") on (unknown)
```

</details>

### Configuration file

<details>

```console
$ cat .go-header.yml
template: MY {{title}}
values:
  const:
    title: TITLE.
```

</details>

### Go environment

<details>

```
$ uname -a            
Darwin MacBook-Pro.local 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:04:44 PST 2022; root:xnu-8792.61.2~4/RELEASE_ARM64_T8103 arm64
```

```console
$ go version && go env
go version go1.20.1 darwin/arm64
GO111MODULE="on"
GOARCH="arm64"
GOBIN="/Users/anthony/golang_workspace/bin"
GOCACHE="/Users/anthony/Library/Caches/go-build"
GOENV="/Users/anthony/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="arm64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/anthony/golang_workspace/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/anthony/golang_workspace"
GOPRIVATE=""
GOPROXY="https://proxy.golang.org,direct"
GOROOT="/usr/local/go"
[...]

PR #3748, opened 2023-03-31
goheader: fix relative template path

thejan2009:
go-header template-path setting is usually a relative filepath. As a result, the linter only worked when golangci-lint was invoked from that folder.

We detect if the template path is relative and prefix it with config path.

go-header_bad.go test case was modified because old files are not checked https://github.com/denis-tingaikin/go-header/blob/v0.4.3/analyzer.go#L59-L63


thejan2009:
Hi, is there anything I can help with to unblock this PR?

I saw https://github.com/golangci/golangci-lint/issues/3682 related to https://github.com/denis-tingaikin/go-header/issues/22. Should that issue be fixed first?

thejan2009:
@ldez bump

test/testdata/configs/go-header.yml
```diff
@@ -1,6 +1,6 @@
 linters-settings:
   goheader:
-    template: MY {{title}}
+    template-path: go-header-template
```
Antonboom, line 3: let's keep both tests
and `template` and `template-path`
(with better naming for config and test files)

pkg/golinters/goheader.go
```diff
@@ -15,16 +16,20 @@ import (
 
 const goHeaderName = "goheader"
 
-func NewGoHeader(settings *config.GoHeaderSettings) *goanalysis.Linter {
+func NewGoHeader(settings *config.GoHeaderSettings, cfg *config.Config) *goanalysis.Linter {
```
Antonboom, line 19: 1) Ambiguous API – you passed the whole config obj, but need the config dir only
2) Looks better to do overrides at `func (m Manager) GetAllSupportedLinterConfigs()` level
- will be consistent with other similar logics (like `GoVersion` update)
- will not require changing the constructor signature

Antonboom (changes requested):

pkg/golinters/goheader.go
thejan2009: I agree, the proposed solution seems better.

[GoCritic](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/gocritic.go#L33) has the same config pattern though. Should that be changed as well?

test/testdata/go-header_bad.go
```diff
@@ -1,4 +1,4 @@
-/*MY TITLE!*/ // want `Expected:TITLE\., Actual: TITLE!`
+/*MY TITLE?*/ // want `Expected:TITLE\., Actual: TITLE?`
```
Antonboom, line 1: Extra change?

test/testdata/go-header-template_good.go
```diff
@@ -0,0 +1,6 @@
+/*MY TITLE.*/
```
Antonboom, line 1: where is `go-header-template_bad`?

test/testdata/go-header-template_good.go
```diff
@@ -0,0 +1,6 @@
...
+
+//golangcitest:args -Egoheader
+//golangcitest:config_path testdata/configs/go-header-template.yml
```
Antonboom, line 4: Just a question – how it works without `test/` prefix?
Looks like your `filepath.Join` is not covered 🤔

pkg/lint/lintersdb/manager.go
```diff
@@ -247,6 +247,10 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 		if stylecheckCfg != nil && stylecheckCfg.GoVersion != "" {
 			stylecheckCfg.GoVersion = trimGoVersion(m.cfg.Run.Go)
 		}
+
+		if goheaderCfg != nil {
+			goheaderCfg.LintConfigDir = m.cfg.GetConfigDir()
```
Antonboom, line 252: I meant you can override `goheaderCfg.TemplatePath` at this level.
But I cannot decide which option is better

Antonboom, line 252: Did you try to run golangci-lint without config, it works as expected?

pkg/golinters/goheader.go
Antonboom: > [GoCritic](https://github.com/golangci/golangci-lint/blob/master/pkg/golinters/gocritic.go#L33) has the same config pattern though. Should that be changed as well?

Good point. But it is not relevant to your PR's goal, please do not add extra changes 👌

Antonboom (changes requested):
👌

PR #3760, opened 2023-04-03
build(deps): bump github.com/alexkohler/nakedret to 2.0.1

### LINKED ISSUE #1317: nakedret doesn't check "return"s in conditionals
The nakedret linter embedded in golangci-lint appears to only check for "return" statements at the top level of the function, outside of conditionals.  My expectation is that all returns are checked in functions above the specified length, regardless of nesting.

In golangci-lint's test suite, should line 6 of test/testdata/nakedret.go have an "// ERROR" marker?  Letting the first "return" pass here seems wrong:
```
func NakedretIssue() (a int, b string) {
	if a > 0 {
		return
	}

	if b == "" {
		return 0, "0"
	}
[...]
	return // ERROR "naked return in func `NakedretIssue` with 31 lines of code"
}
```

Thanks!
-- Aaron

zalenskivolt:
Handling of naked returns inside conditionals as well as from nested functions has been added upstream, in https://github.com/alexkohler/nakedret/pull/12 and more commits.

This should resolve #1317.

### DIFF
--- a/pkg/golinters/nakedret.go
+++ b/pkg/golinters/nakedret.go
@@ -1,7 +1,7 @@
 package golinters
 
 import (
-	"github.com/alexkohler/nakedret"
+	"github.com/alexkohler/nakedret/v2"
 	"golang.org/x/tools/go/analysis"
 
 	"github.com/golangci/golangci-lint/pkg/config"
--- a/pkg/golinters/nakedret.go
+++ b/pkg/golinters/nakedret.go
@@ -1,140 +1,26 @@
 package golinters
 
 import (
-	"fmt"
-	"go/ast"
-	"go/token"
-	"sync"
-
+	"github.com/alexkohler/nakedret"
 	"golang.org/x/tools/go/analysis"
 
 	"github.com/golangci/golangci-lint/pkg/config"
 	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
-	"github.com/golangci/golangci-lint/pkg/lint/linter"
-	"github.com/golangci/golangci-lint/pkg/result"
 )
 
 const nakedretName = "nakedret"
 
-//nolint:dupl
 func NewNakedret(settings *config.NakedretSettings) *goanalysis.Linter {
-	var mu sync.Mutex
-	var resIssues []goanalysis.Issue
-
-	analyzer := &analysis.Analyzer{
-		Name: nakedretName,
-		Doc:  goanalysis.TheOnlyanalyzerDoc,
-		Run: func(pass *analysis.Pass) (any, error) {
-			issues := runNakedRet(pass, settings)
-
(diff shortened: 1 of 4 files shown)


pkg/golinters/nakedret.go
```diff
@@ -72,63 +73,68 @@ type nakedretVisitor struct {
...
-				if ident != nil {
-					namedReturns = append(namedReturns, ident)
-				}
-			}
+	var (
+		funcType *ast.FuncType
+		funcName string
+	)
+	switch s := node.(type) {
+	case *ast.FuncDecl:
```
zalenskivolt, line 104: `Visit()` from upstream
https://github.com/alexkohler/nakedret/blob/d95c37512/nakedret.go#L196-L235
only updated with the issue reporting.

pkg/golinters/nakedret.go
```diff
@@ -72,63 +73,68 @@ type nakedretVisitor struct {
...
-		if funcDecl.Name == nil {
-			continue
+func hasNamedReturns(funcType *ast.FuncType) bool {
+	if funcType == nil || funcType.Results == nil {
+		return false
+	}
+	for _, field := range funcType.Results.List {
+		for _, ident := range field.Names {
+			if ident != nil {
+				return true
```
zalenskivolt, line 91: `hasNamedReturn` from upstream
https://github.com/alexkohler/nakedret/blob/d95c37512/nakedret.go#L182-L194

ldez:
Hello,

the way that we handle this linter is not right, we have to change that.

Instead of "duplicating" the code inside golangci-lint I think it's better to a create PR on nakedret to be able to use it as a dependency.

zalenskivolt:
> Instead of "duplicating" the code inside golangci-lint I think it's better to a create PR on nakedret to be able to use it as a dependency.

This was updated in https://github.com/alexkohler/nakedret/pull/25 and now it uses that analyzer directly!

zalenskivolt:
@ldez linter is updated, and should not be blocked any more! 😄

ldez (approved):
LGTM

test/testdata/nakedret.go
```diff
@@ -1,11 +1,15 @@
...
 
+import "fmt"
+
 func NakedretIssue() (a int, b string) {
 	if a > 0 {
-		return
+		return // want "naked return in func `NakedretIssue` with 33 lines of code"
 	}
 
+	fmt.Println("nakedret")
```
zalenskivolt, line 11: @ldez were these `fmt.Println()` updates intentional?

should it still test the edge case of 31 lines, by removing two `// ...` lines below?

ldez: Yes the println is intentional, because I wanted to have at least one std import.

Issue #3663, opened 2023-03-06
docs: missing image version for the Docker installation 


a-h:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc.).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

Basically, this is a problem again.

https://github.com/golangci/golangci-lint/issues/3103#issue-1340388829

I didn't comment on the closed issue, since it's over a year old. Thought a new issue might be more appropriate.

https://golangci-lint.run/usage/install/#docker

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
# Paste output here
```

</details>

### Configuration file

<details>

```console
$ cat .golangci.yml
# paste output here
```

</details>

### Go environment

<details>

```console
$ go version && go env
# paste output here
```

</details>

### Verbose output of running

<details>

```console
$ golangci-lint cache clean
$ golangci-lint run -v
# paste output here
```

</details>

### Code example or link to a public repository

<details>

```go
// add your code here
```

</details>

erdnaxeli:
Hi, could this be fixed, or could you give us hint about how to fix it? It was opened two months ago.

ldez:
In some cases, the version is missing during the doc generation, but for now, I know which cases.
It works but sometimes the version is missing.

erdnaxeli:
The [current version](https://golangci-lint.run/usage/install/) is broken:
![image](https://github.com/golangci/golangci-lint/assets/479441/13fe637f-007a-4deb-9ec9-7f264bc15299)

ldez:
I know, it's an intermittent problem.

ldez:
The page has been updated, I started a migration from Netlify to GitHub Pages to be able to stop this intermittent problem.

PR #3846, opened 2023-05-24
Fix go install instructions

donatj:
The `{.LatestVersion}` token isn't being populated correctly on https://golangci-lint.run/usage/install#install-from-source

image

A better solution is probably to just to use `@latest` unless there's some reason not to I am not understanding.


ldez:
Hello, the version is filled by a script, but sometimes the script fails.

Related to #3663

So I will decline your PR.

donatj:
I'd argue just using `@latest` would be a better and simpler solution 🤷

ldez:
`latest` is a moving reference, I can recommend it because it can break CI.

PR #3866, opened 2023-06-02
depguard: fix GOROOT detection

### LINKED ISSUE #3862: depguard is enabled and fails on v1.53.0
### Welcome

Running v1.53.0 of golangci-lint seems to enable `depguard` and fails due to a config error. 

### Version of golangci-lint

<details>

```console
$ golangci-lint --version

v1.53.0
```

</details>

### Configuration file

<details></details>

### Go environment

<details>

```console
$ go version && go env

go version go1.20.4 darwin/amd64
```

</details>

### Verbose output of running

<details>

```console
$ golangci-lint cache clean
$ golangci-lint run -v

golangci/golangci-lint info checking GitHub for tag 'v1.53.0'
golangci/golangci-lint info found version: 1.53.0 for v1.53.0/darwin/amd64
STARTING golangci-lint checks
ERRO [linter] depguard: create analyzer: couldn't expand $gostd: could not read GOROOT directory: open src: no such file or directory
must have an Allow and/or Deny package list 
```

</details>

### Code example or link to a public repository

Updating golangci-lint to v1.53.0 in https://github.com/moov-io/infra/blob/master/go/lint-project.sh 

CI run: https://github.com/moov-io/infra/actions/runs/5149020788/jobs/9271486191

ldez:
Following the plan that I explained [here](https://github.com/golangci/golangci-lint/issues/3862#issuecomment-1572907138):
- create a fork inside the golangci org
- apply the fix
- use the fork as a dependency (I will create a dedicated module to avoid `replace` directives, because `replace` directives are not transitive, so users that are using that use the `tools` pattern will not have a problem.)

---

Summary of issue #3862:
depguard uses `build.Default.GOROOT` but this value is filled correctly only if the `GOROOT` is defined through an env var.

```console
$ go env | grep GOROOT
GOROOT="/usr/local/go"

$ env | grep GOROOT

$ export GOROOT=/usr/local/go

$ env | grep GOROOT
GOROOT=/usr/local/go
```

The "expander" cannot be overridden because it's a global variable inside an internal package, the `compile` method that produces the error and the analyzer constructor are not exported.

---

The fix I have done in our `depguard` fork is based on what is done inside [`golang.org/x/tools`](https://github.com/golang/tools/blob/86c93e8732cce300d0270bce23117456ce92bb17/cmd/godoc/goroot.go#L15-L30).

https://github.com/golangci/depguard/compare/v2.0.1...ed68d37

When the PR https://github.com/OpenPeeDeeP/depguard/pull/48 will be merged and when depguard will create a new release, we will drop this dependency (but we will have to keep the fork for compatibility.)

Fixes #3862

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -12,7 +12,6 @@ require (
 	github.com/BurntSushi/toml v1.3.0
 	github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24
 	github.com/GaijinEntertainment/go-exhaustruct/v2 v2.3.0
-	github.com/OpenPeeDeeP/depguard/v2 v2.0.1
 	github.com/alexkohler/nakedret/v2 v2.0.1
 	github.com/alexkohler/prealloc v1.0.0
 	github.com/alingse/asasalint v0.0.11
@@ -37,6 +36,7 @@ require (
 	github.com/go-xmlfmt/xmlfmt v1.1.2
 	github.com/gofrs/flock v0.8.1
 	github.com/golangci/check v0.0.0-20180506172741-cfe4005ccda2
+	github.com/golangci/depguard/v2 v2.0.2-0.20230601235138-ed68d3771f48
 	github.com/golangci/dupl v0.0.0-20180902072040-3e9179ac440a
 	github.com/golangci/go-misc v0.0.0-20220329215616-d24fe342adfe
 	github.com/golangci/gofmt v0.0.0-20220901101216-f2edd75033f2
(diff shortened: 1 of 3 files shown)


gandarez (approved):

bombsimon (approved):
LGTM! 

Would an alternative be to roll back `depguard` instead to not have to keep a fork around forever? It doesn't really matter having it around so I'll approve, just curious.

ldez:
I didn't roll back because the depguard v2 configuration is breaking and I think that we don't have to wait for more to break this, and because this new version of depguard fixes an annoying old bug with concurrency.

elgohr:
@ldez looks like you have to convince the CI that this is okay... https://github.com/golangci/golangci-lint/actions/runs/5153483603/jobs/9280765338

ldez:
It's just because we use the previous version inside the CI.

nhatthm:
@ldez  Do you know how long it takes to get the fix populated in GHA? We are using the `latest` tag

ldez:
the GitHub action is currently generate the GHA files https://github.com/golangci/golangci-lint/actions/runs/5153980997/jobs/9281874877

It will be available quickly.

PR #3872, opened 2023-06-02
docs: Update assets

caveman280:
Following the release of `v1.53.1`, `golangci/golangci-lint-action` still resolves the latest version to `v1.53.0`. This PR changes that to targets the latest, `v1.53.1`, release.

### DIFF
--- a/assets/github-action-config.json
+++ b/assets/github-action-config.json
@@ -1,8 +1,8 @@
 {
   "MinorVersionToConfig": {
     "latest": {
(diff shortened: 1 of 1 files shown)


ldez:
Hello, we have an action to handle that https://github.com/golangci/golangci-lint/blob/master/.github/workflows/post-release.yml

https://github.com/golangci/golangci-lint/actions/runs/5153980997/jobs/9281874877

PR #3882, opened 2023-06-02
codeclimate: less memory allocation

Lasiar:
Use json encoder to write codeclimate report.

### DIFF
--- a/pkg/printers/codeclimate.go
+++ b/pkg/printers/codeclimate.go
@@ -2,7 +2,6 @@ package printers
 
 import (
 	"encoding/json"
-	"fmt"
 	"io"
 
 	"github.com/golangci/golangci-lint/pkg/result"
@@ -52,12 +51,7 @@ func (p CodeClimate) Print(issues []result.Issue) error {
 		codeClimateIssues = append(codeClimateIssues, codeClimateIssue)
 	}
 
-	outputJSON, err := json.Marshal(codeClimateIssues)
-	if err != nil {
-		return err
-	}
-
-	_, err = fmt.Fprint(p.w, string(outputJSON))
+	err := json.NewEncoder(p.w).Encode(codeClimateIssues)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

do you have a problem with the code climate reporter?
Can you explain your problem?

Lasiar:
@ldez

> Hello,
> 
> do you have a problem with the code climate reporter? Can you explain your problem?

Hello, I don't have any problem, this PR just reduce allocation and memory usage.

[Source code of benchmark.](https://gist.github.com/Lasiar/c089288acfab500defda4821f24831cc)

<details>
  <summary>result on my machine:</summary>
  
`go test -run='^$' -bench=. -count=10`

| name       | old time/op | new time/op | delta   |                 |
| ---------- | ----------- | ----------- | ------- | --------------- |
| Climate-10 | 2.79µs ± 1% | 2.56µs ± 1% | \-8.00% | (p=0.000 n=9+8) |

| name       | old alloc/op | new alloc/op | delta    |                  |
| ---------- | ------------ | ------------ | -------- | ---------------- |
| Climate-10 | 3.36kB ± 1%  | 2.23kB ± 3%  | \-33.66% | (p=0.000 n=10+9) |


| name       | old alloc/op | new alloc/op | delta    |                  |
| ---------- | ------------ | ------------ | -------- | ---------------- |
| Climate-10 | 34.0 ± 0%  | 31.0 ± 0%  | \-8.82% | (p=0.000 n=10+10)|

</details>

ldez:
Hello,

The problem with this kind of benchmark is in the real interest of this optimization.

The basic rule of optimization is to do it only when you have a performance problem.
There is a name for that: premature optimization.

In a general way, `json.Marshal` or `json.Unmarshal` produce more useful errors than `json.NewDecoder` and `json.NewEncoder`.

In the current situation, there's little chance of an error, so we can accept this PR.
But I want to be clear that performance optimization should be led by performance problems and not micro-benchmarks.

ldez (approved):

PR #3901, opened 2023-06-12
dev: simplify GetAllSupportedLinterConfigs method

ldez:
- add `WithEnabledByDefault` method to set `EnabledByDefault` 
- remove `enableLinterConfigs` and `enabledByDefault := map[string]bool{...}`. This was introduced because of GolangCI (the SAAS version of golangci-lint). The PR which introduced this code is here: https://github.com/golangci/golangci-lint/pull/2. Now that SAAS is dead, we can make it simpler.
- use the fluent syntax (as the other linter) for the linter related to plugins.

### DIFF
--- a/pkg/lint/linter/config.go
+++ b/pkg/lint/linter/config.go
@@ -50,6 +50,11 @@ type Config struct {
 	Deprecation *Deprecation
 }
 
+func (lc *Config) Default() *Config {
+	lc.EnabledByDefault = true
+	return lc
+}
+
 func (lc *Config) ConsiderSlow() *Config {
 	lc.IsSlow = true
 	return lc
--- a/pkg/lint/linter/config.go
+++ b/pkg/lint/linter/config.go
@@ -33,8 +33,8 @@ type Deprecation struct {
 }
 
 type Config struct {
-	Linter           Linter
-	EnabledByDefault bool
+	Linter  Linter
+	Default bool
 
 	LoadMode packages.LoadMode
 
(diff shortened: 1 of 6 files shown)


pkg/lint/linter/config.go
```diff
@@ -50,6 +50,11 @@ type Config struct {
 	Deprecation *Deprecation
 }
 
+func (lc *Config) Default() *Config {
```
bombsimon, line 53: I get that calling this `Default` makes it shorter when chaining but I'm not sure I see a reason to not call this `EnabledByDefault` other than the conflicting name with the field.

If you don't want to change the field name due to the collision (which we probably shouldn't due to being exported and a breaking change) maybe `SetEnabledByDefault` which is what the function does or maybe `WithEnabledByDefault` to match the other methods on the struct?

bombsimon (approved):
LGTM!

Just a personal pref. about naming so approving if you still don't agree

pkg/lint/linter/config.go
ldez: I don't want to use `SetEnabledByDefault` or `WithEnabledByDefault` because the function doesn't have or need args.

I switch `Default` and `EnabledByDefault` names.

bombsimon: Why does it matter if it takes any arguments? `WithLoadForGoAnalysis` is named with the same pattern and takes no arguments.

ldez: you are right, I will rename it for homogeneity.

PR #3922, opened 2023-06-21
go1.21 support

ldez:
This PR is to evaluate and prepare golangci-lint to [go1.21](https://tip.golang.org/doc/go1.21).

This PR will evolve during the beta and RC phases of [go1.21](https://tip.golang.org/doc/go1.21).

### DIFF
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -11,7 +11,7 @@ jobs:
     name: Build and deploy documentation
     runs-on: ubuntu-latest
     env:
-      GO_VERSION: '1.20'
+      GO_VERSION: '1.21.0-rc.2'
       NODE_VERSION: '20.x'
       CGO_ENABLED: 0
 
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -11,7 +11,7 @@ jobs:
     name: Build and deploy documentation
     runs-on: ubuntu-latest
     env:
-      GO_VERSION: '1.21.0-rc.4'
+      GO_VERSION: '1.21'
       NODE_VERSION: '20.x'
       CGO_ENABLED: 0
 
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -11,7 +11,7 @@ jobs:
     name: Build and deploy documentation
     runs-on: ubuntu-latest
     env:
-      GO_VERSION: '1.21.0-rc.3'
+      GO_VERSION: '1.21.0-rc.4'
       NODE_VERSION: '20.x'
       CGO_ENABLED: 0
 
--- a/.github/workflows/documentation.yml
+++ b/.github/workflows/documentation.yml
@@ -11,7 +11,7 @@ jobs:
     name: Build and deploy documentation
     runs-on: ubuntu-latest
     env:
-      GO_VERSION: '1.21.0-rc.2'
+      GO_VERSION: '1.21.0-rc.3'
       NODE_VERSION: '20.x'
       CGO_ENABLED: 0
 
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -18,7 +18,7 @@ jobs:
           # ex:
           # - 1.18beta1 -> 1.18.0-beta.1
           # - 1.18rc1 -> 1.18.0-rc.1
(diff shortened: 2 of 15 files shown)


ldez:
For now, there is a problem with go-critic, I will investigate to understand the problem.

pohly:
As mentioned on Slack: if Go 1.21rc2 is installed in a module, `gocritic: load embedded ruleguard rules: rules/rules.go:13: can't load fmt: setting an explicit GOROOT can fix this problem.` occurs again.

ldez:
I disabled `ruleguard` (one "rule" inside gocritic), I think it's the best move to be able to create a release quickly.

SVilgelm (approved):
LGTM

pohly:
Disabling ruleguard does not fix [the problem](https://github.com/golangci/golangci-lint/pull/3922#issuecomment-1620650363) in Kubernetes.

I first tried a config change:
```
linters-settings:
  gocritic:
    disabled-checks:
      - ruleguard               # unused
```

Then I bumped the golangci-lint in Kubernetes to ldez:feat/go1.21 and also still got the same problem:
```
ERROR: level=warning msg="plugin: 'AnalyzerPlugin' plugins are deprecated, please use the new plugin signature: https://golangci-lint.run/contributing/new-linters/#create-a-plugin"
ERROR: level=error msg="[linters_context] gocritic: load embedded ruleguard rules: rules/rules.go:13: can't load fmt: setting an explicit GOROOT can fix this problem."
```

If I had to guess, I would say that gocritic loads ruleguard rules regardless whether the ruleguard check is enabled or not (haven't checked).

pohly:
Right, I also just found that. So to me it looks like all of gocritic is broken for Kubernetes because of this problem.

pohly:
Disabling gocritic entirely makes golangci-lint usable with 1.21 in Kubernetes, but that would be a pity.

pohly:
> From my memories, the problem was already here before go1.21.

There was a different problem with having the Go toolchain inside the Go package. That got solved. Now go1.21 broke that again, with no solution so far.

ldez:
Have you tried with go-critic itself?

pohly:
go-critic works with go1.20.4 and go1.21rc4 when those are installed outside of Kubernetes. It also works with _output/local/.gimme/versions/go1.21rc3.linux.amd64/bin inside the Kubernetes repo but only if go-critic was compiled with go1.20.4.

However, go-critic fails once it is compiled with go1.21rc3, no matter where the Go installation is located when invoking it. I'm a bit surprised by that because I thought I had seen a difference before when trying it with golangi-lint - perhaps I had made a mistake when testing.

ldez:
I would like to propose moving this discussion outside of this PR, maybe on Slack.

Dasio:
Just FYI regarding https://github.com/golangci/golangci-lint/pull/3922/files#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160R44

With latest version it is still built with 1.20
```
golangci-lint run --version
golangci-lint has version 1.54.1 built with go1.20.7 from a9378d9 on 2023-08-11T10:14:36Z
```
Because of this I got errors such as
```
../../../../../../../../opt/homebrew/Cellar/go/1.21.0/libexec/src/slices/sort.go:67:7: undefined: min (typecheck)
		m = min(m, x[i])
```

It's fixed if I compile it myself using go1.21

ldez:
> Just FYI regarding https://github.com/golangci/golangci-lint/pull/3922/files#diff-15c806aa509538190832852f439e9921a23bec2da81f95ed0e4bf13c14e5b160R44

This line is to run some tests with the previous version of golangci-lint.
It's not related to the build.

The build use go1.21: https://github.com/golangci/golangci-lint/blob/ed8eb4ea7b773993f0b1a78dec1dc0179146ddf4/.github/workflows/tag.yml#L19

In your log I can see `opt/homebrew/Cellar/`, I guess you are on Darwin, and maybe you are using Brew: Brew doesn't use our binary, it built the binary so it depends on how Brew built it.

The official binaries:
```console
$ golangci-lint version
golangci-lint has version 1.54.1 built with go1.21.0 from a9378d9b on 2023-08-11T12:49:16Z
```

Dasio:
Oh, you are right, I'm sorry, thanks

PR #3929, opened 2023-06-28
feat: explain typecheck and remove it from the linter list

### LINKED ISSUE #3460: Improve documentation about `typecheck`

In in March 2021 we landed #1861 which changed how compilation errors was handled by `golangci-lint` (see #1859 for an example of the improvement). This work was done as a followup of people not understanding why they couldn't lint their code when it didn't compile.

The result was, simply explained, to introduce `typecheck` which acts as a linter but is more of a front-end of the compilation errors. This helped users to see where the compiler pointed at in the code for such compilation errors. Of course, this is just as good as the compiler itself and a lot of compilation issues will not properly show where in the code your error lies. Furthermore one compilation error may often lead to many others making the output look like it contains a lot of issues.

Even though this made us able to see the compilation errors, many users are still not aware about the fact that that's what they are - errors compiling the code. Many users gets frustrated and try to disable the linter (which obviously isn't possible) and a common response from the maintainers is that "[`typecheck` is not a real linter](https://github.com/golangci/golangci-lint/issues?q=is%3Aissue+%22typecheck+is+not+a+real+linter%22%22)"

I think we should start by improving the documentation and better explain what `typecheck` is. We should also clarify that it's not possible to disable this "linter". It may also be a good idea to remove it from the list of linters.

* Pin an issue on the topic
* Add an answer under the [discussions](https://github.com/golangci/golangci-lint/discussions) Q&A
* Continuing replying that "`typecheck` is not a real linter"

### Additional context.

_No response_

### Suggested troubleshooting documentation

In addition to describing what `typecheck` is and why it can't be disabled it would be useful to list some common things to try to find the root cause. Some examples would be

- [ ] Ensure building works with `go build ./...` - whole package
- [ ] Ensure all C dependencies are installed if using CGO
- [ ] Ensure the version of `golangci-lint` is built with a compatible version of the repo - always try latest

ldez:
- add a documentation section about `typecheck`
- remove `typecheck` from the lists of linters (doc, help, etc.)
- enforce the activation of `typecheck`

Fixes #3460


ldez:
Something is weird: some tests don't fail locally but fail on the CI, and the failures on the CI are right.
I don't understand something :thinking:

docs/src/docs/usage/faq.mdx
```diff
@@ -32,3 +32,22 @@ The same as the Go team (the 2 last minor versions)
...
+
+`typecheck` is like the front-end of a Go compiler, parses and type-checks Go code, it manages compilation errors.
+
+It cannot be disabled because of that.
+
+Of course, this is just as good as the compiler itself and a lot of compilation issues will not properly show where in the code your error lies.
+
+`typecheck` is not a real linter, it's just a way to parse compiling errors (produced by the `types.Checker`) and some linter errors.
+
+A linter that reports `typecheck` errors is not able to report other issues for a package because that kind of error doesn't allow it to perform its own analysis.
```
bombsimon, line 46:
```suggestion
A result that reports `typecheck` errors is not able to report other issues for a package because that kind of error doesn't allow it to perform its own analysis.
```

bombsimon (approved):
Nice! Should we add a link to this section in the issue template so we increase the chances of people finding this before creating an issue?

docs/src/docs/usage/faq.mdx
ldez: In this context, it's the case where a linter (not typecheck) reports a typecheck errors.
It's related to the previous sentence "and some linter errors".

IMO, we can keep the sentence or the sentence must rewritten because the word "result" cannot be a substitute of "linter" in the context of the sentence.

ldez:
> so we increase the chances of people finding this before creating an issue?

I will do that inside PR #3930 after the merge of this PR.

docs/src/docs/usage/faq.mdx
ldez: I rewrote the sentence.

bombsimon: Amazing, very clear!

bombsimon (approved):
LGTM!

PR #3930, opened 2023-06-28
dev: improve bug report template

ldez:
Move "Yes, I've included all information below (version, config, etc.)." to the end of the form to enforce the need to include all the required information.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yml
@@ -11,6 +11,8 @@ body:
           required: true
         - label: Yes, I've searched similar issues on GitHub and didn't find any.
           required: true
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -103,3 +101,11 @@ body:
         </details>
     validations:
       required: true
+
+  - type: checkboxes
+    id: validation
+    attributes:
+      label: Validation
+      options:
+        - label: Yes, I've included all information below (version, config, etc.).
```
bombsimon, line 110: I guess this should now be "above" if this will end up at the bottom?

```suggestion
        - label: Yes, I've included all information above (version, config, etc.).
```

bombsimon (approved):
LGTM, pre-approving before adding the `typecheck` note discussed in #3929

PR #3941, opened 2023-07-08
unexportedglobal: Linter to require '_' prefix on globals

abhinav:
This adds a unexportedglobal to golangci-lint.
[unexportedglobal][1] is a linter that requires that unexported global variables and constants are prefixed with '_'.

  [1]: https://github.com/abhinav/unexportedglobal

```go
package foo

// Bad
var pool = sync.Pool{ /* ... */ }

// Good
var _pool = sync.Pool{ /* ... */ }
```

The idea is to eliminate the risk of conflict between names of local variables and names of global variables.

The linter is inspired by the [Prefix Unexported Globals with `_`][2] guidance in [Uber's Go Style Guide][3].

  [2]: https://github.com/uber-go/guide/blob/master/style.md#prefix-unexported-globals-with-_
  [3]: https://github.com/uber-go/guide/blob/master/style.md

The linter is not enabled by default.
I have also not added it to the 'style' preset because not everyone may agree with this change.

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -2175,6 +2175,7 @@ linters:
     - tparallel
     - typecheck
     - unconvert
+    - unexportedglobal
     - unparam
     - unused
     - usestdlibvars
@@ -2289,6 +2290,7 @@ linters:
     - tparallel
     - typecheck
     - unconvert
+    - unexportedglobal
     - unparam
     - unused
     - usestdlibvars
--- a/go.mod
+++ b/go.mod
@@ -115,6 +115,7 @@ require (
 	github.com/yeya24/promlinter v0.2.0
 	github.com/ykadowak/zerologlint v0.1.3
 	gitlab.com/bosi/decorder v0.2.3
+	go.abhg.dev/unexportedglobal v0.2.1
 	go.tmz.dev/musttag v0.7.1
 	golang.org/x/exp v0.0.0-20230510235704-dd950f8aeaea
 	golang.org/x/tools v0.11.0
--- a/go.sum
+++ b/go.sum
@@ -579,6 +579,8 @@ github.com/yusufpapurcu/wmi v1.2.3/go.mod h1:SBZ9tNy3G9/m5Oi98Zks0QjeHVDvuK0qfxQ
 gitlab.com/bosi/decorder v0.2.3 h1:gX4/RgK16ijY8V+BRQHAySfQAb354T7/xQpDB2n10P0=
 gitlab.com/bosi/decorder v0.2.3/go.mod h1:9K1RB5+VPNQYtXtTDAzd2OEftsZb1oV0IrJrzChSdGE=
 go-simpler.org/assert v0.5.0 h1:+5L/lajuQtzmbtEfh69sr5cRf2/xZzyJhFjoOz/PPqs=
+go.abhg.dev/unexportedglobal v0.2.1 h1:uxlNjpBk7WvCaTryyqj2h4aS0t5264cQGrURVbUVLGg=
+go.abhg.dev/unexportedglobal v0.2.1/go.mod h1:cgoKzc+6zAgv8FiRZ1xkZS+OrQcwjrtztP4A0ZJvgRg=
 go.opencensus.io v0.21.0/go.mod h1:mSImk1erAIZhrmZN+AvHh14ztQfjbGwt4TtuofqLduU=
 go.opencensus.io v0.22.0/go.mod h1:+kGneAE2xo2IficOXnaByMWTGM9T73dGwxeWcUqIpI8=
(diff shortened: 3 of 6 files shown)


ldez:

- [ ] The CLA must be signed

### Pull Request Description

### Linter

### `.golangci.reference.yml`

### Others Requirements

### Recommendations

- [ ] The linter should not use SSA. (currently, SSA does not fully support generics)

---

ldez:
> The linter is not enabled by default.

This sentence is partially wrong because all the linters are enabled when using `enable-all`.

I have seen different behavior with the handle of a new linter:
- if a linter doesn't have the "right" default behavior they just ignore it.
- some people just follow the rules because "linters say the truth".

This is why for me the default behavior of a linter (and not default linters) is really important.

> I have also not added it to the 'style' preset because not everyone may agree with this change.

I should be a part of the style preset because it's a linter about style, nothing else is required to be in this preset.

---

I personally dislike global variables: it's a source of confusion, and bugs, and can produce side effects. IMO only a few cases really need global variables.
It's the same thing for `init` function, especially inside a library.

I'm also not a fan of prefixing things: it feels like a rule from another epoch or language.

The fact that the rule comes from a company is just an authority argument, so it's not an argument that interests me.

As you can understand, for now, I disagree with this rule and this linter, but I'm not alone on this project, I will wait for feedback. Also, I need to think more about this rule.

So I'm not making any decision at this time.

abhinav:
Thanks for looking over this, @ldez!

> I should be a part of the style preset because it's a linter about style, nothing else is required to be in this preset.

Makes sense. I'll add it to the preset.

---

> I personally dislike global variables

Same! But are legitimate cases where you need them.
Buffer pools, default values, static collections (e.g., list of known subcommands), etc.
Valid uses are rare, but present.

> I'm also not a fan of prefixing things: it feels like a rule from another epoch or language.

Fair. The prefixing is specifically to separate the namespaces -- similar to how exported variables are already a different namespace (because they start with uppercase letters).

> The fact that the rule comes from a company is just an authority argument, so it's not an argument that interests me.

It was not my intent to use the company's practices as an authority, only as an example.
The argument in favor of the pattern is made in the linked style guide, and in the README for the linter, but in short:

The idea is to prevent mistakes where you accidentally use or modify the global variable when you intended to use the local variable.

By prefixing globals with `_`, we separate the namespaces (because idiomatic Go code will never use a `_` prefixed local variable), and avoid these conflicts while still being able to use short and concise names for them. As global variables are meant to be rare anyway, this makes the choice to refer to the global quite explicit.

This also has a nice effect on readability while reviewing code on a website (GitHub, Phabricator, whatever): when you see a `_`-prefixed variable, you can assume it's a global and don't need to scan upwards in the function to find where that variable was declared.

ldez:
> Same! But are legitimate cases where you need them.
> Buffer pools, default values, static collections (e.g., list of known subcommands), etc.
> Valid uses are rare, but present.

Your examples are the illustration of the problem because they are not valid use cases for me:
- buffer pools could be fields
- default values should be constants or handled in a constructor
- static collections could be functions

Another concrete illustration: you removed a global variable and an `init` function inside your linter. In the context of your linter, and mainly because your linter is a library, those elements are not needed and should be forbidden. We forbid those kinds of elements because, for example, the `init` will be called even if your linter is not used (note it's the same thing for global variables). If there is an error, the error will block golangci-lint (we added this rule because the problem already happened).

Valid use cases are, for example, analyzers that share their reports with other analyzers (like `inspect.Analyzer`) because the instance of the analyzer is used as a key inside `pass.ResultOf`.
But an analyzer/linter inside golangci-lint, because its reports are not shared, doesn't need to be a global variable.

The discussion is a bit off-topic but it's interesting to talk about that :smile_cat:.

Antonboom:
JFYI: some of Uber style guide rules were covered by `ruleguard` – https://github.com/quasilyte/uber-rules
Usually such simple linters do not require be a separate Go `module`.

Also [gochecknoglobals](https://github.com/leighmcculloch/gochecknoglobals) (included in `golangci-lint`) already handles `_` prefix.

ldez:
I don't want to be responsible for a new convention about global variable naming.

This can be implemented as a plugin or as ruleguard rule.

So we will decline the PR.

thank you anyway, the discussion was fun.

abhinav:
Reasonable. Thanks for reviewing.

Thanks for the rule-guard tip, @Antonboom.
(Small correction: gochecknoglobals allows global variables named `_`, not variables named `_foo`.)

Antonboom:
@abhinav I respect your work and don't want to belittle it.
But maybe it will be better to do contribution in `gochecknoglobals`? 🤔

abhinav:
> But maybe it will be better to do configurable contribution in gochecknoglobals? 🤔

No offense taken. Yep, an option in gochecknoglobals was indeed what I was considering next. 😄

PR #3970, opened 2023-07-23
Add tests for the new ginkgolinter's configuration

nunnatsa:
PR #3962 added the new `suppress-focus-container` configuration for ginkgolinter.

This PR adds some tests to check the new configuration.

/cc @ldez


ldez (approved):
We don't really need test cases for every option of a linter: the tests inside your linter repository are enough.
I never refuse tests :smile:

PR #3986, opened 2023-08-03
lll: ignore comments containing only URL

### LINKED ISSUE #3983: lll: Ignore long URLs, strings, and struct tags

There are upstream issues in lll about those with very good arguments:

* https://github.com/walle/lll/issues/12
* https://github.com/walle/lll/issues/11

I think ignoring those should at least be possible through options, but I would even claim that those should be defaults. There is simply no easy way to wrap URLs and struct tags. And wrapping long strings just makes it less readable.

I currently have to manually add lint ignore pragmas in those cases.

### Additional context.

Upstream lll repository seems not maintained. So I am opening an issue here. There is a precedent for this, see upstream issue https://github.com/walle/lll/issues/15 and fix here https://github.com/golangci/golangci-lint/pull/3288.

ldez:
Fixes #3983

As illustrated in the following example, to ignore long URLs the `nolint` directive should be placed on the top of the comments but this ignores `lll` on the whole function.

<details><summary>.golangci.yml</summary>

```yml
linters:
  disable-all: true
  enable:
    - lll
    - typecheck

linters-settings:
  lll:
    line-length: 50
```

```go
package main

import "fmt"

// main runs the app.
// https://github.com/walle/lll/blob/4438bccd245f7e87a5a69023625f01e7035c05c0/utils.go#L15
//
//nolint:lll
func main() {
	fmt.Println("Foo")
}
```

```console
$ golangci-lint run 
$
```

</details>

### DIFF
--- a/pkg/golinters/lll.go
+++ b/pkg/golinters/lll.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"go/token"
 	"os"
+	"regexp"
 	"strings"
 	"sync"
 	"unicode/utf8"
@@ -89,6 +90,8 @@ func getLLLIssuesForFile(filename string, maxLineLen int, tabSpaces string) ([]r
 	lineNumber := 0
 	multiImportEnabled := false
 
+	urlComment := regexp.MustCompile(`\s*//\s*http(s)?://[^ ]+$`)
+
 	scanner := bufio.NewScanner(f)
 	for scanner.Scan() {
 		lineNumber++
@@ -97,6 +100,10 @@ func getLLLIssuesForFile(filename string, maxLineLen int, tabSpaces string) ([]r
 		line = strings.ReplaceAll(line, "\t", tabSpaces)
 
 		if strings.HasPrefix(line, goCommentDirectivePrefix) {
+			continue
+		}
+
+		if urlComment.MatchString(line) {
 			continue
 		}
 
--- a/test/testdata/lll_comment_url.go
+++ b/test/testdata/lll_comment_url.go
@@ -1,0 +1,20 @@
+//golangcitest:args -Elll
+//golangcitest:config_path testdata/configs/lll.yml
+package testdata
+
+import (
+	"fmt"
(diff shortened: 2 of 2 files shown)


test/testdata/lll_comment_url.go
```diff
@@ -0,0 +1,20 @@
+//golangcitest:args -Elll
+//golangcitest:config_path testdata/configs/lll.yml
+package testdata
+
+import (
+	"fmt"
+	_ "unsafe"
+)
+
+// https://github.com/golangci/golangci-lint/blob/master/pkg/result/processors/testdata/autogen_exclude_block_comment.go
```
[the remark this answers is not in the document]
ldez: I think that ignoring all lines with a link is a bad idea

pkg/golinters/lll.go
```diff
@@ -89,6 +90,8 @@ func getLLLIssuesForFile(filename string, maxLineLen int, tabSpaces string) ([]r
 	lineNumber := 0
 	multiImportEnabled := false
 
+	urlComment := regexp.MustCompile(`\s*//\s*http(s)?://[^ ]+$`)
```
mattdowdell, line 93: Would it be possible to also exclude URLs used as reference links? e.g.

```markdown
Here's a [link] and [another][1]!

[link]: https://example.com
[1]: https://example.com
```

I something use these references to keep Go docs readable (inserting long URLs makes following text hard), and given modern godoc uses markdown, this would likely be universally helpful.

batazor:
Sometimes in the code you have to leave links to tasks in the tracker or links to discussion of some bugs - to understand why it was done this way.

And often they are long links, it would be cool to be able to skip such cases by default.

pkg/golinters/lll.go
nieomylnieja: For that I specifically use:

```yaml
  exclude-rules:
    - linters:
        - lll
      source: "^// \\[.*\\]: \\w+"
```

PR #3993, opened 2023-08-09
add goimportgroups linter

kmirzavaziri:
We, in our organization, want to have the following structure for our imports

```
import (
    // standard

    // external

    // company

    // internal
)
```

which is not possible with current linter which uses `goimports`. So I added another linter which can be configured to accept a list of groups and will force the imports to follow the configured pattern.

Each group is configured with a boolean expression of regexps, for example

```
pattern1,pattern2:pattern3
```

means

```
match(s, pattern1) && match(s, pattern2) || match(s, pattern3)
```

where s is the import path.


ldez:
Duplicate of [gci](https://golangci-lint.run/usage/linters/#gci)

PR #4003, opened 2023-08-10
whitespace: update after moving to the `analysis` package

bombsimon:
The `whitespace` linter is rewritten to use `annalysis.Analyzer`. This PR will make use of the new code.

---

~~This is a draft PR until https://github.com/ultraware/whitespace/pull/8 is merged and a new release has been made.~~

### DIFF
--- a/pkg/golinters/whitespace.go
+++ b/pkg/golinters/whitespace.go
@@ -13,7 +13,7 @@ import (
 	"github.com/golangci/golangci-lint/pkg/result"
 )
 
-//nolint:dupl
+
 func NewWhitespace(settings *config.WhitespaceSettings) *goanalysis.Linter {
 	var mu sync.Mutex
 	var resIssues []goanalysis.Issue
--- a/pkg/golinters/whitespace.go
+++ b/pkg/golinters/whitespace.go
@@ -49,8 +49,8 @@ func NewWhitespace(settings *config.WhitespaceSettings) *goanalysis.Linter {
 				switch issue.MessageType {
 				case whitespace.MessageTypeRemove:
 					report.LineRange = &result.Range{
-						From: issue.LineNumber,
-						To:   issue.LineNumber,
+						From: issue.LineNumbers[0],
+						To:   issue.LineNumbers[len(issue.LineNumbers)-1],
 					}
 
 					report.Replacement = &result.Replacement{
--- a/pkg/golinters/whitespace.go
+++ b/pkg/golinters/whitespace.go
@@ -48,6 +48,10 @@ func NewWhitespace(settings *config.WhitespaceSettings) *goanalysis.Linter {
 
 				switch issue.MessageType {
 				case whitespace.MessageTypeRemove:
+					if len(issue.LineNumbers) == 0 {
+						continue
+					}
+
 					report.LineRange = &result.Range{
 						From: issue.LineNumbers[0],
 						To:   issue.LineNumbers[len(issue.LineNumbers)-1],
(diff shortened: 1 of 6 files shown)


pkg/golinters/whitespace.go
```diff
@@ -24,27 +21,58 @@ func NewWhitespace(settings *config.WhitespaceSettings) *goanalysis.Linter {
...
+						To:   issue.LineNumber,
+					}
+
+					report.Replacement = &result.Replacement{
+						NeedOnlyDelete: true,
+					}
+
+				case whitespace.MessageTypeAdd:
+					position := pass.Fset.PositionFor(issue.Diagnostic, false)
+					bracketLine, err := lintCtx.LineCache.GetLine(position.Filename, position.Line)
```
bombsimon, line 62: @ldez Can you help me out here? I would like to get rid of this as well but I must be missing something obvious.

I was hoping I could do just like with `SuggestedFixes` and get the column of the statement and add a `"\n"` of length 1 but this doesn't work as I thought. I was trying to do something like this:

```go
report.Pos = pass.Fset.PositionFor(issue.FixStart, false)
report.Replacement = &result.Replacement{
	Inline: &result.InlineFix{
		StartCol:  0,
		Length:    1,
		NewString: "\n",
	},
}
```

but this makes the tests fail with:

```diff
Diff:
--- Expected
+++ Actual
@@ -50,3 +50,3 @@

-       fmt.Println("Hello world")
+fmt.Println("Hello world")
 }
@@ -57,3 +57,3 @@

-               fmt.Println("Hello multi-line world")
+       fmt.Println("Hello multi-line world")
        }
```

One fix seems to be to always add an extra tab to `NewString` (set it to `\n\t`) and still keep length of 1. This makes the tests pass and it also works for a deeper example like this one:

```go
	if true {
		if true {
			if true {
				if 1 == 1 &&
					2 == 2 {
					fmt.Println("Hello multi-line world")
				}
			}
		}
	}
```

but is it reasonable to rely on `\t` and that the file uses tabs for the fixer? Maybe the fixer will always convert spaces to tabs even if `gofmt` isn't enabled?

ldez: Our internal fixes are really different than SuggestedFixes.
I don't remember exactly but there is a trick and a limitation, I will try it locally.

ldez: from my memories, inline cannot work with line**s**, so `\n` will not work.

bombsimon: Thanks, sounds good to me! I'll go with the `\n\t` fix then because that way we don't need to use the lint cache to read the file.

test/testdata/fix/out/whitespace.go
```diff
@@ -56,4 +56,20 @@ func multiIfFunc() {
...
+
+						fmt.Println("Hello nested multi-line world")
+				}
+			}
+		}
+	}
+}
+
+func notGoFmted() {
+         fmt.Println("Hello world")
```
bombsimon, line 74: Noticed here that we have the same situation as in https://github.com/golangci/golangci-lint/pull/4003#discussion_r1290737071. Basically this results in the standalone linter and `golangci-lint` not fixing the same way, `golangci-lint` will leave the sapces before `fmt` whereas the linter itself will replace it with a tab.

I don't think there's anything we should do extra in this specific linter for this scenario (and also that's the same behaviour as before, we just didn't have a standalone linter) but yet another reason to try to integrate the suggested fixes with the `golangci-lint` fixer.

ldez (approved):
LGTM

PR #4008, opened 2023-08-11
dev: CI cleanups

scop:
Some CI config cleanups, noticed while working on something else. `GITHUB_TOKEN` env scoping not fully tested, but I believe it could work like this.

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -114,9 +114,6 @@ jobs:
   check_generated:
     needs: golangci-lint # run after golangci-lint action to not produce duplicated errors
     runs-on: ubuntu-latest
-    env:
-      # needed for github-action-config.json generation
-      GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
     steps:
       - uses: actions/checkout@v3
       - name: Unshallow
@@ -131,3 +128,6 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
       - name: Check generated files are up to date
         run: make fast_check_generated
+        env:
+          # needed for github-action-config.json generation
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -7,8 +7,6 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    env:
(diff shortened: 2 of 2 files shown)


.github/workflows/tag.yml
```diff
@@ -19,14 +19,12 @@ jobs:
           # - 1.18beta1 -> 1.18.0-beta.1
           # - 1.18rc1 -> 1.18.0-rc.1
           go-version: '1.21'
-      - name: Unshallow
-        run: git fetch --prune --unshallow
```
ldez, line 23: can you restore this lines ? (related to my previous comment)

.github/workflows/tag.yml
```diff
@@ -7,10 +7,10 @@ on:
 jobs:
   release:
     runs-on: ubuntu-latest
-    env:
-      GITHUB_TOKEN: ${{ secrets.GOLANGCI_LINT_TOKEN }}
     steps:
       - uses: actions/checkout@v3
+        with:
+          fetch-depth: 0
```
ldez, line 13: this will not work: for tags we need the full graph

ldez (changes requested):

.github/workflows/pr.yml
```diff
@@ -131,3 +128,6 @@ jobs:
           go-version: ${{ env.GO_VERSION }}
       - name: Check generated files are up to date
         run: make fast_check_generated
+        env:
+          # needed for github-action-config.json generation
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
ldez, line 133: this decreases the readability: the token's scope is for the job.

ldez (changes requested):

.github/workflows/tag.yml
scop: Removed that change.

.github/workflows/pr.yml
scop: Maybe. On the other hand, the token is a security sensitive thing, and it seems prudent to make it available to as few steps as possible. In a sense, the scope of the token can be said to be the step(s) of a job where it's needed, only one here that I can tell. But sure, can revert this one if you want.

What about the corresponding change in tag.yml, it contained the same env/secret setting both on the job and the step level where it's needed? I removed the apparently redundant job level one.

ldez (approved):
Note: It's better to address review as a commit, it's better to follow changes.

PR #4009, opened 2023-08-12
dev: packaging improvements

scop:
Include section, group, soft dependency on (packaged) Go, and LICENSE and README.md in docs.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -106,7 +106,7 @@ nfpms:
     maintainer: "golangci-lint Team <golangci-releaser@users.noreply.github.com>"
     description: Fast linters Runner for Go
     license: GPLv3
-    section: devel
+    section: golang
     formats:
       - deb
       - rpm
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -106,6 +106,29 @@ nfpms:
     maintainer: "golangci-lint Team <golangci-releaser@users.noreply.github.com>"
     description: Fast linters Runner for Go
     license: GPLv3
+    section: devel
     formats:
       - deb
       - rpm
+    umask: 0o022
+    overrides:
+      deb:
+        contents:
+          - src: LICENSE
+            dst: /usr/share/doc/golangci-lint/copyright
+          - src: README.md
+            dst: /usr/share/doc/golangci-lint/README.md
+        recommends:
+          - golang-go
+      rpm:
+        contents:
+          - src: LICENSE
+            dst: /usr/share/doc/golangci-lint/LICENSE
+            type: license
+          - src: README.md
+            dst: /usr/share/doc/golangci-lint/README.md
+            type: readme
+        recommends:
+          - /usr/bin/go
+    rpm:
+      group: Development/Tools


.goreleaser.yml
```diff
@@ -106,6 +106,29 @@ nfpms:
     maintainer: "golangci-lint Team <golangci-releaser@users.noreply.github.com>"
     description: Fast linters Runner for Go
     license: GPLv3
+    section: devel
```
ldez, line 109:
```suggestion
    section: golang
```

Based on Debian documentation, the section must be `golang`:
- https://www.debian.org/doc/debian-policy/ch-archive.html#s-subsections
- https://packages.debian.org/unstable/
- https://packages.debian.org/unstable/golang/

.goreleaser.yml
```diff
@@ -106,6 +106,29 @@ nfpms:
...
+      rpm:
+        contents:
+          - src: LICENSE
+            dst: /usr/share/doc/golangci-lint/LICENSE
+            type: license
+          - src: README.md
+            dst: /usr/share/doc/golangci-lint/README.md
+            type: readme
+        recommends:
+          - /usr/bin/go
```
ldez, line 132:
```suggestion
    contents:
      - src: LICENSE
        dst: /usr/share/doc/golangci-lint/copyright
      - src: README.md
        dst: /usr/share/doc/golangci-lint/README.md
    overrides:
      deb:
        recommends:
          - golang-go
      rpm:
        recommends:
          - /usr/bin/go
```

Based on the [JSON schema](https://github.com/goreleaser/goreleaser/blob/main/www/docs/static/schema.json), the property `type` inside `content` cannot be "license" or  "readme":
https://github.com/goreleaser/goreleaser/blob/4526c5b6813ade905f9658467c3a85aa25bdc9b9/www/docs/static/schema.json#L682-L692

ldez (changes requested):

.goreleaser.yml
scop: Looks like the schema is out of date. Or maybe they're not there because currently they can only be used with the `rpm` packager; anyway they can be those values: https://github.com/goreleaser/nfpm/blob/d309eb60151adcc3df3b73a78250d77fe469558c/files/files.go#L43-L47, and setting those values do take effect (see #4010 for what kind of a build the example package below originates from):

```shellsession
$ rpm -qLp "https://packagecloud.io/scop/golangci-lint/packages/rpm_any/rpm_any/golangci-lint-0.0.42-1.x86_64.rpm/download.rpm?distro_version_id=227"
/usr/share/doc/golangci-lint/LICENSE
```

I don't know/remember offhand a way to query rpms for readme files, but that's off less importance. Files labeled as such do end up in the package anyway, so I don't think it hurts. But at least labeling license files properly as such is quite desirable, that I recommend keeping even if removing the marker from the readme.

(FWIW see also https://github.com/goreleaser/nfpm/pull/698 for an upstream RFE/WIP PR to make use of these types less of a chore.)

.goreleaser.yml
scop: Interestingly, `golang` is not listed in the policy manual sections link. But seems appropriate anyway, done in 9c1cf7ef72896e682e58b9775f3851fdb8691fb9

ldez (approved):

PR #4010, opened 2023-08-12
dev: upload debs and rpms to Packagecloud

scop:
Keeping local installs of golangci-lint on Linux systems up to date is a bit of a chore, because building from source is not recommended (so tools like `gup` won't help), and the binaries are not easily available for automatic installation, need to manually download and install from GH releases.

Since the .deb and .rpm packages are already available, it would be nice to have them in a public repo somewhere. This PR implements uploading them to https://packagecloud.io (I have no affiliation with them).

Demo release and repository created by this PR's content (along with some tweaks to avoid golangci specific things):
- Demo release including the footer note: https://github.com/scop/golangci-lint/releases/tag/v0.0.42 (do not mind the huge changelog, see below it)
- Populated packagecloud.io repository: https://packagecloud.io/scop/golangci-lint (packagecloud renames the rpms and debs to their liking)
(The packages in the repo are built off a branch that has also #4009 applied, but that's not a blocker for this.)

To make it work for the golangci/golangci-lint project, the PR requires two things set up by golangci-lint project admins:
1) A packagecloud.io account. This PR assumes the account name over there will be `golangci`, and the repo name `golangci-lint`. If they end up being something else, we can obviously adjust here. But I think they would make sense.
2) Packagecloud.io API token in the GH golangci-lint project's [actions secrets](https://github.com/golangci/golangci-lint/settings/secrets/actions) -> repository secrets, as `PACKAGECLOUD_TOKEN`.

### DIFF
--- a/.github/workflows/tag.yml
+++ b/.github/workflows/tag.yml
@@ -30,6 +30,16 @@ jobs:
         env:
           GITHUB_TOKEN: ${{ secrets.GOLANGCI_LINT_TOKEN }}
 
+      - name: Upload nfpms
+        run: |
+          gem install --no-document --user-install --bindir ~/.local/bin package_cloud
+          export CLICOLOR_FORCE=1
+          # deb: armv6 clashes with armv7, no loong64/mips64/riscv64 support (mips64le is ok)
+          ~/.local/bin/package_cloud push ${{ github.repository }}/any/any $(ls -1 dist/*.deb | grep -vwE 'armv6|loong64|mips64|riscv64')
+          ~/.local/bin/package_cloud push ${{ github.repository }}/rpm_any/rpm_any dist/*.rpm
+        env:
+          PACKAGECLOUD_TOKEN: ${{ secrets.PACKAGECLOUD_TOKEN }}
+
   docker-release:
     needs: release
     runs-on: ubuntu-latest
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -5,6 +5,9 @@ release:
   github:
     owner: golangci
     name: golangci-lint
+  footer: >-
+    apt and dnf/yum package repositories are available at
+    [Packagecloud](https://packagecloud.io/golangci/golangci-lint).
 
 builds:
   - binary: golangci-lint


ldez:
have you a paid or trial account?

scop:
Free account works just fine for this purpose.

ldez:
Free account but with a trial.

scop:
I think that's just misleadingly put. Free with a trial does not make much sense to me in the first place, they say "we offer a limited Free Plan *and also* offer 14 day trials of any of our paid plans". I've had mine for 2+ months now; it does have a "Free trial" label in the UI, and they've "charged" me $0.00 twice each month for it (no credit card required).

scop:
Cloudsmith and Balto have free repo offerings for OSS projects, too, in case Packagecloud does not seem right. I don't have experience with either: https://cloudsmith.com/pricing, https://www.getbalto.com/pricing (scroll down/ search for "Open source" in both).

scop:
I've sent mail to support@packagecloud.io, asking them to chime in here to clarify.

scop:
Got a response, and a permission to copy it here, so here goes. Note that I'm just relaying their message here, and not taking any responsibility for its content, nor what if anything the golangci-lint project decides to do based on it nor the consequences of those decisions. If further clarifications/discussion is required, I suggest contacting their support directly.

> Paula from Packagecloud.io <paula@[redacted-by-scop]> | Fri, 8 Sept, 00:40 (2 days ago)
>
> Hello again Ville!
>
> Thanks for your patience!
> When you sign up, you have a 14-day trial on the Pro Plan. Once you hit those 14 days, you'll need to add a payment method to continue on the Pro plan (or switch to a different plan). This is the trial period.
>
> We offer a Free Plan, which is independent of the Trial period. You can check from your Subscription section that you are on a Free Plan, and that one includes 2 GB of storage and 10 GB of bandwith per month.
>
> May I ask for your feedback on the following sentence that we have in our [price page](https://packagecloud.io/pricing/?_ga=2.266544459.418383823.1694028698-1964155450.1674071809)? :
> "Yes, we offer a limited [Free Plan](https://packagecloud.io/users/new?plan=free_usage_plan) and also offer 14 day trials of any of our paid plans."
> 
> Do you think is confusing?
>
> Thanks!

(There was a "We offered a Free Plan" typo in their original which I confirmed with them it was a typo indeed; the corrected/intended form "We offer a Free Plan" is modified by me inline in the above.)

scop:
Anything I can do to help with nudging this forward?

If for example Cloudsmith sounds more like it than Packagecloud (their quotas are larger) I could have a look into implementing that. There's a "Contact us for details" button at the end of the page at https://cloudsmith.com/pricing that should probably be followed by a project representative here to clarify the terms first, though. Pushing itself looks trivial: https://help.cloudsmith.io/docs/integrating-with-github-actions#push-debian-package

PR #4028, opened 2023-08-19
govet: add directive and slog passes

johejo:
ref
- https://github.com/golang/go/issues/59407
- https://github.com/golang/go/issues/56986


ldez (approved):
LGTM

Issue #4027, opened 2023-08-18
Possibility to override/modify issue texts


zak-pawel:
### Your feature request related to a problem? Please describe.

Would be great to have a configurable feature which will allow to override/modify `Text` from `Issue`.
Mainly in order to better guide the developers towards the best way to fix issues found by golangci-lint.

Example for `gocritic`->`appendAssign`:
Original message: `appendAssign: append result not assigned to the same slice`
Adjusted message: ``appendAssign: append result not assigned to the same slice, use `slices.copyAndAppend` to safely concatenate two slices``

### Describe the solution you'd like.

`IdentifierMarker` can be used for that. 
Possibility of passing slice of `replacePattern`s to `NewIdentifierMarker()` can be added.

User will be able to provide list of pairs (regexp and string) in configuration file.
To achieve adjustment for mentioned case (`gocritic`->`appendAssign`), following config can be provided in `issues` section:
```
  override-messages:
    - regexp: "^(appendAssign: append result not assigned to the same slice)$"
      replacement: "${1}, use `slices.copyAndAppend` to safely concatenate two slices"
```

### Describe alternatives you've considered.

Didn't find any alternative.

ldez:
Hello,

I think it's better to improve at the root (linter messages) instead of trying to "fix the consequences".

zak-pawel:
> I think it's better to improve at the root (linter messages) instead of trying to "fix the consequences".

@ldez 
It makes sense for some messages. 

For case I provided for **CUSTOM** function `slices.copyAndAppend` it is not possible to improve at the root (because linter doesn't have any knowledge about my function(s) in my repo).
I can provide few more real-life examples if you want :)

ldez:
I understand your topic but I think it will introduce complexity and performance regression only for a niche.

Avoid UPPERCASE and **bold** or **BOTH**, it feels a bit aggressive.

zak-pawel:
> Avoid UPPERCASE and **bold** or **BOTH**, it feels a bit aggressive.

@ldez 
Sorry, it wasn't my intention to be aggressive, only wanted to emphasize type of function :)


> I think it will introduce complexity and performance regression only for a niche.

Yes, I agree that some complexity and performance regression may be introduced. But only when this feature is used. This will not change anything for projects that do not use it.

ldez:
> This will not change anything for projects that do not use it.

the problem is not exactly here: to evaluate messages we will need to evaluate all the messages with regexp for all the reports.

[...]

PR #4031, opened 2023-08-20
dev: Add MessageAdjuster processor

zak-pawel:
This PR adds `MessageAdjuster` processor which will allow to override/modify issue message (`Text` from `Issue`).
Mainly in order to better guide the developers towards the best way to fix issues found by `golangci-lint`.

```
  # Changing the content of messages.
  # You can specify the existing-message by regular expression,
  # and new-message by regular expression expansion syntax like below
  # (see https://pkg.go.dev/regexp#Regexp.Expand for details).
  # Specifying (optional) linter will speed up processing.
  adjust-messages:
    - linter: gocritic
      existing-message: "^(appendAssign: append result not assigned to the same slice)$"
      new-message: "${1}, use `slices.copyAndAppend` to safely concatenate two slices"
```

resolves: https://github.com/golangci/golangci-lint/issues/4027

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -2434,6 +2434,16 @@ issues:
     - EXC0014
     - EXC0015
 
+  # Changing the content of messages.
+  # You can specify the existing-message by regular expression,
+  # and new-message by regular expression expansion syntax like below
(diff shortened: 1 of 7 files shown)


ldez:
I declined the issue so the PR will not be accepted.

PR #3521, opened 2023-01-30
feat: check only changed files in pre-commit hook

davidjb:
Fixes #1245
by passing filenames from pre-commit to `golangci-lint run` (which supports this since that issue was originally opened) and when used with `--new-from-rev HEAD` allows for partial commit support.

Issue #1245, opened 2020-07-15
pre-commit hook runs on all files instead of just changed files in the repo


coderavels:
I want to run the tool only on the changed files since I am integrating it to an existing repo which is huge. Fixing all the errors at once is not practical. 
I am using the `run: new` config in the `.golangci.yml` but it still runs on all the files not just the changed files.

I have created this small repo to mock the issue: https://github.com/heisfullstacked/golintci-test
If I uncomment the comment in the file `service/should_be_lint.go` and try to commit, I am expecting the pre-commit to only show errors in that file. 
Instead I see errors from both `main.go` and `service/should_be_lint.go`
```
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
Check Yaml...........................................(no files to check)Skipped
Check for added large files..............................................Passed
golangci-lint............................................................Failed
- hook id: golangci-lint
- exit code: 1

service/should_be_lint.go:7:6: `change` is unused (deadcode)
type change string
     ^
main.go:14:1: commentFormatting: put a space between `//` and comment text (gocritic)
//type change string
^

```

### Additional Information

- [x ] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [ x] Yes, I've searched similar issues on GitHub and didn't find any.
- [ x] Yes, I've included all information below (version, config, etc).

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
# golangci-lint has version 1.27.0 built from fb74c2e on 2020-05-13T18:45:55Z
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
linters-settings:
  depguard:
    list-type: blacklist
    packages:
      # logging is allowed only by logutils.Log, logrus
      # is allowed to use only in logutils package
      - github.com/sirupsen/logrus
    packages-with-error-message:
      - github.com/sirupsen/logrus: "logging is allowed only by logutils.Log"
  dupl:
    threshold: 100
  funlen:
    lines: 100
    statements: 50
  goconst:
    min-len: 2
    min-occurrences: 2
  gocritic:
    enabled-tags:
      - diagnostic
      - experimental
      - opinionated
      - performance
      - style
    disabled-checks:
      - dupImport # https://github.com/go-critic/go-critic/issues/845
      - ifElseChain
      - octalLiteral
      - whyNoLint
      - wrapperFunc
  gocyclo:
    min-complexity: 15
  goimports:
[...]

PR #4046, opened 2023-08-29
Add a pre-commit hook to check all files

hanzo:
https://github.com/golangci/golangci-lint/pull/3521 added `--new-from-rev HEAD` to the pre-commit hook with the goal of making it easier to integrate golangci-lint into an existing code base without needing to address all existing lint errors. The problem with only checking changed files is that it makes it far too easy for new lint errors to be introduced: as soon as a developer runs `git commit --no-verify` they will never again be warned about the lint errors they just added. It's common to commit some incomplete code changes to save your progress before you're ready to polish the code and clean up every lint error, so you run `git commit --no-verify` and now the lint errors are permanently ignored.

The same is true for CI: a unit test that runs golangci-lint via pre-commit can now only warn about errors that were introduced in the most recent commit instead of the entire pull request. This also prevents linters from running against the entire codebase post-merge to catch lint errors that were caused by interactions between multiple PRs.

Instead of hardcoding `--new-from-rev HEAD` in the pre-commit hook, developers who desire this behavior can achieve this via configuration as described in [this comment](https://github.com/golangci/golangci-lint/issues/1245#issuecomment-983686674).

cc @davidjb @ldez - curious to hear your thoughts on this


davidjb:
This PR shouldn't be merged.

@hanzo `pre-commit` is typically intended to run only on changed files (and the specific staged changes from those files) by default before a commit, so what you're describing as a concern is `pre-commit`'s intended behaviour. If a developer is bypassing hooks in order to commit invalid code, that goes against the intent of `pre-commit` which is to catch issues ahead of commit (https://pre-commit.com/#introduction).

By forcing a check of _all_ files in a repo during every pre-commit, you'll end up seeing unrealted linting errors which then prevent otherwise valid commits from being permitted. For example, if my commit changes only file A but unrelated file B has linting errors, checking all files means my commit will be blocked until I either fix file B, or use `--no-verify` -- both of which are non-solutions.  If your intent is to avoid devs using `--no-verify`, then forcing a repo-wide check of unrelated files will see it used _more_ (and/or your commits becoming less atomic).

`pre-commit` has an option for running hooks against all files, see https://pre-commit.com/#4-optional-run-against-all-the-files, but it sounds as though you just want to run `golangci-lint` over your whole code base. If so, then do this separately - just call `golangci-lint` directly, either locally or in CI; `pre-commit` is, as the name suggests, something that happens before a commit, rather than after.

If you _really_ want your entire code base to be checked ahead of every commit, then I'd suggest creating a local hook with what's in your PR.

hanzo:
Thanks for your reply, @davidjb. I'm aware of pre-commit's intended behavior. 

To describe my organization's setup a little better: we have a microservice architecture with over 300 Go repos, and we want our linters to run the same way on every repo, whether you're running locally before committing or whether it's a CI check. It turns out that pre-commit works great for both, and we can easily manage the versions of golangci-lint and other linters in the .pre-commit-config.yaml file in each repo via automation. In every case we just use `pre-commit run --all-files` to run golangci-lint and other linters, and the same configuration is used both locally and in CI. We've already adopted a standard set of golangci-lint linters fleet-wide, and each of those 300 Go repos is fully conformant with the latest linter requirements. This setup has been working quite well for years. I recommend you try it! 

I understand that this might not be the most common scenario for users of golangci-lint. The part that's not clear to me is why the behavior added by https://github.com/golangci/golangci-lint/pull/3521 needs to be hardcoded in the pre-commit hook, when it could easily be achieved by configuration instead?

aneeshusa (approved):
When I originally added the .pre-commit-hooks.yaml in this repo (https://github.com/golangci/golangci-lint/pull/311) I intentionally did not include `--new-from-rev`. pre-commit should catch the majority of issues locally, but it's explicitly designed to be [used in CI](https://pre-commit.com/#usage-in-continuous-integration) - see their first-class https://pre-commit.ci/ service - and running it in CI lets you reuse the same linter definition instead of having a separate definition (e.g. running golangci-lint directly) which will get out of sync.

I'd also agree that enabling new-from-rev is an option downstream customers can enable easily - either in .golangci.yml, or by adding `args: [--new-from-rev, HEAD]` in their .pre-commit-config.yaml. Conversely, removing this flag is much harder (.golangci.yml config won't work as CLI args take precedence; can't use `args` in .pre-commit-config.yaml, must override the whole `entry` so blocked from getting future upstream fixes). I'd suggest we prefer correctness for varied workflows over optimizing for a specific workflow.

ldez:
`pre-commit` is here to check the code before a commit, but this does not imply directly the scope of the check.
I can agree with @davidjb, a restricted scope to the commit can be expected.
But I can also understand that a restricted scope is not necessarily implied by the use of `pre-commit`.

Using `new` inside the configuration doesn't really seem a good approach because it will not be conditional.

The 2 points of view about `pre-commit` are understandable and don't allow a compromise.

I don't use `pre-commit`, maybe it's possible to define 2 hooks inside the `pre-commit-hooks.yaml`, one for each use case.

Note: using `pre-commit` is, for me, a bad idea because this tool is a local tool to use before a commit.

hanzo:
Thanks for your input, @ldez. I agree that it seems like the best path forward is to define two hooks. 

Even ignoring the CI use case, `--new-from-rev HEAD` breaks linters like `unused` which need to run against the entire repo. I've updated my PR to create a new hook named `golangci-lint-diff` which uses `--new-from-rev HEAD` for those who desire this behavior. Any suggestions for a better name for the hook?

ldez:
I would prefer the opposite:
- default: with `new-from-rev` same as today
- `golangci-lint-full`: without `new-from-rev`

hanzo:
I've updated the PR as you suggested to add a hook named `golangci-lint-full` without `new-from-rev`. 

I've also added descriptions for both hooks to make their behavior more obvious to future users. My team had to spend weeks undoing all of the lint errors that were introduced after https://github.com/golangci/golangci-lint/pull/3521 silently broke our CI for months.

hanzo:
@ldez thanks for updating the title. Anything else you'd like to see changed before this is ready to merge?

ldez (approved):
LGTM

hanzo:
Hi @ldez, would it be possible to create a new release that includes this and the rest of the recent commits? I can't tell from the [docs](https://golangci-lint.run/contributing/workflow/#new-releases) if anyone can create a release, or if this is only done by maintainers.

ldez:
I will not create a release only for pre-commit.
I have some PRs to review, I will create a release after that.

PR #4069, opened 2023-09-06
add linter: protogolint

ghostiam:
[Protogolint](https://github.com/ghostiam/protogolint) analyzes code and helps detect direct `protobuf` field accesses that could give rise to panic.

The linter suggests using getters:
```go
m.GetFoo().GetBar().GetBaz()
```
instead of direct field access:
```go
m.Foo.Bar.Baz
```

A little more information: https://github.com/golang/protobuf/issues/54#issuecomment-125386932


ldez:

- [ ] The CLA must be signed

### Pull Request Description

### Linter

### `.golangci.reference.yml`

### Others Requirements

### Recommendations

- [ ] The linter should not use SSA. (SSA can be a problem with generics)

---

ldez:
FYI 0.0.x are not semver complaint because a bug fix cannot happen before a minor release or an impossible initial version v0.0.0.

> PATCH version when you make backward compatible bug fixes

https://semver.org/

> How should I deal with revisions in the 0.y.z initial development phase?
>
> The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.

https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

ghostiam:
@ldez corrected the version

go.mod
```diff
@@ -135,6 +135,7 @@ require (
 	github.com/ettle/strcase v0.1.1 // indirect
 	github.com/fatih/structtag v1.2.0 // indirect
 	github.com/fsnotify/fsnotify v1.5.4 // indirect
+	github.com/ghostiam/protogolint v0.1.0 // indirect
```
ldez, line 138: your linter cannot be an indirect dependencies.

ldez (changes requested):
some elements don't fit our checklist (see my previous message)

go.mod
ghostiam: This happened due to the call `go get -u`, fixed it using `go mod tidy`

ghostiam:
> The files (tests and linter) inside golangci-lint must have the same name as the linter.

Because my linter tests require a dependency on `protobuf`, the test is in
`test/testdata/protogolint/protogolint.go` with `go.mod` and `go.sum`
instead of
`test/testdata/protogolint.go`

> They must have at least one std lib import.

I'm importing `fmt`

> They must work with T=<name of the linter test file>.go make test_linters. (the team will help to verify that)

<details>
  <summary>T=protogolint make test_linters</summary>
GL_TEST_RUN=1 go test -v ./test -count 1 -run TestSourcesFromTestdata/protogolint
=== RUN   TestSourcesFromTestdata
    linters_test.go:21: testdata/*.go
--- PASS: TestSourcesFromTestdata (4.09s)
=== RUN   TestSourcesFromTestdataSubDir
=== RUN   TestSourcesFromTestdataSubDir/protogolint
    linters_test.go:40: testdata/protogolint/*.go
=== RUN   TestSourcesFromTestdataSubDir/protogolint/protogolint.go
=== PAUSE TestSourcesFromTestdataSubDir/protogolint/protogolint.go
=== CONT  TestSourcesFromTestdataSubDir/protogolint/protogolint.go
level=info msg="[test] ran [<redacted>/src/github.com/golangci/golangci-lint/golangci-lint run --internal-cmd-test --no-config --allow-parallel-runners --disable-all --out-format=json --max-same-issues=100 -Eprotogolint protogolint.go] in 1.020834709s"
level=info msg="[test] ran [<redacted>/src/github.com/golangci/golangci-lint/golangci-lint run --internal-cmd-test --no-config --allow-parallel-runners --disable-all --out-format=json --max-same-issues=100 -Etypecheck -Eprotogolint protogolint.go] in 251.07975ms"
--- PASS: TestSourcesFromTestdataSubDir (1.47s)
    --- PASS: TestSourcesFromTestdataSubDir/protogolint (0.19s)
        --- PASS: TestSourcesFromTestdataSubDir/protogolint/protogolint.go (1.27s)
PASS
ok      github.com/golangci/golangci-lint/test  6.983s
</details>

Antonboom:
@ghostiam, hi!

Thank you for linter.
I deeply reviewed it.

1) Naming proposal – `protogolint` is too general and conflicts with [protolint](https://github.com/yoheimuta/protolint) and similar. If in the future will appear the new proto-related linter when we going to question – why is it not part of your linter? IMHO, better to narrow naming to specific functionality. In your case – `protoaccess`, `protoderefchecker`, `protogetter`, `protodeferlint`, `protousegetter`, `protoshield`, `protonodirectread`, `protoreadguard` (help me 🙂). Inspiration can be drawn from https://golangci-lint.run/usage/linters/

2) 
```go
// want "proto field read without getter:"
```
What is purpose of `:` in the end?
Based on implementation looks like a bug / typo.

3)
```go
// want "proto field read without getter:" "proto field read without getter:" "proto field read without getter:" "proto field read without getter:"
```
IMHO, it is cool to merge the similar messages ([example](https://github.com/kulti/thelper/blob/1ed2e864ac8709945d823d882736c3931c178740/pkg/analyzer/report.go#L49)) or make it more informative (see below).

4) 
It is cool to make message more informative and without tautology, e.g.
```go
// want "proto field `Embedded` read without getter"
// want "direct access to proto `Embedded` field" (golangci-lint appends linter name as prefix)
// want "direct access to `Embedded` field of proto type `proto.Test`"
``` 
(help me 🙂)

5) 
Great idea with different suggested fixes modes:
```go
const (
	StandaloneMode Mode = iota
	GolangciLintMode
)
```
My respect!

6)
You wrote
```go
Requires: []*analysis.Analyzer{inspect.Analyzer},
```

but dont use this dependency.

7)
You ignore linting for linter project (see check list).
E.g. I see preallocation cases.

8)
IMHO, try to avoid strings comparison like
```
return sct.Field(0).Type().String() == messageState
```
(but in this case I cannot suggest better approach)

Also, old projects uses `github.com/golang/protobuf`.


9)
```
// If getter exists, use it.
```

In such cases it doesn't exist?

10)
**IMHO**

```
func (c *Checker) Check
```
is not the best naming, it's more about suggested fixing, not checking itself.
or `Checker` type itself just does not match the letter S from SOLID.

In addition `check` looks over-engineered a lot.

I expected simple logic like:
- find `SelectorExpr`
- if not Getter then print warning and add suggest fix with simple ("Get"+value), or with help of `go/format.Node`

11)

Reporting linter errors via diagnostic doesn't look as common and nice practice
```go
if err != nil {
	pass.Report(analysis.Diagnostic{
		Pos:     n.Pos(),
		End:     n.End(),
		Message: fmt.Sprintf("error: %v", err),
	})
	return nil
}
```
(@ldez knows better)

🙂

ldez (approved):
LGTM

pkg/lint/lintersdb/manager.go
```diff
@@ -715,6 +715,13 @@ func (m Manager) GetAllSupportedLinterConfigs() []*linter.Config {
 			WithPresets(linter.PresetStyle).
 			WithURL("https://github.com/yeya24/promlinter"),
 
+		linter.NewConfig(golinters.NewProtoGetter()).
+			WithSince("v1.55.0").
+			WithPresets(linter.PresetBugs).
+			WithLoadForGoAnalysis().
+			WithAutoFix().
```
ash2k, line 722: brew updated my golangci-lint and I suddenly started getting all my proto field accesses rewritten. This is not cool, they are not bugs and I didn't ask for this "fix". I suggest this behavior should not be the default.

ghostiam: This linter is not enabled by default, you should have turned it on yourself.
If you have all linters enabled, you can disable `protogetter` through the configuration by adding a linter to the `disable` array.
https://golangci-lint.run/usage/configuration/#linters-configuration

Antonboom: @ash2k hi!

`WithAutoFix()` affects documentation only.
Do you sure that you run `golangci-lint` without `--fix` flag?

ash2k: Looks like I have `fix: true` in my config. I used [the reference config](https://github.com/golangci/golangci-lint/blob/5a0f059f450ce822d2e0ad3167331f5bcaa9559b/.golangci.reference.yml#L2594) (a long time ago) as a template and this behavior was and still is the default. I didn't know fixing was enabled as no other linter ever made any fixes. In fact, I didn't know linters can autofix issues they find.

I've disabled the linter and autofixing for now.

> If you have all linters enabled

Yes, I do. I don't have time to inspect each version and learn about the new linters and their configuration so this is convenient because that's how I discover there is a new linter in a new version and that it found something suspicious. I just didn't expect it to mess with me code :)

Thank you for educating me. I've corrected my config.

Antonboom: @ash2k Also it may not be obvious, but this config describes default values in field comments, and uses an example of non-default values in the fields themselves 🙂 

Example:
```yml
   # Independently of option `exclude` we use default exclude patterns, it can be disabled by this option.
   # Default: true.
   exclude-use-default: false
```

The default `fix` option value is `false` but comment was missed 😢  We'll fix it

PR #4098, opened 2023-09-21
depguard: add current module name to allow list

mirecl:
**AS-IS**: `depguard/v2` automatic add current module name (from go.mod) in deny list.
**TO-BE**: `depguard/v2` automatic add current module name (from go.mod) in allow list.

### DIFF
--- a/pkg/golinters/depguard.go
+++ b/pkg/golinters/depguard.go
@@ -1,7 +1,10 @@
 package golinters
 
 import (
+	"os"
+
 	"github.com/OpenPeeDeeP/depguard/v2"
+	"golang.org/x/mod/modfile"
 	"golang.org/x/tools/go/analysis"
 
 	"github.com/golangci/golangci-lint/pkg/config"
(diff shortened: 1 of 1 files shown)


mirecl:
@alexandear, i'm sure this will be a useful addition, what do you think?

ldez:
Hello,

I think it will create confusion: it's better to let the user create it's own configuration.
Also, the way to get the go.mod is too naive and it will not work in some context.

I will decline this PR.

mirecl:
@ldez, your logic is strange, all will continue to do this setting, add current module to allow list!

ldez:
An example of a real configuration:

```yml
  depguard:
    rules:
      main:
        deny:
          - pkg: "github.com/instana/testify"
            desc: not allowed
          - pkg: "github.com/pkg/errors"
            desc: Should be replaced by standard lib errors package
```

As you can see the current module doesn't need to be added.

mirecl:
But if you add at least 1 rule to the allowed list, then the current module becomes prohibited.

ldez:
Yes because it's the way that depguard v2 works.

PR #4105, opened 2023-09-25
docs: add source options to exclude-rules docs

### LINKED ISSUE #3409: Document the "source" option in exclude-rules

There is apparently a `source` option which is available under `exclude-rules` in order to exclude a rule from code lines that match a particular regular expression.  After some amount of digging around and first considering filing an issue about exclusions at least for the `lll` linter, I found it mentioned here: https://github.com/golangci/golangci-lint/issues/207#issuecomment-534771981

Unfortunately this option is not documented on the False Positives page at https://golangci-lint.run/usage/false-positives/ which is where I would have expected it: searching the page for `source` yields zero results.  This was actually one of the first places I looked to try to find out if such a feature might exist.

Instead, the only mention is at https://golangci-lint.run/usage/configuration/#issues-configuration in a passing example that doesn't go into the depth of explanation that the False Positives page does.

Add documentation for the `source` option to the False Positives documentation page.

none

### Additional context.

_No response_

hendrywiranto:
Add `source` options to `exclude-rules` docs.

Fix #3409

### DIFF
--- a/docs/src/docs/usage/false-positives.mdx
+++ b/docs/src/docs/usage/false-positives.mdx
@@ -31,7 +31,7 @@ linters-settings:
 
 Exclude issue by text using command-line option `-e` or config option `issues.exclude`.
 It's helpful when you decided to ignore all issues of this type.
-Also, you can use `issues.exclude-rules` config option for per-path or per-linter configuration.
+Also, you can use `issues.exclude-rules` config option for per-path, per-linter, per-text or per-source configuration.
 
 In the following example, all the reports that contains the sentences defined in `exclude` are excluded:
 
@@ -62,7 +62,6 @@ issues:
         - lll
       source: "^//go:generate "
 ```
-
 
 In the following example, all the reports that contains the text (`text`) in the path (`path`) are excluded:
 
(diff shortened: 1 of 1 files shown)


hendrywiranto:
Hi @ldez sorry for mentioning you, is there anything that I can help to get this PR merged? or maybe the documentation is not needed anymore? if that's the case I can just close the PR  
thank you

alexandear (approved):

docs/src/docs/usage/false-positives.mdx
```diff
@@ -31,7 +31,7 @@ linters-settings:
 
 Exclude issue by text using command-line option `-e` or config option `issues.exclude`.
 It's helpful when you decided to ignore all issues of this type.
-Also, you can use `issues.exclude-rules` config option for per-path or per-linter configuration.
+Also, you can use `issues.exclude-rules` config option for per-path, per-linter, per-text or per-source configuration.
```
ldez, line 34:
```suggestion
Also, you can use `issues.exclude-rules` config option for per-path or per-linter configuration.
```
text and source are extra filters, they cannot be used standalone.

ldez (changes requested):

docs/src/docs/usage/false-positives.mdx
hendrywiranto: oh okay, applied, thank you!

ldez (approved):
LGTM

Issue #3408, opened 2022-12-09
Errors during golangci operation


ids79:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc.).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

I am using the last version golangco 1.50.1 on local maschine (windows). 
When I try to perform linter, I get following errors: 
```
level=warning msg="[runner] Can't run linter goanalysis_metalinter: gci: 4:12: expected ';', found \"testing\" (and 3 more errors)"
level=error msg="Running error: 1 error occurred:\n\t* can't run linter goanalysis_metalinter: gci: 4:12: expected ';', found \"testing\" (and 3 more errors)\n\n"
```

What can they be related to?
How can I fix them?

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
# Paste output here
```
1.50.1
</details>

### Configuration file

<details>

```console
$ cat .golangci.yml
# paste output here
```
```yml
run:
  tests: true
  build-tags:
    - bench
    - !bench

linters-settings:
  funlen:
    lines: 150
    statements: 80

issues:
  exclude-rules:
    - path: _test\.go
      linters:
        - errcheck
        - dupl
        - gocyclo
        - gosec

linters:
  disable-all: true
  enable:
    - asciicheck
    - bodyclose
    - depguard
    - dogsled
    - dupl
    - durationcheck
    - errorlint
    - exhaustive
    - exportloopref
    - funlen
    - gci
    - gocognit
    - goconst
    - gocritic
    - gocyclo
    - godot
    - gofmt
    - gofumpt
    - goheader
    - goprintffuncname
    - gosec
    - gosimple
    - govet
    - importas
    - ineffassign
    - lll
    - makezero
    - misspell
    - nestif
    - nilerr
    - noctx
    - nolintlint
    - prealloc
    - predeclared
    - revive
    - rowserrcheck
    - sqlclosecheck
    - staticcheck
    - stylecheck
    - tagliatelle
    - thelper
    - typecheck
    - unconvert
    - unparam
    - unused
    - wastedassign
    - whitespace
```

</details>

### Go environment

<details>

```console
$ go version && go env
# paste output here
```
windows
</details>

[...]

PR #4113, opened 2023-10-04
fix: search for diff.exe on Windows

trollixx:
On Windows searching for `diff` returns PowerShell cmdlet, rather than diff binary even when it is installed.

Related to #3408.

### DIFF
--- a/pkg/result/processors/diff.go
+++ b/pkg/result/processors/diff.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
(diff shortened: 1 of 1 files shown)


ldez:
This change is not in the official goimports https://github.com/golang/tools/blob/e8722c010395d3c3528b2ea232476a2d15ac4af3/cmd/goimports/goimports.go#L326-L329

ldez:
Hello,

I prefer to follow what the Go team does, and they don't allow `diff.exe`.

We will decline this PR.

PR #4127, opened 2023-10-12
govet: add `appends` analyzer

Antonboom:


ldez (approved):

ldez:
I understand why you added default values for `enable` and `disable` but it's not their default values: if you `disable-all` default passes are not the same as if you `enable-all`.

The default passes are unrelated to `enable` and `disable`.

Antonboom:
Any other idea how to highlight default values?

I was confused with empty list.
And for me it is more unobvious than different sets at `disable` and `disable-all`.

We can write
```yml
    # Disable all analyzers (not just those that are disabled by default).
    # Default: false
    disable-all: true
```

PR #4128, opened 2023-10-14
protogetter: fix false positive

ghostiam:
🫣
https://github.com/ghostiam/protogetter/compare/v0.2.2...v0.2.3


ldez:
The update is handled by the dependabot you don't need to create a PR.

Issue #4134, opened 2023-10-17
doc: check default lists configuration reference


Antonboom:
### Your feature request related to a problem? Please describe.

```
$ grep ': \[\]' .golangci.reference.yml
  # Default: [].
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
      # Default: []
      # Default: []
      # Default: []
      # Default: []
    # Default: [] - means include all rules
    # Default: []
    # Default: []
        # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
    # Default: []
  # Default: []
  # Default: []
```

### Describe the solution you'd like.

1. Review configuration.
2. If `default` is really empty – no changes.
3. If `default` is not empty – add value and context in the comment.
4. Change rule about this field in the new-linter-checklist

### Describe alternatives you've considered.

Let it be as it is.

### Additional context.

https://github.com/golangci/golangci-lint/pull/4127#issuecomment-1760904555

ldez:
The majority of those empty defaults are real.

PR #4146, opened 2023-10-19
docs: describe enabled and disabled by default analyzers of govet

Antonboom:
Closes https://github.com/golangci/golangci-lint/issues/4134

@ldez, hi!

I reviewed all linters and yes, you are right:

> The majority of those empty defaults are real.

If you don't see the point in this pr, you can close it, I won't mind.

Thanks


ldez (approved):
LGTM

PR #4149, opened 2023-10-22
output: convert backslashes to forward slashes for GitHub Action annotations printer

surik:
This commit addresses an issue in the GitHub Actions printer where file paths were using backslashes on Windows systems.
As result we didn't see correct annotations in the PRs. To ensure proper formatting we always convert to forward slash separator in filepathes.

### DIFF
--- a/pkg/printers/github.go
+++ b/pkg/printers/github.go
@@ -3,6 +3,8 @@ package printers
 import (
 	"fmt"
 	"io"
+	"path/filepath"
+	"strings"
 
 	"github.com/golangci/golangci-lint/pkg/result"
 )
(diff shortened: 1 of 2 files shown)


surik:
Example of the issue I am trying to solve here: 

There is [an error annotation](https://github.com/surik/test-github-annotations/actions/runs/6604602768) which is not shown in [the PR changes](https://github.com/surik/test-github-annotations/pull/1/files) when running on Windows. With the forward slashes on Linux [it works fine](https://github.com/surik/test-github-annotations/pull/2/files). 

I'm assuming the GitHub always wants it with forward slashes.

ldez (approved):

PR #4190, opened 2023-11-09
Note that `exhaustruct` struct regular expressions are expected to match the entire `package/name/structname`

magnetikonline:
Quick one that tripped me up recently.

With the `exhaustruct` linter - the list of structs to `include/exclude` _must_ match the full `package/name/structname` - otherwise no match will be made/considered.

E.g. if I wish to target a struct of:

```
github.com/this/is/my/package/CoolStruct
```

These will work:

```
github\.com/this/is/my/package/CoolStruct
github\.com/this/is/my/package/Cool.+
```

This won't work:

```
github\.com/this/is/my/package/Cool
^github\.com/this/is/my/package/Cool
```

Hopefully this comment helps in that usage!

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -356,13 +356,15 @@ linters-settings:
       - 'example.com/package.ExampleStruct'
 
   exhaustruct:
-    # List of regular expressions to match struct packages and names.
+    # List of regular expressions to match struct packages and their names.
+    # Regular expressions must match complete canonical struct package/name/structname.
     # If this list is empty, all structs are tested.
     # Default: []
     include:
       - '.*\.Test'
       - 'example\.com/package\.ExampleStruct[\d]{1,2}'
-    # List of regular expressions to exclude struct packages and names from check.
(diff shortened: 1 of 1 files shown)


.golangci.reference.yml
```diff
@@ -356,13 +356,15 @@ linters-settings:
...
     # Default: []
     include:
       - '.*\.Test'
       - 'example\.com/package\.ExampleStruct[\d]{1,2}'
-    # List of regular expressions to exclude struct packages and names from check.
+    # List of regular expressions to exclude struct packages and their names from checks.
+    # Regular expressions must match complete canonical struct package/name/structname.
     # Default: []
     exclude:
       - 'cobra\.Command$'
```
magnetikonline, line 370: Actually - now I look at it - I don't think `'cobra\.Command$'` could ever be a valid match either?!

magnetikonline: Just added 7df401d2b58e11518570770b7173204c7c84221f to fix that fact.

ldez (approved):

PR #4271, opened 2023-12-19
docs: improve .golangci.reference.yml defaults

mateusoliveira43:
Add some defaults to `.golangci.reference.yml` file.

Add color to all flag descriptions.


pkg/commands/root.go
```diff
@@ -119,8 +119,9 @@ func formatMemory(memBytes uint64) string {
 
 func getDefaultConcurrency() int {
 	if os.Getenv(envHelpRun) == "1" {
-		// Make default concurrency stable: don't depend on machine CPU number, for docs/src/docs/usage/configuration.mdx help generating builds.
-		return 0
+		// Make stable concurrency for README help generating builds.
```
mateusoliveira43: I think this change is misleading. Earlier, the script would populate README file, but now it only populates files in `docs/` folder

And changing returning to non zero (non nil value) will print default (8). User may think the default is 8 when seeing the the website docs

pkg/commands/root.go
```diff
@@ -170,5 +171,4 @@ func initRootFlagSet(fs *pflag.FlagSet, cfg *config.Config, needVersionOption bo
 	}
 
 	fs.StringVar(&cfg.Output.Color, "color", "auto", wh("Use color when printing; can be 'always', 'auto', or 'never'"))
-	fs.BoolP("help", "h", false, wh("Help for this command"))
```
mateusoliveira43: Without this, help commands does not get colored
Screenshot from 2023-12-19 17-06-04

ldez: Yes It's expected.

ldez (approved):
2 topics = 2 PRs
Please don't create PR with multiple topics.
I will merge the PR in this state (exceptionally) but please avoid that.

PR #4277, opened 2023-12-20
docs: fix Go version fallback 

mateusoliveira43:
Fix documentation to follow what code does. Reference https://github.com/golangci/golangci-lint/blob/master/pkg/config/config.go#L70


ldez (changes requested):
the problem is more related to the doc than the code.

ldez (approved):

PR #4303, opened 2024-01-07
Improve `typecheck` docs

bombsimon:
Add a step in the troubleshooting guide to ensure users are not linting individual files. Even though this _might_ work, it's often the root cause of `typecheck` reports when the package consists of multiple files.

---

This seems to be a common mistake and mentioned in https://github.com/golangci/golangci-lint/discussions/3759. This step ensures users targeting single files tries to lint the whole package first (and not necessarily `./...`).

### DIFF
--- a/docs/src/docs/usage/faq.mdx
+++ b/docs/src/docs/usage/faq.mdx
@@ -49,5 +49,6 @@ How to troubleshoot:
 
 - [ ] Ensure the version of `golangci-lint` is built with a compatible version of Go.
 - [ ] Ensure dependencies are up-to-date with `go mod tidy`.
+- [ ] Ensure you're linting packages and not single files. I.e. `golangci-lint run ./path/to/pkg` (or `./...`), **not** `./path/to/pkg/file.go`.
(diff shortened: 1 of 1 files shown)


ldez (changes requested):
The sentence is ambiguous because:
- A single file or package can be linted if there are no dependencies with other files/packages
- package has the same "problem" as a single file, if there are dependencies with other files/packages it will fail.

Your suggestion is the same topic as `Ensure building works with go run ./.../go build ./... - whole package.`

I think I will create PR to explain with more details what is `typecheck`, it will not be fun but I think it's the best approach we can take.

bombsimon:
Yes, the sentence wasn't meant to be an absolute truth for every time `typecheck` is seen, it was just one thing to try in the troubleshooting process. Same as `go mod tidy` won't _always_ solve the issue but it _might_ solve the issue and worth having the user test.

I think it would be nice to extend the list to ensure that `golangci-lint` is not ran on a single file since that's pretty common mistake (either with the CLI tool or with editor settings). However you're explanation in #4306 is better so I'll close this.

Issue #4197, opened 2023-11-13
Sort by linter and not just line number


x528491x:
### Your feature request related to a problem? Please describe.

When using the CLI with --sort-results, it only sorts by line number. I usually run --enable-all and to catch all lint errs and I want to be able to sort by linter in an alphabetical order. Even using the tabbed output format is not straightforward because the spaces between the columns are arbitrary and not an actual single tab.

### Describe the solution you'd like.

An inbuilt `--sort-results-by-linter` would be great.

### Describe alternatives you've considered.

`golangci-lint run --enable-all --out-format=tab | sed "s/ \+ /\t/g" | sort -t"	" -k2`

PR #4325, opened 2024-01-18
feat: Add output option to group results by linter name

marcciosilva:
Hi! Newbie to the project here, took a stab at trying to implement the enhancement from #4197.
Hope it makes sense to you to include the grouping code in the sorting processor (made sense to me since I wanted to preserve the eventual sorting within the linter 'group' - it could be done in a separate processor adding extra processing but keeping concerns apart, so I'm obviously open to suggestions!).

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -117,6 +117,10 @@ output:
   # Default: false
   sort-results: true
 
+  # Group results by linter name.
+  # Default: false
+  group-results-by-linter: false
+
 
 # All available settings of specific linters.
 linters-settings:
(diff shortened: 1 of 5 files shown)


ldez:
Hello,

I think an option only to "group" by linter name is too specific.

Otherwise, I don't understand the need for this kind of sorting, what is the usage of this?

marcciosilva:
> Hello,
> 
> I think an option only to "group" by linter name is too specific.
> 
> Otherwise, I don't understand the need for this kind of sorting, what is the usage of this?

I was aiming to address the #4197 issue - I'm guessing this could be useful if e.g. you're solving misspellings, you might as well just fix all misspellings at once. Might be inefficient in terms of file traversing but it's the same type of work.

Same thing for other stuff like refactoring code to reduce cyclomatic complexity in functions. Might be annoying to 'context switch' between different types of work while fixing issues? That's what comes to mind.

But yeah it'd be perfectly reasonable to just close this (along with the issue) since it's too specific.

ldez:
As we don't understand the usage and the option proposal is not enough "generic", we will close this PR, but the issue will stay open.

In any case, thank you for your implementation proposal.

PR #4326, opened 2024-01-19
docs: adjust wrapcheck ignoreSigs to new defaults

sirkrypt0:
This adds `errors.Join(` to the signatures ignored by default, which has been added to the wrapcheck linter in v2.8.0 a while ago. See https://github.com/tomarrell/wrapcheck/releases/tag/v2.8.0


ldez (approved):
LGTM

PR #4333, opened 2024-01-24
build(deps): bump mvdan.cc/unparam to HEAD

alexandear:
https://github.com/mvdan/unparam/compare/7455f1af531d...c549a3470d14

**This update is done by hand because dependabot is not able to update non-semver dependency.**


ldez:
It was intentional to not update this linter because there is no change.

ldez (approved):

Issue #4108, opened 2023-09-26
go-critic: support to enable/disable all checks


Dentrax:
### Your feature request related to a problem? Please describe.

We have to manually define all the checks in the config: https://go-critic.com/overview.html It's a bit annoying to manage all of the _checks_ individually.

Ref: https://golangci-lint.run/usage/linters/#gocritic

### Describe the solution you'd like.

Add support for `enable-all: true` and `disable-all: true` for `gocritic`. So that I can drop unwanted checks in the `disabled-checks` section.

### Describe alternatives you've considered.

As-is. Manually define and manage all the checks.

nemre:
+1

ferhatelmas:
As a workaround, you can add [all tags](https://github.com/go-critic/go-critic/blob/master/linter/linter.go#L14) (which are only 6 of them but `security` not working yet) and then disable what you don't want.

GreyXor:
Doing it like so:
```yaml
linters-settings:
  gocritic:
    enabled-tags:
      - diagnostic
      - style
      - performance
      - experimental
      - opinionated
```

PR #4335, opened 2024-01-25
gocritic: support of enable-all and disable-all options

Antonboom:
Closes https://github.com/golangci/golangci-lint/issues/4108


test/testdata/configs/gocritic.yml
```diff
@@ -2,9 +2,9 @@ linters-settings:
   gocritic:
     enabled-checks:
       - rangeValCopy
-      - flagDeref
```
Antonboom, line 5: Else we received warning in stderr about enabling already enabled checks

test/testdata/configs/gocritic.yml
```diff
@@ -2,9 +2,9 @@ linters-settings:
...
-      - wrapperFunc
       - ruleguard
+    disabled-checks:
+      - dupSubExpr # So as not to overlap `ruleguard`.
```
Antonboom, line 7: Couldnot understand this is regress or not 🤔

pkg/golinters/gocritic.go
```diff
@@ -254,377 +261,344 @@ func (w *goCriticWrapper) normalizeCheckerParamsValue(p any) any {
...
 
 	logger logutils.Log
 
-	allCheckers   []*gocriticlinter.CheckerInfo
-	allCheckerMap map[string]*gocriticlinter.CheckerInfo
+	allCheckers []*gocriticlinter.CheckerInfo
 
-	inferredEnabledChecks map[string]bool
+	allChecks           map[string]struct{}
+	allChecksLowerCased map[string]struct{}
```
Antonboom, line 278: `*LowerCased` used for `SettingsPerCheck` validation only

ldez: this should be added in a comment inside the code.

test/testdata/configs/gocritic.yml
ldez: I think it's related to the order of the rules (based on name)
I think it's better to change the test instead of the configuration.

test/testdata/configs/gocritic.yml
ldez: feel like a problem because this modifies the configuration and the expectation.

pkg/golinters/gocritic.go
```diff
@@ -31,13 +32,13 @@ var (
 	isGoCriticDebug = logutils.HaveDebugTag(logutils.DebugKeyGoCritic)
 )
 
-func NewGoCritic(settings *config.GoCriticSettings, cfg *config.Config) *goanalysis.Linter {
+func NewGoCritic(settings *config.GoCriticSettings, lintConfigDirGetter func() string) *goanalysis.Linter {
```
ldez, line 35: not sure about this change: a struct seems better than a closure.

pkg/golinters/gocritic.go
```diff
@@ -200,13 +205,17 @@ func runGocriticOnFile(linterCtx *gocriticlinter.Context, f *ast.File, checks []
...
+	info *gocriticlinter.CheckerInfo,
+	allLowerCasedParams map[string]config.GoCriticCheckSettings,
+) error {
+	params := allLowerCasedParams[strings.ToLower(info.Name)]
 	if params == nil { // no config for this checker
 		return nil
 	}
 
-	infoParams := normalizeCheckerInfoParams(info)
+	// NOTE(ldez): lowercase info param keys here because golangci-lint's config parser lowercases all strings.
```
ldez, line 217:
```suggestion
	// lowercase info param keys here because golangci-lint's config parser lowercases all strings.
```

pkg/golinters/gocritic.go
```diff
@@ -254,377 +261,344 @@ func (w *goCriticWrapper) normalizeCheckerParamsValue(p any) any {
...
-		tagToCheckers := s.buildTagToCheckersMap()
-		for _, tag := range s.EnabledTags {
-			enabledChecks = append(enabledChecks, tagToCheckers[tag]...)
+	if s.EnableAll {
+		enabledChecks = make(map[string]struct{}, len(s.allCheckers))
+		for _, info := range s.allCheckers {
+			enabledChecks[info.Name] = struct{}{}
+		}
+	} else if !s.DisableAll {
+		// NOTE(Antonboom): enable-all/disable-all revokes the default settings.
```
ldez, line 337:
```suggestion
		// enable-all/disable-all revokes the default settings.
```

pkg/golinters/gocritic.go
```diff
@@ -92,15 +94,15 @@ func (w *goCriticWrapper) init(settings *config.GoCriticSettings, logger logutil
...
 		}
 	})
 
 	settingsWrapper := newGoCriticSettingsWrapper(settings, logger)
-
-	settingsWrapper.inferEnabledChecks()
-
-	if err := settingsWrapper.validate(); err != nil {
+	settingsWrapper.InferEnabledChecks()
+	// NOTE(Antonboom): Validate must be after InferEnabledChecks, not before.
```
ldez, line 103:
```suggestion
	// Validate must be after InferEnabledChecks, not before.
```

pkg/golinters/gocritic.go
```diff
@@ -156,8 +158,11 @@ func (w *goCriticWrapper) buildEnabledCheckers(linterCtx *gocriticlinter.Context
 	return enabledCheckers, nil
 }
 
-func runGocriticOnPackage(linterCtx *gocriticlinter.Context, checks []*gocriticlinter.Checker,
-	files []*ast.File) []result.Issue {
+func runGocriticOnPackage(
+	linterCtx *gocriticlinter.Context,
+	checks []*gocriticlinter.Checker,
+	files []*ast.File,
+) []result.Issue {
```
ldez, line 165:
```suggestion
func runGocriticOnPackage(linterCtx *gocriticlinter.Context, checks []*gocriticlinter.Checker, files []*ast.File) []result.Issue {
```

ldez (changes requested):
There are a lot of changes not exactly related to the feature.
I can understand why those changes but it's complex to review the PR.

pkg/golinters/gocritic.go
Antonboom: history of change:

1. confused by ambiguous API – what fields of config are needed by GoCritic oO?
2. changed to passing `lintConfigDir` only
3. realized that `Config.cfgDir` filled after `Manager.GetAllSupportedLinterConfigs` call
4. changed to factory

do you think it's better to return it back?

pkg/golinters/gocritic.go
Antonboom: Why you don't like noted comments? :)
helps to find the author in case of misunderstanding..

pkg/golinters/gocritic.go
Antonboom: 131 symbol per line

UPD: oh, ok, we have the 140 limit oO

pkg/golinters/gocritic.go
Antonboom: done

pkg/golinters/gocritic.go
Antonboom: done

pkg/golinters/gocritic.go
Antonboom: done

test/testdata/configs/gocritic.yml
Antonboom: under "problem" do you mean such kind of warnings?

I repeated the [existed logic](https://github.com/golangci/golangci-lint/pull/4335/files#diff-05b9162497599694184af35eedcc22cffa1179be8eff6935a6d082753349ba17L380):

```go
s.logger.Warnf("%s: no need to enable check %q: it's already enabled", goCriticName, enabledCheck)
```

but did not investigate why this is not working before 🤷‍♂️

test/testdata/configs/gocritic.yml
Antonboom: I refactored tests

- no intersection between default checkers and custom rules
- no intersection between gocritic-fixes tests (I was confused by `rangeValCopy` and `rangeExprVal`)
- little more border cases

Antonboom (commented):
@ldez thank u for review
see answers in comments above

> but it's complex to review the PR.

Yes, it's true :( 

I wasted a lot of time to review & test it too, but to be honest the actual state is easier and simpler to review – I tried to simplify code and make it more robust and strict

I was hoping that this feature would look good in the upcoming release.

What your proposal? Next steps?

I see it as:
1) Request review from @alexandear, @SVilgelm, @bombsimon 
2) Test (independently from me) this feature on real project with different config combinations.
3) Merge it and to be ready to ASAP patches if some issues found.

ldez:
> What your proposal? Next steps?

My comment was just to say that the review will not be quick.

pkg/golinters/gocritic.go
ldez: Notes are just comments, we don't need to know who writes a comment.
This is different for FIXME and TODO.

pkg/golinters/gocritic.go
```diff
@@ -254,377 +257,346 @@ func (w *goCriticWrapper) normalizeCheckerParamsValue(p any) any {
...
+		!info.HasTag(gocriticlinter.PerformanceTag) &&
+		!info.HasTag(gocriticlinter.SecurityTag)
+}
 
-	if err := validateStringsUniq(s.EnabledChecks); err != nil {
-		return fmt.Errorf("validate enabled checks: %w", err)
+func (s *goCriticSettingsWrapper) expandTagsToChecks(tags []string) []string {
+	var checks []string
+	for _, t := range tags {
+		checks = append(checks, s.allChecksByTag[t]...)
```
ldez, line 411:
```suggestion
	for _, tag := range tags {
		checks = append(checks, s.allChecksByTag[tag]...)
```
better for the readability

pkg/golinters/gocritic.go
```diff
@@ -254,377 +257,346 @@ func (w *goCriticWrapper) normalizeCheckerParamsValue(p any) any {
...
-	return enabled
+	goCriticDebugf("%s checks (%d): %s", fmt.Sprintf(format, args...), len(checks), sprintStrings(checks))
 }
 
-func (s *goCriticSettingsWrapper) getDefaultDisabledCheckersNames() []string {
-	var disabled []string
+func sprintStrings(ss []string) string {
+	sort.Strings(ss)
+	return fmt.Sprint(ss)
+}
```
ldez, line 464:
```suggestion
func sprintStrings(v []string) string {
	sort.Strings(v)
	return fmt.Sprint(v)
}
```

ldez (changes requested):
Instead of creating methods like `isLowerCasedCheckEnabled`, `IsCheckEnabled`, `isKnownLowerCasedCheck`, `isKnownCheck`, etc. I propose use a type:

```go
type goCriticMap[T any] map[string]T

func (f goCriticMap[T]) has(name string) bool {
	_, ok := f[name]
	return ok
}

type goCriticSettingsWrapper struct {
	*config.GoCriticSettings

	logger logutils.Log

	allCheckers []*gocriticlinter.CheckerInfo

	allChecks             goCriticMap[struct{}]
	allChecksByTag        goCriticMap[[]string]
	allTagsSorted         []string
	inferredEnabledChecks goCriticMap[struct{}]

	// *LowerCased* fields are used for GoCriticSettings.SettingsPerCheck validation only.

	allChecksLowerCased             goCriticMap[struct{}]
	inferredEnabledLowerCasedChecks goCriticMap[struct{}]
}

// ...

func (s *goCriticSettingsWrapper) validateCheckerTags() error {
	for _, tag := range s.EnabledTags {
		if !s.allChecksByTag.has(tag) {
			return fmt.Errorf("enabled tag %q doesn't exist, see %s's documentation", tag, goCriticName)
		}
	}

	for _, tag := range s.DisabledTags {
		if !s.allChecksByTag.has(tag) {
			return fmt.Errorf("disabled tag %q doesn't exist, see %s's documentation", tag, goCriticName)
		}
	}

	return nil
}

```

pkg/golinters/gocritic.go
ldez: yes, I prefer the usage of `cfg` because it's more homogenous with the other linters when they need to use the Config.

Antonboom: I implemented an option between - keep `cfg` in constructor signature, but narrow usage of its inside
(more obvious for newcomers)

Antonboom:
@ldez , hi

thank you for review

>  I propose use a type

cool idea! simplified.

the type is candidate to be a more common type, but I left `goCritic`-prefix
- so as not to generalize ahead of time
- to isolate linters from each other

P.S. God bless the new tests :D

ldez (approved):
LGTM

PR #4272, opened 2023-12-19
go1.22 support

ldez:
This PR is to evaluate and prepare golangci-lint to [go1.22](https://tip.golang.org/doc/go1.22).

This PR will evolve during the beta and RC phases of [go1.22](https://tip.golang.org/doc/go1.22).

**Questions and problems related to go1.22 should be reported in the issue #4273.**

---

_Sponsoring is a good way to sustain open source maintainers: [sponsor me](https://github.com/sponsors/ldez)_

PR #4356, opened 2024-02-08
feat: support Golang v1.22

TomerJLevy:


bombsimon:
We released [1.56.0](https://github.com/golangci/golangci-lint/releases/tag/v1.56.0) yesterday that supports Go 1.22. It was added in https://github.com/golangci/golangci-lint/pull/4272

PR #4365, opened 2024-02-09
revive: add exclude option

### LINKED ISSUE #4364: revive: forward Exclude rule config

Right now I'm using issues.exclude-rules to achieve what I want to do, but it would be nice if there was a way to just provide the exclusion to revive directly via [the Exclude config option](https://github.com/mgechev/revive/blob/5c5d6c1075a82337d9840e7f8195c0eaccdb959c/lint/config.go#L15)

```yaml
linters-settings:
  revive:
    rules:
      - name: unused-parameter
        exclude: ["TEST"]
        # TEST is a magic word in revive to filter out test files,
        # no special handling needed in golangci-lint
```

issues.exclude-rules

### Additional context.

_No response_

ldez:
Closes #4364

<details>

To test the configuration locally, you can use this script:

```bash
#!/bin/sh -e

## setup module
rm -rf reviveexclude
mkdir reviveexclude && cd $_
go mod init helloworld

## main.go
cat > main.go  <<EOF
package main

// Foo is foo.
func Foo(a string, b string) string {
	return a
}
EOF

## main_test.go
cat > main_test.go  <<EOF
package main

// Bar is bar.
func Bar(a string, b string) string {
	return a
}
EOF

## .golangci.yml
cat > .golangci.yml  <<EOF
linters:
  disable-all: true
  enable:
    - revive

linters-settings:
  revive:
    enable-all-rules: true
    rules:
      - name: package-comments
        disabled: true
      - name: unused-parameter
#        exclude: ["~main_test.go"]
        exclude: ["TEST"]

issues:
  exclude-use-default: false
  max-issues-per-linter: 0
  max-same-issues: 0
EOF
```

</details>

### DIFF
--- a/.golangci.reference.yml
+++ b/.golangci.reference.yml
@@ -1534,6 +1534,7 @@ linters-settings:
       - name: add-constant
         severity: warning
         disabled: false
+        exclude: [""]
         arguments:
           - maxLitCount: "3"
             allowStrs: '""'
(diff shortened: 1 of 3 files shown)


.golangci.reference.yml
```diff
@@ -1543,145 +1544,176 @@ linters-settings:
...
       # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#atomic
       - name: atomic
         severity: warning
+        exclude: [""]
         disabled: false
       # https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#banned-characters
       - name: banned-characters
         severity: warning
         disabled: false
+        exclude: [""]
```
Antonboom, line 1558: inconsistent with `arguments` format, e.g. 

```
[ "Ω","Σ","σ", "7" ]
[ 4 ]
```

`[ "" ]` expected

pkg/golinters/revive.go
```diff
@@ -182,6 +183,14 @@ func getReviveConfig(cfg *config.ReviveSettings) (*lint.Config, error) {
 
 	normalizeConfig(conf)
 
+	for k, r := range conf.Rules {
+		err := r.Initialize()
+		if err != nil {
+			return nil, fmt.Errorf("error in config of rule [%s] : [%w]", k, err)
```
Antonboom, line 189:
```suggestion
			return nil, fmt.Errorf("error in config of rule %q: %w", k, err)
```

Antonboom (approved):

.golangci.reference.yml
ldez: yes because there is no example inside the revive doc, and it's complex to suggest right values for this option.

alexandear (approved):

PR #4366, opened 2024-02-09
docs: improve `govet` description

silverwind:
As per https://github.com/golangci/golangci-lint/discussions/4363#discussioncomment-8413385, I think it's good to clarify this. Aim is to get this on the [docs page](https://golangci-lint.run/usage/linters/#govet). I did not add a `.` at the end of the sentence as it seems the docs build does that.

### DIFF
--- a/pkg/golinters/govet.go
+++ b/pkg/golinters/govet.go
@@ -145,7 +145,8 @@ func NewGovet(settings *config.GovetSettings) *goanalysis.Linter {
 	return goanalysis.NewLinter(
 		"govet",
 		"Vet examines Go source code and reports suspicious constructs, "+
-			"such as Printf calls whose arguments do not align with the format string",
+			"such as Printf calls whose arguments do not align with the format string. "+
+			"This is the same analysis as the 'go vet' tool",
 		analyzersFromConfig(settings),
 		conf,
 	).WithLoadMode(goanalysis.LoadModeTypesInfo)
--- a/pkg/golinters/govet.go
+++ b/pkg/golinters/govet.go
@@ -144,8 +144,7 @@ func NewGovet(settings *config.GovetSettings) *goanalysis.Linter {
 
 	return goanalysis.NewLinter(
 		"govet",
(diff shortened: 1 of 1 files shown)


pkg/golinters/govet.go
```diff
@@ -143,7 +143,8 @@ func NewGovet(settings *config.GovetSettings) *goanalysis.Linter {
 	return goanalysis.NewLinter(
 		"govet",
 		"Vet examines Go source code and reports suspicious constructs, "+
-			"such as Printf calls whose arguments do not align with the format string",
+			"such as Printf calls whose arguments do not align with the format string. "+
+			"This is the same analysis as the 'go vet' tool",
```
ldez, line 147:
```suggestion
			"It uses the same passes as 'go vet'.",
```

The sentence inside my [comment](https://github.com/golangci/golangci-lint/discussions/4363#discussioncomment-8413385) was a popularization for you but it's not something that we should write in a doc because it's not the reality.

Golancgi-lint is based on the internal tooling of `go vet`, and it also uses the passes.
"Analysis" is a specific term inside the technical context of golangci-lint and linters, which is not adapted to this situation.

ldez (changes requested):

silverwind:
As per https://github.com/golangci/golangci-lint/discussions/4363#discussioncomment-8415288, I think I will change it to this, highlighting the fact that they are not exactly the same (for example when `go vet` adds new passes but golangci has not updated yet:

`It is roughly the same as 'go vet' and uses its passes.`

pkg/golinters/govet.go
silverwind, line 147:
```suggestion
			"It is roughly the same as 'go vet' and uses its passes.",
```

silverwind: Thanks, I used a variant of your suggestion below.

ldez (approved):
LGTM

Antonboom (approved):

Issue #4368, opened 2024-02-09
Version missing from golangci-lint in Docker image since v1.56.0


ashmckenzie:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've read the typecheck section of the FAQ (https://golangci-lint.run/usage/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

Since v1.56.0, the version component when running `golangci-lint` in the official Docker images is missing.

I believe https://github.com/golangci/golangci-lint/pull/4279 is connected with the missing version.

### Version of golangci-lint

<details>
N/A
</details>

### Configuration

<details>
N/A
</details>

### Go environment

<details>
N/A
</details>

### Verbose output of running

<details>
N/A
</details>

### A minimal reproducible example or link to a public repository

```sh
$ docker run --rm -ti golangci/golangci-lint:v1.55.2-alpine golangci-lint version
golangci-lint has version v1.55.2 built with go1.21.3 from e3c2265f on 2023-11-03T13:27:48Z
```

vs.

```sh
$ docker run --rm -ti golangci/golangci-lint:v1.56.0-alpine golangci-lint version
golangci-lint has version  built with go1.22.0 from 1a00b409 on 2024-02-07T14:52:31Z
```

### Validation

- [X] Yes, I've included all information above (version, config, etc.).

PR #4369, opened 2024-02-09
dev: fix variables incorrectly passed to 'go build'

ashmckenzie:
This PR aims to solve https://github.com/golangci/golangci-lint/issues/4368 by ensuring the `APP_VERSION` and `CGO_ENABLED` variables are correctly set prior to being used by `go build`:

**Before** (from d7513984d5b44c7cd49d98931a5fcb1c1a4231b8)

```sh
$ docker build --quiet -f build/Dockerfile --build-arg VERSION=9.9.9 --build-arg SHORT_COMMIT=abc123 --build-arg DATE=now . -t test/golang-ci-lint && docker run --rm -ti test/golang-ci-lint golangci-lint version
sha256:6749ff93236299ceac1d04987f3384415fa7f081cfc9ea8695412a2a86f30690
golangci-lint has version  built with go1.22.0 from abc123 on now
```

**After** (from `ashmckenzie/fix-missing-version-in-docker-build`)

```sh
$ docker build --quiet -f build/Dockerfile --build-arg VERSION=9.9.9 --build-arg SHORT_COMMIT=abc123 --build-arg DATE=now . -t test/golang-ci-lint && docker run --rm -ti test/golang-ci-lint golangci-lint version
sha256:2f2607b9836e56d75b7aca69363aa5329619bcaf8a46d2448957d450645550ce
golangci-lint has version 9.9.9 built with go1.22.0 from abc123 on now
```

Closes: https://github.com/golangci/golangci-lint/issues/4368

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -7,8 +7,8 @@ ARG DATE
 
 COPY / /golangci
 WORKDIR /golangci
-RUN APP_VERSION=${VERSION#v} \
-    CGO_ENABLED=0 \
+RUN APP_VERSION=${VERSION#v}; \
+    CGO_ENABLED=0; \
     go build -trimpath -ldflags "-s -w -X main.version=$APP_VERSION -X main.commit=$SHORT_COMMIT -X main.date=$DATE" -o golangci-lint ./cmd/golangci-lint/main.go
 
 # stage 2


ldez:
Hello @ashmckenzie, thank you for the issue and the PR.

I think will solve this problem by stopping the old approach with a "manual" build, and just integrating the build of the images inside the gorelease process.
See #4370.

ashmckenzie:
Hi @ldez, thanks for the quick response! 

https://github.com/golangci/golangci-lint/pull/4370 looks like a more scalable approach for the future which is great but the diff is quite large in comparison to the two line fix here 🙂  

Would you consider merging this PR as a bugfix and https://github.com/golangci/golangci-lint/pull/4370 as a feature improvement ?

ldez:
There is a large diff but it's really easy to review and test it.

I don't close your PR for now, I will see with the other maintainers if they have the time to review my PR.

In all cases, your PR or my PR will be a part of the next bug-fix release.

build/Dockerfile
```diff
@@ -7,8 +7,8 @@ ARG DATE
 
 COPY / /golangci
 WORKDIR /golangci
-RUN APP_VERSION=${VERSION#v} \
-    CGO_ENABLED=0 \
+RUN APP_VERSION=${VERSION#v}; \
```
Antonboom, line 10: @ashmckenzie just for interest – how it works? :D

Antonboom (approved):

build/Dockerfile
ashmckenzie: The previous logic never worked as intended as passing variables to commands like this just doesn't work, for example:

```sh
$ VAR="test" echo "VAR=[${VAR}]"
VAR=[]

$ VAR="test"; echo "VAR=[${VAR}]"
VAR=[test]
```

Antonboom: this is `sh` magic?
I remember the common pattern `FOO=bar cmd args` (or using `export`)

bombsimon: This is more how the variable is scoped to `echo` only and not `sh`. 

```sh
› VAR="test" echo "VAR=[${VAR}]"
VAR=[]

› VAR="test" sh -c 'echo "VAR=[${VAR}]"'
VAR=[test]
```
In the first example `$VAR` is only visible to `echo` (not the assignment) but in the second example it's scoeped to `sh` so it will be in the scope for everything `sh` calls.

So in this case `APP_VERSION` is visible to `go` but not to the command line flags. Here's an example where you can read it as an env but not use it as an arg:

```sh
› VAR=test perl -E 'say $ENV{"VAR"}; say @ARGV' some-arg $VAR
test
some-arg
```

If you use `export` you make it globally visible so that's not the same.

PR #4404, opened 2024-02-21
dev: clean up Executor 

ldez:
This is the first part of the rewrite of the command line (the `Executor`).
I split the changes into several PRs to ease the review.

This PR mainly moves, organizes, and splits the current code.
There are no major changes, but I spent a lot of time analyzing the code to be able to make the right changes to prepare for the new step.

I also removed all the deprecated flags.
They have been deprecated for multiple years (2018), so it is safe to remove them.

I made small commits to ease the review: each commit contains one change.
I recommend reviewing each commit first instead of the full change because it's easier to follow changes.

I already have the next PR, but this PR must be merged first.

related to https://github.com/golangci/golangci-lint/pull/4397#discussion_r1494525093

FYI with changes inside this PR, I can just move one line to fix the problem related to the configuration loading, but this will not fix the core of the problem: the maintainability of the code.
Also, this PR improves a bit the performance by removing one of the multiple builds of the linters (I will also fix that in another PR).

EDIT: [--update-refs](https://git-scm.com/docs/git-rebase/2.43.0#Documentation/git-rebase.txt---update-refs) is the best rebase option ever!

### DIFF
--- a/pkg/commands/executor.go
+++ b/pkg/commands/executor.go
@@ -1,14 +1,7 @@
 package commands
 
 import (
-	"bytes"
-	"context"
-	"crypto/sha256"
 	"errors"
-	"fmt"
-	"io"
-	"os"
-	"path/filepath"
 	"strings"
 	"time"
 
@@ -16,9 +9,7 @@ import (
 	"github.com/gofrs/flock"
 	"github.com/spf13/cobra"
 	"github.com/spf13/pflag"
-	"gopkg.in/yaml.v3"
 
-	"github.com/golangci/golangci-lint/internal/cache"
 	"github.com/golangci/golangci-lint/internal/pkgcache"
 	"github.com/golangci/golangci-lint/pkg/config"
 	"github.com/golangci/golangci-lint/pkg/fsutils"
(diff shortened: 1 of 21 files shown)


pkg/commands/run.go
```diff
@@ -358,11 +358,13 @@ func (e *Executor) releaseFileLock() {
...
 func initRunFlagSet(fs *pflag.FlagSet, cfg *config.Config) {
-	// Config file config
-	rc := &cfg.Run
-	initConfigFileFlagSet(fs, rc)
+	fs.BoolVar(&cfg.InternalCmdTest, "internal-cmd-test", false, wh("Option is used only for testing golangci-lint command, don't use it"))
+	if err := fs.MarkHidden("internal-cmd-test"); err != nil {
+		panic(err)
+	}
+
+	// --- Output config
```
bombsimon, line 366: Not a huge fan of these divider comments, should `initRunFlagSet` rather be split into multiple functions, one for each config section?

bombsimon (approved):
Exceptionally good PR, really easy to review with the separate commits with small changes and the fact that this PR only includes refactoring. 5/5, would review again.

Added some comments but approving since I don't see them as blocking.

pkg/commands/executor.go
```diff
@@ -97,19 +107,8 @@ func NewExecutor(buildInfo BuildInfo) *Executor {
...
-	e.initLinters()
-	e.initConfig()
-	e.initVersion()
-	e.initCache()
-
-	// init e.cfg by values from config: flags parse will see these values
-	// like the default ones. It will overwrite them only if the same option
-	// is found in command-line: it's ok, command-line has higher priority.
+	// init e.cfg by values from config: flags parse will see these values like the default ones.
+	// It will overwrite them only if the same option is found in command-line: it's ok, command-line has higher priority.
```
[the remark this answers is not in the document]
ldez: Yes, I follow https://sembr.org/ because for me the number of character limit is coming from the past, I think we already discussed that inside an issue or PR about `lll` :smile:

pkg/commands/run.go
ldez: I added the divider only to help me on the next step.
If possible I would prefer to keep the divider because it will impact my current branch.
In the next step, all the flag sets will be revamped.

pkg/lint/lintersdb/manager.go
```diff
@@ -992,6 +956,42 @@ func (m Manager) GetAllLinterConfigsForPreset(p string) []*linter.Config {
...
+		linter.PresetPerformance,
+		linter.PresetSQL,
+		linter.PresetStyle,
+		linter.PresetTest,
+		linter.PresetUnused,
+	}
+}
+
+func allPresetsSet() map[string]bool {
+	ret := map[string]bool{}
```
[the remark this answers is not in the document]
ldez: I think this method is mainly useless because it only uses by only one method:
https://github.com/golangci/golangci-lint/blob/2a2eebdbea18a191837c1603dbc3fda59361da0e/pkg/lint/lintersdb/validator.go#L41-L48

Inside my PR, the method is now a function.

I think I will just remove the function `allPresetsSet()` (map) and use `AllPresets()` (slice), we don't need extreme performance optimization to browse 10 presets :smile:.

pkg/commands/executor.go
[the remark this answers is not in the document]
ldez: I don't know if we should recommend that but I share the opinion of the article you provided.
For me, readability is the key, I often reformat a comment only when I try to read it and it feels painful to read.

pkg/commands/run.go
bombsimon: Go for it, keep it here for now!

pkg/commands/executor.go
```diff
@@ -163,103 +154,36 @@ func (e *Executor) Execute() error {
...
-		return
-	}
-
-	if err := e.flock.Unlock(); err != nil {
-		e.debugf("Failed to unlock on file: %s", err)
-	}
-	if err := os.Remove(e.flock.Path()); err != nil {
-		e.debugf("Failed to remove lock file: %s", err)
-	}
+func wh(text string) string {
```
Antonboom, line 228: question: what does it mean "wh"?

pkg/commands/root.go
```diff
@@ -89,6 +106,26 @@ func (e *Executor) persistentPostRun(_ *cobra.Command, _ []string) error {
...
+	fs.StringVar(&cfg.Output.Color, "color", "auto", wh("Use color when printing; can be 'always', 'auto', or 'never'"))
+
+	fs.StringVar(&cfg.Run.CPUProfilePath, "cpu-profile-path", "", wh("Path to CPU profile output file"))
+	fs.StringVar(&cfg.Run.MemProfilePath, "mem-profile-path", "", wh("Path to memory profile output file"))
+	fs.StringVar(&cfg.Run.TracePath, "trace-path", "", wh("Path to trace output file"))
+
+	fs.IntVarP(&cfg.Run.Concurrency, "concurrency", "j", getDefaultConcurrency(),
+		wh("Number of CPUs to use (Default: number of logical CPUs)"))
+
+	if needVersionOption {
```
Antonboom: question: why do not print version always? and if no build just print `(devel)` or something

pkg/lint/lintersdb/manager.go
```diff
@@ -992,6 +956,42 @@ func (m Manager) GetAllLinterConfigsForPreset(p string) []*linter.Config {
 	return ret
 }
 
+func linterConfigsToMap(lcs []*linter.Config) map[string]*linter.Config {
+	ret := map[string]*linter.Config{}
+	for _, lc := range lcs {
+		lc := lc // local copy
```
Antonboom, line 962: this is copy of pointer – no make sense?

pkg/commands/executor.go
```diff
@@ -54,7 +54,6 @@ func NewExecutor(buildInfo BuildInfo) *Executor {
 	e := &Executor{
 		cfg:       config.NewDefault(),
 		buildInfo: buildInfo,
-		dbManager: lintersdb.NewManager(nil, nil),
```
Antonboom: it must help to replace *cfg.Config in linters constructor with specific fields
(and remove my `lintConfigDirGetter`

Antonboom (commented):
Sorry for 🐌 
Fantastic job!

pkg/commands/root.go
Antonboom: UPD: condition will be removed

pkg/commands/executor.go
bombsimon: Given how it's used I read it as "with help" but I also suggested a better name in a different PR.

pkg/lint/lintersdb/manager.go
ldez: it was existing code, fixed in the next PR.

PR #4412, opened 2024-02-23
dev: new commands system

ldez:
## Part 2 of the rewrite and the death of the `Executor`.

First, this PR will be more complex to review than #4404 because of the complexity of the `Executor`.

I did my best to split changes into commits:
- a commit by one new command
- a commit by removing one old command

But as the previous code was very tangled, it's really complex to provide something easy to read.

### What were the problems?

It will be complex to explain all the problems with the previous design of the `Executor`, so I will try to make a summary.

I was not implying inside golangci-lint dev when the `Executor` has been created, but based on git history I can give a short history.
The `Executor` was created when golangci-lint was just a single run command and this has a major importance.
At the moment, in this context, this was a good design.

The `NewExecutor` did a lot of things, but I will focus on 3 majors:
- the configuration loading
- the creation of the commands
- a (partial) building of the linters

The `NewExecutor` was called before everything.
This means that every command, even the version command, before running, will load the configuration and build linters.
In the past, this has led to breaking all the commands because of a linter panic or a configuration error.

The configuration structures were a mix of real configuration (`Run`, `Linters`, etc.) and specific needs of each command (`Version`, `IsVerbose`, etc.).

The configuration was filled with 2 different techniques (Cobra and Viper), after 3 parsing phases.
In detail, the first parsing (Cobra) was mainly for the logger (early access), the data obtained were used to initialize the configuration.
And then Viper was reading the configuration files by overriding the previous data.
The resulting extracted data was used as default values (because it was done before the creation of the commands) for the next and last parsing.
The flags were parsed (Cobra), through the commands, and partially overridden the previous data (I will explain later why I say "partially").

The linters were built before the full configuration was loaded.

And I didn't talk about the complexity of `Executor` itself: everything was linked together, and it was tough to know when something was used or loaded.

I think based on my previous explanation, you can now see the problems:
- performance issues
- very complex and unreliable configuration parsing
- timing issues that lead to load linters without the proper configuration
- impossible to fix something quickly, so very hard to maintain

### PR Content

Now I will summarize the PR: it kills the `Executor` :smile:

- the configuration structures contain only user configuration (except some test elements)
- the configuration is loaded only for commands that need the configuration
- the configuration is loaded only by Viper (via a simple binding)
    - replacement of `config.Reader` with `config.Loader`
- the linters are loaded only for commands that need linters
- the specific options of each command are handled by each command itself
- the flag sets are organized with simple functions (and without a link to the configuration)
- each command is independent of the other commands and manages its dependencies (`lintersdb.Manager`, `lintersdb.EnabledSet`, etc.)

### Side Effect

I implemented a hack, no side effects now.

<details><summary>previous explanation</summary>

Now, all flags override the configuration from the file, not partially:
before, for some slice of string options (those usable with a flag), the values were appended instead of overridden.

So it's breaking.

Based on the code, this hack has been created to being able to merge linters from the files (`linters.enable`, `linters.disable`) with the flags one (`-E`, `--enable`, `-D`, `--disable`).
It was possible to add/append a linter to the current configuration (obtained by the file) with the command line.
I think it was interesting when only a few linters existed but the number of linter has increased since the beginning of golangci-lint.

Example:

| file    | flag   | Old behavior | New behavior |
|---------|--------|--------------|--------------|
| govet   | gofmt  | govet,gofmt  | gofmt        |

So I have several proposals:
- consider that as a breaking change -> major version
- find a hack to keep the same behavior:
    - the previous hack (`fixSlicesFlags`) used inside golangci-lint doesn't work anymore with the new way to handle the configuration because it was specific to Cobra.
    - based on issues I found, Viper doesn't support that.
      I tried several "pure" Viper hacks (ex: `UnmarshalText`) but it doesn't work.
    - I found a dirty hack: the duplication of the fields and the merge of the 2 fields during the load of the configuration:
    ```go
    type Linters struct {
  	  FlagsEnable []string `mapstructure:"flags-enable"` // should never be used inside a configuration file
  	  Enable      []string
      // ...
    }
    ```
    - I found a less dirty hack but still not perfect: Viper loads the configuration and flags that are not slices. For the slices, we don't bind them with Viper, so we can have the value via Cobra.
    ```go
    fs.StringSliceP("enable", "E", nil, wh("Enable specific linter"))
  
    // ...
  
    if cmd.Flags().Changed("enable") {
      val, err := cmd.Flags().GetStringSlice("enable")
      // ...
    }
    ```
- a mix of both previous proposals -> major version
- something else?

</details>

### Next step

The rewrite of `lintersdb.Manager` to avoid the build of the linters 5 times by run.

[--update-refs](https://git-scm.com/docs/git-rebase/2.43.0#Documentation/git-rebase.txt---update-refs) is still the best git rebase option ever!


.golangci.reference.yml
```diff
@@ -68,17 +68,13 @@ run:
   # Allow multiple parallel golangci-lint instances running.
   # If false, golangci-lint acquires file lock on start.
   # Default: false
-  allow-parallel-runners: false
+  allow-parallel-runners: true
```
bombsimon, line 71: Do you wanna keep this as `false` for default even though this change? I noticed #4100 but didn't have a good answer so didn't respond to it. Thinking since there are more minor breaking changes and the reference is changed.

pkg/commands/flagsets.go
```diff
@@ -0,0 +1,100 @@
...
+	internal.Vibra(v, fs, fs.String, "new-from-rev", "issues.new-from-rev", "",
+		wh("Show only new issues created after git revision `REV`"))
+	internal.Vibra(v, fs, fs.String, "new-from-patch", "issues.new-from-patch", "",
+		wh("Show only new issues created in git patch with file path `PATH`"))
+	internal.Vibra(v, fs, fs.Bool, "whole-files", "issues.whole-files", false,
+		wh("Show issues in any part of update files (requires new-from-rev or new-from-patch)"))
+	internal.Vibra(v, fs, fs.Bool, "fix", "issues.fix", false, wh("Fix found issues (if it's supported by the linter)"))
+}
+
+func wh(text string) string {
```
bombsimon, line 98: Maybe take the opportunity to improve the naming, `wh` doesn't give much info without context.

pkg/commands/internal/vibra.go
```diff
@@ -0,0 +1,32 @@
...
+func Vibra[T any](v *viper.Viper, fs *pflag.FlagSet, pfn FlagFunc[T], name, bind string, value T, usage string) {
+	pfn(name, value, usage)
+
+	err := v.BindPFlag(bind, fs.Lookup(name))
+	if err != nil {
+		panic(fmt.Sprintf("failed to bind flag %s: %v", name, err))
+	}
+}
+
+// VibraP adds a Cobra flag and binds it with Viper.
```
bombsimon, line 24: What does the `P` stand for?

pkg/commands/internal/vibra.go
```diff
@@ -0,0 +1,32 @@
...
+
+	"github.com/spf13/pflag"
+	"github.com/spf13/viper"
+)
+
+type FlagFunc[T any] func(name string, value T, usage string) *T
+
+type FlagPFunc[T any] func(name, shorthand string, value T, usage string) *T
+
+// Vibra adds a Cobra flag and binds it with Viper.
```
bombsimon, line 14: Can't we just call this what it does so it gives some context even if you're not fully familiar with Viper and Cobra (which I guess this is some combination of)?

Suggestion:
- `AddFlagAndBind
- `AddAndBindFlag`

Also maybe I'm missing something but where is Cobra in this picture? Can't see it here or from the caller in `flagset`, seems to be using `pflag` which doesn't look to be a part of Cobra?

pkg/commands/root.go
```diff
@@ -1,160 +1,168 @@
...
-	envMemProfileRate = "GL_MEM_PROFILE_RATE"
-)
+func Execute(info BuildInfo) error {
+	return newRootCommand(info).Execute()
+}
+
+type rootOptions struct {
+	PrintVersion bool
+
+	Verbose bool   // Persistent flag.
```
bombsimon, line 25: What does persistent mean here? That it is in affect for all sub commands as well?

pkg/commands/root.go
```diff
@@ -1,160 +1,168 @@
...
-}
+	fs := rootCmd.Flags()
+	fs.BoolVar(&c.opts.PrintVersion, "version", false, wh("Print version"))
 
-func (e *Executor) persistentPreRun(_ *cobra.Command, _ []string) error {
-	if e.cfg.Run.PrintVersion {
-		_ = printVersion(logutils.StdOut, e.buildInfo)
-		os.Exit(exitcodes.Success) // a return nil is not enough to stop the process because we are inside the `preRun`.
-	}
+	setupRootPersistentFlags(rootCmd.PersistentFlags(), &c.opts)
```
bombsimon, line 57: Ah, here it's pretty clear what persistent flags means, maybe the comments I commented about can be removed since they add more confusion than clarification.

pkg/commands/root.go
```diff
@@ -1,160 +1,168 @@
...
+func safeArgs(fs *pflag.FlagSet, args []string) []string {
+	var shorthands []string
+	fs.VisitAll(func(flag *pflag.Flag) {
+		shorthands = append(shorthands, flag.Shorthand)
+	})
+
+	var cleanArgs []string
+	for _, arg := range args {
+		if len(arg) > 1 && arg[0] == '-' && arg[1] != '-' && !slices.Contains(shorthands, string(arg[1])) {
+			cleanArgs = append(cleanArgs, "--potato")
```
bombsimon, line 160: 🥔

pkg/commands/root.go
```diff
@@ -1,160 +1,168 @@
...
 }
 
-func getDefaultConcurrency() int {
-	if os.Getenv(envHelpRun) == "1" {
-		// Make stable concurrency for generating help documentation.
-		const prettyConcurrency = 8
-		return prettyConcurrency
+// Shorthands are a problem because pflag, with UnknownFlags, will try to parse all the letters as options.
+// A shorthand can aggregate several letters (ex `ps -aux`)
+// The function replaces non-supported shorthands by a dumb flag.
```
bombsimon, line 150: Is this the same as the current behavior? What's the rationale behind silently ignoring such faulty usage?

pkg/config/loader.go
```diff
@@ -10,33 +10,64 @@ import (
...
+		log:   log,
+		cfg:   cfg,
 	}
 }
 
-func (r *FileReader) Read() error {
-	// XXX: hack with double parsing for 2 purposes:
-	// 1. to access "config" option here.
-	// 2. to give config less priority than command line.
+func (r *Loader) Load() error {
```
bombsimon, line 49: Should we update the receiver variable since `Reader` got renamed to `Loader`?

```suggestion
func (l *Loader) Load() error {
```

(Probably benefits from LSP capabilities to do proper renaming, I won't suggest changes in more places)

pkg/config/run.go
```diff
@@ -25,20 +25,6 @@ type Run struct {
...
 
-	// --- Flags only section.
-
-	IsVerbose bool `mapstructure:"verbose"` // Flag only
-
-	PrintVersion bool // Flag only. (used by the root command)
-
-	CPUProfilePath string // Flag only.
-	MemProfilePath string // Flag only.
-	TracePath      string // Flag only.
```
bombsimon, line 36: Didn't see any comment on this but I see you ripped out all the profiling parts. Just because it wasn't used and mostly aims for development of `golangci-lint`?

bombsimon (commented):
Sorry for slow reaction, took a first quick pass! Mostly added some questions and some minor suggestions. Seems like a good improvement and split of tangled code. 👍

pkg/config/run.go
ldez: It didn't remove the profiling, I just moved the profiling to the only command that needs profiling: the `run` command.

.golangci.reference.yml
ldez: I already changed a lot of things, I will not change everything in this PR :smile: 
Here, it's just a wrong value in the reference.

pkg/commands/internal/vibra.go
ldez: It's the same as `VarP`, `BoolP`, etc. it's when you want to add a shorthand.

pkg/commands/root.go
ldez: > Is this the same as the current behavior?

Yes, before the "pre-parsing" was done with all the flags but only some were used.

The function works in isolation, there are no impact on the effective flag parsing.

pkg/commands/internal/vibra.go
bombsimon: Sorry to be a party 🎉 pooper 💩! If you like the name, keep it! I guess you won't read this code anyways if you're not familiar with the concepts and the comments explain the name!

.golangci.reference.yml
bombsimon: No and that's appreciated! Just didn't follow why the reference changes this value that seems to be untouched in the rest of the PR.

pkg/config/run.go
bombsimon: Am I blind? I can't see the flags `cpu-profile-path` and `mem-profile-path` added back in any of your additions?

ldez: `setupRunPersistentFlags` in the `run` command

.golangci.reference.yml
ldez: I reverted this change, I will create another PR.

pkg/commands/internal/vibra.go
ldez: I'm not attached to this name :smile: it was just fun, so I will keep your suggestion in can be better for readability.

pkg/config/run.go
bombsimon: Omg so silly, missed it was collapsed and tried to sesrch for the term in the browser. 🤦🏽‍♀️ Sorry about that!

bombsimon (approved):

ldez:
It's emotional for me to merge this PR, there is a lot of work behind it.
And the next step will also be great! The PR will happen quickly!

pkg/commands/version.go
```diff
@@ -26,65 +24,75 @@ type versionInfo struct {
 	BuildInfo *debug.BuildInfo
 }
 
-func (e *Executor) initVersion() {
+type versionOptions struct {
+	Format string `mapstructure:"format"`
```
Antonboom, line 28: don't see these fields in config reference 🤔 
or what reason of mapstructrure?

pkg/commands/root.go
```diff
@@ -1,160 +1,168 @@
...
-		}
+	return c.cmd.Execute()
+}
 
-		var ms runtime.MemStats
-		runtime.ReadMemStats(&ms)
-		printMemStats(&ms, e.log)
+func setupRootPersistentFlags(fs *pflag.FlagSet, opts *rootOptions) {
+	fs.BoolVarP(&opts.Verbose, "verbose", "v", false, color.GreenString("Verbose output"))
+	fs.StringVar(&opts.Color, "color", "auto", color.GreenString("Use color when printing; can be 'always', 'auto', or 'never'"))
```
Antonboom, line 90: are `verbose` and `color` actual for not `run` commands?

pkg/commands/internal/vibra.go
```diff
@@ -0,0 +1,32 @@
+package internal
```
Antonboom, line 1: `vibra` 😂

pkg/commands/config.go
```diff
@@ -5,15 +5,27 @@ import (
...
 
 	"github.com/golangci/golangci-lint/pkg/config"
 	"github.com/golangci/golangci-lint/pkg/exitcodes"
 	"github.com/golangci/golangci-lint/pkg/fsutils"
+	"github.com/golangci/golangci-lint/pkg/logutils"
 )
 
-func (e *Executor) initConfig() {
+type configCommand struct {
+	viper *viper.Viper
```
Antonboom, line 17: some offtop: do we really need the `viper` dependency?

why not just declare the strict ways to use golangci-lint:

1. flags for command itself settings
2. config for linter specific settings

without intersection

Antonboom (commented):
image

pkg/commands/root.go
ldez: they are global, but they are always used by the root command like before.

pkg/commands/config.go
ldez: > some offtop: do we really need the viper dependency?

I think at some point I will remove viper.

> why not just declare the strict ways to use golangci-lint:
>
>    flags for command itself settings
>    config for linter specific settings
>
> without intersection

It's useful to be be the mix both on linters (ex: `-E`, `-D`)

pkg/commands/version.go
ldez: copy-pasting :smile:

ldez: Removed in the next PR

Issue #1909, opened 2021-04-12
The --fast flag does not overwrite the configuration file


rluders:
- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [ ] Yes, I've included all information below (version, config, etc).
- [ ] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

<details><summary>Description of the problem</summary>

It seems that the flag `--fast` kind of doesn´t overwrite the lining configuration that lives in the configuration file. At the same time, I would like to keep the file configurations, but at the same time, IMHO the flag `--fast` should have priority over the configuration file.

</details>

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
# golangci-lint has version 1.39.0 built from 9aea4aee on 2021-03-26T08:02:53Z
```

</details>

<details><summary>Config file</summary>

```console
$ cat .golangci.yml
You can try to use the one that is running on this own repository.
```

</details>

<details><summary>Go environment</summary>

```console
$ go version && go env
# paste output here
```

</details>

<details><summary>Verbose output of running</summary>

```console
$ golangci-lint cache clean
$ golangci-lint run -v
# paste output here
```

</details>

<details><summary>Code example or link to a public repository</summary>

```go
// add your code here
```

</details>

ldez:
Hello,

The `--fast` works as expected with `enable-all` but with `disable-all` this option seems ignored.

Currently, I don't know if the behavior related to `disable-all` is expected.
It's possible to think, as when you are using `disable-all` you have to use `enable`, that `enable` is a filter that overrides the `--fast` filter.
 
<details>
<summary>main.go</summary>

```go
package main

import (
	"net/http"
)

func main() {
	println("foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo foo")

	req, err := http.NewRequest(http.MethodGet, "http://localhost:8080", nil)
	if err != nil {
		return
	}

	response, err := http.DefaultClient.Do(req)
	if err != nil {
		return
	}

	println(response.StatusCode)
}
```

</details>

<details>
<summary>disable-all</summary>

```yml
# .golangci.yml
linters:
  disable-all: true
  enable:
    - lll # fast
    - noctx # slow
```


```console
$ golangci-lint run
main.go:9: line is 127 characters (lll)
[...]

PR #4416, opened 2024-02-25
dev: add tests on linterdb package

ldez:
- add tests on Validator
- add tests on EnableSet

I don't add tests on the Manager because I will add them in the PR with the rewrite of the Manager (this is the step next after #4412).

### DIFF
--- a/pkg/lint/lintersdb/enabled_set_test.go
+++ b/pkg/lint/lintersdb/enabled_set_test.go
@@ -4,12 +4,91 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 
 	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
 	"github.com/golangci/golangci-lint/pkg/lint/linter"
+	"github.com/golangci/golangci-lint/pkg/logutils"
 )
 
-func TestGetEnabledLintersSet(t *testing.T) {
+type dummyLogger struct{}
+
+func (d dummyLogger) Fatalf(_ string, _ ...any) {}
+
+func (d dummyLogger) Panicf(_ string, _ ...any) {}
+
+func (d dummyLogger) Errorf(_ string, _ ...any) {}
+
+func (d dummyLogger) Warnf(_ string, _ ...any) {}
+
+func (d dummyLogger) Infof(_ string, _ ...any) {}
+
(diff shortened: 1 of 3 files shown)


bombsimon (approved):
🥳

pkg/lint/lintersdb/enabled_set_test.go
```diff
@@ -4,12 +4,91 @@ import (
...
+func (d dummyLogger) Panicf(_ string, _ ...any) {}
+
+func (d dummyLogger) Errorf(_ string, _ ...any) {}
+
+func (d dummyLogger) Warnf(_ string, _ ...any) {}
+
+func (d dummyLogger) Infof(_ string, _ ...any) {}
+
+func (d dummyLogger) Child(_ string) logutils.Log {
+	return nil
```
Antonboom, line 28: return `dummyLogger{}`?

pkg/lint/lintersdb/enabled_set_test.go
```diff
@@ -117,3 +196,68 @@ func TestGetEnabledLintersSet(t *testing.T) {
 		})
 	}
 }
+
+func TestEnabledSet_combineGoAnalysisLinters(t *testing.T) {
+	m := NewManager(nil, nil)
```
Antonboom, line 201: proposal: restrict zero value for `NewManager` args and require at least some config (default or empty struct) or some logger

pkg/lint/lintersdb/validator_test.go
```diff
@@ -0,0 +1,410 @@
...
+		desc: "enable-all and disable",
+		cfg: &config.Linters{
+			Enable:     nil,
+			EnableAll:  true,
+			Disable:    []string{"goimports", "gosec", "nolintlint"},
+			DisableAll: false,
+		},
+	},
+	{
+		desc: "enable-all and enable and fast",
```
Antonboom, line 239: offtop: could you explain why `Fast` in this condition at all?

```
if cfg.EnableAll && len(cfg.Enable) != 0 && !cfg.Fast {
```

Antonboom (commented):
🐌

pkg/lint/lintersdb/enabled_set_test.go
ldez: the new Manager from the next PR resolve that

pkg/lint/lintersdb/validator_test.go
ldez: Because of the way that fast is handled.
Cannot be changed for now: https://github.com/golangci/golangci-lint/issues/1909

PR #4419, opened 2024-02-26
dev: rewrite linters Manager

ldez:
I just put a log inside the methods used to build/load linters/plugins, and validate linters in the configuration.
You can see the problem easily.

Before the PR #4404:

<details>

```console
$ go run ./cmd/golangci-lint run 
Build all plugins.
Build all linters.
Build all plugins.
Build all linters.
Validate EnabledDisabledLinters.
Build all linters.
Validate EnabledDisabledLinters.
Build all linters.
Build all linters.
Validate EnabledDisabledLinters.
Build all linters.
$
```

| Build       | Count |
|-------------|-------|
| all linters | 6     |
| all plugins | 2     |
| validate    | 3     |

</details>

After the PR #4404:

<details>

```console
go run ./cmd/golangci-lint run 
Build all plugins.
Build all linters.
Validate EnabledDisabledLinters.
Build all linters.
Validate EnabledDisabledLinters.
Build all linters.
Build all linters.
Validate EnabledDisabledLinters.
Build all linters.
$
```

| Build       | Count |
|-------------|-------|
| all linters | 5     |
| all plugins | 1     |
| validate    | 3     |

</details>

With this PR:

<details>

```console
$ go run ./cmd/golangci-lint run
Build all linters.
Build all plugins.
Validate EnabledDisabledLinters.
$
```

| Build       | Count |
|-------------|-------|
| all linters | 1     |
| all plugins | 1     |
| validate    | 1     |

</details>

This behavior has been here since nearly the beginning of golangci-lint but the problem has increased with time.

Summary of the execution after #4404, #4412, and this PR:

- Set up logger.
- Load the configuration from file and flags (`config.Loader.Load`).
- Initiate the linter database based on the effective configuration (`linterdb.NewManager`).
  - Build the linters (internals and plugins) (`linterdb.LinterBuilder.Build`, `linterdb.PluginBuilder.Build`)
  - Validate configuration (`linterdb.Validator.Validate`)
- Runs the analysis.

Everything is done once and only once.

The improvement of the performance is not high, but the impact on debugging and maintainability is huge!

### Next Step

A proposal for a new flag option and a bit of cleaning.

PR #4420, opened 2024-02-27
dev: add test on ExcludeRule and SeverityRule

ldez:
When working on #4419, I saw that those validation methods were not tested.
And I find a bug when using `path` and `path-except` at the same time.

### DIFF
--- a/pkg/config/issues_test.go
+++ b/pkg/config/issues_test.go
@@ -4,11 +4,16 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestGetExcludePatterns(t *testing.T) {
-	assert.Equal(t, GetExcludePatterns(nil), DefaultExcludePatterns)
+	patterns := GetExcludePatterns(nil)
 
+	assert.Equal(t, patterns, DefaultExcludePatterns)
+}
+
+func TestGetExcludePatterns_includes(t *testing.T) {
 	include := []string{DefaultExcludePatterns[0].ID, DefaultExcludePatterns[1].ID}
 
 	exclude := GetExcludePatterns(include)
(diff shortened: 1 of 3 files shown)


pkg/config/issues_test.go
```diff
@@ -4,11 +4,16 @@ import (
...
 
 	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
 )
 
 func TestGetExcludePatterns(t *testing.T) {
-	assert.Equal(t, GetExcludePatterns(nil), DefaultExcludePatterns)
+	patterns := GetExcludePatterns(nil)
 
+	assert.Equal(t, patterns, DefaultExcludePatterns)
```
Antonboom, line 13:
```suggestion
	assert.Equal(t, DefaultExcludePatterns, patterns)
```

pkg/config/issues_test.go
```diff
@@ -19,3 +24,109 @@ func TestGetExcludePatterns(t *testing.T) {
 		assert.Contains(t, DefaultExcludePatterns, p)
 	}
 }
+
+func TestExcludeRule_Validate(t *testing.T) {
```
Antonboom, line 28: the same comments from me like for `TestSeverity_Validate`

Antonboom (approved):

pkg/config/severity_test.go
```diff
@@ -0,0 +1,77 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSeverity_Validate(t *testing.T) {
+	testCases := []struct {
+		desc     string
+		rule     *SeverityRule
+		expected string
```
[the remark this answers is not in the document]
ldez: it's not really error, it's the message, I prefer `expected` in this context.

Issue #3826, opened 2023-05-08
more flexible loading of custom plugins


pohly:
### Your feature request related to a problem? Please describe.

In Kubernetes, we have a `hack/golangci-lint.yaml` with a path for a logcheck.so which is relative to that config:
https://github.com/kubernetes/kubernetes/blob/b09b7e5bdb1c763c6d7e47e2c792d54f37f7d9a2/hack/golangci.yaml#L42

That works because normally, `_output` contains build artifacts. It fails when using containerized builds where build artifacts are located elsewhere:
https://github.com/kubernetes/kubernetes/issues/117831

### Describe the solution you'd like.

When loading a plugin, golangci-lint could first try the absolute or relative path and if that file doesn't exist, fall back to searching for the plugin in directories specified via env variables. This could be PATH (kind of convenient because the plugin gets compiled together with the golangci-lint binary to ensure that both use the same Go and compile options and already gets placed in the same directory) or a custom variable like GOLANGCI_LINT_CUSTOM_PLUGIN_PATH.

### Describe alternatives you've considered.

- An absolute path doesn't work, developers check out there source in different locations.
- Patching the config file is ugly (dirty repository during the build).
- Copying the config file elsewhere and modifying it there messes with other relative paths.

ldez:
Hello,

> When loading a plugin, golangci-lint could first try the absolute or relative path and if that file doesn't exist, fall back to searching for the plugin in directories specified via env variables.

Do you mean: if the path is a relative path then use the env var as a root?
ex: `./foo/foo.so` => `filepath.Join("GOLANGCI_LINT_CUSTOM_PLUGIN_PATH", "./foo/foo.so")`


Side note: maybe [logcheck](https://github.com/kubernetes-sigs/logtools/tree/main/logcheck) can be "merged" with [loggercheck](https://github.com/timonwong/loggercheck)

pohly:
I would:
- Completely strip the file path: `./foo/foo.so` -> `foo.so`. Otherwise re-creating the existing path elsewhere becomes awkward.
- Treat `GOLANGCI_LINT_CUSTOM_PLUGIN_PATH` like `PATH` in a shell and support multiple directories. That way, multiple different custom plugins could get built in different directories. Might be overkill, though.

`logcheck` is heavily biased towards Kubernetes logging conventions. I don't think it can or should get merged with other linters, even if some aspects of it are similar.

pohly:
[...]

Issue #2566, opened 2022-02-13
support automatically building plugin linters


ashanbrown:
### Your feature request related to a problem? Please describe.

Currently you have to build a plugin locally at a known path prior to running golangci-lint, which creates a challenge for setting up dev and CI environments that use this plugin.

### Describe the solution you'd like.

I think one way to address this would be to add a new field, perhaps called "build" to the custom configuration.  For example, it might look like this:

```
  custom:
    gofmts:
      build: github.com/ashanbrown/gofmts/golangci-lint
      description: gofmts sort formatter
      original-url: github.com/ashanbrown/gofmts
```

What I think this would do is run:

```
go build -o <some path in the cache>/plugin.so  --buildmode=plugin github.com/ashanbrown/gofmts/golangci-lint
```
and then auto-determine the `path` field based on where the plugin is built.

This appears to work within a module-based environment that also includes golangci-lint in the go.mod file.

The one question I have is where the plugins would be stored.  We use a cache directory that is shared between versions of golang ci lint (on my mac its ~/Library/Caches/golangci-lint/) and the plugin.so files need to be built for a precise set of dependencies, so the path to the plugin file should probably be some hash including all of the dependencies.  Despite the shared cache, it might make sense just to have a plugin per package/plugin pair and ignore the dependencies in the path, so that we don't leak old plugins.

### Describe alternatives you've considered.

I think think the alternative is to instrument dev and CI environments to build the plugin themselves.

dnephin:
I wrote a custom plugin for `golangci-lint` recently and ran into all the same problems. 

I was able to script the module versions using the output of `go version -m` and the `golang.org/x/mod/modfile` package. You can see an example of that [here](https://github.com/infrahq/infra/pull/3201/files).

Unless  the developer has built `golangci-lint` locally, I don't see a way to automatically deal with the Go version. 

Is part of this proposal to also built `golangci-lint` locally? It seems like if the binary for `golangci-lint` comes from a package manager or a pre-build binary, there's no way to build the plugin with the correct Go version.

ldez:
I think the "new" plugin system addressed this issue.

https://golangci-lint.run/docs/plugins/module-plugins/

PR #4437, opened 2024-03-02
poc: new custom linters system

### LINKED ISSUE #1276: Debug why a private linter was not added to the list
Hi, I am using a private linter but it doesn't appear in the list of linters.

- I have checked the dependencies list to make sure all versions are the same as the golang ci 1.29 (http://github.com/alexisvisco/logwercase).
- My .golang-ci.yml is like that:

```yml
linters:
 enable:
   - unparam
   - misspell
   - godox
   - gocritic
   - scopelint
   - golint
   - unconvert
   - gofmt
   - exportloopref
   - govet
   - bodyclose
   - gosimple
   - goconst
   - gosec
   - logwercase

linters-settings:
  custom:
    logwercase:
      path: /home/aviscogl/go/bin/logwercase.so
      description: Analyze case of log message and WithField keys
      original-url: github.com/alexisvisco/logwercase
```

There is an option to known why a private linter is not added? https://golangci-lint.run/contributing/debug/ Does not help me ...

### LINKED ISSUE #3669: Allow usage for custom linters without a need to recomplie `golangci` binary.

Current requirement to recompile everything to use for a custom private linter is sub-optimal.

Would be great to have an Interpreter built-into into golangci, something like [yaegi](https://github.com/traefik/yaegi) so that custom linters could be just a bunch of code in local repo.

Recompiling custom linter as plugin is not an option here.

### Additional context.

_No response_

### LINKED ISSUE #4427: Allow custom binaries with custom linters

Adding a custom linter is a massive headache.

Provide a mechanism for creating wrapper binaries that register custom linters.

- Loading custom linters as plugins. This is a massive headache.
- Create a custom wrapper that uses hacks to register a custom linter. I did this, and it works but it's a horrible hack relying on unsafe and it breaks with the latest version.

### Additional context.

I already [implemented this](https://github.com/firelizzard18/golangci-lint/tree/custom-linters). If this issue is received positively I will submit a PR.

ldez:
This PR is a POC to discuss the custom linters.

The idea comes during the rewrite of the commands and the `Manager`.

We know the complaints about custom linter with the [Go plugin system](https://pkg.go.dev/plugin).

Today if you want to use a plugin, you have to:

- build `golangci-lint` from sources with `CGO_ENABLED=1`
- build the plugins with the same stack (Go, dependencies, etc.) as `golangci-lint`
- copy your plugins somewhere

The problems are:

- the requirement to have `golangci-lint` and the plugins build with the same stack.
- the requirement to store and rebuild the plugins when you need to share them (with other devs or CI).
- limited OS support

And I will cite the Go documentation:

> Together, these restrictions mean that, in practice, the application and its plugins must all be built together by a single person or component of a system.
> https://pkg.go.dev/plugin#hdr-Warnings

For me, it's clear that Go plugin system is not adapted to extend golangci-lint.

> In that case, it may be simpler for that person or component to generate Go source files that blank-import the desired set of plugins and then compile a static executable in the usual way.
> https://pkg.go.dev/plugin#hdr-Warnings

I propose to use a different approach that follows the Go recommendation, the principle is similar to SQL drivers in Go:

- the plugins are simple Go dependencies that follow a contract and are imported with a blank-import.
- a plugin uses `init` to register itself.
- inside golangci-lint we read the register to load the plugins.

FYI I already created something similar on https://github.com/kvtools/valkeyrie.

I put `how-to.md` and `readme.md` inside the directory `pkg/experimental` and its subdirectories.

<details>

```
pkg/experimental
├── commands
│   ├── custom                    # The command to build a custom version of golangci-lint with custom linters.
│   │   ├── custom.go
│   │   └── internal
│   │       ├── builder.go
│   │       ├── configuration.go
│   │       └── imports.go
│   ├── .mygcl.reference.yml      # The configuration reference.
│   └── mygcl.schema.json         # The JSONSchema of the configuration.
├── db
│   ├── builder_plugin_module.go  # the new plugin loader.
│   ├── builder_plugin_wrapper.go # temporary: it's just because it was easy to dev with that.
│   └── readme.md
├── how-to.md                     # How to use this POC.
├── modules
│   ├── myplugin                  # A plugin example (should be seen as an external module).
│   │   ├── myplugin.go
│   │   └── readme.md
│   └── register                  # the register library (should be seen as an external module).
│       ├── readme.md
│       └── register.go
└── todo.md                       # Just personal notes.
```

</details>

There is a new command: `golangci-lint custom`

If we want to follow this way we can either keep the command inside `golangci-lint` or use it as an external tool (inside the org).

The idea is to place the configuration file `.mygcl.yml` into a project that requires a plugins and then just run `golangci-lint custom`.
A binary `gcl-custom` will be automatically built and placed where the command has been run.

You should read the file `pkg/experimental/how-to.md` for more details.

Fixes #4427
Fixes #3669
Fixes #1276

Maybe a fix for #3826, #2566


ldez:
FYI The failure of the CI is expected because it's just a POC.

ldez:
@pohly I'm interested in your feedback about this new way to use plugins.
I think it can simplify your workflow inside Kubernetes.

firelizzard18:
I like this design. I've been using a custom binary (the manual way, more or less) that I hacked to load my own plugins. From reading this PR, I'm confident it will cover my current use case without having to resort to hacks.

It would be nice to have a way for `golangci-lint run` to automatically build and execute a custom binary. I have been using a custom binary and that works fine, but the vscode-go plugin likes rebuilding golangci-lint, which clobbers my custom binary. Of course I could (and probably should) rename my custom binary something else and configure vscode-go to use that instead, but it would be rather convenient if `golangci-lint run` worked (built and executed a custom binary) with just a bit of configuration.

ldez:
With this POC you can generate a binary called `golangci-lint` and use it to generate another binary also called `golangci-lint`, etc.

I prefer to keep the thing separated: one command to run, one command to build.
Custom plugins are a niche, I think it's not worth the cost of creating an "auto-magical" run command.

Maybe the vscode-go plugin can offer an option to handle the binary (name, path, etc.)

firelizzard18:
That makes sense. Once this change lands I can submit a PR to vscode-go to detect `.mygcl.yml` and prompt the user to set a setting that will automatically build with `golangci-lint custom`. That will allow users with custom plugins to opt-in to using custom-built binary mode.

pkg/experimental/how-to.md
```diff
@@ -0,0 +1,63 @@
+## The Manual Way
+
+- add a blank import of module inside `cmd/golangci-lint/plugins.go`
```
pohly, line 3: It's worth mentioning here that this is a new file and doesn't need to be called exactly like that. When I first read this, it looked like it already exists and needs to be extended.

pohly (commented):
This looks good to me. I've been proposing a similar mechanism myself elsewhere, without knowing about this SQL driver precedent.

I'm not sure whether we would use this in Kubernetes, though: we already rebuild in the CI from source, so building the Go plugin in the same environment is not a burden. Because we build with `go install`, adding a file to the build would be more complicated than what we do now.

I'm aware that `go install` is not recommended. It has been working so far.

pkg/experimental/how-to.md
ldez: This file already exists but only in my PR :wink:

ldez:
@bombsimon do you have an opinion about this PR?
Should I start to work on the cleaning of this PR?

bombsimon:
I don't have much experience with how a plugin interface would look like without re-building so with that in mind this looks fine to me. But it also makes me feel, if shelling out to `git` and `go` replacing directives to then re-build `golangci-lint` from source is the proposed solution, it's not too far from adding my linter as a regular `git` patch and apply them and re-build `golangci-lint`. Obviously this would be a more stable interface and better user experience but yeah...

I'm also curious about the decision to bootstrap and build `golangci-lint`. I know there's been requested to use a specific version of `golangci-lint` based on the configuration and one of the reasons we didn't want to do that was to avoid external network calls to fetch and build/install `golangci-lint`. Has that changed or is there a difference in your opinion?

Either way, the current plugin system is crap and this is a great initiative! I don't have any better suggestion and it seems to satisfy users who requested this! 👍

ldez:
> it's not too far from adding my linter as a regular git patch and apply them and re-build golangci-lint.

The goal of an official way to build plugins is just to simplify the usage and provide something stable, but you can do the same manually with different approaches.

> Has that changed or is there a difference in your opinion?

I don't remember the "check the version" topic :thinking: 

From my memory, the call to external resources was to get something not controlled by golangci-lint or a public source of trust, and updatable at runtime.

Here, we rely on GitHub (only for the golangci-lint repo) and Go proxies, they are not perfect sources of trust but at least it's manageable from the point of view of the security.
Also, this cannot be changed at runtime and the resources are called only on running explicitly a specific command.

At least, this approach is more secure than the current Go plugin system https://twitter.com/dominikhonnef/status/1394766501157167112

bombsimon:
Thanks for elaborating, it sounds like you've given this a lot of though and considered the alternatives!

I'm mostly asking because once this lands I guess it will be a part of `golangci-lint` for quite some time, even if there are progress with an alternative solution based on interpreters or wasm. 

I think a wasm solution could be interesting to explore as well. I've seen projects like https://github.com/knqyf263/go-plugin which seems pretty cool although I didn't dig much deeper to see how relevant it is. But also speaking of the linked inspiration by HashiCorp that seems pretty well tested, have you had any look to see if something like https://github.com/hashicorp/go-plugin would fit? It looks pretty nice to run the plugin in a subprocess and use rpc communication.

But looking at this poc implementation there doesn't seem to be a lot to maintain and I don't have any concrete alternative or relevant experience so don't interpret my questions as if I'm not for this suggestion - I'm all for it and just want to bounce some alternatives.

Btw do you plan to deprecate the current plugin system closely after or will they live in parallel?

ldez:
> But also speaking of the linked inspiration by HashiCorp that seems pretty well tested

This kind of plugin is extremely slow (from the POV of golangci-lint) but it's not the main problem.
It's RPC communication, then the communication is based on network calls and so requires that everything be serializable, I'm not sure if it's possible or if I want to create and maintain a binding for the goanalysis package.

Like the other ways to create plugins, the pros and cons depend on the usage context.

golangci-lint needs a high-performance extension system and easy maintainance.

> Btw do you plan to deprecate the current plugin system closely after or will they live in parallel?

I don't think so: the 2 plugin systems can live together, and they don't require a lot of maintenance.
But this also doesn't mean that we will never deprecate one of those systems: we don't know the future :smile:

ldez:
The PR is now ready for review.

The 2 first commits are the POC and the revert of the POC commit, then you can ignore them.

The other commits contain one focused change by commit ("easy" to review).

The doc is not perfect but I think it's a good start.

The new module `github.com/golangci/plugin-module-register` will be released at the end of the review phase before the merge.

.mygcl.reference.yml
```diff
@@ -0,0 +1,37 @@
+# The golangci-lint version used to build the custom binary.
```
bombsimon, line 1: I'm note sure how I feel about `mygcl` as a name for this. It's short and nice but not obvious or an obvious abbreviation, I guess because of the org I more often see `golangci` than `gcl`.

Since we use `golangci.yml` for the other configuration file, would something like `my-golangci.yml` or maybe `custom-golangci.yml` make sense? Or since it's purely for plugins, maybe `golangci-plugins.yml`?

pkg/commands/internal/configuration.go
```diff
@@ -0,0 +1,138 @@
...
+	"os"
+	"path/filepath"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+const base = ".mygcl"
+
+const defaultBinaryName = "gcl-custom"
```
bombsimon, line 15: I think the name in the reference yaml is much better default name; `custom-golangci-lint`.

.golangci.yml
```diff
@@ -33,6 +33,9 @@ linters-settings:
       - whyNoLint
   gocyclo:
     min-complexity: 15
+  godox:
+    keywords:
+      - FIXME
```
bombsimon, line 38: Unrelated to this PR?

docs/src/docs/contributing/new-linters.mdx
```diff
@@ -46,75 +46,7 @@ After that:
...
-See [plugin/example.go](https://github.com/golangci/example-plugin-linter/blob/master/plugin/example.go) for more info.
-
-To build the plugin, from the root project directory, run:
-```bash
-go build -buildmode=plugin plugin/example.go
-```
-
-This will create a plugin `*.so` file that can be copied into your project or another well known location for usage in `golangci-lint`.
+- [Go Plugin System](/plugins/module-plugins)
+- [Module Plugin System](/plugins/go-plugins)
```
bombsimon, line 52: 🎉 Nice splitting this out

docs/src/docs/plugins/module-plugins.mdx
```diff
@@ -0,0 +1,71 @@
+---
+title: Module Plugin System
+---
+
+An example linter can be found at [here](https://github.com/golangci/example-plugin-module-linter/settings).
+
+## The Automatic Way
+
+- define your building configuration into `.mygcl.yml`
+- run the command `golangci-lint custom` (`go run ./cmd/golangci-lint custom -v`)
```
bombsimon, line 10: Don't think we need to add this to the module docs, doesn't feel like a common way to run `golangci-lint`? 

```suggestion
- run the command `golangci-lint custom`
```

pkg/commands/internal/builder.go
```diff
@@ -0,0 +1,210 @@
...
+
+	err = b.copyBinary(binaryName)
+	if err != nil {
+		return fmt.Errorf("move golangci-lint binary: %w", err)
+	}
+
+	return nil
+}
+
+func (b Builder) clone(ctx context.Context) error {
```
bombsimon, line 86: Should we make use of `os.MkdirTemp` to not clone in the current directory? Always feels more safe and easier to clean up that way imo.

pkg/commands/internal/builder.go
```diff
@@ -0,0 +1,210 @@
...
+		b.log.Warnf(string(output))
+
+		return fmt.Errorf("%s: %w", strings.Join(cmd.Args, " "), err)
+	}
+
+	return nil
+}
+
+func (b Builder) goBuild(ctx context.Context, binaryName string) error {
+	//nolint:gosec
```
bombsimon, line 144: Can we be an exemplary user here and add a reason to why we disable this (not clear what `gosec` is complaning about here). That way it would also pass `nolintlint` (even though I know we don't use it).

Having a clear reason to why linters are disabled is always nice and helps future us (and anyone reading the code).

bombsimon (commented):
⭐

.golangci.yml
ldez: It's related :smile: because during the dev I put a lot of FIXME and I forgot to remove them, so I enabled godox.
So it's unrelated but related.

docs/src/docs/plugins/module-plugins.mdx
ldez: I will add the verbose version but with the binary :smile:

pkg/commands/internal/builder.go
ldez: it's already the case:

```
tmp, err := os.MkdirTemp(os.TempDir(), "mygcl")
```

pkg/commands/internal/builder.go
ldez: G204 is always the same thing: it's impossible to fix it.
If you are using a variable inside an `exec.Command`, even if sanitized, the report will still be here...
I sanitized the variable and added a comment.

pkg/commands/internal/builder.go
ldez: but maybe you suggest cloning inside the current directory :thinking: 
It doesn't seem like a good idea.

.mygcl.reference.yml
ldez: I didn't use the word `golangci` because I think it would be confusing with the configuration file `.golangci`.

pkg/commands/internal/configuration.go
ldez: I preferred a name consistent with the file configuration name `mygcl` -> `gcl` but if someone uses an alias `gcl` for `golangci-lint it can be confusing.
So I use a "verbose" non-ambiguous name.

pkg/commands/internal/builder.go
bombsimon: Yeah exactly, I was thinking not clone straight in the current directory. Reason mostly being that if I'm in my repository that I want to lint with my custom linter I don't want a clone of the repo in it. Maybe I have a directory called `golangci-lint` already and it will fail. Not likely, but also easy to avoid.

But since we clean up the clone after the build I guess there's no harm no foul here.

pkg/commands/internal/configuration.go
bombsimon: If you want to keep `mygcl.yml` and `gcl` is to be seen as an abbreviation of `golangci-lint`, wouldn't `mygcl` (or `my-gcl`) be a better name? Otherwise the definition of not upstream for config would be `my` and for binary `custom`. I think using one of them would make sense. 

And seems like this is very related to my first comment about the name of `.mygcl.yml` so let's keep it all in one thread.

We call the feature `custom` so using that instead of `my` would avoid having two references of the custom built tool (my _and_ custom). `.gcl-custom.yml` (or `.custom-gcl.yml`) and generate a binary named `gcl-custom` or keep the name `.mygcl.yml` and generate `mygcl` (or `my-gcl`).

Do you think that's a bad idea?

pkg/commands/internal/builder.go
ldez: Why using a temp dir inside the current dir instead of a temp dir inside the global temp dir?

pkg/commands/internal/configuration.go
ldez: Seems right, `.custom-gcl.yml` for the configuration and `custom-gcl` for the binary name.

bombsimon (approved):
🆒 🆒 🆒

silverwind:
I endorse this as well. It sounds like a similar systems that other linters like `eslint` also use and it scales very well there, with over a thousand of plugins available in the ecosystem.

One of my primary concerns with golangci-lint is that it takes ages to compile because of all the plugins that ship with it. With such a system a user will only compile what they actually use.

So I recommend, make a slim core with a well-defined API for plugins to access the AST and to report problems, basically exactly what `eslint` is doing today.

.custom-gcl.reference.yml
```diff
@@ -0,0 +1,37 @@
...
+# The directory path used to store the custom binary.
+# Optional.
+# Default: .
+destination: ./my/path/
+
+# The list of the plugins to integrate inside the custom binary.
+plugins:
+  # a plugin from a Go proxy
+  - module: 'github.com/example/plugin3'
+    version: v1.2.3
```
silverwind, line 19: How about syntax `github.com/example/plugin3@v1.2.3` instead of splitting into two properties? This is what `go install` and `go run` [already use](https://go.dev/ref/mod#go-install).

silverwind: A local `path` is easily detectable when it starts with `/` or `./`. A remote path can be part of the specifier.

I don't mean to recommend `go install` but to use the common syntax among go tooling.

silverwind: `go.mod` does indeed not use the new syntax, but I guess it may do so eventually. Having a single string to specify a module is very beneficial to the whole ecosystem.

ldez: We need the module name and the path.

I don't share your opinion about `go.mod` and the CLI syntax.

silverwind: Well, then complicate it however you like.

By the way, `version` is technically a incorrect term because it's a git tag that not necessarily has to hold a version. I hope it supports `latest`, which the go 1.16 syntax does.

ldez: > Well, then complicate it however you like.

Please.

> By the way, version is technically a incorrect term because it's a git tag that not necessarily has to hold a version.

Technically speaking, it's not a git tag because `latest` is not a git tag.
Version is a generic term.

ldez:
It's time to merge this PR :tada:

docs/src/docs/plugins/module-plugins.mdx
```diff
@@ -0,0 +1,71 @@
+---
+title: Module Plugin System
+---
+
+An example linter can be found at [here](https://github.com/golangci/example-plugin-module-linter/settings).
```
Antonboom, line 5: 404 (no settings dir)

Antonboom (commented):
Sorry for noise 🙏 
Cool job.

pkg/commands/custom.go
```diff
@@ -0,0 +1,81 @@
...
+	ctx := context.Background()
+
+	tmp, err := os.MkdirTemp(os.TempDir(), "custom-gcl")
+	if err != nil {
+		return fmt.Errorf("create temporary directory: %w", err)
+	}
+
+	defer func() {
+		if os.Getenv(envKeepTempFiles) != "" {
+			log.Printf("WARN: The env var %s has been dectected: the temporary directory is preserved: %s", envKeepTempFiles, tmp)
```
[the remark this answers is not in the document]
ldez: already fixed

pkg/commands/internal/builder.go
```diff
@@ -0,0 +1,219 @@
...
+		return fmt.Errorf("move golangci-lint binary: %w", err)
+	}
+
+	return nil
+}
+
+func (b Builder) clone(ctx context.Context) error {
+	//nolint:gosec // the variable is sanitized.
+	cmd := exec.CommandContext(ctx,
+		"git", "clone", "--branch", sanitizeVersion(b.cfg.Version),
```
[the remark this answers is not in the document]
ldez: I documented the requirements https://golangci-lint.run/plugins/module-plugins/#the-automatic-way (`go` and `git`).

The `git` and the fact that uses modules seem enough clear.

pkg/commands/internal/configuration.go
```diff
@@ -0,0 +1,138 @@
...
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+)
+
+const base = ".custom-gcl"
```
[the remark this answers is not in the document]
ldez: Because someone named his linter `gci`... so I tried to find a non-confusing name.

pkg/commands/internal/configuration.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// Import to use.
+	Import string `yaml:"import,omitempty"`
+
+	// Version of the module.
+	// Only for module available through a Go proxy.
+	Version string `yaml:"version,omitempty"`
+
+	// Path to the local module.
+	// Only for local module.
```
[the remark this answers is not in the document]
ldez: it can be relative or absolute.

docs/src/docs/plugins/module-plugins.mdx
```diff
@@ -0,0 +1,71 @@
...
+  enable:
+    - foo
+```
+
+## The Manual Way
+
+- add a blank-import of your module inside `cmd/golangci-lint/plugins.go`
+- run `go mod tidy`. (the module containing the plugin will be imported)
+- run `make build`
+- define the plugin inside the configuration `linters-settings.custom` section with the type `module`.
```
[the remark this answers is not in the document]
ldez: already fixed

docs/src/docs/plugins/module-plugins.mdx
ldez: https://github.com/golangci/golangci-lint/pull/4508

ccoVeille:
I just found you implemented this @ldez, it's amazing

it will change the way we had to work with the home-made plugin.

We might have to rewrite a lot of things, but at least we will avoid the iterative fix we had

You make my day @ldez 

Thanks !!!

I will have a look at https://github.com/golangci/example-plugin-module-linter

PR #4439, opened 2024-03-02
fix: move show-stats field from run to output

ldez:
During the review of the PR about `show-stats` I missed that the value was not in the right structure.

This PR deprecated the old `run.show-stats`, and replaced it with `output.show-stats`.

The flag `--show-stats` is not impacted, it's only about the file configuration.

The old `run.show-stats` still works but with a deprecation log message.

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -397,7 +397,11 @@ func (c *runCommand) setExitCodeIfIssuesFound(issues []result.Issue) {
 }
 
 func (c *runCommand) printStats(issues []result.Issue) {
-	if !c.cfg.Run.ShowStats {
+	if c.cfg.Run.ShowStats {
+		c.log.Warnf("run.show-stats is deprecated, please use output.show-stats")
+	}
+
+	if !c.cfg.Run.ShowStats && !c.cfg.Output.ShowStats {
 		return
 	}
 
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -398,7 +398,7 @@ func (c *runCommand) setExitCodeIfIssuesFound(issues []result.Issue) {
 
 func (c *runCommand) printStats(issues []result.Issue) {
 	if c.cfg.Run.ShowStats {
-		c.log.Warnf("run.show-stats is deprecated, please use output.show-stats")
+		c.log.Warnf("The configuration option `run.show-stats` is deprecated, please use `output.show-stats`")
 	}
 
 	if !c.cfg.Run.ShowStats && !c.cfg.Output.ShowStats {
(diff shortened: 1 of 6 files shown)


pkg/config/run.go
```diff
@@ -23,7 +23,7 @@ type Run struct {
 	AllowParallelRunners bool `mapstructure:"allow-parallel-runners"`
 	AllowSerialRunners   bool `mapstructure:"allow-serial-runners"`
 
-	ShowStats bool `mapstructure:"show-stats"`
+	ShowStats bool `mapstructure:"show-stats"` // Deprecated use Output.ShowStats instead.
```
alexandear, line 26:
```suggestion
	// Deprecated: use Output.ShowStats instead.
	ShowStats bool `mapstructure:"show-stats"`
```

Did you intentionally write the [deprecated comment](https://go.dev/wiki/Deprecated) incorrectly to prevent `staticcheck` from raising the issue `SA1019`?

pkg/commands/run.go
```diff
@@ -466,7 +466,11 @@ func (c *runCommand) createPrinter(format string, w io.Writer) (printers.Printer
 }
 
 func (c *runCommand) printStats(issues []result.Issue) {
-	if !c.cfg.Run.ShowStats {
+	if c.cfg.Run.ShowStats {
+		c.log.Warnf("run.show-stats is deprecated, please use output.show-stats")
```
alexandear, line 470: Maybe it makes sense to add the prefix "The config option " for better clarity and embrace with backticks for consistency with this [warning](https://github.com/golangci/golangci-lint/blob/b14d05cdb451c9b51c6705401ffad5c7a6d11a46/pkg/golinters/unparam.go#L52C1-L53C1):

So, we get the following warning message:
image

alexandear (approved):

pkg/config/run.go
ldez: It was just a mistake.

pkg/commands/run.go
```diff
@@ -397,7 +397,11 @@ func (c *runCommand) setExitCodeIfIssuesFound(issues []result.Issue) {
 }
 
 func (c *runCommand) printStats(issues []result.Issue) {
-	if !c.cfg.Run.ShowStats {
+	if c.cfg.Run.ShowStats {
+		c.log.Warnf("The configuration option `run.show-stats` is deprecated, please use `output.show-stats`")
+	}
+
+	if !c.cfg.Run.ShowStats && !c.cfg.Output.ShowStats {
```
Antonboom, line 404: de morgan cosmetic:

```go
if !(c.cfg.Run.ShowStats || c.cfg.Output.ShowStats) {
```

PR #4440, opened 2024-03-02
dev: split post-release workflow

ldez:
The post-release workflow does 2 things:
- generate contributors list for the readme
- generate GitHub Action assets

And then create a PR with all the generated content.

The generation of the contributor list takes more than 1 hour and can fail.
https://github.com/golangci/golangci-lint/actions/workflows/post-release.yml

The generation of GitHub Action assets is really quick, and those assets are required to be able to install the new version from GitHub Action.

Each time it fails, I'm forced to do the work by hand.

I changed the workflow
- to not fail on a failure of the update of the readme.
- to create 2 PRs: one for the readme, and one for the assets.

So even if the readme generation fails, the assets are updated.

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -5,7 +5,46 @@ on:
       - published
 
 jobs:
-  update-docs-and-assets:
+  update-docs:
+    name: "Update readme"
+    continue-on-error: true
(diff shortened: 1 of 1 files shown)


bombsimon (approved):
> The generation of the contributor list takes more than 1 hour and can fail.

🤦

PR #4457, opened 2024-03-04
docs: use 2 references files

ldez:
use 2 references files:
- one for the reference of the released version (`.golangci.reference.yml`)
- one for the next release (`.golangci.next.reference.yml`)

`.golangci.reference.yml` has been restored from v1.56.2 tag.

`.golangci.next.reference.yml` contains the new elements related to the next release.

I working on a workflow to update automatically `.golangci.reference.yml` with `.golangci.next.reference.yml` after/before a release.


.github/workflows/post-release.yml
```diff
@@ -62,12 +62,20 @@ jobs:
       - name: Update GitHub action config
         run: make assets/github-action-config.json
 
+      - name: Update reference files
+        run: |
+          rm .golangci.reference.yml
+          cp .golangci.next.reference.yml .golangci.reference.yml
```
alexandear, line 68: Was `rm + cp` intentionally used instead of just `mv`?"

```suggestion
          mv .golangci.next.reference.yml .golangci.reference.yml
```

scripts/website/dump_info/main.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	defer func() { _ = file.Close() }()
+
+	encoder := json.NewEncoder(file)
+	encoder.SetIndent("", "  ")
+
+	err = encoder.Encode(data)
+	if err != nil {
+		return fmt.Errorf("JSON encode (%s): %w", dst, err)
```
alexandear, line 87: Let's avoid starting the error message with  [capital letter](https://go.dev/wiki/CodeReviewComments#error-strings):
```suggestion
		return fmt.Errorf("encode JSON (%s): %w", dst, err)
```

scripts/website/expand_templates/exclusions.go
```diff
@@ -0,0 +1,30 @@
...
+	"path/filepath"
+	"strings"
+
+	"github.com/golangci/golangci-lint/pkg/config"
+)
+
+func getDefaultExclusions() string {
+	defaultExcludePatterns, err := readJSONFile[[]config.ExcludePattern](filepath.Join("assets", "default-exclusions.json"))
+	if err != nil {
+		panic(err)
```
alexandear, line 15: Maybe it would be better for `getDefaultExclusions()` to return an error instead of causing a panic? After all, `getDefaultExclusions()` is invoked within `buildTemplateContext()`, which is designed to handle errors.

scripts/website/expand_templates/linters.go
```diff
@@ -0,0 +1,315 @@
...
+
+	snippets.ConfigurationFile = overview + builder.String()
+
+	return &snippets, nil
+}
+
+func getLintersSettingSections(node, nextNode *yaml.Node) (string, error) {
+	linters, err := readJSONFile[[]*shared.LinterWrapper](filepath.Join("assets", "linters-info.json"))
+	if err != nil {
+		panic(err)
```
alexandear, line 241: Can be:
```suggestion
		return "", err
```

scripts/website/dump_info/main.go
```diff
@@ -0,0 +1,91 @@
...
+	linters := lintersdb.NewLinterBuilder().Build(config.NewDefault())
+
+	var wraps []shared.LinterWrapper
+	for _, l := range linters {
+		wraps = append(wraps, shared.LinterWrapper{Config: l, Name: l.Linter.Name(), Desc: l.Linter.Desc()})
+	}
+
+	err := saveToJSONFile(filepath.Join("assets", "linters-info.json"), wraps)
+	if err != nil {
+		log.Fatalf("save linters: %v", err)
```
alexandear, line 27: I propose to be consistent and start `log` messages from the capital letter:

```suggestion
		log.Fatalf("Save linters: %v", err)
```

The comment applies to log messages in the whole file.

scripts/website/shared/shared.go
```diff
@@ -0,0 +1,16 @@
+package shared
```
alexandear, line 1: Could we consider a more descriptive package name? According to the Go blog on [package names](https://go.dev/blog/package-names), `shared` might be perceived as somewhat generic.

scripts/website/expand_templates/linters.go
ldez: In this context, an error or panic is the same thing.
This method is used to fill a map form the template, it's easier to not have to handle error.

.golangci.next.reference.yml
mateusoliveira43, line 1: what the idea behind this?

the goal is to have the latest release reference (in the old file) and master reference (this file)?

Is not easier to add a link in documentation to `.golangci.reference.yml` in latest release? In this case, this https://github.com/golangci/golangci-lint/blob/v1.56.2/.golangci.reference.yml

ldez: The explanation is inside the description:

> If we need to fix something in the current doc, we should update the file .golangci.reference.yml.
> If we need to add new options or linters, we should update the file .golangci.next.reference.yml.

mateusoliveira43: what you believe can be done in this category "If we need to fix something in the current doc, we should update the file .golangci.reference.yml."?

ldez: I know where you would like to go.
I think about the comments, the validation will not change anything here.

bombsimon (approved):
I hope I didn't miss anything but this seems to simply be a split of the old code in a single `main.go` file to several files and the big diff (additions) is due to the copy of the reference and the new JSON files?

If so, feel free to ignore my comments about any code, this LGTM. Nice with a file that targets the latest release separately!

scripts/website/dump_info/main.go
```diff
@@ -0,0 +1,133 @@
...
+
+	err = saveDefaultExclusions()
+	if err != nil {
+		log.Fatalf("Save default exclusions: %v", err)
+	}
+
+	err = saveCLIHelp(filepath.Join("assets", "cli-help.json"))
+	if err != nil {
+		log.Fatalf("Save CLI help: %v", err)
+	}
```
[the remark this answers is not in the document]
ldez: I'm not a fan of this kind of "oneliner". Sorry but I will not do that :smile_cat:

scripts/website/expand_templates/exclusions.go
```diff
@@ -0,0 +1,30 @@
...
+	bufferString := bytes.NewBufferString("")
+
+	for _, pattern := range defaultExcludePatterns {
+		_, _ = fmt.Fprintln(bufferString)
+		_, _ = fmt.Fprintf(bufferString, "### %s\n", pattern.ID)
+		_, _ = fmt.Fprintln(bufferString)
+		_, _ = fmt.Fprintf(bufferString, "- linter: `%s`\n", pattern.Linter)
+		_, _ = fmt.Fprintf(bufferString, "- pattern: `%s`\n", strings.ReplaceAll(pattern.Pattern, "`", "&grave;"))
+		_, _ = fmt.Fprintf(bufferString, "- why: %s\n", pattern.Why)
+	}
```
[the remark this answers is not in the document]
bombsimon: Yes, there were a bunch of `_, _ =` which just felt like noise and decreases readability. But as I wrote in my comment I realised later that this was just moved code and wrote you can discard my comments.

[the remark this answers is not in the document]
ldez: I replaced this section with a template.

.github/workflows/post-release.yml
ldez: I think it's because I have a zsh (default?) `cp` alias: 
```console
$ alias        
cp='cp -i'
```

ldez:
> the big diff (additions) is due to the copy of the reference and the new JSON files?

yes exactly

scripts/website/expand_templates/main.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	helps, err := readJSONFile[types.CLIHelp](filepath.Join("assets", "cli-help.json"))
+	if err != nil {
+		return nil, err
+	}
+
+	changeLog, err := os.ReadFile("CHANGELOG.md")
+	if err != nil {
+		return nil, err
```
Antonboom, line 139: fmt.Errorf

scripts/website/expand_templates/main.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	snippets, err := extractExampleSnippets(golangciYamlExample)
+	if err != nil {
+		return nil, fmt.Errorf("can't read .golangci.reference.yml: %w", err)
+	}
+
+	helps, err := readJSONFile[types.CLIHelp](filepath.Join("assets", "cli-help.json"))
+	if err != nil {
+		return nil, err
```
Antonboom, line 134: fmt.Errorf

scripts/website/expand_templates/main.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	changeLog, err := os.ReadFile("CHANGELOG.md")
+	if err != nil {
+		return nil, err
+	}
+
+	latestVersion, err := getLatestVersion()
+	if err != nil {
+		return nil, fmt.Errorf("failed to get the latest version: %w", err)
```
Antonboom, line 144: inconsistency: `failed to`, `can't `, no verb (as below)

scripts/website/expand_templates/linters.go
```diff
@@ -0,0 +1,315 @@
...
+
+	snippets.ConfigurationFile = overview + builder.String()
+
+	return &snippets, nil
+}
+
+func getLintersSettingSections(node, nextNode *yaml.Node) (string, error) {
+	linters, err := readJSONFile[[]*types.LinterWrapper](filepath.Join("assets", "linters-info.json"))
+	if err != nil {
+		return "", err
```
Antonboom, line 241: fmt.Errorf

Antonboom (commented):
a lot of code 💥 
thank you for your job!

scripts/website/expand_templates/linters.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+
+	return formatDesc(desc)
+}
+
+func formatDesc(desc string) string {
+	runes := []rune(desc)
+
+	r, _ := utf8.DecodeRuneInString(desc)
+	runes[0] = unicode.ToUpper(r)
```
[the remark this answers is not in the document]
ldez: The description is required, so no problem

scripts/website/expand_templates/linters.go
```diff
@@ -0,0 +1,315 @@
...
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/scripts/website/types"
+)
+
+const listItemPrefix = "list-item-"
+
+func getLintersListMarkdown(enabled bool) string {
+	linters, err := readJSONFile[[]*types.LinterWrapper](filepath.Join("assets", "linters-info.json"))
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
ldez: yes it's a script, not an app.

PR #4469, opened 2024-03-08
fix: rule severity is required and different than the default one

ldez:
```yml
severity:
  default-severity: error
  rules:
    - linters:
        - lll
      severity: warning
```

The `default-severity` applies to all linters, the rules are optional.

Then a rule can be sensed only if defining a severity and if the severity is different than the default.


pkg/config/severity.go
```diff
@@ -20,6 +20,11 @@ func (s *Severity) Validate() error {
 	}
 
 	for i, rule := range s.Rules {
+		if s.Default == rule.Severity {
+			return fmt.Errorf("error in severity rule #%d: the default severity (%q) is the same as the rule severity, this rule can be removed",
+				i, s.Default)
+		}
```
bombsimon, line 26: Why do we take the responsibility to warn about this? I think this might just be a preference and can be used to be explicit. It can also be used to represent that a specific linter should always have a specific severity, not necessarily the same as default - they just happened to be the same at the point of writing. Meaning avoiding faulty configuration by mistake if default severity ever changes.

And if I understand it correctly you can't even just slap a comment on the severity saying that it should always be a specific value if the default change because the severity is required.

Pretty minor thing so I won't block on it but it doesn't feel like a job for `golangci-lint` as long as the configuration is valid, working and gives the expected result.

bombsimon (approved):

pkg/config/severity.go
ldez: I was thinking that using the same severity for the rule and the default was related to misunderstanding of the severity system.
But your argument about documentation is valid.
I remove this constraint.

PR #4470, opened 2024-03-08
feat: use gosec severities

ldez:
gosec has severities, then we can use them as value.

### DIFF
--- a/pkg/golinters/gosec.go
+++ b/pkg/golinters/gosec.go
@@ -97,7 +97,7 @@ func runGoSec(lintCtx *linter.Context, pass *analysis.Pass, settings *config.GoS
 
 	issues := make([]goanalysis.Issue, 0, len(secIssues))
 	for _, i := range secIssues {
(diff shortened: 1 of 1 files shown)


bombsimon (approved):

pkg/golinters/gosec.go
```diff
@@ -149,6 +150,19 @@ func toGosecConfig(settings *config.GoSecSettings) gosec.Config {
...
 
+func convertScoreToString(score issue.Score) string {
+	switch score {
+	case issue.Low:
+		return "low"
+	case issue.Medium:
+		return "medium"
+	case issue.High:
+		return "high"
+	default:
```
Antonboom, line 161: enable in project `exhaustive` linter? 🤔

Antonboom (commented):
👍

pkg/golinters/gosec.go
ldez: It's a false positive producer, I'm sure that is a good idea.

Antonboom: not sure* 
? :)

PR #4475, opened 2024-03-09
Revise fallback on Go1.18

sivchari:
go option of [run configuration](https://golangci-lint.run/usage/configuration/) describes that if both go.mod file and GOVERSION doesn't exists, it fallbacks on Go1.18. So I've modified it.

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -86,7 +86,7 @@ func detectGoVersion() string {
 		return v
 	}
 
-	return "1.17"
+	return "1.18"
 }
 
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

>  # Default: use Go version from the go.mod file, fallback on the env var `GOVERSION`, fallback on 1.17

https://golangci-lint.run/usage/configuration/#run-configuration

The documentation says that we fallback on 1.17

sivchari:
Sorry, I saw wrong.

PR #4479, opened 2024-03-10
dev: the printer just needs Output configuration

ldez:
the printer doesn't need the full configuration, the Output configuration is enough.

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -186,7 +186,7 @@ func (c *runCommand) preRunE(_ *cobra.Command, _ []string) error {
 
 	c.dbManager = dbManager
 
-	printer, err := printers.NewPrinter(c.log, c.cfg, c.reportData)
+	printer, err := printers.NewPrinter(c.log, &c.cfg.Output, c.reportData)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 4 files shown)


alexandear (approved):

pkg/commands/run.go
```diff
@@ -327,11 +328,24 @@ func (c *runCommand) runAndPrint(ctx context.Context, args []string) error {
...
+	c.printDeprecatedLinterMessages(enabledLintersMap)
+
 	issues, err := c.runAnalysis(ctx, args)
 	if err != nil {
 		return err // XXX: don't lose type
 	}
 
+	// Fills linters information for the JSON printer.
+	for _, lc := range c.dbManager.GetAllSupportedLinterConfigs() {
+		isEnabled := enabledLintersMap[lc.Name()] != nil
```
Antonboom, line 345: more go way?
```
_, isEnabled := enabledLintersMap[lc.Name()]
```

Antonboom (commented):
👍

pkg/commands/run.go
ldez: it was existing code.

ldez: It's not the same thing: the key can exist but the value can be nil.

PR #4481, opened 2024-03-10
dev: remove unused release target

ldez:
It was introduced here #724 (when Travis was the CI).

At some point, I thought this could be useful as a fallback, but now I know we will never use it.


bombsimon (approved):
I see 🔴, I approve ✅ 

image

PR #4492, opened 2024-03-12
dev: minor clean up

ldez:
Related to @Antonboom reviews.

### DIFF
--- a/cmd/golangci-lint/main.go
+++ b/cmd/golangci-lint/main.go
@@ -35,41 +35,44 @@ func createBuildInfo() commands.BuildInfo {
 		Date:      date,
 	}
 
-	if buildInfo, available := debug.ReadBuildInfo(); available {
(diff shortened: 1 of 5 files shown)


cmd/golangci-lint/main.go
```diff
@@ -35,41 +35,44 @@ func createBuildInfo() commands.BuildInfo {
...
 
-			if revision == "" {
-				revision = "unknown"
-			}
+	info.GoVersion = buildInfo.GoVersion
 
-			if modified == "" {
-				modified = "?"
-			}
+	if date == "" {
```
bombsimon, line 45: Can we use early return here as well?

```go
if date != "" {
    return info
}

// Non indented code
```

alexandear (approved):

bombsimon (approved):

Antonboom (commented):
Thank u

Issue #4360, opened 2024-02-08
Option to exclude folders from scanning


falco467:
### Your feature request related to a problem? Please describe.

`golang-lint-ci --fast` is slowed down, because it traverses all directories (even unrelated big directories like a node_modules folder) and there is no option to exclude unrelated folders from scanning/traversal.

### Describe the solution you'd like.

Provide an option to ignore/exclude folders from traversal/scanning.

### Describe alternatives you've considered.

I tried excluding the folders with the `--skip-dirs` flag, but golang-lint-ci will still traverse the subfolders and take a lot of time on each run.

### Additional context.

I'm using golang-lint-ci as my integrated linter in VS Code (running on file-save), fast execution time is critical. I measured performance with --verbose and `Go packages loading at mode 575` takes up the most time (even with caching enabled) it takes almost 9 seconds every run, even though my project has only ~1000 Lines of Go-Code in 6 files.

We have the frontend project in a subfolder of the go project (/frontend) which includes a huge node_modules folder. Deleting this folder speeds up the scan from 12 to 4 seconds. Is there any option to hide this folder from golang-lint-ci ?

falco467:
This could also probably be fixed by caching. On multiple runs `golang-lint-ci` could recognize by the modification time on the folder that nothing has changed inside the folder and should not need to scan anything inside the folder. I'm working with go build cache and golang-lint-ci cache. But it is still scanning all folders again even when I don't change any file between runs.

ldez:
Hello,

A package, if it's used by another part of the code, cannot be skipped.

falco467:
> Hello,
> 
> A package, if it's used by another part of the code, cannot be skipped.
> 
> Related to #4217

This is not about skipping packages, but skipping unrelated Folders without any .go files in it (like a folder with PDFs or JS files)

falco467:
Here is the output which illustrates the problem:

```
golangci-lint run --verbose
level=info msg="[loader] Go packages loading at mode 575 (types_sizes|files|imports|name|compiled_files|deps|exports_file) took 10.0107529s"
level=info msg="Execution took 12.4308233s"
```
After deleting the big folder without any .go files:
```
golangci-lint run --verbose
level=info msg="[loader] Go packages loading at mode 575 (compiled_files|imports|name|types_sizes|deps|exports_file|files) took 2.3652813s"
level=info msg="Execution took 4.488697s"
```

marcoshuck:
[...]

Issue #4084, opened 2023-09-14
Should skip_dirs skip analyzing, not just reporting?


alexbozhenko:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've read the typecheck section of the FAQ (https://golangci-lint.run/usage/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

Hi.
It seems like that presence of some invalid code in thrid_party breaks the golangci-lint (see reproducer below).
Would it be possible to change the behavior of skip_dirs to not even look at those dirs?

### Version of golangci-lint

<details>

```console
golangci-lint --version
golangci-lint has version v1.54.2 built with go1.21.0 from (unknown, mod sum: "h1:oR9zxfWYxt7hFqk6+fw6Enr+E7F0SN2nqHhJYyIb0yo=") on (unknown)
```

</details>

### Configuration

Fails even with empty config.

### Go environment

<details>

```console
$ go version && go env
go version go1.21.0 linux/amd64
GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/abozhenko/.cache/go-build'
GOENV='/home/abozhenko/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/abozhenko/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/abozhenko/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.21.0'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='gcc'
CXX='g++'
CGO_ENABLED='1'
GOMOD='/home/abozhenko/reproducer/go.mod'
GOWORK=''
CGO_CFLAGS='-O2 -g'
[...]

Issue #3775, opened 2023-04-11
skip-dirs configuration field ot --skip-dirs not working


axetroy:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc.).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

I set skip-dirs in the configuration file.

But It still lint thie file in the dir.

See the configuration file below.

`ent` is a generated folder, so I set it to configuration file and `--skip-dirs`. none of one works

ref: #3711

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version 1.52.2 built with go1.20.2 from da04413a on 2023-03-25T18:11:28Z
```

</details>

### Configuration file

<details>

```console
$ cat .golangci.yml
linters:
  timeout: 10m
  skip-dirs:
    - ent
  enable:
    - nilerr
  modules-download-mode: vendor

issues:
  fix: true
```

</details>

### Go environment

<details>

```console
$ go version && go env
go version go1.20 darwin/amd64
GO111MODULE="on"
GOARCH="amd64"
GOBIN=""
GOCACHE="/Users/axetroy/Library/Caches/go-build"
GOENV="/Users/axetroy/Library/Application Support/go/env"
GOEXE=""
GOEXPERIMENT=""
GOFLAGS=""
GOHOSTARCH="amd64"
GOHOSTOS="darwin"
GOINSECURE=""
GOMODCACHE="/Users/axetroy/go/pkg/mod"
GONOPROXY=""
GONOSUMDB=""
GOOS="darwin"
GOPATH="/Users/axetroy/go"
GOPRIVATE=""
GOPROXY="https://goproxy.io,direct"
GOROOT="/usr/local/go"
GOSUMDB="sum.golang.org"
GOTMPDIR=""
GOTOOLDIR="/usr/local/go/pkg/tool/darwin_amd64"
GOVCS=""
GOVERSION="go1.20"
GCCGO="gccgo"
GOAMD64="v1"
AR="ar"
CC="clang"
CXX="clang++"
CGO_ENABLED="1"
GOMOD="/dev/null"
GOWORK=""
CGO_CFLAGS="-O2 -g"
CGO_CPPFLAGS=""
[...]

Issue #2748, opened 2022-04-06
skip-dirs not skipping the file


mayank558:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [X] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

### Description of the problem

I am trying to skip some directories in project from golangci-lint but using the skip-dirs is not helping

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
# golangci-lint has version 1.43.0 built from 861262b on 2021-11-02T20:54:42Z

```

</details>

### Configuration file

<details>

```console
$ cat .golangci.yml
run:
  # which dirs to skip: issues from them won't be reported;
  # can use regexp here: generated.*, regexp is applied on full path;
  # default value is empty list, but default dirs are skipped independently
  # from this option's value (see skip-dirs-use-default).
  skip-dirs:
    - <path to the file>
```

</details>

### Go environment

<details>

```console
$ go version && go env
go version go1.17.5 darwin/arm64
```

</details>

### Verbose output of running

<details>

```console
$ golangci-lint cache clean
$ golangci-lint run -v
# paste output here
```

</details>

### Code example or link to a public repository

<details>

```go
run:
  # which dirs to skip: issues from them won't be reported;
  # can use regexp here: generated.*, regexp is applied on full path;
  # default value is empty list, but default dirs are skipped independently
  # from this option's value (see skip-dirs-use-default).
  skip-dirs:
    - <path to the file>
```

</details>

ldez:
duplicate of #1832

PR #4509, opened 2024-03-15
feat: replace `run.skip-xxx` options by `issues.exclude-xxx` options

ldez:
The option name `skip-xxx` is a source of confusion.
And to increase the confusion, those options are inside the section `run`.

This sample of issues illustrate the problem: #1912, #2748, #4084, #3775, #1832, #4360

My proposal is move and rename those options:
- `run.skip-dirs` deprecation: replaced by `issues.exclude-dirs`
- `run.skip-dirs-use-default` deprecation: replaced by  `issues.exclude-dirs-use-default`
- `run.skip-files` deprecation: replaced by `issues.exclude-files`

This is a non-breaking change: the `skip-xxx` will still work but with a warning, and `issues.exclude-xxx` options do the same things as `skip-xxx`.

<details><summary>log output</summary>

Using a flag:
```console
$ go run ./cmd/golangci-lint run --skip-dirs=foo
Flag --skip-dirs has been deprecated, check the documentation for more information.
WARN [config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`. 
```

Using the file:
```console
$ go run ./cmd/golangci-lint run
WARN [config_reader] The configuration option `run.skip-dirs` is deprecated, please use `issues.exclude-dirs`. 
```

</details>

one commit, one change.


.golangci.next.reference.yml
```diff
@@ -2810,10 +2785,35 @@ issues:
...
+  # "/" will be replaced by current OS file path separator to properly work on Windows.
+  # Default: []
+  exclude-dirs:
+    - src/external_libs
+    - autogenerated_by_my_lib
+
+  # Enables exclude of directories:
+  # - vendor$, third_party$, testdata$, examples$, Godeps$, builtin$
+  # Default: true
+  exclude-dirs-use-default: false
```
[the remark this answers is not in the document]
ldez: The name `exclude-dirs-use-default` is related to the other option `exclude-dirs`.
So I think it's better to keep the current pattern.

.golangci.next.reference.yml
```diff
@@ -2810,10 +2785,35 @@ issues:
...
   # Default: false
   exclude-case-sensitive: false
 
+  # Which dirs to exclude: issues from them won't be reported.
+  # Can use regexp here: `generated.*`, regexp is applied on full path,
+  # including the path prefix if one is set.
+  # Default dirs are skipped independently of this option's value (see exclude-dirs-use-default).
+  # "/" will be replaced by current OS file path separator to properly work on Windows.
+  # Default: []
+  exclude-dirs:
```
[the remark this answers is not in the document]
ldez: It's 2 different processors. Their implementations are completely different.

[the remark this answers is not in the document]
alexandear: Yes, you're right about the different implementation.

When suggesting `exclude-paths`, I'm thinking from the user's perspective.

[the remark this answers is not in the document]
ldez: For now, I think it's easier for users to keep the same options during this deprecation cycle.

pkg/commands/flagsets.go
```diff
@@ -47,9 +48,13 @@ func setupRunFlagSet(v *viper.Viper, fs *pflag.FlagSet) {
...
 
 	internal.AddFlagAndBind(v, fs, fs.Bool, "tests", "run.tests", true, color.GreenString("Analyze tests (*_test.go)"))
+
+	fs.StringSlice("skip-files", nil, color.GreenString("Regexps of files to skip")) // Hack see Loader.applyStringSliceHack
+	deprecateFlag(fs, "skip-files")
 	fs.StringSlice("skip-dirs", nil, color.GreenString("Regexps of directories to skip")) // Hack see Loader.applyStringSliceHack
+	deprecateFlag(fs, "skip-dirs")
 	internal.AddFlagAndBind(v, fs, fs.Bool, "skip-dirs-use-default", "run.skip-dirs-use-default", true, getDefaultDirectoryExcludeHelp())
-	fs.StringSlice("skip-files", nil, color.GreenString("Regexps of files to skip")) // Hack see Loader.applyStringSliceHack
+	deprecateFlag(fs, "skip-dirs-use-default")
```
alexandear, line 57: Perhaps the following is better because it reduces duplication of name:
```go
	internal.AddDeprecatedFlag(fs, fs.StringSlice, "skip-files", nil, color.GreenString("Regexps of files to skip"))
	internal.AddDeprecatedFlag(fs, fs.StringSlice, "skip-dirs", nil, color.GreenString("Regexps of directories to skip"))
	internal.AddDeprecatedFlagAndBind(v, fs, fs.Bool, "skip-dirs-use-default", "run.skip-dirs-use-default", true,
		getDefaultDirectoryExcludeHelp())
```

But we need to extend `pkg/commands/internal/vibra.go` with the following functions:
```
func AddDeprecatedFlag[T any](fs *pflag.FlagSet, pfn FlagFunc[T], name string, value T, usage string) {
	pfn(name, value, usage)
	deprecateFlag(fs, name)
}

func AddDeprecatedFlagAndBind[T any](v *viper.Viper, fs *pflag.FlagSet, pfn FlagFunc[T], name, bind string, value T, usage string) {
	AddFlagAndBind(v, fs, pfn, name, bind, value, usage)
	deprecateFlag(fs, name)
}

func deprecateFlag(fs *pflag.FlagSet, name string) {
	_ = fs.MarkHidden(name)
	_ = fs.MarkDeprecated(name, "check the documentation for more information.")
}
```

alexandear (approved):

pkg/commands/flagsets.go
ldez: ok for `AddDeprecatedFlagAndBind` but I don't want to factorize around "hacked" flags.

ldez: I found a difference approach that allows both: visible Hack and factorization.

.golangci.next.reference.yml
```diff
@@ -2810,10 +2785,35 @@ issues:
...
```
bombsimon, line 2798: I think moving this to issues is the key thing here and "exclude" makes more sense than "skip" in that context!

I wonder if "exclude" is the best word to use here? It still hints that something might not be run or processed. What do you feel about the term "ignore", `ignore-dirs` and `ignore-files`? I think for me that clearer signals that they will be processed and added as issues but they will be ignored. This is also in line with the documentation saying that issues won't be reported meaning the issues will be there (not excluded) but not reported (ignored).

ldez: the term "ignore" is also ambiguous and in the future, it can be more confusing: https://github.com/golang/go/issues/42965#issuecomment-1974089649

the term "exclude" is already used for the other options in this section, so it seems more appropriate.

jsonschema/golangci.next.jsonschema.json
```diff
@@ -422,33 +422,6 @@
           "default": [],
           "examples": [["mytag"]]
         },
-        "skip-dirs": {
```
bombsimon, line 425: There's a section in the JSON schema regarding deprecations: https://json-schema.org/draft/2019-09/json-schema-validation#rfc.section.9.3

Should we use that instead of just removing the configuration?

ldez: The deprecation doesn't work very well: there is no warning during validation, so people will still use them.

ldez: FYI we are using Draft-07 (as recommended by the schema store)

https://json-schema.org/draft-07/json-schema-validation

There is no `deprecated` properties.

ldez: FYI I tried `deprecated` and `deprecationMessage`, and there are no effects.

bombsimon (approved):

PR #4516, opened 2024-03-15
dev: split ContextLoader

ldez:
`ContextLoader` was an ambiguous name to hide that it was doing 2 things:
- load packages
- build a linter context

The PR splits `ContextLoader` into `ContextBuilder` and `PackageLoader`:
- `ContextBuilder` builds a linter context
- `PackageLoader` loads packages

There are no fundamental changes here, the goal is to ease maintenance.

Also, it removes the usage of `cfg.Run.Args`.

I will create another PR to remove completely the usage of `cfg.Run.Args`.

One commit, one change.


pkg/lint/package.go
```diff
@@ -0,0 +1,297 @@
...
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/exitcodes"
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis/load"
+	"github.com/golangci/golangci-lint/pkg/goutil"
+	"github.com/golangci/golangci-lint/pkg/lint/linter"
+	"github.com/golangci/golangci-lint/pkg/logutils"
+)
+
+// PackageLoader loads packages based on `golang.org/x/tools/go/packages.Load`.
```
alexandear, line 24: We can change to the following:
```suggestion
// PackageLoader loads packages based on [golang.org/x/tools/go/packages.Load].
```
to show active links in the doc:
image

pkg/lint/package.go
```diff
@@ -318,3 +186,96 @@ func (cl *ContextLoader) filterDuplicatePackages(pkgs []*packages.Package) []*pa
...
+	var buildFlags []string
+
+	if len(l.cfg.Run.BuildTags) != 0 {
+		// go help build
+		buildFlags = append(buildFlags, "-tags", strings.Join(l.cfg.Run.BuildTags, " "))
+		l.log.Infof("Using build tags: %v", l.cfg.Run.BuildTags)
+	}
+
+	if l.cfg.Run.ModulesDownloadMode != "" {
+		// go help modules
```
alexandear, line 243: Looks like we should move this comment right above the line in `Run.Validate`:
```
allowedMods := []string{"mod", "readonly", "vendor"}
```

Because it previously was there https://github.com/golangci/golangci-lint/blob/8f3027c631ffdf590af5d930af2c89726d1f16f3/pkg/lint/load.go#L137-L138.

alexandear (approved):

pkg/lint/package.go
ldez: `go help modules` is related to both: the values and the flag `-mod=`.

bombsimon (approved):

PR #4517, opened 2024-03-15
dev: review config path usage inside gocritic

ldez:
Since the previous refactor of the commands, the configuration is fully loaded before the build of the linters, so we don't need a closure.

The configuration is inside the linter context, we can use it instead of passing it as a parameter.


Antonboom (approved):
🔥

PR #4519, opened 2024-03-16
docs: add more tagliatelle examples

pershin-daniil:
Sync settings here and in https://github.com/ldez/tagliatelle?tab=readme-ov-file#as-a-golangci-lint-linter

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -2097,6 +2097,7 @@ linters-settings:
         json: camel
         yaml: camel
         xml: camel
+        toml: camel
         bson: camel
         avro: snake
         mapstructure: kebab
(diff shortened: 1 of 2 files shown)


ldez:
Hello,

The goal is not to be exhaustive because you can define all the tag names you want.

This is just an example.

pershin-daniil:
I've just came up with an issue that there is no info about `toml` in `golangci` docs but it is in readme.md. It is confusing.

Is there any place with all tags or with all possible settings?

ldez:
> Is there any place with all tags or with all possible settings?

It's impossible to do because everybody can create a custom tag name and new "official" tag names can appear.

The only setting with a limited set of values is the case.

ldez:
We can add the `toml` example but the exhaustivity is impossible, the tagliatelle doc and golangci-lint doc don't need to be synced.

ldez (approved):
LGTM

PR #4520, opened 2024-03-16
dev: validate test configurations

ldez:
Another step related to Schema JSON validation: checking the configuration used by tests.

The PR fixes the configuration but also fixes the JSON schema.

one commit, one change.


test/configuration_file_test.go
```diff
@@ -0,0 +1,115 @@
...
+	"strings"
+	"testing"
+
+	"github.com/santhosh-tekuri/jsonschema/v5"
+	"github.com/stretchr/testify/require"
+	"gopkg.in/yaml.v3"
+)
+
+func Test_validateTestConfigurationFiles(t *testing.T) {
+	err := validateTestConfigurationFiles("../jsonschema/golangci.next.jsonschema.json", ".")
```
Antonboom, line 18: does it depend on directory where `go test` was executed?

I remember 
```go
 _, filename, _, _ := runtime.Caller(0)
```
pattern for tests using related paths

test/configuration_file_test.go
```diff
@@ -0,0 +1,115 @@
...
+	require.NoError(t, err)
+}
+
+func validateTestConfigurationFiles(schemaPath, targetDir string) error {
+	schema, err := loadSchema(filepath.FromSlash(schemaPath))
+	if err != nil {
+		return err
```
Antonboom, line 25: no wrapping everywhere :(

Antonboom (approved):
👍

test/configuration_file_test.go
ldez: It depends on the test file position.

PR #4521, opened 2024-03-17
feat: new output.formats file configuration syntax

ldez:
This PR improves the syntax inside the configuration file to specify output formats.

The previous syntax has been defined when only the format type was an option.
However, the syntax has evolved to support output paths and multiple formats.

```yml
output:
  format: "checkstyle:report.xml,json:stdout,colored-line-number"
```

This is a flag syntax friendly but not a file syntax friendly.

I propose this new syntax:

```yml
output:
  formats:
    - format: checkstyle
      path: report.xml
    - format: json
      path: stdout
    - format: colored-line-number
```

The CLI flag keeps the same syntax.

```
--out-format=checkstyle:report.xml,json:stdout,colored-line-number
```

To keep this new option compatible with the GitHub Action, the CLI must override the configuration file.

The previous field `output.format` is deprecated and works as before but with a warning.

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -59,15 +59,22 @@ run:
 
 # output configuration options
 output:
+  # The formats used to render issues.
   # Format: colored-line-number|line-number|json|colored-tab|tab|checkstyle|code-climate|junit-xml|github-actions|teamcity
-  #
-  # Multiple can be specified by separating them by comma, output can be provided
-  # for each of them by separating format name and path by colon symbol.
   # Output path can be either `stdout`, `stderr` or path to the file to write to.
-  # Example: "checkstyle:report.xml,json:stdout,colored-line-number"
+  #
+  # For the CLI flag (`--out-format`), multiple formats can be specified by separating them by comma.
+  # The output can be specified for each of them by separating format name and path by colon symbol.
+  # Example: "--out-format=checkstyle:report.xml,json:stdout,colored-line-number"
+  # The CLI flag (`--out-format`) override the configuration file.
   #
   # Default: colored-line-number
-  format: json
+  formats:
+    - format: json
+      path: stderr
+    - format: checkstyle
+      path: report.xml
+    - format: colored-line-number
 
   # Print lines of code with issue.
   # Default: true
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -60,7 +60,7 @@ run:
 # output configuration options
 output:
   # The formats used to render issues.
-  # Format: colored-line-number|line-number|json|colored-tab|tab|checkstyle|code-climate|junit-xml|github-actions|teamcity
+  # Format: `colored-line-number`, `line-number`, `json`, `colored-tab`, `tab`, `checkstyle`, `code-climate`, `junit-xml`, `github-actions`, `teamcity`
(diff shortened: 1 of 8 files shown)


pkg/config/output.go
```diff
@@ -35,14 +35,15 @@ var OutFormats = []string{
...
 type Output struct {
-	Format          string   `mapstructure:"format"`
-	PrintIssuedLine bool     `mapstructure:"print-issued-lines"`
-	PrintLinterName bool     `mapstructure:"print-linter-name"`
-	UniqByLine      bool     `mapstructure:"uniq-by-line"`
-	SortResults     bool     `mapstructure:"sort-results"`
-	SortOrder       []string `mapstructure:"sort-order"`
-	PathPrefix      string   `mapstructure:"path-prefix"`
-	ShowStats       bool     `mapstructure:"show-stats"`
+	Format          string        `mapstructure:"format"` // Deprecated: use Formats instead.
```
bombsimon, line 38: Should we follow deprecation cycle and add in what version this got deprecated so we remember to remove it? Or are we planning to support this one a bit longer?

.golangci.next.reference.yml
```diff
@@ -59,15 +59,22 @@ run:
...
-  # Example: "checkstyle:report.xml,json:stdout,colored-line-number"
   #
-  # Default: colored-line-number
-  format: json
+  # For the CLI flag (`--out-format`), multiple formats can be specified by separating them by comma.
+  # The output can be specified for each of them by separating format name and path by colon symbol.
+  # Example: "--out-format=checkstyle:report.xml,json:stdout,colored-line-number"
+  # The CLI flag (`--out-format`) override the configuration file.
+  #
+  # Default: `format: colored-line-number` and `path: stdout`
```
bombsimon, line 71: Maybe a bit verbose but I think it's worth having a value that actually works by just uncomment or copy.

```suggestion
  # Default:
  # formats:
  #   - format: colored-line-number
  #   - path: stdout
```

bombsimon (approved):

pkg/config/output.go
ldez: We have only defined the deprecation cycle for the linters, for the options, I don't know what is the best deprecation cycle.
So we will keep it as long as needed, the implementation is simple to maintain.

pkg/config/output.go
```diff
@@ -64,5 +67,46 @@ func (o *Output) Validate() error {
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+type OutputFormat struct {
+	Format string `mapstructure:"format"`
+	Path   string `mapstructure:"filepath"`
```
Antonboom, line 82:
```suggestion
	Path   string `mapstructure:"path"`
```

pkg/config/output_test.go
```diff
@@ -81,6 +81,79 @@ func TestOutput_Validate_error(t *testing.T) {
...
+			desc: "only format",
+			settings: &OutputFormat{
+				Format: "json",
+			},
+		},
+		{
+			desc: "format and path",
+			settings: &OutputFormat{
+				Format: "json",
+				Path:   "./example.json",
```
Antonboom, line 123: \+

```
Path: "example.json",
```

```
Path: "/tmp/example.json" // or any other absolute path
````

Antonboom (approved):
👍

PR #4526, opened 2024-03-18
dev: run all commands on the CI

ldez:
Runs all golangci-lint commands.

The command `run` is tested by the other steps (`make test`).

### DIFF
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -130,3 +130,44 @@ jobs:
         env:
           # needed for github-action-config.json generation
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  # Note: the command `run` is tested by the previous steps (`make test`).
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yml
```diff
@@ -130,3 +130,44 @@ jobs:
         env:
           # needed for github-action-config.json generation
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+  # Note: the command `run` is tested by the previous steps (`make test`).
+  commands:
```
Antonboom, line 135: how to not forget to update this :D 

maybe we need bash magic like

1. run `golangci-lint 42`
2. parse `Available Commands:`
3. loop: run command from 2, parse `Available Commands` again
4. etc

But, of course, explicit is better than implicit 🤷‍♂️

Antonboom (approved):
👍

PR #4543, opened 2024-03-19
Refactor nolintlint handling: Enforce directive and linter list formatting

apradhan12:
The rationale for this PR is to enforce the formatting of the spacing in the nolint directive via the nolintlint linter. The syntax of a Go directive [should roughly conform to the following format](https://github.com/golangci/golangci-lint/issues/1658): ^[a-z]+:[a-z]+. For nolint directive comments that have leading whitespace, the nolintlint linter currently gives a warning of the form “directive `// nolint: lll // ignore line length warnings` should be written without leading space as `//nolint: lll // ignore line length warnings`", when it should actually suggest `//nolint:lll // ignore line length warnings` (with no space between nolint: and lll) instead. Broadly, this PR makes the nolintlint warnings more strictly conform to the syntax of a Go directive and avoids providing inaccurate suggestions such as the aforementioned. This should reduce confusion for developers and should make it easier to conform to the requirements of go fmt as of go1.19.

Change list:
- Remove `directiveWithOptionalLeadingSpace` field - leading spaces in the directive are [no longer allowed as of go1.19](https://github.com/golangci/golangci-lint/pull/3002), so the directive will now always be `//nolint`. It cannot be any directive other than `nolint` (e.g., `go:generate`) due to the guarantee provided by matching the comment against the `commentPattern` regular expression in the `Run` function. Therefore, we can just hardcode `//nolint` as the directive. (The directive, as the term is generally used in Go, *could* be something like `//nolint:lll` or `//nolint:all` with a colon and specific linter(s) after `nolint`, but the `directiveWithOptionalLeadingSpace` variable would not currently include the optional colon and anything following it, so for the purposes of this field, the directive will always be `//nolint`).
- Update the `fullDirectivePattern` to disallow spaces in the linter list. The updated regular expression also disallows spaces at the beginning of the string in order to avoid providing an error message that could potentially contain an inaccurate suggestion for how to update the `nolint` directive, as would have previously been done by the `NotMachine` issue. The regular expression continues to allow empty explanation comments because they will be caught by the `NoExplanation` issue if the `NeedsExplanation` flag is set.
- Unexport `BaseIssue` struct, since it should not be used outside nolintlint.go.
- Deprecate `NeedsMachineOnly` and remove all references in code.
- Remove the `NotMachine` issue, since it has been superseded by the `ParserError`, which gives a more accurate error message (specifically, it does not provide a suggestion for how to update the `nolint` directive that could potentially still have spaces in the linter list, but rather a more generic suggestion of the format of a well-formed `nolint` comment).
- Update `NoExplanation` details to provide a generic directive template rather than a specific suggestion; update unit tests accordingly. Similar to the previous bullet point, this avoids inadvertently providing an inaccurate error message.
- Remove unused `//nolint:gocritic` comments.
- Update test cases in the unit test file that should be marked as invalid due to whitespace before/between/after the list of linters.
- Update test cases that previously resulted in the `NotMachine` issue to reflect that they now result in the `ParseError` issue instead.
- Add test cases that would have previously erroneously resulted in `directiveWithOptionalLeadingSpace` values not equal to `//nolint` or `// nolint` (namely, `directiveWithOptionalLeadingSpace` values containing additional characters after the directive, such as `//nolint / description`).

### DIFF
--- a/pkg/golinters/nolintlint/nolintlint.go
+++ b/pkg/golinters/nolintlint/nolintlint.go
@@ -113,7 +113,7 @@ const (
 var commentPattern = regexp.MustCompile(`^//\s*(nolint)(:\s*[\w-]+\s*(?:,\s*[\w-]+\s*)*)?\b`)
 
 // matches a complete nolint directive
-var fullDirectivePattern = regexp.MustCompile(`^//\s*nolint(?::([\w-]+(?:,[\w-]+)*))?(?: (//.*))?\n?$`)
+var fullDirectivePattern = regexp.MustCompile(`^//nolint(?::([\w-]+(?:,[\w-]+)*))?(?: (//.*))?\n?$`)
 
 type Linter struct {
 	needs           Needs // indicates which linter checks to perform
--- a/pkg/golinters/nolintlint/nolintlint.go
+++ b/pkg/golinters/nolintlint/nolintlint.go
@@ -12,28 +12,24 @@ import (
 	"github.com/golangci/golangci-lint/pkg/result"
 )
 
-type BaseIssue struct {
-	fullDirective                     string
-	directiveWithOptionalLeadingSpace string
-	position                          token.Position
-	replacement                       *result.Replacement
+type baseIssue struct {
+	fullDirective string
+	position      token.Position
+	replacement   *result.Replacement
 }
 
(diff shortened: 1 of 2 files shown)


ldez:
FYI, the "lax" formatting is intentional not to break existing usage.

PR #4556, opened 2024-03-20
feat: Add `.config` directories to `configSearchPaths`

### LINKED ISSUE #4554: Support .config

golangci uses `.golangci.yml` file on top of project. This clutters root directory of projects.

I suggest to search for `.config/golangci.yml` too as described at https://github.com/pi0/config-dir

Using environment variable. This work on CI but is not easy for mates.

### Additional context.

_No response_

bombsimon:
Support for `.config` proposal from https://github.com/pi0/config-dir. While walking the tree, check if a directory named `.config` exist. If so, add it to `configSearchPaths` to include when searching for `.golangci.yaml`.

Fixes #4554

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -138,6 +138,13 @@ func (l *Loader) setupConfigFileSearch() {
 	for {
 		configSearchPaths = append(configSearchPaths, curDir)
 
+		dotConfig := filepath.Join(curDir, ".config")
+
+		info, err := os.Stat(dotConfig)
+		if err == nil && info.IsDir() {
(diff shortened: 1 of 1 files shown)


ldez:
Based on my comment [here](https://github.com/golangci/golangci-lint/issues/4554#issuecomment-2010896211), I will close the PR for now, more investigations are needed.

bombsimon:
It was mostly to not print a list with directories that doesn't exist when using `-v` rather than avoiding issues. 

Makes sense, lets not keep the PR open 👍🏼

PR #4567, opened 2024-03-24
feat: err113 analyzer name

ldez:
When a linter name is different from the analyzer name, the analyzer name is added to the report message.
In the context of err113, it's redundant.

<details><summary>Before</summary>

```console
$ golangci-lint run 
main.go:13:10: err113: do not compare errors directly "err == e", use "errors.Is(err, e)" instead (goerr113)
                return err == e
                       ^
```

</details>

<details><summary>With the PR</summary>

```console
$ golangci-lint run 
main.go:13:10: do not compare errors directly "err == e", use "errors.Is(err, e)" instead (err113)
                return err == e
                       ^
```

</details>


pkg/golinters/gomnd.go
```diff
@@ -9,10 +9,23 @@ import (
 )
 
 func NewGoMND(settings *config.GoMndSettings) *goanalysis.Linter {
+	// The constant is only used to force the analyzer name to use the same name as the linter.
+	// This is required to avoid displaying the analyzer name inside the issue text.
```
bombsimon, line 13: Not sure I follow the rationale here. Is it OK to change the output from `err113` because there are no settings but for linters with settings it would be confusing to configure it as `gomnd` but only see `(mnd)` in the output?

Since we use `analyzer.Name` for most linters, what happens if someone changes the name of their linter?

pkg/golinters/gomnd.go
```diff
@@ -9,10 +9,23 @@ import (
...
+	//
+	// Alternative names cannot help here because of the linter configuration that uses `gomnd` as a name.
+	// The complexity of handling alternative names at a lower level (i.e. `goanalysis.Linter`) isn't worth the cost.
+	// The only way to handle it properly is to deprecate and "duplicate" the linter and its configuration,
```
bombsimon, line 17: I think it would be enough to duplicate the configuration and pick the one that's non empty or the newest one. Then we can use the same deprecation cycle as with linters to show warnings if you use the `gomnd` key in the config.

I know they're public but I suspect we don't see `NewGoMND` as a public API (given that `NewErr113` was just changed so we could just accept multiple configuration:

```go
func NewGoMND(settings *config.MndSettings, deprecatedSettings *config.GoMndSettings) *goanalysis.Linter
```

pkg/golinters/gomnd.go
ldez: > Is it OK to change the output from err113 because there are no settings but for linters with settings it would be confusing to configure it as gomnd but only see (mnd) in the output?

The problem is not the output but the name of the linter inside `enable`/`disable` section and the name of the linter inside the configuration.

> Since we use analyzer.Name for most linters, what happens if someone changes the name of their linter?

First, you have to know that, currently, we have linters that are multiple analyzers (ex: govet)

If someone changes the name of an analyzer, we can inside golangci-lint, override it to keep the compatibility.

pkg/golinters/gomnd.go
ldez: In all cases, the public API of golangci-lint can be broken, because golangci-lint should not be used as a library.

I put quotes around `duplicate`, because, yes I was not suggesting to duplicate the code but to "register" a new linter with the same code.

ldez: In the line after, I said "for now, I don't know if it's worth the cost."

I feel like you thing it is worth the cost, so I will "duplicate" the linter.

ldez: I will split the PR into 2 PRs because the deprecation of gomnd is not the same topic as the simple rename of an analyzer. I don't feel comfortable having 2 differents topics inside the one PR.

ldez: https://github.com/golangci/golangci-lint/pull/4572

pkg/golinters/gomnd.go
bombsimon: Ah, right! Thanks for the explanation!

bombsimon (approved):

Antonboom (approved):

Antonboom (changes requested):
block this due to https://github.com/golangci/golangci-lint/pull/4572#discussion_r1537115847

alexandear (approved):

Antonboom (approved):

PR #4572, opened 2024-03-24
feat: replace gomnd by mnd

ldez:
Rename gomnd to mnd:
- Deprecate gomnd.
- Add mnd.

This PR also fixes the problem of consistency between the analyzer name and linter name like #4567.

<details><summary>Before</summary>

```console
$ golangci-lint run 
main.go:28:23: mnd: Magic number: 200, in <condition> detected (gomnd)
        if res.StatusCode != 200 {
                             ^
main.go:21:12: mnd: Magic number: 2, in <assign> detected (gomnd)
                Timeout: 2 * time.Second,
                         ^
```

</details>

<details><summary>With the PR</summary>

```console
$ golangci-lint run 
main.go:28:23: Magic number: 200, in <condition> detected (mnd)
        if res.StatusCode != 200 {
                             ^
main.go:21:12: Magic number: 2, in <assign> detected (mnd)
                Timeout: 2 * time.Second,
                         ^
```

</details>


bombsimon (approved):
LGTM, as long as the name actually _is_ just `mnd`. I see it's set to `mnd` in the code but most of the references refer to it as `go-mnd` and so does the gopher image in the repo.

@tommy-muehle Can you confirm that the name of the linter should be just `mnd` or would you rather set the name to `gomnd` on the analyzer? I guess `mnd` makes more sense since it's the name of the binary as well but might as well ask to clarify.

ldez:
> I see it's set to mnd in the code but most of the references refer to it as go-mnd and so does the gopher image in the repo.

So you are saying that only my first commit (coming from the previous PR) was enough, it's because you need more PRs to review? :smile_cat:

ldez:
> I'm sorry if this yielded extra work for you. 

I was just kidding :smile_cat: 

I was torn between the name change and the deprecation, so I used your comment as justification.

ldez:
More seriously, the readme is a mix of `go-mnd` and `mnd` but the command, the binary, the analyzer, and the brew-tap are called `mnd`.
Only the Docker image and the repository are named `go-mnd`.

We will wait for the feedback of the author but `mnd` seems to be the right name.

bombsimon:
> We will wait for the feedback of the author but `mnd` seems to be the right name.

Yeah 100% agree! I pinged to see if we got some response within a day or two max but I wouldn't block on it.

pkg/lint/lintersdb/builder_linter.go
```diff
@@ -286,9 +286,15 @@ func (b LinterBuilder) Build(cfg *config.Config) ([]*linter.Config, error) {
...
 
+		linter.NewConfig(golinters.NewMND(&cfg.LintersSettings.Mnd)).
+			WithSince("v1.22.0").
+			WithPresets(linter.PresetStyle).
+			WithURL("https://github.com/tommy-muehle/go-mnd"),
+
 		linter.NewConfig(golinters.NewGoMND(&cfg.LintersSettings.Gomnd)).
 			WithSince("v1.22.0").
 			WithPresets(linter.PresetStyle).
+			Deprecated("The linter has been renamed.", "v1.58.0", "mnd").
```
Antonboom, line 297: Why for `goerr13` you used alternative name without deprecation, but here you deprecated the old name? oO

pkg/golinters/mnd.go
```diff
@@ -0,0 +1,61 @@
...
+
+import (
+	mnd "github.com/tommy-muehle/go-mnd/v2"
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/golangci/golangci-lint/pkg/config"
+	"github.com/golangci/golangci-lint/pkg/golinters/goanalysis"
+)
+
+func NewMND(settings *config.MndSettings) *goanalysis.Linter {
```
Antonboom, line 11: I thought MND means "monad" (like in functional programming) :D 
really not obvious that mnd is for "magic number detector"

why not use name like "magic-number"?

Antonboom (changes requested):

pkg/lint/lintersdb/builder_linter.go
ldez: https://github.com/golangci/golangci-lint/pull/4562

pkg/golinters/mnd.go
ldez: We will not change the name of linters, golangci-lint is a wrapper.

pkg/lint/lintersdb/builder_linter.go
Antonboom: but why not `.WithAlternativeNames("gomd").` here?
I try to understand the reason of approaches difference

ldez: https://github.com/golangci/golangci-lint/pull/4572/commits/623d457238670ee79767f1dcd63f5e9b82e0b519#diff-23eb637b7fd1abddf7f975546966746b84c387b8294f08c9a9a19cd7dee59082R12-R19

Antonboom (approved):

PR #4573, opened 2024-03-25
fix: the TeamCity inspectionType service message

namco1992:
The [Inspection Type](https://www.jetbrains.com/help/teamcity/service-messages.html#Inspection+Type) is used in the TeamCity printer. The message type should be `inspectionType` as the doc specified, but the golangci-lint uses `InspectionType`(capitalized "I").

I'm not sure if it works with previous TeamCity versions, I'm currenly on "TeamCity Enterprise 2023.11.3 (build 147512)" and it doesn't work for me. Changing the `InspectionType` to `inspectionType` did the trick.

Update the TeamCity printer to have it align with the specification, and avoid any surprise in the future.

### DIFF
--- a/pkg/printers/teamcity.go
+++ b/pkg/printers/teamcity.go
@@ -88,7 +88,7 @@ type InspectionType struct {
 }
 
 func (i InspectionType) Print(w io.Writer, escaper *strings.Replacer) (int, error) {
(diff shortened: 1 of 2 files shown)


ldez (approved):
LGTM

Issue #1779, opened 2021-02-25
Add support for SuggestedFixes


ldez:
**Is your feature request related to a problem? Please describe.**

Add support for SuggestedFixes.

https://github.com/golang/tools/blob/b4639ccb830b1c9602f1c96eba624bbbe20650ea/go/analysis/doc/suggested_fixes.md

**Describe the solution you'd like**

Be able to apply automatically the fixes via `--fix`.

**Describe alternatives you've considered**

Display the suggestion as extra information.

**Additional context**

```go
type Diagnostic struct {
	// ...

	// SuggestedFixes contains suggested fixes for a diagnostic which can be used to perform
	// edits to a file that address the diagnostic.
	// TODO(matloob): Should multiple SuggestedFixes be allowed for a diagnostic?
	// Diagnostics should not contain SuggestedFixes that overlap.
	// Experimental: This API is experimental and may change in the future.
	SuggestedFixes []SuggestedFix // optional


	// ...
}
```
https://github.com/golang/tools/blob/b4639ccb830b1c9602f1c96eba624bbbe20650ea/go/analysis/diagnostic.go#L23-L28

Current linters that support `SuggestedFixes`:
- `goerr113`
- `ruleguard`
- `staticcheck`
- `nlreturn`
- `exportloopref`
- `exhaustive`
- `govet`

Related to #1728

If you want to help me:
- https://bento.me/golangci-lint
- https://bento.me/ldez

Latest update: https://github.com/golangci/golangci-lint/issues/1779#issuecomment-2545783471

ldez:
I'm working on and for now, I'm able to apply suggestions from `goerr113` because the suggestion is only on one line.

For the other, the problem is a bit more complex because the fix must be applied on multiple lines.

SVilgelm:
I think we need to print the suggested fixes if they are not applied (without `--fix`)

ldez:
In some cases, it will br very weird to display the suggestions, because a suggestion can be a `\n` and `TextEdits` is a slice.

SVilgelm:
we can print a `diff`, anyway you are going to create a diff, so we can print replacement, maybe in verbose output

SVilgelm:
@ldez 
I think we need to replace the `Replacement` with `SuggestedFixes` in `golangci-lint`, to avoid a transformation.

ldez:
I think we can support both as the first step, but I'm thinking about I need more time.

SVilgelm:
[...]

PR #4583, opened 2024-03-27
feat: add fatcontext linter

Crocmagnon:
As described in a blog post I wrote (https://gabnotes.org/fat-contexts/), we've been bitten by contexts not shadowed inside loops.
This linter addresses this issue.

Github repo: https://github.com/Crocmagnon/fatcontext

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -2532,6 +2532,7 @@ linters:
     - exhaustive
     - exhaustruct
     - exportloopref
+    - fatcontext
     - forbidigo
     - forcetypeassert
     - funlen
@@ -2645,6 +2646,7 @@ linters:
     - exhaustive
     - exhaustruct
     - exportloopref
+    - fatcontext
     - forbidigo
     - forcetypeassert
     - funlen
--- a/go.mod
+++ b/go.mod
@@ -11,6 +11,7 @@ require (
 	github.com/Antonboom/nilnil v0.1.7
 	github.com/Antonboom/testifylint v1.2.0
 	github.com/BurntSushi/toml v1.3.2
+	github.com/Crocmagnon/fatcontext v0.2.2
 	github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24
 	github.com/GaijinEntertainment/go-exhaustruct/v3 v3.2.0
 	github.com/OpenPeeDeeP/depguard/v2 v2.2.0
--- a/go.sum
+++ b/go.sum
@@ -49,6 +49,8 @@ github.com/BurntSushi/toml v0.3.1/go.mod h1:xHWCNGjB5oqiDr8zfno3MHue2Ht5sIBksp03
 github.com/BurntSushi/toml v1.3.2 h1:o7IhLm0Msx3BaB+n3Ag7L8EVlByGnpq14C4YWiu/gL8=
 github.com/BurntSushi/toml v1.3.2/go.mod h1:CxXYINrC8qIiEnFrOxCa7Jy5BFHlXnUU2pbicEuybxQ=
 github.com/BurntSushi/xgb v0.0.0-20160522181843-27f122750802/go.mod h1:IVnqGOEym/WlBOVXweHU+Q+/VP0lqqI8lqeDx9IjBqo=
+github.com/Crocmagnon/fatcontext v0.2.2 h1:OrFlsDdOj9hW/oBEJBNSuH7QWf+E9WPVHw+x52bXVbk=
+github.com/Crocmagnon/fatcontext v0.2.2/go.mod h1:WSn/c/+MMNiD8Pri0ahRj0o9jVpeowzavOQplBJw6u0=
 github.com/Djarvur/go-err113 v0.0.0-20210108212216-aea10b59be24 h1:sHglBQTwgx+rWPdisA5ynNEsoARbiCBOyGcJM4/OzsM=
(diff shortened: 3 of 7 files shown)


ccoVeille:
Thanks for this linter. I got same issue with a very long ctx in a loop last week

ldez:

- [ ] The CLA must be signed

### Pull Request Description

### Linter

### Others Requirements

    - `goanalysis.LoadModeTypesInfo` required `WithLoadForGoAnalysis()` in the `lintersdb/builder_linter.go`

### Recommendations

---

ccoVeille:
> As described in a blog post I wrote (https://gabnotes.org/fat-contexts/), we've been bitten by contexts not shadowed inside loops. This linter addresses this issue.
> 
> Github repo: https://github.com/Crocmagnon/foreshadow

I can tell you your linter helped me to spot 4 errors in my code. Thanks

ldez:
First, I understand the goal of this linter.

But I have a problem with the fact to "fight" against one of the classical govet rules: shadow

The shadowing is a well-known problem, it's weird to suggest shadowing variables instead of just using a dedicated name.

Crocmagnon:
> it's weird to suggest shadowing variables instead of just using a dedicated one

Contexts are often named `ctx` more than anything else. The first reaction when encountering the shadowing error from govet is to remove the shadowing while keeping the same variable name, leading to the kind of context misuse described in my blog post.

E.g.:
```go
ctx := context.Background()
for /* cond */ {
    ctx := context.WithValue(ctx, "key", "val")
    // ...
}
```
will likely become
```go
ctx := context.Background()
for /* cond */ {
    ctx = context.WithValue(ctx, "key", "val")
    // ...
}
```
or even

```go
ctx := context.Background()
for /* cond */ {
    ctx := context.WithValue(ctx, "key", "val") //nolint
    // ...
}
```
where the shadowing would be more appropriate (or the definition of a new variable).

In our case, we disabled the shadowing linter on contexts specifically because developers in the team found it too noisy/annoying, not realizing we were exposing ourselves to another kind of issue.

Also, we prefer keeping the name `ctx` for context values because it's so ubiquitous it becomes strange to read code that uses another name.

ldez:
As I said, I already understand the goal of this linter.
My remarks are about suggesting shadowing.

> In our case, we disabled the shadowing linter
> ...
> Also, we prefer keeping the name ctx

This is your story with shadowing, context, and naming, I understand.
But from the POV of golangci-lint, having two linters that fight against is not something desirable.

To be clear, I suggest not incited to shadow but to detect the problem, and inside the message not talking about shadow but about nesting context.
The users will decide what they want to take action.

I know that you implemented `suggestedFix` (not supported by golangci-lint) and it's not a problem to create a shadow as a fix because, after the fix, the users will have to "choose" between your linter and shadow reports. (but once again, `suggestedFix` is not supported by golangci-lint).

Crocmagnon:
> To be clear, I suggest not incited to shadow but to fix the problem: inside the message not talking about shadow but about nesting context.

That's much clearer, thanks for taking the time to explain.

I changed the linter according to your suggestions, released v0.2.0 and updated the PR.

ldez:
You still have a usage of `panic`: https://github.com/Crocmagnon/fatcontext/blob/781d68514681a8a8ed2a88cb2e7e8f49e88f5711/pkg/analyzer/analyzer.go#L106

Crocmagnon:
fixed

ldez:
> curious, how to suggest fixes with golangci-lint if SuggestedFix isn't supported?

First, at some point, we will support it.

Currently, auto-fixing is limited to a few linters:
- the linters based on the "gofmt" style (not recommended)
- the linters like `protogetter` which created a specific "mode": https://github.com/golangci/golangci-lint/blob/c0379205d7be0dba28b0cb37e609aaac0c0f95e5/pkg/golinters/protogetter.go#L28

bombsimon:
For reference the tracking issue to support `SuggestedFixes` is #1779

ldez:
@bombsimon the topic is in the top 5 of backlog, for now, I waiting for bug fixes of v1.57 but after that, I will start again with my multiple PRs every day :smile_cat:

Antonboom:
I feel some interception with https://github.com/kkHAIKE/contextcheck 🤔

Crocmagnon:
Tested today on our codebase at work, all reports were real cases except for one intentional (a function which returns the context it modifies). It also properly detected known cases.

ldez:
@Antonboom `contextcheck`, `fatcontext` are complementary: `contextcheck` verifies the context continuity, and `fatcontext` verifies the nested context which doesn't impact the continuity of the context.

Crocmagnon:
Hello!
Quick questions:
* what are the next steps for this PR?
* do you expect something from me to move this forward?
(not impatient, just checking if I have something to do 😀)

ldez (approved):
LGTM

batazor:
I create a context and want to send it to each element, I think this is not a frequent but valid case, is it not?

```golang
// start tracing
newCtx, span := otel.Tracer(fmt.Sprintf("saga: %s", s.name)).Start(ctx, fmt.Sprintf("saga: %s", s.name))
defer span.End()

for _, step := range initSteps {
	step.ctx = newCtx // nested context in loop (fatcontext)
	g.Go(step.Run)
}
```

Crocmagnon:
@batazor in this case, it looks like ˋg.Goˋ is from ˋerrgroup`. I would use ˋWithContext` instead of passing a context to a struct. Or closures.

Issue #4593, opened 2024-03-29
json-schema: missing some properties


nikpivkin:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've read the typecheck section of the FAQ (https://golangci-lint.run/usage/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

VSCode issues the following warnings:
- Property check-shadowing is not allowed.
- Property skip-files is not allowed.
- Property skip-dirs is not allowed.

Scheme source: https://golangci-lint.run/jsonschema/golangci.jsonschema.json

<img width="680" alt="image" src="https://github.com/golangci/golangci-lint/assets/100182843/debb1f22-218e-4c67-9c71-8298f427fa10">

### Version of golangci-lint

-

### Configuration

https://github.com/aquasecurity/trivy/blob/main/.golangci.yaml

### Go environment

-

### Verbose output of running

-

### A minimal reproducible example or link to a public repository

-

### Validation

- [X] Yes, I've included all information above (version, config, etc.).

ldez:
Hello,

it's the expected behavior, those properties have been deprecated.

https://golangci-lint.run/product/changelog/#v1570

nikpivkin:
@ldez Thank you! Why doesn't `golangci` warn about deprecated fields?

akutz:
> Hello,
> 
> it's the expected behavior, those properties have been deprecated, then they are considered as removed.
> 
> https://golangci-lint.run/product/changelog/#v1570

I do not see `skip-files` as deprecated in the linked changelog page.

ldez:
> ⚠️ Replace `run.skip-xxx` options by `issues.exclude-xxx` options

PR #4602, opened 2024-03-30
Add `pre-commit` hook to run `config verify`

matthewhughes934:
Add this as a way to easily ensure configs stay valid. Filenames are not passed to the hook, this means For the default case (config in one of the default searched files), the config is picked up automatically and there's nothing extra to configure. For the case of a non-standard config setup the hook can be configured like

    repos:
    -   repo: https://github.com/golangci/golangci-lint
        rev: v1.57.2
        hooks:
        -   id: golangci-lint-config-verify
            # config is kept at '.golangci-config.yaml'
            files: \.golangci-config\.yaml
            args: ['--config', '.golangci-config.yaml']

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -14,3 +14,10 @@
   language: golang
   require_serial: true
   pass_filenames: false
+- id: golangci-lint-config-verify
+  name: golangci-lint-config-verify
+  description: Verifies the configuration file
+  entry: golangci-lint config verify
+  files: '\.golangci\.(?:yml|yaml|toml|json)'
+  language: golang
+  pass_filenames: false


ccoVeille:
I had to do this once in a commit hooks

I found the best way to do it is to launch

golangci-lint --config whatever.yaml linters

It returns status code 0 when the conf is valid.

matthewhughes934:
I skipped submitting a feature request issue since it seemed as simple to just discuss the change itself

ldez:
The validation of the configuration against the JSON Schema is different than using `run`.
The deprecated options are removed from the JSON Schema (draft-07 doesn't have deprecation support).
The `run` command (and only the `run` command) displays deprecation messages, but the validation will display that the fields are missing and will lead to this kind of misunderstanding: #4593

Also, I don't see the point in systematically checking a configuration which most of the time has not changed.

matthewhughes934:
> Also, I don't see the point in systematically checking a configuration which most of the time has not changed.

This hook will only be executed via `pre-commit run` _if_ there's a change in one of the files matching the pattern in `files`, so it'll only check the config if something has changed (though in CI I expect most people to run with `pre-commit run --all-files` in which case it will always check, but that's just a small bit of extra time in CI which I think is reasonable)

> The deprecated options are removed from the JSON Schema (draft-07 doesn't have deprecation support).
The run command (and only the run command) displays deprecation messages, but the validation will display that the properties are missing and will lead to this kind of misunderstanding: https://github.com/golangci/golangci-lint/issues/4593

My use case is: I just want to keep my config up-to-date: if a change brings deprecates some option I'm using then I want to be aware of if (and remove it) as soon as I can. This could probably also be solved by just running `golangci-lint config verify` in CI, but I like the workflow of working with `pre-commit`

I have no opinion re: potential misunderstandings, if you think this has the potential to be a bother then I have no objections

>  IDEs and text editors allow us to do that live, a call to a pre-commit hook seems overkill.

This would require everyone working in a project to have their (potentially various) IDEs/editors configured to handle this, vs. just having everyone install `pre-commit` and have it managed there.

ldez:
> I have no opinion re: potential misunderstandings, if you think this has the potential to be a bother then I have no objections

It's not a blocker, it's more a warning, if there are too many issues on the topic, we will see how to handle that.

> This would require everyone working in a project to have their (potentially various) IDEs/editors configured to handle this,

At least for the IDEs I use, there is no configuration required, no specific user action or installation, it's just handled automatically based on the file name.

The configuration file names are defined inside the schema store: https://github.com/SchemaStore/schemastore/blob/c42ad08e78f70a0d37f6fb53a204371ef80bc3d4/src/api/json/catalog.json#L2139-L2143

But I'm not a pre-commit user, and your workflow seems right for a pre-commit user.

ldez (approved):
LGTM

matthewhughes934:
> It's not a blocker, it's more a warning, if there are too many issues on the topic, we will see how to handle that.

:+1: I'd be happy to be pinged/included if any such issues do arise

PR #4609, opened 2024-04-03
feat: add linereturn linter

Ak-Army:
Add [linereturn](https://github.com/Ak-Army/linereturn) linter.

Linter requires a new line before return statements if the previous line in a block is a real statement.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	4d63.com/gochecknoglobals v0.2.1
 	github.com/4meepo/tagalign v1.3.3
 	github.com/Abirdcfly/dupword v0.0.14
+	github.com/Ak-Army/linereturn v0.0.0-20240402133102-9b651aa9b3c7
 	github.com/Antonboom/errname v0.1.12
 	github.com/Antonboom/nilnil v0.1.7
 	github.com/Antonboom/testifylint v1.2.0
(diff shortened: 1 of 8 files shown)


bombsimon:
Feels like a duplicate of [`nlreturn`](https://github.com/ssgreg/nlreturn)?

ldez:
Yes it's a duplicate.

Ak-Army:
Not the same as nlreturn, but very similar.
The difference is that this linter expects an enter if there is a real statement in the previous line, but forbids an enter if there is only a statement closure.


Examples of incorrect code:

```go
func fooa() int {
    o := []int{0, 1}
    return o[0]
}

func foob(s string) *a {
    o := &a{
     a: s,
    }
	
    return o
}

func fooc() int {
    defer bar()
    return 0
}

func fooe(s string) interface{} {
    o := foob(
        s,
    )
	
    return o
}
```

Examples of correct code:

```go
func fooa() int {
    o := []int{0, 1}
	
    return o[0]
}

func foob(s string) *a {
    o := &a{
        a: s,
    }
    return o
}

func fooc() int {
    defer bar()
    
    return 0
}

func fooe(s string) interface{} {
    o := foob(
        s,
    )
    return o
}
```

ccoVeille:
Maybe you could contribute to [nlreturn](https://github.com/ssgreg/nlreturn) by opening an PR on it ? it could become a setting of it.

Ak-Army:
Thx, i will try that way.

Issue #4639, opened 2024-04-13
Improve expereince of running linters with auto-fix capabilities on save events in editor.


HartBlanc:
### Your feature request related to a problem? Please describe

A repo I work in uses golangci-lint with the `gci` linter (among other linters).

golangci-lint is run automatically before raising each PR, and the auto-fixes from gci are applied at this point.

As a developer in this repo, I find that these changes introduce a little friction as I need to amend the suggested changes to my latest commit before raising my PR. The change which introduces the import and the change which sorts the imports may then be in different commits which introduces a little unnecessary noise in my changes (or additional effort to reorganise the commits). 

I would prefer to have the autofix applied automatically by my editor (neovim) after each save to ensure that imports are sorted as early as possible.

I can configure my editor to run gci when saving a file relatively easily. However, I would like to ensure that when gci runs on save in my editor it uses the same config that golangci-lint will use without needing to manually ensure that the two stay in sync.

I have attempted the following approaches:

1. Introduce logic in my editor configuration / an external script to parse the .golangci-lint.yml configuration and convert it into the relevant gci CLI arguments.

Limitations:
- This is quite a lot of effort for integration. For every individual linter every user will need to implement this conversion logic. This would likely be better solved once by a linter aggregator.

2. Configure my editor to run golangci-lint on save (`goclangci-lint run --fix --fast --enable-only=gci $FILENAME` flags).

Limitations:
- golangci-lint reports type check errors even in cases where gci would be able to format the file without any issues (e.g. a syntax error in another file in the package, a reference to an undefined type etc.)
- The user needs to explicitly enumerate all of the linters with autofix capabilities

### Describe the solution you'd like

1. Resolve failure modes for golangci-lint where prerequisite checks (i.e. typecheck) fail despite these presenting no issues for the enabled linters.
2. Consider adding support for only running the linters with autofix capabilities (e.g. `run --autofix-linters` flag or something similar)

### Describe alternatives you've considered

- Adding support for querying config for a specific linter (this would make the integration for users a little easier, but still non-trivial.
- Making do with running gci directly and parsing gci config manually (this does seem to have the advantage of being a bit faster)
- Simply not running the autoformatter when there are typecheck errors (loses some of the incrementality benefits of running on save.)

### Additional context

I'm happy to contribute a PR if any of the proposed solutions are acceptable

ldez:
Hello,

in v1.57.0, we introduced the flag `--enable-only`:

[...]

PR #4640, opened 2024-04-13
fix: keep only typecheck issues

ldez:
Related to https://github.com/golangci/golangci-lint/issues/4639#issuecomment-2053660195

### DIFF
--- a/pkg/result/processors/invalid_issue_test.go
+++ b/pkg/result/processors/invalid_issue_test.go
@@ -32,6 +32,21 @@ func TestInvalidIssue_Process(t *testing.T) {
 			},
 		},
 		{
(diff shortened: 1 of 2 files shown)


pkg/result/processors/invalid_issue_test.go
```diff
@@ -31,6 +31,21 @@ func TestInvalidIssue_Process(t *testing.T) {
 				{FromLinter: "typecheck"},
 			},
 		},
+		{
+			desc: "typecheck and Go file",
```
bombsimon, line 35: Might be worth clarifying this
```suggestion
			desc: "keep only typecheck issues if any exist",
```

bombsimon (approved):

PR #4653, opened 2024-04-17
fix: speed up "fast" linters

### LINKED ISSUE #2914: Load Package is not necessary for all linters

Some linters (`lll, misspell, goimports, gocyclo etc`) don't require any package information.  It seems golangci-lint always load packages irrespective of which set of linters are enabled. [[Architecture](https://golangci-lint.run/contributing/architecture/)]. It results performance degradation. Running these linters individually is taking 10X less time than running in the golangci-lint

Smartly determines if "load package" step is necessary or not.

Cache seems no effect here. Running golangci-lint again takes same amount of high time in "load Package" step. 

### Additional context.

_No response_

### LINKED ISSUE #4639: Improve expereince of running linters with auto-fix capabilities on save events in editor.

ldez:
Currently, there are 2 problems:

1. `typecheck` uses `LoadModeTypesInfo` and `WithLoadForGoAnalysis`.
   So the load mode of the linter config for the `metalinter` is always `LoadModeTypesInfo` and never `LoadModeSyntax`.
2. the `metalinter` has a hardcoded call to `WithLoadForGoAnalysis`, then the types are always loaded.  

As a consequence, the "fast" linters (like `lll`, `gofmt`, `misspell`, etc.) are slower than expected when either they are running alone or only "fast" linters are run.

This PR changes the config linter load mode of `typecheck` to `LoadModeNone` and removes `WithLoadForGoAnalysis` for this fake linter.
The `metalinter` goanalysis load mode is computed based on linters.

`typecheck` doesn't need an explicit load mode because it is never run alone, then at the end, the load mode will depend on the load mode of the real linters.
Same thing for the goanalysis load mode.

FYI, there are 2 load modes:
- the config linter load mode (`LoadModeTypesInfo`, `LoadModeSyntax`, etc.)
- the goanalysis load mode (`WithLoadForGoAnalysis`, `WithLoadFiles`)

They are used for different purposes.

---

Based on local tests (verbose mode), the performances are improved (~50%).

<details>

### On golangci-lint code

Current version:
- `golangci-lint cache clean; golangci-lint run --enable-only misspell --enable-only lll  -v`
-  Execution took between 1.4s and 950ms

With the PR:
- `golangci-lint cache clean; ./golangci-lint run --enable-only misspell --enable-only lll  -v`
-  Execution took between 900ms and 800ms

### On Traefik code

Current version:
- `golangci-lint cache clean; golangci-lint run --enable-only gofmt --enable-only misspell -v`
-  Execution took between 1.3s and 1.2ms

Current version:
- `golangci-lint cache clean; ./golangci-lint run --enable-only gofmt --enable-only misspell -v`
-  Execution took between 650ms and 500ms

</details>

Fixes #2914
Fixes #4639

### DIFF
--- a/pkg/lint/lintersdb/builder_linter.go
+++ b/pkg/lint/lintersdb/builder_linter.go
@@ -751,10 +751,7 @@ func (LinterBuilder) Build(cfg *config.Config) ([]*linter.Config, error) {
 		linter.NewConfig(golinters.NewTypecheck()).
 			WithInternal().
 			WithEnabledByDefault().
-			WithSince("v1.3.0").
-			WithLoadForGoAnalysis().
-			WithPresets(linter.PresetBugs).
-			WithURL(""),
+			WithSince("v1.3.0"),
 
 		linter.NewConfig(unconvert.New(&cfg.LintersSettings.Unconvert)).
 			WithSince("v1.0.0").
--- a/pkg/lint/lintersdb/manager.go
+++ b/pkg/lint/lintersdb/manager.go
@@ -204,18 +204,30 @@ func (m *Manager) build(enabledByDefaultLinters []*linter.Config) map[string]*li
 }
 
 func (m *Manager) combineGoAnalysisLinters(linters map[string]*linter.Config) {
+	mlConfig := &linter.Config{}
+
 	var goanalysisLinters []*goanalysis.Linter
 	goanalysisPresets := map[string]bool{}
+
 	for _, lc := range linters {
 		lnt, ok := lc.Linter.(*goanalysis.Linter)
 		if !ok {
 			continue
 		}
+
 		if lnt.LoadMode() == goanalysis.LoadModeWholeProgram {
 			// It's ineffective by CPU and memory to run whole-program and incremental analyzers at once.
 			continue
 		}
+
+		mlConfig.LoadMode |= lc.LoadMode
+
+		if lc.IsSlowLinter() {
+			mlConfig.ConsiderSlow()
+		}
+
 		goanalysisLinters = append(goanalysisLinters, lnt)
+
 		for _, p := range lc.InPresets {
 			goanalysisPresets[p] = true
 		}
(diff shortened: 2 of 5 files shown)


pkg/lint/lintersdb/builder_linter.go
```diff
@@ -751,10 +751,7 @@ func (LinterBuilder) Build(cfg *config.Config) ([]*linter.Config, error) {
 		linter.NewConfig(golinters.NewTypecheck()).
 			WithInternal().
 			WithEnabledByDefault().
-			WithSince("v1.3.0").
-			WithLoadForGoAnalysis().
-			WithPresets(linter.PresetBugs).
-			WithURL(""),
```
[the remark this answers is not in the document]
ldez: no, it's part of the changes related to typecheck.

pkg/lint/lintersdb/manager.go
```diff
@@ -245,22 +257,14 @@ func (m *Manager) combineGoAnalysisLinters(linters map[string]*linter.Config) {
 		return a.Name() <= b.Name()
 	})
 
-	ml := goanalysis.NewMetaLinter(goanalysisLinters)
+	mlConfig.Linter = goanalysis.NewMetaLinter(goanalysisLinters)
```
bombsimon, line 260: Is there a reason this is set after the linters loop like this instead of in the creating of `mlConfig` on L207?

```go
mlConfig := &linter.Config{
    Linter: goanalysis.NewMetaLinter(goanalysisLinters),
}
```

Feels more natural to me but maybe something is preventing this in `ConsiderSlow`?

bombsimon (approved):
LGTM

pkg/lint/lintersdb/manager.go
ldez: The linters loop is on a map: due to alt names, linters can be duplicated (they are deduplicated by using `.Names()`).
 
The computation of `LoadMode` and `IsSlow` is easier and more readable by not using local variables.

I will rewrite the `presets` computation to follow the same "style" as `LoadMode` and `IsSlow`.

Antonboom (approved):

PR #4673, opened 2024-04-22
dev: clean deprecated options

ldez:
Remove deprecated options from `.golangci.next.reference.yml` and `jsonschema/golangci.next.jsonschema.json`


.golangci.next.reference.yml
```diff
@@ -637,9 +622,6 @@ linters-settings:
         # The option is passed to the ruleguard 'debug-group' argument.
         # Default: ""
         debug: 'emptyDecl'
-        # Deprecated, use 'failOn' param.
-        # If set to true, identical to failOn='all', otherwise failOn=''
-        failOnError: false
```
Antonboom, line 642: was noticed that `jsonschema/golangci.next.jsonschema.json` doesn't describe the gocritic settings in details, just the checkers list 

but checkers list is not right solution too, because not all checkers have settings 🤔

pkg/config/linters_settings.go
```diff
@@ -909,11 +909,11 @@ type UseStdlibVarsSettings struct {
 	TimeLayout         bool `mapstructure:"time-layout"`
 	CryptoHash         bool `mapstructure:"crypto-hash"`
 	DefaultRPCPath     bool `mapstructure:"default-rpc-path"`
-	OSDevNull          bool `mapstructure:"os-dev-null"`
+	OSDevNull          bool `mapstructure:"os-dev-null"` // Deprecated
```
Antonboom, line 912: don't see `usestdlibvar` config in reference files 🤔

Antonboom (approved):

pkg/config/linters_settings.go
ldez: `usestdlibvars` is in the reference files:

https://github.com/golangci/golangci-lint/blob/55b2f5ddebb8270a2dac581cd072fbf287bd3174/.golangci.reference.yml#L2612

https://github.com/golangci/golangci-lint/blob/55b2f5ddebb8270a2dac581cd072fbf287bd3174/.golangci.reference.yml#L2261

https://github.com/golangci/golangci-lint/blob/55b2f5ddebb8270a2dac581cd072fbf287bd3174/.golangci.next.reference.yml#L2740

https://github.com/golangci/golangci-lint/blob/55b2f5ddebb8270a2dac581cd072fbf287bd3174/.golangci.next.reference.yml#L2272

.golangci.next.reference.yml
ldez: I updated the JSON schema.

PR #4685, opened 2024-05-01
feat: use problem matchers for GitHub Action format

ldez:
Currently, when running the GitHub Action each annotation creates 2 outputs:
- a "popup"
- a log

The "popup" works well but the log doesn't display the file information (path, position).

The default annotations format `::error file={name},line={line},col={endLine}::{message}` doesn't display the file information (I don't know why).

- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-a-warning-message
- https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message

This PR uses a specific feature called "[Problem Matchers](https://github.com/actions/toolkit/blob/main/docs/problem-matchers.md)" that allows defining a custom format.
With this custom format, the log displays file information.

<details>

Screenshot log

Screenshot popup summary

Screenshot popup code

</details>

Fixes https://github.com/golangci/golangci-lint-action/issues/119


pkg/printers/github.go
```diff
@@ -1,28 +1,102 @@
...
-// print each line as: ::error file=app.js,line=10,col=15::Something went wrong
-func formatIssueAsGithub(issue *result.Issue) string {
-	severity := defaultGithubSeverity
+func (p *GitHub) Print(issues []result.Issue) error {
+	// Note: the file with the problem matcher definition should not be removed.
+	// A sleep can mitigate this problem but this will be flaky.
+	//
+	// Error: Unable to process command '::add-matcher::/tmp/golangci-lint-action-problem-matchers.json' successfully.
+	// Error: Could not find file '/tmp/golangci-lint-action-problem-matchers.json'.
+	//
```
bombsimon, line 76:
```suggestion
	// Result if the file is removed prematurely:
	// Error: Unable to process command '::add-matcher::/tmp/golangci-lint-action-problem-matchers.json' successfully.
	// Error: Could not find file '/tmp/golangci-lint-action-problem-matchers.json'.
```

bombsimon (approved):
💯

pkg/printers/github.go
ldez: Those lines are real logs from GitHub Action, we cannot rephrase them.

ldez: :thinking: the github diff is weird

PR #4719, opened 2024-05-12
docs: improve integrations and install pages

ldez:
- Add LSP integration
- Improve CI page
- Remove archived integration: [syntastic](https://github.com/vim-syntastic/syntastic)


docs/src/docs/welcome/install.mdx
```diff
@@ -6,25 +6,28 @@ title: "Install"
 
 Most installations of `golangci-lint` are performed for CI.
 
+It's important to have reproducible CI: don't start to fail all builds at the same time.
+With golangci-lint this can happen if you use option `--enable-all` and a new linter is added
+or even without `--enable-all` when one upstream linter is upgraded.
+
+**IMPORTANT**: It's highly recommended installing a specific version of golangci-lint available on the [releases page](https://github.com/golangci/golangci-lint/releases).
```
bombsimon, line 13: I see this is just moved text but the examples in https://github.com/golangci/golangci-lint-action/blob/master/README.md uses `latest`, is it worth changing that or mentioning this in that repo to be consistent with the website docs?

bombsimon (approved):
LGTM

docs/src/docs/welcome/install.mdx
ldez: You are right I will change the readme of the action.

Issue #2864, opened 2022-05-18
Option to allow the linting of generated files


0xjac:
### Your feature request related to a problem? Please describe.

I have some Go files generated from a template I control. Some mistakes in the template can result in sub-optimal Go code being generated. For example a `struct` can be packed poorly and `fieldalignment` (from `govet`) would catch it.

Many linters in the bugs category could be useful on generated files which depends on changes made by contributors within the project.

### Describe the solution you'd like.

A  configuration options to include or exclude all or some generated files as part of the set of files being linted.

Something like:
```yaml
run:
    # Include generated files or not.
    # If disabled, specific generated files can still be skipped by adding them 
    # to skip-dirs or skip-files.
    # Default: false
    generated: true
```

The `run` options  `skip-dirs-use-default`, `skip-dirs`, `skip-files` would still be applied such that after including generated files, some specific generated files (for example those generated from another tool) can still be explicitly ignored.

> Most tools which generate Go files will either put them in a specific directory or will add a prefix in the name. 
> E.g. [`stringer`](https://pkg.go.dev/golang.org/x/tools/cmd/stringer) will add a `_string` prefix so running `stringer` on `myfile.go` will produce `myfile_string.go` which is easy to ignore with `skip-files` by adding `".+_string.go".

The `issues` options `exclude`, `exclude-rules` and `include` should also be applied such that some rules can be disabled for generated files that are linted.

*TL;DR this would be similar to the `run` option `tests` but for generated files instead of test files.*

### Describe alternatives you've considered.

1. Removing the "generated" comment from my generated files but that doesn't work as other tools rely on those comments to identify generated files.
2. Make a script to run all the linters I want on the generated files. But this require the installation of multiple individual linters and having to re-configure them. It is essentially a poor-man's implementation of golangci-lint, so what's the point...

### Additional context.

1. The new `generated` option should default to `false` to ignore generated files and preserve the current default behavior.
1. Some people are also having issues with false positive when detecting generated files (c.f. #2254). 
1. Some linters also have their own options which can be toggled to handle generated files—like [exhaustive](https://golangci-lint.run/usage/linters/#exhaustive) and [gosec](https://golangci-lint.run/usage/linters/#gosec)—and make the overall golangci-lint configuration easier when generated files are included. Those options are currently ignored!

wiegell:
Workaround script that can be run after code generation (careful, sed arguments are platform specific, this one is for macOS):
```
[...]

PR #4728, opened 2024-05-18
include generated files

korableg:
Hello! 
I created my own linter which checks proto generated files (I need checking options) and faced with the same issue as here https://github.com/golangci/golangci-lint/issues/2864 that I can't to check the generated proto files.

This pull request fixes this problem by adding new config parameter.

Check plz :) Thank you!


ldez:
Hello,

Your first was better than the current one.

The current implementation is a kind of weird opposite of `exclude-dirs` and `exclude-files` mixed with the management of the generated files.

This implementation is not in the right direction.

Supersedes by #4740

Issue #4735, opened 2024-05-20
gosec: impact on performances (v1.58.2)


ldez:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've read the typecheck section of the FAQ (https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

The way to detect the Go version has been changed since gosec v2.20.0, this has an impact on performances inside golangci-lint.

v1.58.2
```console
$ golangci-lint run --print-issued-lines=false --enable-only gosec
  Time (mean ± σ):      3.185 s ±  0.047 s    [User: 8.912 s, System: 6.396 s]
  Range (min … max):    3.112 s …  3.288 s    10 runs
```

v1.58.1
```console
$ golangci-lint run --print-issued-lines=false --enable-only gosec
  Time (mean ± σ):     606.8 ms ±  18.7 ms    [User: 2388.6 ms, System: 1018.3 ms]
  Range (min … max):   584.4 ms … 648.8 ms    10 runs
```

Related to #4724
Related to https://github.com/securego/gosec/pull/1119

**Workaround:**

```yml
linters-settings:
  gosec:
    excludes:
      - G601
      - G113
```

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version 1.58.2 built with go1.22.3 from 8c4cfb61 on 2024-05-19T18:08:33Z
```

</details>

### Configuration

<details>

```yml
linters:
  disable-all: true
  enable:
    - gosec

```

</details>

### Go environment

<details>

```console
$ go version
go version go1.22.3 linux/amd64
```

</details>

### Verbose output of running

<details>

```console
$ golangci-lint cache clean
$ golangci-lint run -v --enable-only gosec
INFO [config_reader] Config search paths: [./ /home/ldez/sources/go/src/github.com/golangci/golangci-lint /home/ldez/sources/go/src/github.com/golangci /home/ldez/sources/go/src/github.com /home/ldez/sources/go/src /home/ldez/sources/go /home/ldez/sources /home/ldez /home /] 
INFO [config_reader] Used config file .golangci.yml 
[...]

PR #4736, opened 2024-05-21
gosec: exclude G601 and G113

ldez:
Temporary fix.

Related to #4735

### DIFF
--- a/pkg/golinters/gosec/gosec.go
+++ b/pkg/golinters/gosec/gosec.go
@@ -150,19 +150,6 @@ func toGosecConfig(settings *config.GoSecSettings) gosec.Config {
 	return conf
 }
 
-func convertScoreToString(score issue.Score) string {
-	switch score {
-	case issue.Low:
-		return "low"
-	case issue.Medium:
-		return "medium"
-	case issue.High:
-		return "high"
-	default:
-		return ""
-	}
-}
-
 // based on https://github.com/securego/gosec/blob/47bfd4eb6fc7395940933388550b547538b4c946/config.go#L52-L62
 func convertGosecGlobals(globalOptionFromConfig any, conf gosec.Config) {
 	globalOptionMap, ok := globalOptionFromConfig.(map[string]any)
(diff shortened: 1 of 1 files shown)


pkg/golinters/gosec/gosec.go
```diff
@@ -179,17 +166,35 @@ func convertGosecGlobals(globalOptionFromConfig any, conf gosec.Config) {
 func gosecRuleFilters(includes, excludes []string) []rules.RuleFilter {
 	var filters []rules.RuleFilter
 
-	if len(includes) > 0 {
-		filters = append(filters, rules.NewRuleFilter(false, includes...))
+	cleanIncludes := cleanRules(includes)
```
bombsimon, line 169: Should we really remove the possibility to enable them even if a user wants to? I get we remove them from default rules if no config is provided but if explicitly enabled by a config I would expect them to be enabled, the extra time isn't necessarily a blocker for everyone.

ldez: > the extra time isn't necessarily a blocker for everyone.

We are talking about a duration multiplied by at least 4 (milliseconds becomes seconds).

> Should we really remove the possibility to enable them even if a user wants to?

We cannot really play with "default" because the current problem impacts users who already explicitly enable those rules.

This fix is a temporary fix, we will create the real fix when gosec allows us to handle the Go version.

ldez: If you feel uncomfortable with this PR, we can downgrade to v2.19.0.
But technically, the v2.19.0 contains a bug with how to handle the Go version.

bombsimon: > We are talking about a duration multiplied by at least 4 (milliseconds becomes seconds).

Yeah I saw your table but to me 10s isn't a blocker in my CI e.g. so if I was using this check I would be annoyed it wasn't up to me to opt in.

> This fix is a temporary fix, we will create the real fix when gosec allows us to handle the Go version.

Got that, I guess it's not the end of the world and for people who reeeally care they can probably stick to an older version of `golangci-lint`.

bombsimon (approved):

ldez:
Supersedes by #4748

PR #4758, opened 2024-05-27
fix: init empty result slice for SARIF printer

Zxilly:
if no issue is found, the sarif printer will produce a `nil` as a result, and `encoding/json` makes it a `null`, which is required to be an array in sarif format.

see https://github.com/Zxilly/go-size-analyzer/actions/runs/9254489826/job/25456396105 for real-world errors.


ldez (approved):
LGTM

PR #4764, opened 2024-05-27
fix golangci name

ccoVeille:
- **chore: fix golangci-lint name**
- **chore: fix golangci-lint name in teamcity**

From my understanding:
- the tool name is golangci-lint
- the team/organization is GolangCI

golangci-lint should be used everywhere it's not about GolangCI team/org


ldez:
There is no real official convention for golangci-lint.

For now, GolangCI-Lint, golangci-lint, or Golangci-lint are allowed

ccoVeille:
OK then, I understand

PR #4775, opened 2024-05-30
fix: SARIF format require issue column >= 1

Zxilly:
Some linters won't provider column value for issue, under this situation we should set it to 1.

image

For real world error see https://github.com/Zxilly/go-size-analyzer/actions/runs/9300840788/job/25597744381


ldez (approved):
LGTM

PR #4781, opened 2024-05-31
lll: Advertise max line length instead of just reporting failure

blast-hardcheese:
Pardon if this has been discussed before, I couldn't find it in the issue/PR list.

As a user, I find it disorienting when tools inform me that I have made a mistake but do not offer any insights into how to resolve the issue.

In this case in particular, we have all the information, but then require the user to know the defaults and to traverse the config overrides.

It'd be nice to just provide this information. Impact to existing users should be minimal, including data transfer or storage of logs in CI, since teams who are already in compliance would not get the new log component, and teams who are not in compliance yet would get the message once and then be done.

Happy to circle back if there's prior art for how this information should be displayed.

Thanks!

### DIFF
--- a/pkg/golinters/lll/lll.go
+++ b/pkg/golinters/lll/lll.go
@@ -121,7 +121,7 @@ func getLLLIssuesForFile(filename string, maxLineLen int, tabSpaces string) ([]r
 					Filename: filename,
 					Line:     lineNumber,
 				},
-				Text:       fmt.Sprintf("line is %d characters, max %d", lineLen, maxLineLen),
+				Text:       fmt.Sprintf("The line is %d characters long, which exceeds the maximum of %d characters.", lineLen, maxLineLen),
 				FromLinter: linterName,
 			})
 		}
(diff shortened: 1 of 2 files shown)


ldez:
Hello, it's a good suggestion, I rephrased the message to be more explicit.

pkg/golinters/lll/lll.go
```diff
@@ -121,7 +121,7 @@ func getLLLIssuesForFile(filename string, maxLineLen int, tabSpaces string) ([]r
 					Filename: filename,
 					Line:     lineNumber,
 				},
-				Text:       fmt.Sprintf("line is %d characters", lineLen),
+				Text:       fmt.Sprintf("The line is %d characters long, which exceeds the maximum of %d characters.", lineLen, maxLineLen),
```
Antonboom, line 124: Usually linters warnings not ended with period?

Antonboom (approved):

pkg/golinters/lll/lll.go
Antonboom: And not capitalized 🤔

ldez: I think there is no convention about linter messages: some have a period, others do not, and the same thing applies to capitalization.

ldez (approved):

PR #4831, opened 2024-06-19
fix: sanitize level property for SARIF

Zxilly:
https://docs.oasis-open.org/sarif/sarif/v2.1.0/errata01/os/sarif-v2.1.0-errata01-os-complete.html#_Toc141790898

level can only be `none, note, warning, error`, some linter like `gosec` may report something like `high`

For real-world error, see https://github.com/Zxilly/go-size-analyzer/actions/runs/9586342361/job/26434117667

### DIFF
--- a/pkg/printers/sarif.go
+++ b/pkg/printers/sarif.go
@@ -76,8 +76,15 @@ func (p Sarif) Print(issues []result.Issue) error {
 		issue := issues[i]
 
 		severity := issue.Severity
-		if severity == "" {
-			severity = "error"
+
+		// can be only none, note, warning, error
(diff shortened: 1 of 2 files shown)


ldez:
There is no severity convention inside golangci-lint but you can override or define your own inside the `severity` section.

https://golangci-lint.run/usage/configuration/#severity-configuration

Zxilly:
I'm not familiar with the golangci lint codebase, but I was wondering if an e2e test exists? Maybe we can add to these tests for sarif and we can validate the output json file with a json schema.

ldez:
we don't e2e to validate the output, and to be honest I already thinking of adding a test about that.
But those tests will not be more useful than the current tests.

ldez (approved):
LGTM

PR #4848, opened 2024-06-25
dev: rewrite benchmarks

ldez:
This file is not used but I think it's better if it's working :wink: 

I have had those changes locally for years, so it's time to add them.

I'm working on a [better benchmark system](https://github.com/ldez/golangci-lint-bench), I think we may remove this in the future.

### DIFF
--- a/test/bench/bench_test.go
+++ b/test/bench/bench_test.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"slices"
 	"strconv"
 	"strings"
 	"testing"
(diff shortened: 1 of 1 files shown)


test/bench/bench_test.go
```diff
@@ -9,111 +9,257 @@ import (
...
-func prepareGithubProject(owner, name string) func(testing.TB) {
-	return func(b testing.TB) {
-		dir := filepath.Join(build.Default.GOPATH, "src", "github.com", owner, name)
-		_, err := os.Stat(dir)
-		if os.IsNotExist(err) {
-			repo := fmt.Sprintf("https://github.com/%s/%s.git", owner, name)
-			err = exec.Command("git", "clone", repo, dir).Run()
-			if err != nil {
-				b.Fatalf("can't git clone %s/%s: %s", owner, name, err)
+func Benchmark_linters(b *testing.B) {
```
Antonboom, line 38: nitpick: not consistent naming `Benchmark_linters`, `BenchmarkGolangciLint`

Antonboom (approved):

PR #4855, opened 2024-07-02
dev: add action to add checklist for new linter

bombsimon:
This will add the checklist for new linters to the repository and revision control. When the label `linter: new` is added the checklist will automatically be added to the PR instead of having to copy-paste it from an arbitrary source.

Since only collaborators and members can add labels it's not likely that this will be abused, however to mitigate the risk even further the action will reset the checklist comment if the tag is added and removed multiple times.


.github/workflows/new_linter_checklist.yaml
```diff
@@ -0,0 +1,105 @@
...
+            | wc -l)
+          [ "$comment_exist" -gt 0 ] && edit_last="--edit-last"
+
+          # Comment on the PR with the checklist.
+          gh pr comment "$NUMBER" --body "$BODY" $edit_last
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GH_REPO: ${{ github.repository }}
+          NUMBER: ${{ github.event.number }}
+          BODY: |
```
ldez, line 33: I think it's better to use a dedicated markdown file instead of a multiple YAML string inside an env var.

ldez (changes requested):
I'm not a fan of `labeled` events with an `if` because the workflow will be triggered for all the `labeled` events and will create "noise" inside the GitHub action tab.
Sadly, it's not possible to define the type inside the `on` section.

This is the main reason why I never created this workflow.

bombsimon:
> I'm not a fan of `labeled` events with an `if` because the workflow will be triggered for all the `labeled` events and will create "noise" inside the GitHub action tab. Sadly, it's not possible to define the type inside the `on` section.

Makes sense! I can address your concern about the markdown file but if you don't think this is useful I think we can close it instead. Didn't know you considered it and opted not to do it.

> An alternative is to use `workflow_dispatch`.

Yeah maybe, it would add the list to the repo but would require a manual invocation so not sure if it's useful. I also don't know how to target a specific PR from a `workflow_dispatch`, didn't know you could add context for a singular PR.

ldez:
The best solution is to have "saved replies" by repository or organization, but GitHub doesn't seem to plan to add this feature:
- https://github.com/isaacs/github/issues/670
- https://github.com/orgs/community/discussions/4974

ldez:
I modified your PR to use `workflow_dispatch`: the PR number is defined during the launch of the workflow.

example

bombsimon:
> I modified your PR to use `workflow_dispatch`: the PR number is defined during the launch of the workflow.

Nice! Never used `inputs` so TIL. 

Since 100% of the uses of this checklist have been from you so far feel free to decide if you want this as an action or keep doing it the old style! If the latter just close this PR!

Another goal with the action was to add the comment as soon as possible without human interaction for the author of the PR to quickly see what's expected when adding a new linter. That won't be the case now so maybe just more work for you to trigger the workflow. 😕

ldez:
The approach allows sharing the comment content with all maintainers, it's positive.
The workflow dispatch allows all maintainers to set the comment "easily", it's also positive.

Currently, I will still use my "Saved Reply" in most cases because it's easier for me, but I will migrate to the workflow dispatch.

.github/workflows/new-linter-checklist.yaml
```diff
@@ -0,0 +1,35 @@
...
+          # removed and added back multiple times maliciously. It will reset the
+          # checklist but the previous values will be in the edit history.
+          comment_exist=$(gh pr view "$NUMBER" \
+            --json comments \
+            --jq '.comments[].author | select(.login=="github-actions") | .login' \
+            | wc -l)
+          [ "$comment_exist" -gt 0 ] && edit_last="--edit-last"
+
+          # Comment on the PR with the checklist.
+          gh pr comment "$NUMBER" --body "$(cat .github/new-linter-checklist.md)" $edit_last
```
bombsimon, line 31:
```suggestion
          gh pr comment "$NUMBER" --body "$(cat .github/new-linter-checklist.md)"
```

We can remove all this if it's a `workflow_dispatch` and not automatically on tagging.

ldez: we can keep it to prevent 2 launches of the workflow dispatch in parallel.

ldez: but we can update the comment about the reason for this check.

bombsimon: Updated/reduced comments.

ldez (approved):
LGTM

PR #4859, opened 2024-07-05
Add ctxcause linter

sivchari:
I added [ctxcause](https://github.com/sivchari/ctxcause) linter to golangci-lint.
This linter reports the use of context.WithCancel, context.WithTimeout, and context.WithDeadline instead of context.WithCancelCause, context.WithTimeoutCause, and context.WithDeadlineCause.

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -2588,6 +2588,7 @@ linters:
     - containedctx
     - contextcheck
     - copyloopvar
+    - ctxcause
     - cyclop
     - decorder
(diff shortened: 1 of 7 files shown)


Antonboom:
Hi!

You can just use **forbidigo** for such purposes

ccoVeille:
I opened the following issue to get more information

- https://github.com/sivchari/ctxcause/issues/2

ldez:
It can already be done with forbidigo, so we will not accept this linter.

PR #4871, opened 2024-07-14
Add iface linter

uudashr:
Add new linter "iface" to help discover and avoid interface pollution

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -1207,6 +1207,15 @@ linters-settings:
     # Default: false
     var-require-grouping: true
 
+  iface:
+    # By default set to empty. Leave it empty means all analyzers are enabled.
+    # Default: []
+    enable:
+      - unused
+      - empty
+      - duplicate
+      - opaque
+
   importas:
     # Do not allow unaliased imports of aliased packages.
     # Default: false
--- a/add-iface-patchfile
+++ b/add-iface-patchfile
@@ -1,0 +1,440 @@
+diff --git a/.golangci.next.reference.yml b/.golangci.next.reference.yml
+index 62b7fd65..fe85c519 100644
+--- a/.golangci.next.reference.yml
++++ b/.golangci.next.reference.yml
+@@ -1207,6 +1207,15 @@ linters-settings:
+     # Default: false
+     var-require-grouping: true
+ 
++  iface:
++    # By default set to empty. Leave it empty means all analyzers are enabled.
++    # Default: []
++    enable:
++      - unused
++      - empty
++      - duplicate
++      - opaque
++
+   importas:
+     # Do not allow unaliased imports of aliased packages.
+     # Default: false
+diff --git a/go.mod b/go.mod
+index cea54248..7340bb23 100644
+--- a/go.mod
++++ b/go.mod
+@@ -1,6 +1,6 @@
+ module github.com/golangci/golangci-lint
+ 
+-go 1.21.0
++go 1.22.5
+ 
+ require (
+ 	4d63.com/gocheckcompilerdirectives v1.2.1
+@@ -115,6 +115,7 @@ require (
+ 	github.com/ultraware/funlen v0.1.0
+ 	github.com/ultraware/whitespace v0.1.1
+ 	github.com/uudashr/gocognit v1.1.2
++	github.com/uudashr/iface v1.0.0
+ 	github.com/valyala/quicktemplate v1.8.0
+ 	github.com/xen0n/gosmopolitan v1.2.2
+ 	github.com/yagipy/maintidx v1.0.0
+diff --git a/go.sum b/go.sum
+index cd184b5c..d9a4d9e6 100644
+--- a/go.sum
++++ b/go.sum
+@@ -557,6 +557,8 @@ github.com/ultraware/whitespace v0.1.1 h1:bTPOGejYFulW3PkcrqkeQwOd6NKOOXvmGD9bo/
+ github.com/ultraware/whitespace v0.1.1/go.mod h1:XcP1RLD81eV4BW8UhQlpaR+SDc2givTvyI8a586WjW8=
+ github.com/uudashr/gocognit v1.1.2 h1:l6BAEKJqQH2UpKAPKdMfZf5kE4W/2xk8pfU1OVLvniI=
+ github.com/uudashr/gocognit v1.1.2/go.mod h1:aAVdLURqcanke8h3vg35BC++eseDm66Z7KmchI5et4k=
++github.com/uudashr/iface v1.0.0 h1:NoiiU7xzO9C/4xSMHH7Me7V2qZtRslVHKCIXYvA6VR8=
++github.com/uudashr/iface v1.0.0/go.mod h1:uapH12iu4iwziF3F0exELbQSgJuf3WIwKrNwphQ2xIc=
+ github.com/valyala/bytebufferpool v1.0.0 h1:GqA5TC/0021Y/b9FG4Oi9Mr3q7XYx6KllzawFIhcdPw=
+ github.com/valyala/bytebufferpool v1.0.0/go.mod h1:6bBcMArwyJ5K/AmCkWv1jt77kVWyCJ6HpOuEn7z0Csc=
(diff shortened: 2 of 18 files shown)


ldez:

- [ ] The CLA must be signed

### Pull Request Description

### Linter

- [ ] It must use Go version >= 1.21

### Others Requirements

### Recommendations

---

ldez:
> If the author of the PR is a member of the golangci-lint team, he should not edit this message.

@uudashr please don't edit the message.

ldez:
@uudashr stop editing the message.

uudashr:
Sorry, reverting the message

ldez:
You reverted all my changes...

ldez:
As long as the elements required by the checklist (except those related to maintainer actions) are not handled, the review will not start.

Don't edit the checklist!

uudashr:
FYI the linter repo link: https://github.com/uudashr/iface

ldez:
> - `empty`: Finds empty interfaces.

This could lead to false positive because an empty interface can be used on purpose. 

> - `unused`: Finds unused interfaces within the package.

This leads to false positive because you can provide interface shared by other packages to avoid duplication.

Note: it's not "the package" but "a package"

> - `duplicate`: Finds duplicate interfaces within the package.

This leads to false positive with extended interfaces.

```go
type a interface {
	// ...
}

type b interface {
	a
}
```

Note: it's not "the package" but "a package"

> - `opaque`: Find the interfaces that is used to abstract a single concrete implementation only.

This leads to false positives because a library can provide an interface without implementation to extend the behavior.

uudashr:
> `empty`: Finds empty interfaces.
> 
> This could lead to false positive because an empty interface can be used on purpose.

Correct. Thanks for pointing this out.

I've seen some library use `interface{}`, this is allowed.
But still empty "[interface{} says nothing](https://go-proverbs.github.io)", you need to have careful consideration to use it.

If we define `type Payload interface {}`, it still says nothing unless the name tells that it is Payload, but any value can be assumed as Payload.

We can use `type Payload any` as alternative which is allowed if still needed. Or we can make the `empty` linter works only for Go1.18 or above?

Any suggestion?

The linter still able to help us raising awareness so  we can make an extra consideration when using empty interface.

---
> `unused`: Finds unused interfaces within the package.
> 
> This leads to false positive because you can provide interface shared by other packages to avoid duplication.

I'm taking below statements as consideration:
1.  "Go interfaces generally belong in the package that uses values of the interface type, not the package that implements those values. The implementing package should return concrete (usually pointer or struct) types: that way, new methods can be added to implementations without requiring extensive refactoring." [Go Code Review Comments, go.dev](https://go.dev/wiki/CodeReviewComments#interfaces)
2.  "Don’t export any interfaces until you have to." [Interface pollution in Go, rakyll.org](https://rakyll.org/interface-pollution/)

I would love to have `Logger` interface in built-in Go library, but they didn't take this direction, It might because of above statements.

---
> `duplicate`: Finds duplicate interfaces within the package.
> 
> This leads to false positive with extended interfaces.
>
> ```
> type a interface {
> 	// ...
> }
> 
> type b interface {
> 	a
> }
> ```

Why you want to do that?

Technically Go allow us to pollute our code with interfaces, but again the questions is why you want to do that?

If we have very large package that host many concepts, having duplicate package is possible. But then you might want to consider breakdown the package into more specific concept, the duplication is less likely.

---
> `opaque`: Find the interfaces that is used to abstract a single concrete implementation only.
> 
> This leads to false positives because a library can provide an interface without implementation to extend the behavior.

This is related to function that hides the concrete value, for example
```go
type Server interface {
    Server() error
}
func NewServer(addr string) Server { // will report warning
    return &serverImpl{addr: addr}
}
```

So we suggest to just return the concrete exported struct.

But below code is valid, no warning will be reported. It has possibility to return more than one implementation.
```go
type Server interface {
    Server() error
}


func NewServer(addr string) Server { // will not report warning
  if strings.HasPrefix(addr, "secure:") {
    return &secureServer{addr: addr}
  }

  return &server{addr: addr}
}
```

ldez:
My remarks were based on real projects with legitimate usages.

The theory is great but it should not be turned into a dogma.

You can try to justify your opinion but it is still an opinion :wink: 

---

About `empty`:

For me, this rule is the only one in which false positives are niche.

About `unused`:

When you are designing a complex library, providing an interface in a separate package can be required for several reasons:
- avoiding cyclic imports
- have a clear public API
- etc.

`log` has no interface, but `slog` has one because it's a requirement to use a logger.
Not a lot of projects use the `log` because, in the real world, you need to interact with other loggers so you need an interface.

About `duplicate`:

> Why do you want to do that?

Because of readability: complex things require to be named to be easy to maintain.

About `opaque`:

I will repeat but a library can provide an interface without implementation to extend the behavior.
A, not so real, example, but I think it will be clearer: if you are building a SQL-neutral library then you don't want to have a real implementation of the SQL part.

uudashr:
About `empty`:
> For me, this rule is the only one in which false positives are niche.

This looks valid. Made me think to take out this one. Thanks man. 👍 

Not many people will use empty interface unless they might really need it. If this linter really insist to be exist, will not much useful either.

About `unused`:
> - avoiding cyclic imports
> - have a clear public API
> - etc.

If you have reference, a project or real example on this, I would love to continue my research on what should we do to avoid interface pollution.

> `log` has no interface, but `slog` has one because it's a requirement to use a logger.
Yeah... I didn't use `log` either.

FYI the linter won't yield warning to `slog`.

About `duplicate`:

> Because of readability: complex things require to be named to be easy to maintain.

Been thinking of this before to justify whether the linter is not necessary to be exists. But hardly find real example on this.

If you have some, please drop me a the repo link.

About `opaque`:

> I will repeat but a library can provide an interface without implementation to extend the behavior.

The linter would not yield warning due to this.

> A, not so real, example, but I think it will be clearer: if you are building a SQL-neutral library then you don't want to have real implementations of the SQL part.

I'm not sure this is relate to `opaque` analyzer. I would appreciate if we have this SQL-neutral library for me to learn more about it.

Well I might need to rephrase the description of the analyzer, here it is: "identifies functions returning interfaces when the actual return value is always a single concrete implementation".

It is related to "factory function" mentioned [Avoid Interface Pollution](https://www.ardanlabs.com/blog/2016/10/avoid-interface-pollution.html)

ldez:
I updated your PR to ease the detection of false-positives.

uudashr:
A lot of conflict from the master. Any suggestion, I can do new PR based on master so we can have fresh and minimal changes.

> I updated your PR to ease the detection of false-positives.
@ldez  thanks for this, I use this and made some changes with same goal.

uudashr:
> You have reverted all my changes and did a bad rebase.
Sorry of that, my bad.

> I fixed what I could but please avoid that.
Thanks for the fix. Sure do

Need to catch up with huge master diff, I do confuse myself having this big rebase.
If you need fresh PR or anything else.. please let me know.

> I updated your PR to ease the detection of false-positives.
Thanks for this, I use this and made some changes with same goal.

ldez:
> I need to catch up with huge master diff, I do confuse myself having this big rebase.

The problem was not a "huge master diff/big rebase" but the fact that you didn't pull my changes before modifying the code locally despite my message notifying you about this change.

I fixed the rebase but my previous code has been deleted.

ldez:
I fixed:
- configuration management
- JSONSchema
- tests
- implementation
- etc.

**But this is only to illustrate my opinion with real world examples.**

My opinion is still the same:
- `unused` -> more than 90 % of false positives
- `opaque` -> at least 80 % of false positives
- `identical` (previously `duplicate`) -> 20 % of false positives and non-clear report messages.

For real-world examples, see the "Details" section [here](https://github.com/ldez/golangci-lint-bench/issues/14#issuecomment-2359655059).

uudashr:
@ldez thanks, let me check for those each projects

ldez:
You will not be able to fix something because it's related to to how the nature of your linters.

The best and only way is to change the default to disable `unused` and `opaque`.

uudashr:
I see. Let me change those

ldez (approved):
LGTM

PR #4875, opened 2024-07-21
test: validate sarif output with json schema

Zxilly:


ldez:
https://github.com/golangci/golangci-lint/pull/4831#issuecomment-2179496684

Zxilly:
Obviously we still have work to do on the SARIF format, and this kind of detection ensures that the sarifs we generate are always valid. We can avoid the problem that we have generated a valid json, but not a valid sarif.

ldez:
> Obviously we still have work to do on the SARIF format

I don't share your opinion.

> this kind of detection ensures that the sarifs we generate are always valid. We can avoid the problem that we have generated a valid json, but not a valid sarif.

This kind of detection will do nothing more than a simple test.
To be detected a problem should exist inside a test case.
However a problem doesn't exist by itself, it requires a specific test case, so the problem will never be detected by the validation.

PR #4915, opened 2024-08-15
exportloopref: disable linter for Go versions >= Go 1.22

leonklingele:
The problem reported by the linter has been fixed with Go 1.22.
See https://go.dev/blog/loopvar-preview.


ldez:
I prefer to deprecate the linter: #4916

PR #4916, opened 2024-08-15
exportloopref: deprecation

ldez:
Since Go1.22 (and go1.21 with `GOEXPERIMENT=loopvar`), `exportloopref` is no longer relevant.

It should be replaced by `copyloopvar`.

It's a [level 1 deprecation](https://golangci-lint.run/product/roadmap/#linter-deprecation-cycle) (the linter still runs) but we will keep the level 1 until the release of go1.24.

And as golangci-lint requires min go1.22 to compile we can drop variable copies inside the code.

Reminder: the Go team only maintains the 2 latest "minor" versions of Go (go1.22 and go1.23), and go1.22 was released on 2024-02-06, so it's highly recommended not to use go1.21 or lower.


leonklingele:
What’s the benefit of keeping the linter active when using Go >= 1.22? All of its output is then no longer relevant and should be ignored, really all of it.
In addition to the deprecation, it imo thus makes sense to also disable the linter for the relevant versions of Go as done with https://github.com/golangci/golangci-lint/pull/4915.

ldez:
Because the priority is for users to stop using it, they should remove it from their configuration, so the "false-positives" reports will be a reminder.
If we disable the linter, the users will be less insisted on removing it from their configuration.

leonklingele:
What about people using `--enable-all`? Will they really need to explicitly pass `--disable exportloopref` to prevent the false positives from appearing? Looks quite inconvenient to me — we already know no one on Go 1.22+ wants to see these.

ldez:
This is how the depreciation cycle works.

bombsimon (approved):
LGTM. I agree not "hiding" this and asking users to disable deprecated linters is the right way so they're aware they will go away. This one might be special since it's bot needed but better to be consistent. 

I don't know how common enabling all without a config file is but if you have a config file it's a one time-one line fix.

PR #4918, opened 2024-08-16
feat: add junit-xml-extended format

### LINKED ISSUE #4900: Include `file` and `line` attribute for junit report output format
### Welcome

Various tools like [action-junit-report](https://github.com/mikepenz/action-junit-report) rely on the `file` and `line` testcase attributes in order to report errors nicely inline the file. 

 `file` and `line` testcase attributse are also part of the Common JUnit XML Format:
https://github.com/testmoapp/junitxml?tab=readme-ov-file#complete-junit-xml-example

It'd be nice to include that information in golangci-lint reports as well.

Include  `file` and `line` testcase attributes like so:
```xml
<testsuites>
  <testsuite name="go/internal/pkg/set/set_test.go" tests="1" errors="0" failures="1">
    <testcase name="unused" classname="go/internal/pkg/set/set_test.go:15:6" file="go/internal/pkg/set/set_test.go" line="15">
      <failure message="go/internal/pkg/set/set_test.go:15:6: func `unused` is unused" type=""><![CDATA[: func `unused` is unused
Category: unused
File: go/internal/pkg/set/set_test.go
Line: 15
Details: func unused(t *testing.T) {]]></failure>
    </testcase>
  </testsuite>
</testsuites>
```

Parsing the file and line number from the classname is too brittle. 

### Additional context

_No response_

### Supporter

ldez:
Fixes #4900

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -70,6 +70,7 @@ output:
   # - `checkstyle`
   # - `code-climate`
   # - `junit-xml`
+  # - `extended-junit-xml`
   # - `github-actions`
   # - `teamcity`
   # - `sarif`
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -70,7 +70,7 @@ output:
   # - `checkstyle`
   # - `code-climate`
   # - `junit-xml`
-  # - `extended-junit-xml`
+  # - `junit-xml-extended`
   # - `github-actions`
   # - `teamcity`
   # - `sarif`
--- a/pkg/printers/junitxml.go
+++ b/pkg/printers/junitxml.go
@@ -30,6 +30,8 @@ type testCaseXML struct {
 	Name      string     `xml:"name,attr"`
 	ClassName string     `xml:"classname,attr"`
 	Failure   failureXML `xml:"failure"`
+	File      string     `xml:"file,attr,omitempty"`
+	Line      int        `xml:"line,attr,omitempty"`
 }
 
 type failureXML struct {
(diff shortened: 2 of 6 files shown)


bombsimon (approved):
Is the plan to use this extended version for anything that extends the `junit-xml` or would we have to create yet another format if there are other fields that might be conditional? If the latter it doesn't sound like it scales well but otoh we don't seem to extend the format very often so maybe not worth worrying about.

ldez:
This format will be used for everything that deviates from the "basic" JUnit format.

.golangci.next.reference.yml
```diff
@@ -70,6 +70,7 @@ output:
   # - `checkstyle`
   # - `code-climate`
   # - `junit-xml`
+  # - `extended-junit-xml`
```
Antonboom, line 73: maybe `junit-xml-extended`?

just for better ordering

pkg/printers/junitxml.go
```diff
@@ -39,11 +41,15 @@ type failureXML struct {
 }
 
 type JunitXML struct {
-	w io.Writer
+	extended bool
+	w        io.Writer
```
Antonboom, line 45: why not 

```go
type JunitXMLExtended struct {
    JunitXML
}
```

or inheritance no make sense here?

Antonboom (approved):

.golangci.next.reference.yml
ldez: I use the same pattern as `colored-line-number`/`line-number`, `colored-tab`/`tab`.

pkg/printers/junitxml.go
ldez: It is not really useful in this context: the "extended" coded is inside a loop, I will not create a base struct that takes a function as a parameter (to use it inside the loop) + another struct + a method/function, just to avoid using a boolean.

PR #4926, opened 2024-08-19
dev: fix GO_VERSION in post release workflow

silas:
There are multiple jobs in the GitHub post-release workflow that reference the `GO_VERSION` environment variable, but it is only defined for the `update-docs` (aka `Update readme`) job.

This means that the static assets for version `1.60.1` are built with Go `1.22` instead of `1.23`.

You can confirm this by comparing the version of Go used in the `actions/setup-go@5` step of the [Update readme](https://github.com/golangci/golangci-lint/actions/runs/10379882256/job/28738784965) job to the one in [Update GitHub Action assets](https://github.com/golangci/golangci-lint/actions/runs/10379882256/job/28738785585).

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -4,6 +4,13 @@ on:
     types:
       - published
 
+env:
+  # https://github.com/actions/setup-go#supported-version-syntax
+  # ex:
+  # - 1.18beta1 -> 1.18.0-beta.1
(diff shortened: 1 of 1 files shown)


ldez:
> This means that the static assets for version 1.60.1 are built with Go 1.22 instead of 1.23.

This only impacts the generators of docs and JSON files, there is no impact on the documentation or the public assets.

ldez (approved):

PR #4196, opened 2023-11-12
Add`gofactory` linter

maranqz:
The linter checks that the structes are created by the factory, and not directly.
The checking helps to provide invariants without exclusion and helps avoid creating an invalid object.

PR #4937, opened 2024-08-21
Add constructorcheck linter

reflechant:
[constructorcheck](https://github.com/reflechant/constructor-check ) reports places where you create values directly from a type ( `T{}` or `new(T)` ) ignoring a constructor `NewT` defined for this type in the same package as the type.

This is useful to avoid skipping initialization/validation for a type.

Types defined in the standard library are excluded from analysis.

The linter doesn't have any parameters at the moment.


ldez:

- [ ] The CLA must be signed

### Pull Request Description

### Linter

- [ ] It must use Go version <= 1.21

### Others Requirements

### Recommendations

---

ldez:
It feels like the same thing as #4196

reflechant:
Yeah, looks similar indeed.
There are 2 differences:
1) constructorcheck ignores types from the standard library (there are too many exceptions to the rule there like bytes.Buffer whose zero value is safe to use while there is also bytes.NewBuffer)
2) constructorcheck tells you the name of the constructor to use

ldez:
> 1. constructorcheck ignores types from the standard library (there are too many exceptions to the rule there like bytes.Buffer whose zero value is safe to use while there is also bytes.NewBuffer)

Also [this code](https://github.com/reflechant/constructor-check/blob/a1f8106e06918978c9bbbe9cc535d8d1ebe4f105/analyzer/analyzer.go#L266) is a problem.

> constructorcheck tells you the name of the constructor to use

It's a minor difference and I feel this will lead to wrong suggestions when there are several "constructors".

reflechant:
BTW, is there a different way to get the list of standard library packages? This hack with calling `go tool` was the only way I could find.

ldez:
This should never be called at runtime.

reflechant:
It uses a very strict definition of what is a constructor - it's a function NewT, where T is the name of a type, existing in the same package, that returns only one value of the same type T. You can not have two of them in the same package because every function name must be unique within a package. I don't see a potential for a confusion here.

As for the standard library package list - I agree that I need to find a different way to do it.

ldez:
> I don't see a potential for a confusion here.

I can easily see it:

```go
type Foo struct{}

func NewFoo(a string) *Foo           { return nil }
func NewFooWithOption(b string) *Foo { return nil }
```

```go
type Bar struct{}

// Deprecated: use NewBarWithOption
func NewBar(a string) *Bar           { return nil }
func NewBarWithOption(b string) *Bar { return nil }
```

What will be the constructor suggested by your linter?

reflechant:
Finally good feedback :)
All valid points. I need to fix them. Should I close this PR until I do it?

reflechant:
Yes, also sometimes constructors return something like this:
`(T, error)` `(*T, error)`, `(T, bool)` `(*T, bool)`

But I'm much more worried about being misleading and giving false error messages than missing something. The latter makes the linter not 100% efficient, the former makes it unusable.

reflechant:
> [This code](https://github.com/reflechant/constructor-check/blob/a1f8106e06918978c9bbbe9cc535d8d1ebe4f105/analyzer/analyzer.go#L266) is a major problem.

Go itself does it [all over the place](https://github.com/search?q=org%3Agolang+exec.Command%28%22go%22&type=code).

I will try to dig how the go tool itself does it and see if it makes sense to copy the implementation from there.

.golangci.next.reference.yml
Mehdi9865, line 1: go.mod

Mehdi9865 (commented):
Ok

reflechant:
I started working on a complete revamp of the linter after some feedback here. The idea is to make it "opt-in" instead of "opt-out". But when I finish it it will be another PR

Issue #4933, opened 2024-08-21
When running the linter, my PC crashes or the linter dies


PatricioPoncini:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

Hi, for some reason, updating the golint version is causing my PC to crash. Essentially, the PC freezes and I can do nothing but restart it. I have tried this multiple times, and the result is always the same: either the PC crashes or the process is canceled and dies. I have updated the library, golang, and cleared the cache. I don't know what else could be causing this problem. Has anyone else experienced this?

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version v1.60.2 built with go1.22.3 from (unknown, modified: ?, mod sum: "h1:Y8aWnZCMOLY5T7Ga5hcoemyKsZZJCUmIIK3xTD3jIhc=") on (unknown)
```

</details>

### Configuration

<details>

```console
run:
  timeout: 5m
output:
  sort-results: true
linters:
  enable:
    - asciicheck
    - dogsled
    - durationcheck
    - errcheck
    - errorlint
    - gofumpt
    - exportloopref --> i will change this
    - gofmt
    - goimports
    - gosec
    - misspell
    - nakedret
    - nilerr
    - nolintlint
    - revive
    - wastedassign
    - gocritic
    - godot
    - predeclared
    - unused
    - testifylint

linters-settings:
  testifylint:
    enable-all: true
  gofumpt:
    extra-rules: true
  gosec:
    includes:
      - G102
      - G106
      - G108
      - G109
      - G111
      - G112
      - G201
      - G203

issues:
  exclude-rules:
    - linters:
        - structcheck
        - unused
      text: '`data` is unused'
    - linters:
        - revive
      text: 'var-naming:'
    - linters:
        - revive
      text: 'exported:'
[...]

Issue #4932, opened 2024-08-21
staticcheck uses too much memory and crashes


tang-fh:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [X] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

golangci-lint uses too much memory (and is killed) when running against the sources from https://github.com/prometheus/blackbox_exporter/

I have narrowed it down to the `staticcheck` linter.

Workarounds I have found:
- disable staticcheck linter, e.g. commenting out https://github.com/prometheus/blackbox_exporter/blob/3dd5dfeaabc630ca0c2ec722a07f9755159ae0dd/.golangci.yml#L6
-  use go1.22.6

### Version of golangci-lint

<details>

```console
$ ~/go/bin/golangci-lint --version
golangci-lint has version 1.56.2 built with go1.22.0 from 58a724a0 on 2024-02-15T18:01:51Z
```

</details>

### Configuration

<details>

```console
$ cat .golangci.yml 
# Run only staticcheck for now. Additional linters will be enabled one-by-one.
linters:
  disable-all: true
  enable:
    - misspell
    - staticcheck
```

</details>

### Go environment

<details>

```console
$ go version && go env
go version go1.23.0 linux/amd64
GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/fhltang/.cache/go-build'
GOENV='/home/fhltang/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/fhltang/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/fhltang/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
[...]

Issue #4909, opened 2024-08-14
Under go 1.23, golangci-lint uses all available memory


jaevans:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [X] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

Updating the toolchain line to `1.23.0` in `go.mod` causes the linter to run out of memory. I've duplicated this on both linux x86_64 and darwin arm64. I haven't yet identified which linter is causing the problem, but it's one of the default set

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version 1.59.1 built with go1.22.4 from 1a55854 on 2024-06-08T22:05:49Z
```

</details>

### Configuration

<details>

No configuration file or command line arguments passed.

</details>

### Go environment

<details>

```console
$ go version && go env
go version go1.23.0 darwin/arm64
GO111MODULE=''
GOARCH='arm64'
GOBIN=''
GOCACHE='/Users/jaevans/Library/Caches/go-build'
GOENV='/Users/jaevans/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='arm64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/jaevans/go/pkg/mod'
GONOPROXY='gitlab.sicura.dev/*'
GONOSUMDB='gitlab.sicura.dev/*'
GOOS='darwin'
GOPATH='/Users/jaevans/go'
GOPRIVATE='gitlab.sicura.dev/*'
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/Users/jaevans/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.23.0.darwin-arm64'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
[...]

Issue #4908, opened 2024-08-14
gosimple + go 1.23 consume all available RAM


olegrok:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [X] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

It is quite inadequate to use more than 60 GB of RAM.

![изображение](https://github.com/user-attachments/assets/73e54402-baed-4bd4-9167-ba7f21bd5146)

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version 1.60.1 built with go1.22.5 from v1.60.1 on 19700101-00:00:00
```

</details>

### Configuration

<details>

```console
golangci-lint run -j 4 -v --go 1.23 --enable-only gosimple
```

</details>

### Go environment

<details>

```console
$ go version && go env
go version go1.23.0 linux/amd64
GO111MODULE='on'
GOARCH='amd64'
GOBIN=''
GOCACHE='/tmp/sha256-PObLD2gU7qm3qL3G4JqrBtqTw5i08ab+oNSM+GAQu60=/go-cache'
GOENV='/home/oleg/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS='-modcacherw -mod=vendor -trimpath'
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/tmp/sha256-PObLD2gU7qm3qL3G4JqrBtqTw5i08ab+oNSM+GAQu60=/go/mod/cache'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/tmp/nix-shell.S8T84v/go'
GOPRIVATE=''
GOPROXY='off'
GOROOT='/nix/store/0k144pnsvy0wa3dcl1y3df7d4zskylc4-go-1.23.0/share/go'
[...]

PR #4938, opened 2024-08-21
feat: check that Go version use to build is greater or equals to the Go version of the project

ldez:
To avoid issues related to the new version of Go, I would add a check to ensure that the Go version used to build is compatible with the Go version used by a project. 

Related to all the issues after each new release of Go.
A small extract of the related issues: #4909, #4908, #4932, #4933, etc.

Output example:
```console
$ ./golangci-lint run
Error: can't load config: the Go language version (go1.22) used to build golangci-lint is lower than the targeted Go version (1.23)
Failed executing command with error: can't load config: the Go language version (go1.22) used to build golangci-lint is lower than the targeted Go version (1.23)
```

### DIFF
--- a/pkg/config/config.go
+++ b/pkg/config/config.go
@@ -1,8 +1,11 @@
 package config
 
 import (
+	"fmt"
+	"go/version"
 	"os"
 	"regexp"
+	"runtime"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


bombsimon (approved):
🎉

PR #4927, opened 2024-08-20
build(deps): bump github.com/securego/gosec/v2 from 5f0084eb01a9 to 81cda2f91fbe

ldez:
This PR is related to #4906 (and #4904), the analyzers (G602, G115) can be filtered now.

The rules G115, G405, G406, G506, and G507 are added (kind of side effect of the update).

https://github.com/securego/gosec/compare/5f0084eb01a9...81cda2f91fbe

The update is done by hand because gosec is not released yet and we already use a pseudo version because of #4748.

Note: exclusions are defined inside the configuration due to the compatibility requirements with the previous version of golangci-lint inside our tests.

PR #4941, opened 2024-08-22
gosec: add G115 to default exclusions

### LINKED ISSUE #4935: v1.60.2 reports integer overflow conversion int -> int32 for conversion like int32(group) with group is an int
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).

v1.60.2 starts to report "integer overflow conversion int -> int32 for conversion like int32(group) with group is an int", but v1.60.1 does not report this error.

### Version of golangci-lint

v1.60.2

### Configuration

Using v1.60.2

v1.60.1 is fine.

### Go environment

1.22.4

### Verbose output of running

G115: integer overflow conversion int -> int32 (gosec)
	drqsGroup := int32(group)

group is an int defined.

internal project. no link.

### Validation

- [X] Yes, I've included all information above (version, config, etc.).

### Supporter

ldez:
Fixes #4935, #4939

Related to #4927

### DIFF
--- a/pkg/config/issues.go
+++ b/pkg/config/issues.go
@@ -101,6 +101,12 @@ var DefaultExcludePatterns = []ExcludePattern{
 		Linter:  "revive",
 		Why:     "Annoying issue about not having a comment. The rare codebase has such comments.",
 	},
+	{
+		ID:      "EXC0016",
+		Pattern: `G115: integer overflow conversion`, // rule: package-comments
+		Linter:  "gosec",
+		Why:     "Too many false positives.",
+	},
 }
 
 type Issues struct {


ccoVeille:
Thanks, it will help a lot until gosec discussion ends

bombsimon (commented):
Hmm, is this a good idea? In general I really don't like `exclude-use-default` (all rules should be up to the user to configure) and to me that's legacy that should be removed. It's been brought up in previous discussions (such as #4323, #2239 and #456).

As far as I can tell, G115 _is_ a valid check given it doesn't have false positives. So if this gets excluded by default and then gets fixed upstream we'd no longer warn about actual overflow and instead hide them from users that enables `gosec`.

And as far as I can tell from the linked issues, since we don't change existing exclusions due to backwards compatibility reasons that means that if we add this we can't remove it until a new major version. 

Also, isn't this (partly) fixed by https://github.com/securego/gosec/pull/1188 and more fixes are to come upstream?

ldez:
We have no choice if we want to disable a rule inside gosec rather than using default exclusion.
I also always set it to `false` but at least we will reduce the noise for some users.

> As far as I can tell, G115 is a valid check given it doesn't have false positives.

Currently, it has false positives: `uint8 -> int64`, `uint8 -> int`, etc.

But https://github.com/securego/gosec/pull/1188 can reduce the problem.

ldez:
Closed in favor of #4943

bombsimon:
> We have no choice if we want to disable a rule inside gosec rather than using default exclusion.
> I also always set it to false but at least we will reduce the noise for some users.

Alright, thanks for clarifying! I get that the noise can be annoying but afaict `gosec` isn't one of the default enabled linters in `golangci-lint` so it's not something we need to suppress for users running without a config/default config at least. So given that the linter must be enabled by the user (specifically or with `enable-all`), an alternative that is easier to change could be to note this in the reference config.

I'm open to reconsidering this but if the manual bumping you just did solves the _new_ false positives and the old ones stay the same (or will be fixed in the near future) I think it would be nice to avoid extending this list.

ldez:
> We have no choice if we want to disable a rule inside gosec rather than using default exclusion.

I want to add a more detailed explanation of this sentence:
gosec doesn't have a default configuration and doesn't allow a default configuration.
By using default exclusion, we let the user have the possibility to enable this rule without unnecessary complexity for us to handle interaction with `include/exclude`.

But, for now, we have another solution, so we are "safe" :smile_cat:

proton-ab:
Quite ironic that this was marked as `won't fix` while you yourself recommend end-users to disable G115 (and people will - because it still ignores bound checks). In my opinion, telling users to ignore it creates more mess as users will inevitably forget to enable it back whenever its enabled back, while you (`golangci-lint` team) do follow `gosec` development closely (for regular updates) and could easily decide to re-enable it back once https://github.com/securego/gosec/pull/1189/files lands.

I really fail to see your logic in this decision.

ldez:
I recommended disabling it as a workaround because of false positives just after the v1.60.2 release.
It was because there were no other solutions before creating a new release of golangci-lint.

- v1.60.2 -> G115 is enabled by default inside gosec but there are false positives
- AFTER v1.60.2 and BEFORE v1.60.3, there are 2 solutions: gosec fixes or default exclusions
   - I create this PR to add G115 to the default exclusions
   - I closed this PR BECAUSE gosec created a fix
- v1.60.3 -> the false positives of G115 are fixed

The logic in this decision is clear if you watch the chronology of the events.

PR #4957, opened 2024-08-26
update description for fatcontext

Crocmagnon:
Since v0.5.0, fatcontext now detects nested contexts in function literals too.
I'm not updating the linter itself because dependabot will take care of it. Please tell me if I should.

### DIFF
--- a/assets/linters-info.json
+++ b/assets/linters-info.json
@@ -319,7 +319,7 @@
   },
   {
     "name": "fatcontext",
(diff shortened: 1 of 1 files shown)


ldez:
Hello,

this file is generated for linter information, it should not be handled by hand.

You should update your linter description: https://github.com/Crocmagnon/fatcontext/blob/90afb8d3fae26ac02dc10c0e332b63922fda7482/pkg/analyzer/analyzer.go#L19

PR #4963, opened 2024-08-28
Report packages loading failure with a specific  exit-code

podtserkovskiy:
The `packages.Load()` function can use an [external package provider `GOPACKAGESDRIVER`](https://pkg.go.dev/golang.org/x/tools/go/packages#hdr-The_driver_protocol) for third-party build-systems like Bazel or Buck2.

Having a specific exit code when external driver fails would increase observability and will help to distinguish `GOPACKAGESDRIVER`'s failures from other kinds of failures.

### DIFF
--- a/pkg/exitcodes/exitcodes.go
+++ b/pkg/exitcodes/exitcodes.go
@@ -1,5 +1,9 @@
 package exitcodes
 
+import (
+	"fmt"
+)
+
 const (
 	Success = iota
 	IssuesFound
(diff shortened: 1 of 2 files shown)


podtserkovskiy:
@ldez Is there something I can do to get this PR merged?

ldez:
I'm unsure to understand how this "increases observability and will help to distinguish `GOPACKAGESDRIVER`'s failures from other kinds of failures".
The error message is already "unique" :thinking: 

Can you provide more details?

PR #4967, opened 2024-08-30
feat: exclude Swagger Codegen files

### LINKED ISSUE #4965: Autogenerated files from swagger-codegen not ignored
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).

Autogenerated files from `swagger-codegen` are not ignored by golangci-lint. 

Really, this is something that should be fixed on the swagger-codegen side (and there is an open bug [here](https://github.com/swagger-api/swagger-codegen/issues/12358)), because it does not include the standard `^// Code generated .* DO NOT EDIT\.$` pattern.

The golangci-lint [configuration docs](https://golangci-lint.run/usage/configuration/) say:
```
# There is no need to include all autogenerated files,
# we confidently recognize autogenerated files.
# If it's not, please let us know.
```
So I thought I'd report it anyway.

I also tried setting `exclude-generated: lax`, but this does not exclude these files either.

You could add an ignore pattern for `Generated by: Swagger Codegen` to catch these files.

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
v1.59.1
```

</details>

### Configuration

<details>

```console
# paste configuration file or CLI flags here
```

</details>

### Go environment

<details>

```console
$ go version && go env
go version && go env
go version go1.22.1 darwin/amd64
GO111MODULE='on'
GOARCH='amd64'
GOBIN=''
GOCACHE='/Users/user/Library/Caches/go-build'
GOENV='/Users/user/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMODCACHE='/Users/user/go/pkg/mod'
GONOPROXY='gitlab.com/foo/*,github.com/foo/*'
GONOSUMDB='gitlab.com/foo/*,github.com/foo/*'
GOOS='darwin'
GOPATH='/Users/user/go'
GOPRIVATE='gitlab.com/foo/*,github.com/foo/*'
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/darwin_amd64'
GOVCS=''
GOVERSION='go1.22.1'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='clang'
CXX='clang++'
CGO_ENABLED='1'
GOMOD='/Users/user/go/src/github.com/foo/bar/src/baz/go.mod'
GOWORK=''
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
PKG_CONFIG='pkg-config'
GOGCCFLAGS='-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/z4/tjbsqpbj5pz3_mh9jzb4vhxw0000gn/T/go-build336395111=/tmp/go-build -gno-record-gcc-switches -fno-common'
```

</details>

### Verbose output of running

<details>

```console
$ golangci-lint cache clean
$ golangci-lint run -v
# paste output here
```

</details>

<details>

Generate swagger files using swagger-codegen (docker example for portability):
```
docker run --rm -v ${PWD}:/local swaggerapi/swagger-codegen-cli generate \
    -i https://petstore.swagger.io/v2/swagger.json \
    -l go \
    -o /local/out/go
```

Run golangci-lint against those files. It'll probably fail against things like `revive` for missing exported var comments, `lll` for being too long, etc.

</details>

### Validation

- [X] Yes, I've included all information above (version, config, etc.).

### Supporter

ldez:
Fixes #4965

The pattern used by Swagger Codegen is here: https://github.com/swagger-api/swagger-codegen/blob/61cfeac3b9d855b4eb8bffa0d118bece117bcb7d/modules/swagger-codegen/src/main/resources/go/partial_header.mustache#L16

Related issue: https://github.com/swagger-api/swagger-codegen/issues/12358


bombsimon (approved):
Should we make this configurable so the user can set their own pattern?

ldez:
I don't think we need a configuration: the `exclude` options are enough in the majority of the cases. 
The `AutogeneratedExclude` processor is specialized for the generated files that are known patterns, and it's already weird to have the "lax" mode.

This PR is a type of "workaround" before Swagger changes its pattern to follow the standard because Swagger should change its pattern.

bombsimon:
> This PR is a type of "workaround" before Swagger changes its pattern to follow the standard because Swagger should change its pattern.

Why is Swagger different from any other code that diverges from the standard such that we should add workarounds? I agree lax mode is a bit weird, if the user could configure their own patterns we could get rid of it and only support the official pattern. Anyone using code that doesn't follow that (like Swagger) could add their own patterns.

ldez:
> Why is Swagger different from any other code that diverges from the standard such that we should add workarounds?

It's not so different but Swagger/OpenAPI is, I think the most popular format for API documentation.
So it is worth handling that, but to be honest, I think, at some point, we may remove the "lax" mode.

People should follow the Go standard comment pattern for generated code, it's the standard.

Antonboom (approved):

PR #4974, opened 2024-09-03
feat: support searching for config in XDG_CONFIG_HOME and $HOME/.config

k4yt3x:
This code will, in addition to the current directories being searched, also look for the config file in `XDG_CONFIG_HOME/golangci` and `$HOME/.config/golangci`.

For desktop applications, I find it not very elegant to leave the `.golangci.yml` in my home directory. I usually prefer to put config files in `XDG_CONFIG_HOME` (`$HOME/.config`) to avoid making a mess in the home directory. I do not want to manually specify a config path with `--config` since it will overwrite the repository's config if one exists. I reviewed the code and found that `golangci-lint` does not seem to support finding the config in `XDG_CONFIG_HOME`, so I'm submitting this PR in recommendation of adding this support.

## Setup

I have a configuration file located at `$HOME/.config/golangci/golangci.yml`. I am using `golangci-lint`'s repository for the demo below. I have removed the repository's built-in `.golangci.yml` so it will search for the global config.

## Previous Behavior

Previously, `golangci-lint` does not search for the said directories above, only `$HOME`:

image

## New Behavior

Now, if the old recursive search does not yield any results, it will also search the global config directories:

image

...and if we restore the repo config it still finds that first:

image

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -113,7 +113,15 @@ func (l *Loader) setConfigFile() error {
 			l.viper.SetConfigType("yaml")
 		}
 	} else {
-		l.setupConfigFileSearch()
+		l.setupRecursiveConfigFileSearch()
+
+		// If recursive search didn't find any config file, try searching in the global config paths.
+		if err := l.viper.ReadInConfig(); err != nil {
+			var configFileNotFoundError viper.ConfigFileNotFoundError
+			if errors.As(err, &configFileNotFoundError) {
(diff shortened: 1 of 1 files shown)


ldez:
related to #4554

k4yt3x:
> I suggest looking for the issues or creating an issue before opening this type of PR.

I actually did search, for both issues and PRs, neither returned what I was looking for.

PR #4983, opened 2024-09-04
gosec: disable G407

ldez:
G407 has been introduced by gosec v2.21.0 (#4981) but there are problems:
- https://github.com/securego/gosec/issues/1211
- https://github.com/securego/gosec/issues/1209

So I globally disabled this rule (no user bypass like the previous try to disable G115) because I don't want to live the same stories as G115.


bombsimon (approved):
LGTM, this feels more obvious as a temporary workaround that adding it to default exclude list!

ldez:
I agree, the previous time I wanted to be friendly with users who wanted to enable the rule but it was not the right solution.

PR #4985, opened 2024-09-04
docs: add additionnal info about typecheck

ldez:
Since v1.60.1, there have been a lot of issues/questions about `typecheck`.

So, I would like to add a new section inside the FAQ instead of repeating the same thing.

### DIFF
--- a/docs/src/docs/welcome/faq.mdx
+++ b/docs/src/docs/welcome/faq.mdx
@@ -45,7 +45,19 @@ How to troubleshoot:
 - [ ] Ensure you are not running an analysis on code that depends on files/packages outside the scope of the analyzed elements.
 - [ ] If using CGO, ensure all require system libraries are installed.
 
-## Why running with `--fast` is slow on the first run
+## Why is it not possible to skip/ignore `typecheck` errors?
+
+For mainly the same reasons that you cannot compile when you have a compiler error.
+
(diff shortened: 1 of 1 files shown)


docs/src/docs/welcome/faq.mdx
```diff
@@ -45,7 +45,19 @@ How to troubleshoot:
...
+
+For mainly the same reasons that you cannot compile when you have a compiler error.
+
+`typecheck` errors are reported in the same style as linter reports/issues,
+but are completely different because they are related to problems that block the analysis (`typecheck` is not a linter).
+
+When there are `typecheck` errors,
+most linters are not able to perform the analysis,
+and they simply do not produce any reports,
+so it's not possible to skip/ignore `typecheck` errors.
```
ccoVeille, line 58: maybe you could add something about the fact, typecheck errors means errors that must be fixed if you want to use the linter/golangci-lint

or anything like that, because your current text only mention, what they are and how they differ from a linting error

ldez: https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors

ccoVeille: Indeed. Thanks

bombsimon (approved):

PR #4990, opened 2024-09-08
chore: upgrade tetrafro/godot linter

webstradev:
There was recently a new release of godot with some nice fixes which should be included in golangci-lint


ldez:
The update of linter is done automatically by the dependabot: #4993

webstradev:
> The update of linter is done automatically by the dependabot: #4993

Ah that makes a lot of sense actually, I couldn't really imagine that was done manually :).

Just out of curiosity what is the frequency of dependabot triggering new updates?

ldez:
https://github.com/golangci/golangci-lint/blob/master/.github/dependabot.yml

webstradev:
Excellent thank you!

PR #4994, opened 2024-09-09
Update `intrange` to 0.2.0

ckaznocha:
`intrange` 0.2.0 adds a new check with an autofix and now requires type information. I believe that means some additional changes are needed to golangci-lint beyond the regular dependabot PR.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -30,7 +30,7 @@ require (
 	github.com/butuzov/mirror v1.2.0
 	github.com/catenacyber/perfsprint v0.7.1
 	github.com/charithe/durationcheck v0.0.10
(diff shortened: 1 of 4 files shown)


ldez:
Closed in favor of #4996

ckaznocha:
@ldez it doesn't look like the other PR included the autofix config. Are additional changes needed to set that up properly?

ldez:
because the autoofix is not what you think; it's not related to `SuggestedFixes` but to something specific inside golangci-lint.

bombsimon:
You can follow #1779 for autofix support

ckaznocha:
Got it, thanks for the context.

PR #5007, opened 2024-09-12
Add receivernamelen linter

ytnsym:
[receivernamelen](https://github.com/ytnsym/receivernamelen) is a linter that checks whether the length of a receiver variable name follows [Go style](https://google.github.io/styleguide/go/decisions.html#receiver-names)).

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -2689,6 +2689,7 @@ linters:
     - promlinter
     - protogetter
     - reassign
+    - receivernamelen
     - revive
     - rowserrcheck
     - sloglint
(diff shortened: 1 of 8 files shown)


ldez:
I created a linter that aggregates the different needs around receivers.

- https://github.com/ldez/recv
- https://github.com/golangci/golangci-lint/pull/5013

Don't hesitate if you have feeback.

ldez:
After some discussion, this linter can be an improvement of the revive rule [receiver-naming](https://revive.run/r#receiver-naming).

So I recommend contributing to `revive` to add this new rule. (maybe with a configurable size)

ytnsym:
@ldez 
> After some discussion, this linter can be an improvement of the revive rule [receiver-naming](https://revive.run/r#receiver-naming).
So I recommend contributing to revive to add this new rule. (maybe with a configurable size)

Understood. Thanks.

PR #5013, opened 2024-09-13
feat: add recv linter

ldez:
Recv cheks the methods receiver style:
- length of the receiver name
- consistency to the receiver type (pointer/non pointer)
- consistency on the receiver names between the methods of a struct

https://github.com/ldez/recv/

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -1540,6 +1540,14 @@ linters-settings:
     patterns:
       - ".*"
 
+  recv:
+    # Maximum length for a receiver name.
+    # Default: 3
+    max-name-length: 6
+    # Allow to use pointer receiver and non pointer receiver on the same struct.
+    # Default: true
+    type-consistency: false
+
   revive:
     # Maximum number of open files at the same time.
     # See https://github.com/mgechev/revive#command-line-flags
@@ -2689,6 +2697,7 @@ linters:
     - promlinter
     - protogetter
     - reassign
+    - recv
     - revive
     - rowserrcheck
     - sloglint
@@ -2804,6 +2813,7 @@ linters:
     - promlinter
     - protogetter
     - reassign
+    - recv
     - revive
     - rowserrcheck
     - sloglint
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -1544,7 +1544,7 @@ linters-settings:
     # Maximum length for a receiver name.
     # Default: 3
     max-name-length: 6
-    # Allow to use pointer receiver and non pointer receiver on the same struct.
+    # Allow using pointer receiver and non-pointer receiver on the same struct.
     # Default: true
     type-consistency: false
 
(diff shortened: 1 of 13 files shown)


.golangci.next.reference.yml
```diff
@@ -1540,6 +1540,14 @@ linters-settings:
     patterns:
       - ".*"
 
+  recv:
+    # Maximum length for a receiver name.
+    # Default: 3
+    max-name-length: 6
+    # Allow to use pointer receiver and non pointer receiver on the same struct.
```
ccoVeille, line 1547: This would be better I think

```suggestion
    # Allow using pointer receiver and non pointer receiver on the same struct.
```

ccoVeille: also you may consider to use 

```suggestion
    # Allow using pointer receiver and non-pointer receiver on the same struct.
```

if you agree with this PR https://github.com/ldez/recv/pull/1

pkg/golinters/recv/testdata/recv.go
```diff
@@ -0,0 +1,22 @@
...
+import "fmt"
+
+type Foo struct { // want `the methods of "Foo" use different receiver names: f, fo.`
+	Name string
+}
+
+func (f Foo) A()  {}
+func (fo Foo) B() {}
+
+type Bar struct{} // want `the methods of "Bar" use pointer receiver and non pointer receiver.`
```
ccoVeille, line 13:
```suggestion
type Bar struct{} // want `the methods of "Bar" use pointer receiver and non-pointer receiver.`
```

if you agree with https://github.com/ldez/recv/pull/1

bombsimon:
`revive` has [receiver-naming](https://revive.run/r#receiver-naming) and `stylecheck` has [ST1016](https://staticcheck.dev/docs/checks#ST1016) for the third check for consistent naming of receiver name. Is it worth adding a third linter with the overlap?

`revives` `receiver-naming` also checks for names like `self` and `this` but not an arbitrary length, would extending that check for length be an alternative?

ccoVeille:
> `revive` has [receiver-naming](https://revive.run/r#receiver-naming) and `stylecheck` has [ST1016](https://staticcheck.dev/docs/checks#ST1016) for the third check for consistent naming of receiver name. Is it worth adding a third linter with the overlap?
> 
> `revives` `receiver-naming` also checks for names like `self` and `this` but not an arbitrary length, would extending that check for length be an alternative?

I share your concern, I raised the point earlier there

https://github.com/ldez/recv/issues/2#issue-2525450594

ldez:
You are right, I see the revive rule but I missed the consistency check, and I also missed the staticheck rule.

The check of length can be an addition inside revive.
The receiver name consistency is already checked by revive and staticcheck.

So:
- I close this PR and we will go with another linter focused on type consistency only.
- I close also PR #5007.

ldez:
@ccoVeille can you delete the repo https://github.com/ccoVeille/recv ?

ccoVeille:
> @ccoVeille can you delete the repo https://github.com/ccoVeille/recv ?

Sure. It's done

PR #5022, opened 2024-09-15
docs: consistent WithSince versions

raeperd:
Hi I found inconsistent usage in Since in [linters page](https://golangci-lint.run/usage/linters/)

image

So I fixed it with simple v prefix if not exists


ldez (approved):
LGTM

PR #5036, opened 2024-09-22
docs: improve linter/rule/check docs

ldez:
It's verbose but it's easier to find if a linter already implements a rule.

It's not perfect, but it's good enough for now.

- `govet`: add descriptions for all passes.
- `gocritic`: add descriptions for all checks.
- `staticcheck`/`stylecheck`/`gosimple`: add description for all analyzers.
- `promlinter`: add descriptions for all rules.


Antonboom:
We definitely need checkboxes in PR like "please grep gocritic, revive, staticcheck, govet", "be sure that it cannot be implemented by ruleguard and forbidigo" 🙂

bombsimon (approved):

Issue #5044, opened 2024-09-27
Linter does not run on examples/


gucio321:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [X] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

I've noticed that golangci-lint does notrun on files in examples/ folder.
I didn't notice, however, any configuration for that in my .golangci.yam

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version v1.61.0 built with go1.23.1 from (unknown, modified: ?, mod sum: "h1:VvbOLaRVWmyxCnUIMTbf1kDsaJbTzH20FAMXTAlQGu8=") on (unknown)
```

</details>

### Configuration

<details>

```yaml
---
linters:
  disable-all: true
  enable:
    - wsl

```

</details>

### Go environment

<details>

```console
$ go version && go env
GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/me/.cache/go-build'
GOENV='/home/me/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/me/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/me/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.23.1'
GODEBUG=''
GOTELEMETRY='local'
[...]

PR #5045, opened 2024-09-27
docs: adds a section about `exclude-dirs-use-default`

ldez:
Related to https://github.com/golangci/golangci-lint/issues/5044#issuecomment-2379997625


ldez:
The "Extra" check fails because of a problem with a Docker image.
https://github.com/golangci/golangci-lint/pull/5047

bombsimon (approved):

PR #5047, opened 2024-09-27
dev: remove Nancy

ldez:
As we are using dependabot, Nancy (which focuses on dependency vulnerabilities) is not required anymore.


bombsimon (approved):
Never used Nancy so don't know about the overlap but if dependabot alerts about vulnerabilities even when no new version is available this sounds reasonable.

ldez:
In fact, Nancy never reported something useful, mainly because dependabot handles the dependency updates.

But Nancy has reported a lot of false positives, as you can see inside the `.nancy-ignore`, which breaks the CI on PRs.

PR #5048, opened 2024-09-27
docs: reorganize configuration files

ldez:
Reorganize the configuration references (and the configuration) to improve readability.

Before:

- `run`
- `output`
- `linters-settings`
- `linters`
- `issues`
- `severity`

After:

-  `linters`
-  `linters-settings`
-  `issues`
-  `output`
-  `run`
-  `severity`

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -4,135 +4,269 @@
 # This file is not a configuration example,
 # it contains the exhaustive configuration with explanations of the options.
 
-# Options for analysis running.
-run:
-  # Number of operating system threads (`GOMAXPROCS`) that can execute golangci-lint simultaneously.
-  # If it is explicitly set to 0 (i.e. not the default) then golangci-lint will automatically set the value to match Linux container CPU quota.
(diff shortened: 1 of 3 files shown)


bombsimon (approved):
Having linters before linters settings sounds very reasonable although for me personally I think general `golangci-lint` specific configuration starting at LOC 2000 is very far down and that linters and linters setting should come last.

But it's not a very strong opinion so go for it!

ldez:
My logic was "new user" related:
- first, you choose your linters
- then you configure linters
- then you apply exclusions
- then you customize the output
- then you increase your timeout
- then you don't use severities :smile_cat: 

But each section will still be split inside the doc of the configuration https://golangci-lint.run/usage/configuration/

So this has just a light impact.

bombsimon:
> * then you don't use severities 😸

Haha, this one was the best one!

I get your reasoning around adding and configuring linters, linters settings and exclusions first. And for proper documentation I don't think reading the reference config with thousands of lines is the best anyway.

PR #5053, opened 2024-10-03
fix: type sizing when cross-compiling (32-bit)

gh2o:
`golangci-lint` should respect `GOARCH` when performing typechecking.

For example, the code below fails to build with `GOARCH=arm`, but succeeds with `GOARCH=amd64`:

```
// -uintptr(x-y) creates a static assertion of x==y
const _ = -uintptr(unsafe.Sizeof((*byte)(nil)) - 8)
```

`golangci-lint` currently does not respect `GOARCH` for typechecking because it does not set `types.Config.Sizes`, which tells the typecheck package how to determine the size of machine-specific types (like pointers and `uintptr`). By default, [if `types.Config.Sizes` is `nil`](https://pkg.go.dev/go/types#Config), `SizesFor("gc", "amd64")` is used by default. This will notably break typechecking for 32-bit platforms.

This PR sets `types.Config.Sizes` from `build.Default`, which is set by the `GOARCH` environment variable.


ldez (approved):
LGTM

PR #5086, opened 2024-10-25
code-climate: add check_name field

Wani4ka:
CodeClimateIssue declares to contain just enough fields to support Gitlab CI Code Quality. But there's a field called `check_name`, which is not generated by golangci-lint, but is stated as required in both [Code Climate spec](https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types) and [Gitlab CI Code Quality reference](https://docs.gitlab.com/ee/ci/testing/code_quality.html#implement-a-custom-tool). I moved linter's name from the `description` field to the `check_name` field. I'm not sure if there can be any issue made by an unknown linter, but I made a default value for this field as well, just to be sure.

I also updated the corresponding test and the link to the Gitlab page (the previous one reported 404).

### DIFF
--- a/pkg/printers/codeclimate.go
+++ b/pkg/printers/codeclimate.go
@@ -7,14 +7,18 @@ import (
 	"github.com/golangci/golangci-lint/pkg/result"
 )
 
-const defaultCodeClimateSeverity = "critical"
+const (
+	defaultCheckName           = "golangci-lint"
+	defaultCodeClimateSeverity = "critical"
+)
 
 // CodeClimateIssue is a subset of the Code Climate spec.
 // https://github.com/codeclimate/platform/blob/master/spec/analyzers/SPEC.md#data-types
 // It is just enough to support GitLab CI Code Quality.
-// https://docs.gitlab.com/ee/user/project/merge_requests/code_quality.html
+// https://docs.gitlab.com/ee/ci/testing/code_quality.html#implement-a-custom-tool
 type CodeClimateIssue struct {
 	Description string `json:"description"`
+	CheckName   string `json:"check_name"`
 	Severity    string `json:"severity,omitempty"`
 	Fingerprint string `json:"fingerprint"`
 	Location    struct {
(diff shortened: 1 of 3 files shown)


ldez:
The linter name cannot be empty, so there's no need for default.

The severity is optional, and it should be controlled with the severity configuration: https://golangci-lint.run/usage/configuration/#severity-configuration

The description should not be changed to avoid breaking changes.

ldez (approved):
LGTM

ccoVeille:
@Wani4ka thanks for this

I'm curious, did you see any change in the way GitLab reports the issues?

Wani4ka:
GitLab self-hosted is used in my work, and I tried to make an integration of golangci-lint and ci code quality by uploading linter job artifacts, but it wasn't interpreted as a correct report. I may presume that the reason could be in the job misconfiguration, but I decided to start investigating the issue by checking the code report and found this field mismatch. I'll check the job configuration later, but anyways I think this change might be useful

ccoVeille:
I checked gitlab code and it will have no effect (the example JSON in the link you quoted confirms it). Maybe it would one day. But not for now.

But it's great that golangci-lint code climate report respects the expected format 👍

ldez:
> I checked gitlab code and it will have no effect 

Are you sure?

The Ruby code doesn't seem to handle it but JS code handles it.

- https://gitlab.com/gitlab-org/gitlab/-/blob/master/lib/gitlab/ci/parsers/codequality/code_climate.rb
- https://gitlab.com/gitlab-org/gitlab/-/blob/master/ee/spec/frontend/codequality_report/mock_data.js
- https://gitlab.com/gitlab-org/gitlab/-/blob/master/app/assets/javascripts/ci/reports/codequality_report/utils/codequality_parser.js#L9

ccoVeille:
I was talking about the fact the check_name field is not used in the way GitLab represents the code climate results.

But maybe, I don't get what you just said, or I missed something. If then, I'm sorry about it.

PR #5087, opened 2024-10-26
Revert "Add filen linter (#5081)"

ldez:
As the PR https://github.com/mgechev/revive/pull/1072 has been merged into revive, and as `filen` is not in a released version of golangci-lint, we should remove this linter because it becomes a duplicate linter.

If we don't remove it now, we will deprecate it after the release, so it's better to remove it now.

@DanilXO I'm sorry, it was unexpected to me that, after 2 years without a real reaction to the [discussion](https://github.com/golangci/golangci-lint/discussions/2881), someone created a PR to add this rule to revive.

I'm not really happy that someone uses your linter and your PR as references to create a rule inside revive. Communication should be a priority and it should be improved.
But anyway, what is done is done.

Related to #5081

This reverts commit 29c5529e93f641b6e385b525badb165643f84e10.


alexandear (approved):

bombsimon (approved):
LGTM

ccoVeille:
The fact the revive PR would lead to duplicate rules is the reason why I posted my comment.

I agree with you

DanilXO:
The main thing is that the need for file size limitation is closed. It's okay that PR was reverted, the experience of creating a linter was interesting and curious for me.

PR #5131, opened 2024-11-13
docs: improve testifylint description of require-f-funcs flag

Antonboom:
To decrease common confusing. People often expect that they can continue use not-f assertions if only format string is used (without args).


ldez (approved):
LGTM

PR #5147, opened 2024-11-21
docs: Update README generation

bombsimon:
- Reword `Team` section to `Contributors` since the list is only partly team members.
- Indicate in the contributors list who are a team member of the `golangci` organization
- Update styling on the table to ensure equal sized image and name alignment
- Always link to the user's GitHub profile instead of external web page

---

I don't think linking to a user's external web page is what's expected when clicking on a user's picture here on GitHub. On top of that we also don't have control on what potential sites we're linking to by doing so. Instead I think we should always link to the GitHub profile page and if someone want's to visit the web page from there they can click on the link in the profile.

While exploring this I noticed a few other things that I was addressing since they felt related. I reworded the list to contributors since it's not actually the team or team members. To address that I also made an extra call to the organization's team page to flag contributors who's also team members. This is not optimal and could be handled with a single call and some more code changes but given that the pipeline takes [1h 12m](https://github.com/golangci/golangci-lint/actions/runs/11767888489/job/32777077352) I don't think an extra roundtrip and loop over 500 items really matters.

Lastly since the images and alignment was rendered differently if the name was long enough to cause a line break I added some styling for that as well.


ldez:
I agree with your PR.

I also think replacing this section with something simpler, because, as you said, it takes more than 1 hour to generate.

I think we should keep contributors but with something simpler, like inside the homepage https://golangci-lint.run/.

And keep the "sort by activities" only for team members.

bombsimon:
> I also think replacing this section with something simpler, because, as you said, it takes more than 1 hour to generate.

Yeah I was thinking of removing `name-your-contributors` because that seems to be what's taking time but I didn't explore much since I didn't want to make this too big of an change right now.

I guess if we only take contributors from the [contributors API](https://docs.github.com/en/rest/repos/repos?apiVersion=2022-11-28#list-repository-contributors) we wouldn't need any more stats for them, however `name-your-contributors` doesn't seem to allow specifying a specific set of users so we would have to poll PRs and issues and reactions etc for each team member to sort them accordingly.

I would probably have to spend way more time figuring this out hence I filed this PR to fix styling and title short term.

ldez:
<a href="https://github.com/golangci/golangci-lint/graphs/contributors"></a>

ldez:
FYI I already tried to hack with `name-your-contributors` and it was deceptive.

ldez:
If you agree, I propose to remove all the sections (and the CI process) to just use the contributors list https://github.com/golangci/golangci-lint/pull/5147#issuecomment-2492294420

In the future, we can improve or extend that, but it can be a good first step.

bombsimon:
> If you agree, I propose to remove all the sections 

This sounds way better and simpler! Sure, I'll miss my face there, but I'll survive!

Closing this in favor of #5148

PR #5148, opened 2024-11-21
docs: Simplify team and contributors section

bombsimon:
This will remove the slow and cumbersome process of generating team and contributor section in the README. The section is replaced with a simplified svg image.

Supersedes #5147


.github/contributors/generate.ts
```diff
@@ -1,143 +0,0 @@
...
-    weightedContributors.sort((a, b) => b.weight - a.weight)
-    const coreTeamInfo = await buildCoreTeamInfo()
-    const contributorsInfo = await buildContributorInfo(weightedContributors)
-    const exclude: any = {
-      golangcidev: true,
-      CLAassistant: true,
-      renovate: true,
-      fossabot: true,
-      golangcibot: true,
-      kortschak: true,
```
bombsimon, line 126: Is this problematic? I guess the other contributors doesn't really matter since they're bots but this one specifically asked not to be mentioned. Can you exclude contributors somehow?

ldez: I didn't find options to filter the list, but now there is no link to a profile, so it is less problematic.

ldez (approved):
LGTM

.github/contributors/generate.ts
bombsimon: I opened https://github.com/opencollective/contributors-svg/pull/68 which works for me locally. The repo has been dead for 2,5 years so not likely they'll merge but who knows! 🤞

PR #5150, opened 2024-11-23
dev: simplify sort results processors

ldez:
The package `cmp` has been added in go1.22 and allows the `SortResults` processor to be simplified.

Related to #4467


pkg/result/processors/sort_results.go
```diff
@@ -57,171 +62,56 @@ func (p SortResults) Process(issues []result.Issue) ([]result.Issue, error) {
...
 
-type compareResult int
-
-const (
-	less compareResult = iota - 1
-	equal
-	greater
-	none
-)
+type issueComparator func(a, b *result.Issue) int
```
bombsimon, line 87: nit: Now that the `cosnt` block is at the top and `issueComparator` is used in the first type `SortResults` on LOC 32, can we move this to the top as well on L31 or something?

bombsimon (approved):
Nice! LGTM! 🎉

pkg/result/processors/sort_results.go
ldez: I moved it :+1:

PR #5200, opened 2024-12-06
docs: Ignore previously ignored contributors

bombsimon:
Followup from #5148. It's now supported to filter out users since https://github.com/opencollective/contributors-svg/pull/68 got merged so this will restore the previously ignored users + add `dependabot[bot]`.


ldez (approved):
LGTM :+1:

PR #5201, opened 2024-12-07
dev: sync checker with x/tools@v0.28.0

ldez:
Updates the code borrowed from x/tools

Continuation of the work done inside #5109

I still have not found the Lost Ark but I am near the Temple of Doom :smile_cat:

### DIFF
--- a/pkg/goanalysis/runner_checker.go
+++ b/pkg/goanalysis/runner_checker.go
@@ -2,7 +2,7 @@
 // Use of this source code is governed by a BSD-style
 // license that can be found in the LICENSE file.
 //
-// Partial copy of https://github.com/golang/tools/blob/dba5486c2a1d03519930812112b23ed2c45c04fc/go/analysis/internal/checker/checker.go
+// Altered copy of https://github.com/golang/tools/blob/v0.28.0/go/analysis/internal/checker/checker.go
 
 package goanalysis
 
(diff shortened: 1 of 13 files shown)


pkg/goanalysis/runner_checker.go
```diff
@@ -94,106 +104,140 @@ func (act *action) analyze() {
...
-		analyzedIn := time.Since(startedAt)
-		if analyzedIn > time.Millisecond*10 {
+		analyzedIn := time.Since(t1)
+		if analyzedIn > 10*time.Millisecond {
 			debugf("%s: run analyzer in %s", act, analyzedIn)
 		}
-	}
 
-	// disallow calls after Run
+		// correct result type?
```
bombsimon, line 192: What is the question here? Why they're different? The code seems to properly check the difference and return a well format error. Should the comment indicate that it's not known why they would be different?

bombsimon (approved):
LGTM! 🚢

pkg/goanalysis/runner_checker.go
ldez: it's a copy of the original comment :wink: 

https://github.com/golang/tools/blob/v0.28.0/go/analysis/checker/checker.go#L360

bombsimon: 🤦 Oh sorry. I thought it was you since it was a lone diff, should've checked upstream first!

PR #5202, opened 2024-12-07
docs: Make contributors URL shorter

bombsimon:
It's possible to not specify the `skip` query string for each element but instead use a `,` so this changes the URL to be a bit shorted. Also there was a typo in the previous URL missing an `=` for one element.

---

Sorry for the noise with multiple PRs. Was going to fix the missing `=` but found they also added support for comma separated input making the URL a bit shorted so switched to that.


ldez (approved):
LGTM

PR #5222, opened 2024-12-13
importas: allow multiple empty aliases

ldez:
The identical aliases are forbidden to avoid package conflicts but an empty alias is just a standard import so it's possible to have several ones.

It's less restrictive than the option `no-extra-aliases`.

Fixes #5221


pkg/golinters/importas/importas.go
```diff
@@ -51,11 +51,15 @@ func New(settings *config.ImportAsSettings) *goanalysis.Linter {
...
-				lintCtx.Log.Errorf("invalid configuration, multiple packages with the same alias: alias=%s packages=[%s,%s]", a.Alias, a.Pkg, v.Pkg)
-			} else {
-				uniqAliases[a.Alias] = a
+			// Ignore duplication for empty aliases.
+			if a.Alias != "" {
+				// skip the duplication check when the alias is a regular expression replacement pattern (ie. contains `$`).
+				v, ok := uniqAliases[a.Alias]
+				if ok && !strings.Contains(a.Alias, "$") {
+					lintCtx.Log.Errorf("invalid configuration, multiple packages with the same alias: alias=%s packages=[%s,%s]", a.Alias, a.Pkg, v.Pkg)
+				} else {
```
denisvmedia, line 60: An alternative solution (will read from map and never get the value, but this reduces the indentation).

```go
			if v, ok := uniqAliases[a.Alias]; ok && !strings.Contains(a.Alias, "$") {
 				lintCtx.Log.Errorf("invalid configuration, multiple packages with the same alias: alias=%s packages=[%s,%s]", a.Alias, a.Pkg, v.Pkg)
 			} else if a.Alias != "" {
 				uniqAliases[a.Alias] = a
 			}
```

pkg/golinters/importas/testdata/importas_several_empty_aliases.yml
```diff
@@ -0,0 +1,9 @@
+linters-settings:
+  importas:
+    alias:
```
denisvmedia, line 3: I'd maybe add `no-unaliased: true` to make sure the rule actually works.

ldez: I prefer to keep the tests focused on one element because it's easier for maintenance.

denisvmedia:
LGTM, thanks! I'm not sure about the tests, up to you.

pkg/golinters/importas/importas.go
```diff
@@ -51,8 +51,11 @@ func New(settings *config.ImportAsSettings) *goanalysis.Linter {
 				uniqPackages[a.Pkg] = a
 			}
 
-			// skip the duplication check when the alias is a regular expression replacement pattern (ie. contains `$`).
-			if v, ok := uniqAliases[a.Alias]; ok && !strings.Contains(a.Alias, "$") {
+			// Skips the duplication check when:
+			// - the alias is empty.
+			// - the alias is a regular expression replacement pattern (ie. contains `$`).
+			v, ok := uniqAliases[a.Alias]
+			if ok && a.Alias != "" && !strings.Contains(a.Alias, "$") {
```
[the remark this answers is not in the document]
ldez: I didn't miss it, but I think it is clearer to add the skip conditions in the same place. There is no major impact on performance but this is more readable, so this is better for maintenance.

alexandear (approved):

PR #5226, opened 2024-12-13
docs: empty alias with importas

ccoVeille:
I discovered the feature of enforcing empty aliases with 

- #5221 

A fix was provided

- #5222

I think it could be interesting to document it.


ldez:
The reference file should be (when possible) exhaustive about options but not about usage.
This is not exactly a source of examples or "how to".
We should limit them or the reference file will quickly become unreadable.

So I reduced the example to only one explicit empty alias.

ccoVeille:
Sure, no problems

PR #5242, opened 2024-12-21
fix: remove decoder, sloglint, tagalin from 'format' preset

ldez:
decoder, sloglint, tagalin are not formatters but just linters related to "style".


alexandear (approved):
It's a breaking change. We should probably add the "linter: update" label.

PR #5252, opened 2024-12-25
fix(tenv): disable for Go versions < 1.17

scop:


ldez:
We don't support Go versions lower than go1.22.

https://golangci-lint.run/welcome/faq/#which-go-versions-are-supported

ccoVeille:
> We don't support Go versions lower than go1.22 (with a slight tolerance for go1.21).
> 
> https://golangci-lint.run/welcome/faq/#which-go-versions-are-supported
> 

I have a suggestion. Maybe I'm wrong and it's already the case but:

Maybe the app could simply report an error message then, maybe return an exitcode when lower than 1.21.

Because I mean it's not supported, but it can be launched, so people expect things.

But maybe it's too extreme and a breaking changes.

But it's not that uncommon to see people reporting about blah blah version older than x, and you close "but x is not supported"

I don't have a computer to check what it does and reports now, also I'm always migrating to latest Go version.

ldez:
For me, this is useless `t.Env` doesn't compile before go1.17.
Golancgi-lint can be use for the old Go versions, we will not limit artificially that.

PR #5267, opened 2024-12-31
feat: format fixed files with the same formatters as the issues related to formatting

ldez:
Currently (since the PR about formatters) the reports related to formatting are handled by a dedicated processor, but the suggested fixes can produce formatting issues.
Inside the fixer, golangci-lint always calls the std formatter but it's better if it calls the formatters from the configuration too.

This PR also improves performance by formatting the files only one time, and by using `fileCache` instead of `os.ReadFile` to get the file content.

Note: the formatter processor is merged with the fixer.

This improves cases that can produce missing imports.

```yml
linters:
  disable-all: true
  enable:
    - err113
    - goimports
```

```go
package main

func Foo(e1, e2 error) bool {
	return e1 == e2
}
```

The command:
```
./golangci-lint run --fix
```

Before this PR, the result is:

```go
package testdata

func Foo(e1, e2 error) bool {
	return errors.Is(e1, e2)
}
```

with this PR:

```go
package testdata

import "errors"

func Foo(e1, e2 error) bool {
	return errors.Is(e1, e2)
}
```

because `goimports` fixes imports.

</details> 

Related to #5232, #5246


pkg/goformatters/meta_formatter.go
```diff
@@ -0,0 +1,74 @@
...
+		if err != nil {
+			m.log.Warnf("(%s) formatting file %s: %v", formatter.Name(), filename, err)
+			continue
+		}
+
+		data = formatted
+	}
+
+	return data
+}
```
ccoVeille, line 74: Maybe my approach is naive, but the std go/format could be added directly in NewMetaFormatter


```suggestion
func NewMetaFormatter(log logutils.Log, cfg *config.Config, enabledLinters map[string]*linter.Config) (*MetaFormatter, error) {
	m := &MetaFormatter{log: log}

	if _, ok := enabledLinters[gofmt.Name]; ok {
		m.formatters = append(m.formatters, gofmt.New(cfg.LintersSettings.Gofmt))
	}

	if _, ok := enabledLinters[gofumpt.Name]; ok {
		m.formatters = append(m.formatters, gofumpt.New(cfg.LintersSettings.Gofumpt, cfg.Run.Go))
	}

	if _, ok := enabledLinters[goimports.Name]; ok {
		m.formatters = append(m.formatters, goimports.New())
	}

	// gci is a last because the only goal of gci is to handle imports.
	if _, ok := enabledLinters[gci.Name]; ok {
		formatter, err := gci.New(cfg.LintersSettings.Gci)
		if err != nil {
			return nil, fmt.Errorf("gci: creating formatter: %w", err)
		}

		m.formatters = append(m.formatters, formatter)
	}

	if len(m.formatters) == 0 {
		 // this is pseudo code, it won't work without adding a Format method
		m.formatters = append(m.formatters, format.Source)
	}

	return m, nil
}

func (m *MetaFormatter) Format(filename string, src []byte) []byte {
	data := bytes.Clone(src)

	for _, formatter := range m.formatters {
		formatted, err := formatter.Format(filename, data)
		if err != nil {
			m.log.Warnf("(%s) formatting file %s: %v", formatter.Name(), filename, err)
			continue
		}

		data = formatted
	}

	return data
}
```

I'm suggesting this because we could consider to do that to isolate gci even more

```suggestion
func NewMetaFormatter(log logutils.Log, cfg *config.Config, enabledLinters map[string]*linter.Config) (*MetaFormatter, error) {
	m := &MetaFormatter{log: log}

	if _, ok := enabledLinters[gofmt.Name]; ok {
		m.formatters = append(m.formatters, gofmt.New(cfg.LintersSettings.Gofmt))
	}

	if _, ok := enabledLinters[gofumpt.Name]; ok {
		m.formatters = append(m.formatters, gofumpt.New(cfg.LintersSettings.Gofumpt, cfg.Run.Go))
	}

	if _, ok := enabledLinters[goimports.Name]; ok {
		m.formatters = append(m.formatters, goimports.New())
	}

	if len(m.formatters) == 0 {
		 // this is pseudo code, it won't work without adding a Format method
		m.formatters = append(m.formatters, format.Source)
	}

	// gci is a last because the only goal of gci is to handle imports.
	if _, ok := enabledLinters[gci.Name]; ok {
		formatter, err := gci.New(cfg.LintersSettings.Gci)
		if err != nil {
			return nil, fmt.Errorf("gci: creating formatter: %w", err)
		}

		m.formatters = append(m.formatters, formatter)
	}

	return m, nil
}

func (m *MetaFormatter) Format(filename string, src []byte) []byte {
	data := bytes.Clone(src)

	for _, formatter := range m.formatters {
		formatted, err := formatter.Format(filename, data)
		if err != nil {
			m.log.Warnf("(%s) formatting file %s: %v", formatter.Name(), filename, err)
			continue
		}

		data = formatted
	}

	return data
}
```

ldez: gci doesn't need special treatment or to be more isolated: gci calls the the goimports tooling internally. gci is a formatter like the other.

ldez:
To follow the release plan, I will exceptionally merge this PR.

Issue #5313, opened 2025-01-10
Invalid config file silently ignored


alexbozhenko:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

If golangci-lint found a config file, but the file is invalid, it still runs...

### Version of golangci-lint

<details>

```console
# golangci-lint --version
golangci-lint has version 1.63.4 built with go1.23.4 from c1149695 on 2025-01-03T19:49:42Z
```

</details>

### Configuration

<details>

```console
# paste configuration file or CLI flags here
```

</details>

### Go environment

<details>

```console
# go version && go env
go version go1.23.4 linux/amd64
GO111MODULE=''
GOARCH='amd64'
GOBIN=''
GOCACHE='/home/alex/.cache/go-build'
GOENV='/home/alex/.config/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFLAGS=''
GOHOSTARCH='amd64'
GOHOSTOS='linux'
GOINSECURE=''
GOMODCACHE='/home/alex/go/pkg/mod'
GONOPROXY=''
GONOSUMDB=''
GOOS='linux'
GOPATH='/home/alex/go'
GOPRIVATE=''
GOPROXY='https://proxy.golang.org,direct'
GOROOT='/usr/lib/go'
GOSUMDB='sum.golang.org'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/lib/go/pkg/tool/linux_amd64'
GOVCS=''
GOVERSION='go1.23.4'
GODEBUG=''
GOTELEMETRY='local'
GOTELEMETRYDIR='/home/alex/.config/go/telemetry'
GCCGO='gccgo'
GOAMD64='v1'
AR='ar'
CC='gcc'
CXX='g++'
[...]

PR #5314, opened 2025-01-10
Add yaml language server config

alexbozhenko:
For those who use yaml-language-server in their IDE, the IDE can verify the config, which is nice:
https://github.com/redhat-developer/yaml-language-server?tab=readme-ov-file#using-inlined-schema
https://github.com/redhat-developer/yaml-language-server?tab=readme-ov-file#clients

And can help prevent issues like https://github.com/golangci/golangci-lint/issues/5313

image

Let's add it to the default/example config files?

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -1,3 +1,4 @@
+# yaml-language-server: $schema=https://golangci-lint.run/jsonschema/golangci.jsonschema.json
 # This file contains all available configuration options
 # with their default values (in comments).
 #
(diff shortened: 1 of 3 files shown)


ldez:
I don't want things only for specific IDE inside the code.
And those elements (file name and JSONschema matching) are mainly already handled by the [JSONSchema store](https://github.com/SchemaStore/schemastore).

alexbozhenko:
@ldez may I at least document this behavior somewhere? Either here, or in gh action repo?

Issue #5297, opened 2025-01-06
🌟 Let's talk about exclusion options inside issues section


ldez:
> [!IMPORTANT]
> This is a proposal: I don't know if it is possible and what the impact could be inside the code.
> The proposal may evolve.

---

The current configuration:

<details>

```yml
issues:
  # ...
  
  exclude-generated: strict
  
  # If set to true, `exclude` and `exclude-rules` regular expressions become case-sensitive.
  exclude-case-sensitive: false
  exclude:
    - abcdef
  exclude-rules:
    - path: _test\.go
      linters:
        - gocyclo
        - errcheck
        - dupl
        - gosec
    - path-except: _test\.go
      linters:
        - forbidigo
    - path: internal/hmac/
      text: "weak cryptographic primitive"
      linters:
        - gosec
    - linters:
        - staticcheck
      text: "SA9003:"
    - linters:
        - lll
      source: "^//go:generate "

  exclude-dirs-use-default: false
  exclude-dirs:
    - src/external_libs
    - autogenerated_by_my_lib
  exclude-files:
    - ".*\\.my\\.go$"
    - lib/bad.gO

  exclude-use-default: false
  include:
    - EXC0001
    - EXC0002
    - ...

  # ...
```

</details>

### Useless Option

The option `exclude-dirs-use-default` comes from the "prehistorical" world of `GOPATH`.

- `vendor$`, `testdata$` are useless inside this default configuration because the Go tooling already excludes them.
- Since go modules, `Godeps$` is useless.
- `examples$`: This is unexpected because it excludes all directories ending with `examples`.
- `third_party$`, `builtin$`: we should not create "custom conventions", so those elements should not be excluded.

### Duplicated Options

The goals of `exclude-dirs` and `exclude-files` are the same.

The matching can be complex in some cases:
- https://github.com/golangci/golangci-lint/issues/4635
- https://github.com/golangci/golangci-lint/issues/1178

`exclude` can be expressed with an `exclude-rules`.

### Unclear Section

At the 1st level of the `issues` section, there are options about filtering (ex: `max-issues-per-linter`, `new-from-rev`), and about exclusions (ex: `exclude-rules`).

I don't think we should create a subsection for each topic but we should have at least one for exclusions.

### Default Exclusions

There is a dedicated proposal for this topic:

- xxx

## 💭 Proposal

- All the options related to exclusions should be grouped inside a dedicated subsection of `issues`:
[...]

PR #5316, opened 2025-01-11
dev: processor optimizations

ldez:
I was working on the new exclusion configuration #5297, and I found some "optimizations". 

- `IdentifierMarker` processor update:
  - removes unused regexp
  - checks the match only if the issue is from a linter with the associated pattern instead of trying to match all regex on all issue texts.
  - adds one test by pattern.
  - Note: I'm unsure if keeping/removing this processor in the future is a good idea. 
- new `PathAbsoluter` processor:
  - the absolute paths were obtained by the `Cgo` processor and the `FilenameUnadjuster` then now the `PathAbsoluter` does that once.
  - in theory, all the paths from `analysis.Diagnostic` are absolute.
  - FYI the processor `PathPrettifier` (it already exists) converts absolute path to relative (related to the different elements but mainly to exclusions)

This is not only "optimizations", the `PathAbsoluter` is something I will use with the new exclusion system.

I found another optimization: the usage of `FileCache` inside `misspell`.
The `FileCache` is used by `SourceCode`, `ExcludesRules`, `severity`, `Fixer` processors, and `misspell`.
The `FileCache` is loaded with file content and the keys are paths, but `misspell` is the only one that uses absolute paths to load/read data from/to the cache.
So the cache is filled but never used.
This allows to reduce the memory usage (ex: on k8s this 50MiB)

These are minor optimizations: a few seconds, but they can impact large projects.

To avoid creating a PR for exclusions with some off-topics, and as those elements are standalone, I prefer to create a PR with just these processors.


bombsimon (approved):
Nice!

Re `IdentifierMarker`, I think it's a bit weird. It feels like it's from early days where the linters and their diagnostics was more controlled. Surely I'd like to have identifiers within backtiks just as much as the next person, but having a processor doing search replace on hand written regexes for each linter feels a bit unnecessary and hard to maintain. I'm sure there are more identifiers out there we don't replace, or replacements that are removed from the upstream linters anyways.

ldez:
I agree, we will not remove it for now, I want to avoid "breaking changes" but we will not add new regexp unless there is a real need.

Issue #5298, opened 2025-01-06
🌟 Let's talk about default exclusions


ldez:
> [!IMPORTANT]
> This is a proposal: I don't know if it is possible and what the impact could be inside the code.
> The proposal may evolve.

---

The current configuration:

```yml
issues:
  exclude-use-default: true
  include:
    - EXC0001
    - EXC0002
    - EXC0003
    - EXC0004
```

### Useful vs Good Practice

I agree with Denis: most projects don't add docs and want to add docs.
https://github.com/golangci/golangci-lint/issues/456#issuecomment-534938262

But I also agree that is not a good practice.
https://github.com/golangci/golangci-lint/issues/456#issuecomment-922102219

### The Names of the Embed Exclusions

The codes of the default exclusions are machine-oriented but we are humans.

- [EXC0001](https://golangci-lint.run/usage/false-positives/#EXC0001): (errcheck) std error handling (ex: `os.Close`)
- [EXC0002](https://golangci-lint.run/usage/false-positives/#EXC0002): (golint) comment
- [EXC0003](https://golangci-lint.run/usage/false-positives/#EXC0003): (golint) func name will be used as test\.Test.* by other packages
- [EXC0004](https://golangci-lint.run/usage/false-positives/#EXC0004): (govet) possible misuse of unsafe.Pointer|should have signature
- [EXC0005](https://golangci-lint.run/usage/false-positives/#EXC0005): (staticcheck/SA4011) C-style with an explicit 'break' in a 'switch'
- [EXC0006](https://golangci-lint.run/usage/false-positives/#EXC0006): (gosec/G103) Use of unsafe calls should be audited
- [EXC0007](https://golangci-lint.run/usage/false-positives/#EXC0007): (gosec/G204) Subprocess launched with variable
- [EXC0008](https://golangci-lint.run/usage/false-positives/#EXC0008): (gosec/G104) Duplicated errcheck checks.
- [EXC0009](https://golangci-lint.run/usage/false-positives/#EXC0009): (gosec/G301|G302|G307) (G301|G302|G307): Expect (directory permissions to be 0750|file permissions to be 0600) or less
- [EXC0010](https://golangci-lint.run/usage/false-positives/#EXC0010): (gosec/G304) False positive is triggered by 'src, err := ioutil.ReadFile(filename)'.
- [EXC0011](https://golangci-lint.run/usage/false-positives/#EXC0011): (stylecheck/ST1000|ST1020|ST1021|ST1022) comments
[...]

PR #5338, opened 2025-01-20
docs: add default exclusion names to the JSONSchema

ccoVeille:
I would like to suggest this simple addition

It could avoid loosing time as I did.
I was using EXC014 (instead of EXC0014) and I had to debug everything to figure our the problem was between keyboard and chair.

I validated it worked with the following command:
```shell
go run cmd/golangci-lint/main.go config verify --schema jsonschema/golangci.next.jsonschema.json --config /path/golangci.yml
```

I also checked at the code and it should be safe

https://github.com/golangci/golangci-lint/blob/v1.63.4/pkg/config/issues.go#L233-L247

### DIFF
--- a/jsonschema/golangci.next.jsonschema.json
+++ b/jsonschema/golangci.next.jsonschema.json
@@ -3958,7 +3958,9 @@
           "description": "The list of ids of default excludes to include or disable.",
           "type": "array",
           "items": {
-            "type": "string"
+            "type": "string",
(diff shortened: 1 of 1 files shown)


ldez:
There is not need for regexp, the default exclusion names are known.

Also, those names will be removed: https://github.com/golangci/golangci-lint/issues/5298

ldez (approved):
LGTM

Issue #5300, opened 2025-01-06
🪩 It's time to create a v2 🚀


ldez:
There are at least 6 years since the idea of a v2 is here, we should pass this step now.

> [!IMPORTANT]
> The focus will be on structural configuration changes and default values changes only (and removing all deprecated options). Because a long tunnel with a massive wishlist to create this release is something we should avoid for multiple reasons.

## 💭 Main Proposals

- new `fmt` command (+ new dedicated formatters configuration section) #5296
- new linter configuration (and default set of linters) #5299
- new default exclusion configuration #5298
- new exclusion configuration #5297
- new output format configuration #5364
- changes some default values (timeout, cpu quota, etc.)
- fixes some field names and types
- removes some "prehistorical" flags

## 🗺️ Plan

There are 3 ways to achieve this v2:
1. a deprecation phase: this will be slower and more complex (because of the compatibility).
2. a breaking change (**with a migration command**): this will be faster and easier but the impact on users could be more important.
3. a mix of the 2 previous ways.

The choice of the way to achieve this v2 will not be defined based on opinions about those ways: we will know the pros and cons between all those ways.
The choice will be based on technical constraints and how much time the community is ready to give us to do it (cf the Sustainability section).

In all cases, v1 will still be maintained during the v2 development phase.

Work on v2 begins with the opening of this issue.

## ❤️ Sustainability
This will require a lot of work, and so a lot of time, and where there is time there is money.

First, I would say thanks to the people, who already donated :heart:

Asking for money feels very uncomfortable to me, but I should do it because without that nothing will happen magically.

I will work full-time on this new version, I'm not working for a company, my only revenues are donations.
I've applied to several OSS funding programs but so far without success.
I will continue to look for OSS funding programs, and encourage users to donate, and I will try to find other ways.

Also, I'm not alone: all the active maintainers are a major part of this plan.

As golangci-lint's user what you can do:

1. Give your feedback on the proposals
2. Donate :heart:: even 2€ / $2 / 350円 / 15元 / 200₽ per month is important
	- [golangci-lint](https://donate.golangci.org)
	- [ldez](https://donate.ldez.dev)
3. Ask your company to donate (if needed we can exchange by email: info at golangci dot com)
	- [golangci-lint](https://donate.golangci.org)
	- [ldez](https://donate.ldez.dev)
4. Help us by spreading the message:
   - [on Bluesky](https://bsky.app/profile/golangci-lint.run)
[...]

PR #5339, opened 2025-01-20
feat: new linter exclusions system

### LINKED ISSUE #1178: Use path exclude rules on absolute paths instead of relative paths
Using path exclude rules is difficult because the issue path depends on which directory I run `golangci-lint` from.

I'd like path exclude rules to be evaluated on the _absolute_ path of the issue, not the relative path that is output.

Can't think of a great alternative here.

**Additional context**
Example: If I have an exclude rule on a file `a/b/c.go`, this rule will not apply if I run `golangci-lint` from directory `a/b/`.

### LINKED ISSUE #3717: Relative paths handling is inconsistent between configuration options, hence unable to run golangci-lint from a subdirectory
### Welcome

There are several places in `golangci-lint` where paths are specified, including:

- `run.skip-dirs` paths
- `issues.exclude-rules.*.path`
- `linters-settings.gocritic.ruleguard.rules`
- `linters-settings.depguard.additional-guards.*.ignore-file-rules`
- `custom.*.path`

The handling of these paths is inconsistent, leading to various issues when `golangci-lint run` is run from the directory other than the one containing the `.golangci-lint.yml`:

- `run.skip-dirs`, `issues.exclude-rules.*.path` are matched textually, so if `golangci-lint run` is called from a subdirectory, these matches are silently skipped
- `linters-settings.gocritic.ruleguard.rules` accepts `{configDir}` to make it look up proper ruleguard config file. However this directive does not seem to be accepted in any other places of the configuration file.
- `linters-settings.depguard.additional-guards.*.ignore-file-rules` appear to match  _full_ file paths.
- `custom.*.path` seem to always be relative to the configuration file, so at least this configuration works.

As it stands, the only way to run `golangci-lint run` from a subdirectory is to avoid any paths in configuration file, except `ruleguard` and custom linters.

It would be very useful to make all paths in configuration file behave the same. Treating them as relative to the location of the configuration file appears to be the most reasonable option.

There is already one issue for a specific case of path excludes: #1178, however this issue is more general.

### Version of golangci-lint

<details>

```console
golangci-lint has version v1.51.0 built from (unknown, mod sum: "h1:M1bpDymgdaPKNzPwQdebCGki/nzvVkr2f/eUfk9C9oU=") on (unknown)
```

</details>

### Configuration file

<details>

```console
run:
  skip-dirs:
   - some/path # matched textually, won't be matched if run as `cd some && golangci-lint run`

issues:
  exclude-rules:
      - path: some/other/path # matched textually, won't be matched if run as `cd some && golangci-lint run`
        linters:
          - errcheck

linters-settings:
  gocritic:
    settings:
      ruleguard:
        rules: "${configDir}/ruleguard.go" # this works due to ${configDir}
  depguard:
    additional-guards:
      - list-type: denylist
        ignore-file-rules:
          # There is no way to anchor the match to the root of the repository!
          - "**/some/third/path/*.go" # this seems to match full paths, unlike everything else
  custom:
    mycustomlinter:
      path: ./customlinter.so # this seems to be relative to the configuration file
```

</details>

### LINKED ISSUE #3953: Report useless entry in exclude-rules

From time to time I need to add something to `exclude-rules`. For example I'm introducing new rule to old codebase so I exclude the  rule in some files/folders and iteratively fixing it.

But sometimes I get to the point, that I fix all issues, but forgot to remove that entry in `exclude-rules` section.

It would be nice, if golangci-lint would complain that I have some rule to exclude, which is not applied at all.

I see that similar thing is now done by nolintlint when I have `//nolint:xxx` in the code. But not in the `.golangci.yml` file.

No alternative considered. I saw it in linters for different languages than Go.

### Additional context.

_No response_

### LINKED ISSUE #4635: Add options to include/exclude paths

I use https://entgo.io. It generates an /ent directory, and expects you to put your schema definitions in /ent/schema. It then generates code in /ent/foo, /ent/bar, etc. I have no control over the generated code, so I don't want to see linter errors for it. I don't see a way to exclude /ent/\*\*, but include /ent/schema.

run --exclude-path '^ent/.\*$' --include-path '^ent/schema/.\*$'

--exclude-dirs only works on directory names, which won't work if you want to skip some directories and exclude others with the same name.

### Additional context

_No response_

### LINKED ISSUE #5297: 🌟 Let's talk about exclusion options inside issues section

ldez:
This PR is a part of the v2: #5300

---

> [!IMPORTANT]
> This PR deliberately omits documentation about the new options and section (but the JSONSchema is updated).
> This is because I don't want to deprecate the previous section for now: I think it is better to handle that when all the v2 proposals are managed to avoid multiple migrations.

The main goal is to replace `issue.exclude-xxx` options with `linters.exclusions` section.

---

There is a new option `linters.exclusions.warn-unused` that emits a warning if a `linters.exclusions.rules` or `linters.exclusions.paths` is unused.

There is a new option `run.relative-path-mode` with 4 modes:
- `gomod`: paths are relative to the `go.mod`
- `gitroot`: paths are relative to the git root.
- `cfg`: paths are relative to the configuration position.
- `wd` (v1 default): paths are relative to the location where golangci-lint is run.

This option controls all the path-related options:
- `issues.exclude-rules[].path`
- `issues.exclude-rules[].path-except`
- `issues.exclude-dirs`
- `issues.exclude-files`
- `severity.rules[].path`
- `severity.rules[].path-except`
- `custom[].path`
- `linters.exclusions.paths`
- `linters.exclusions.paths-except`
- `linters.exclusions.rules[].path`
- `linters.exclusions.rules[].path-except`
- linters

The linter configuration related to paths should use the placeholder `${base-path}`:
- `gocritic` (`gocritic.ruleguard.rules`)
- `depguard` (`rules[].files`)
- `goheader` (`goheader.template-path`)

`errcheck`: the option `errcheck.exclude` has been deprecated since golangci-lint v1.42 so I didn't add the support of the placeholder.

With the new option `run.relative-path-mode`, you can now run golangci-lint inside a package and the configuration related to file paths will work.

In v2, `output.path-prefix` will be used only for output and not for matching, and the default relative path mode will be `gomod` or `cfg` (not `wd` like today).

---

More details about the topic are inside the issues:
- https://github.com/golangci/golangci-lint/issues/5297
- https://github.com/golangci/golangci-lint/issues/5298

---

<details><summary>configuration overview</summary>

```yml
linters:
  # ...
  
  exclusions:
    presets:
      - comments
      - stdErrorHandling
      - commonFalsePositives
      - legacy
    generated: strict
    warn-unused: true
    rules:
      - path: _test\.go
        linters:
          - gocyclo
          - errcheck
          - dupl
          - gosec
      - path-except: _test\.go
        linters:
          - forbidigo
      - path: internal/hmac/
        linters:
          - gosec
      - linters:
          - staticcheck
        text: "SA9003:"
      - linters:
          - lll
        source: "^//go:generate "
    paths:
      - src/external_libs
      - autogenerated_by_my_lib
      - ".*\\.my\\.go$"
      - lib/bad.go
    paths-except:
      - '^base/a/.+'
```

</details>

---

Fixes #5297
Related to #5298
Fixes #4635
Fixes #3953
Fixes #3717
Fixes #1178
Closes #3748


pkg/result/processors/exclusion_rules.go
```diff
@@ -0,0 +1,190 @@
...
+	if oldCfg.ExcludeCaseSensitive {
+		prefix = ""
+	}
+
+	excludeRules := slices.Clone(cfg.Rules)
+	excludeRules = append(excludeRules, filterInclude(getDefaultLintersExclusions(cfg.Default), oldCfg.IncludeDefaultExcludes)...)
+
+	p.rules = createRules(excludeRules, prefix)
+
+	// TODO(ldez): should be remove in v2.
```
ccoVeille, line 76: Nit

```suggestion
	// TODO(ldez): should be removed in v2.
```

You wrote it randomly at a few place

.golangci.next.reference.yml
```diff
@@ -1254,7 +1254,7 @@ linters-settings:
         failOn: dsl,import
         # Comma-separated list of file paths containing ruleguard rules.
         # If a path is relative, it is relative to the directory where the golangci-lint command is executed.
-        # The special '${configDir}' variable is substituted with the absolute directory containing the golangci config file.
+        # The special '${configDir}' variable is substituted with the absolute directory containing the golangci-lint config file.
```
alexandear, line 1257: I propose moving all typo-related commits to a separate PR (so we can merge it quicker). This will make this PR easier to review because of fewer changes.

ldez: Most of the typos are related to the topic of the PR, also there is one commit by changes to ease the review.

So I will not move them.

pkg/result/processors/exclusion_paths.go
```diff
@@ -0,0 +1,118 @@
...
+		return true
+	}
+
+	matched := false
+	for _, pattern := range p.pathExceptPatterns {
+		if !pattern.MatchString(issue.RelativePath) {
+			continue
+		}
+
+		p.excludedPathExceptCounter[pattern] += 1
```
alexandear, line 113:
```suggestion
		p.excludedPathExceptCounter[pattern]++
```

pkg/result/processors/exclusion_paths.go
```diff
@@ -0,0 +1,118 @@
...
+		if p.warnUnused && count == 0 {
+			p.log.Warnf("The pattern %q match %d issues", pattern, count)
+		}
+	}
+}
+
+func (p *ExclusionPaths) shouldPassIssue(issue *result.Issue) bool {
+	for _, pattern := range p.pathPatterns {
+		if pattern.MatchString(issue.RelativePath) {
+			p.excludedPathCounter[pattern] += 1
```
alexandear, line 98:
```suggestion
			p.excludedPathCounter[pattern]++
```

ldez:
@bombsimon @alexandear Can I do something to ease the review?

bombsimon:
> @bombsimon @alexandear Can I do something to ease the review?
> 

Sorry for slow review! I don't think so, mostly just ben busy and given your _very_ thorough research on this topic in the issue (thanks for that amazing work) I wanted to also catch up with all your findings to better understand these changes. 

I'll try to get to this asap during the week!

pkg/result/processors/uniq_by_line.go
```diff
@@ -12,13 +11,13 @@ var _ Processor = (*UniqByLine)(nil)
 // UniqByLine filters reports to keep only one report by line of code.
 type UniqByLine struct {
 	fileLineCounter fileLineCounter
-	cfg             *config.Config
+	enable          bool
```
alexandear, line 14: Perhaps better:
```suggestion
	enabled         bool
```

pkg/config/loader.go
```diff
@@ -68,6 +68,28 @@ func (l *Loader) Load(opts LoadOptions) error {
 
 	l.applyStringSliceHack()
 
+	if l.cfg.Linters.LinterExclusions.Generated == "" {
+		// This is always non-empty because of the flag default value.
+		if l.cfg.Issues.ExcludeGenerated != "" {
+			l.cfg.Linters.LinterExclusions.Generated = l.cfg.Issues.ExcludeGenerated
+		} else {
+			l.cfg.Linters.LinterExclusions.Generated = "strict"
+		}
```
alexandear, line 77: We can avoid `else`:

```suggestion
		l.cfg.Linters.LinterExclusions.Generated = cmp.Or(l.cfg.Issues.ExcludeGenerated, "strict")
```

pkg/result/processors/exclusion_rules.go
```diff
@@ -0,0 +1,167 @@
...
+	"github.com/golangci/golangci-lint/pkg/result"
+)
+
+var _ Processor = (*ExclusionRules)(nil)
+
+type excludeRule struct {
+	baseRule
+}
+
+func (e excludeRule) String() string {
```
alexandear, line 21: I propose to move this unexported type below after the `ExclusionRules` to increase readability. What really matters is `ExclusionRules` struct, so it should be first.

pkg/result/processors/exclusion_rules.go
```diff
@@ -0,0 +1,190 @@
...
+	}
+
+	// TODO(ldez) remove prefix in v2: the matching must be case sensitive, users can add `(?i)` inside the patterns if needed.
+	prefix := caseInsensitivePrefix
+	if oldCfg.ExcludeCaseSensitive {
+		prefix = ""
+	}
+
+	excludeRules := slices.Clone(cfg.Rules)
+	excludeRules = append(excludeRules, filterInclude(getDefaultLintersExclusions(cfg.Default), oldCfg.IncludeDefaultExcludes)...)
```
alexandear, line 72: You are using `slices.Clone`, so we can use `slices.Concat` instead of `append`.

Perhaps this is more readable (but it's up to you):

```go
	excludeRules := slices.Concat(
		slices.Clone(cfg.Rules),
		filterInclude(getDefaultLintersExclusions(cfg.Default), oldCfg.IncludeDefaultExcludes),
	)
```

pkg/result/processors/exclusion_rules.go
```diff
@@ -0,0 +1,190 @@
...
+	parsedRules := make([]excludeRule, 0, len(rules))
+
+	for _, rule := range rules {
+		parsedRules = append(parsedRules, createRule(&rule, prefix))
+	}
+
+	return parsedRules
+}
+
+func createRule(rule *config.ExcludeRule, prefix string) excludeRule {
```
alexandear, line 152: This function `createRule` can be a constructor for the `excludeRule`.

pkg/result/processors/exclusion_rules.go
```diff
@@ -0,0 +1,190 @@
...
+	for rule, count := range p.skippedCounter {
+		if p.warnUnused && count == 0 {
+			p.log.Warnf("Skipped %d issues by rules: [%s]", count, rule)
+		} else {
+			p.log.Infof("Skipped %d issues by rules: [%s]", count, rule)
+		}
+	}
+}
+
+func createRules(rules []config.ExcludeRule, prefix string) []excludeRule {
```
alexandear, line 138: It makes sense to rename `createRules` to `createExcludeRules`. This helps to avoid possible name collisions in the future.

pkg/result/processors/exclusion_rules.go
```diff
@@ -0,0 +1,190 @@
...
+	parsedRule := excludeRule{}
+	parsedRule.linters = rule.Linters
+	parsedRule.internalReference = rule.InternalReference
+
+	if rule.Text != "" {
+		parsedRule.text = regexp.MustCompile(prefix + rule.Text)
+	}
+
+	if rule.Source != "" {
+		parsedRule.source = regexp.MustCompile(prefix + rule.Source)
```
alexandear, line 162: Don't know, but maybe it's better to return an error instead of panicking.

pkg/result/processors/exclusion_paths.go
```diff
@@ -0,0 +1,118 @@
...
+	for _, pattern := range p.pathPatterns {
+		if pattern.MatchString(issue.RelativePath) {
+			p.excludedPathCounter[pattern]++
+			return false
+		}
+	}
+
+	if len(p.pathExceptPatterns) == 0 {
+		return true
+	}
```
alexandear, line 105: This check seems redundant. The following loop will not iterate if `p.pathExceptPatterns` is empty, `matched` will be `false`, and returned value is `true`.

alexandear (approved):

pkg/result/processors/exclusion_paths.go
ldez: this check is for readability.

pkg/fsutils/basepath.go
```diff
@@ -0,0 +1,77 @@
...
+const (
+	RelativePathModeGoMod   = "gomod"
+	RelativePathModeGitRoot = "gitroot"
+	RelativePathModeCfg     = "cfg"
+	RelativePathModeWd      = "wd"
+)
+
+func AllRelativePathModes() []string {
+	return []string{RelativePathModeGoMod, RelativePathModeGitRoot, RelativePathModeCfg, RelativePathModeWd}
+}
```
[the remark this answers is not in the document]
ldez: Not really because the mode values are used by the error messages.

pkg/result/processors/exclusion_rules.go
ldez: It is an existing code, I just moved it.
I will not change it inside this PR, this has too much impact on my next PRs, I will change that in another PR.

pkg/golinters/goheader/goheader.go
```diff
@@ -9,17 +9,18 @@ import (
...
 
-func New(settings *config.GoHeaderSettings) *goanalysis.Linter {
+func New(settings *config.GoHeaderSettings, basePath string) *goanalysis.Linter {
 	conf := &goheader.Configuration{}
 	if settings != nil {
 		conf = &goheader.Configuration{
 			Values:       settings.Values,
 			Template:     settings.Template,
-			TemplatePath: settings.TemplatePath,
+			TemplatePath: strings.ReplaceAll(settings.TemplatePath, internal.PlaceholderBasePath, basePath),
```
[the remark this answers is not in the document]
ldez: It makes sense if you don't know the complexity of doing that because playing with the paths inside integration tests is a bit hacky.
Also, adding an integration test, to check if `strings.ReplaceAll` works, seems overkill.

pkg/config/linters_exclusions.go
```diff
@@ -0,0 +1,41 @@
...
+type LinterExclusions struct {
+	Generated   string        `mapstructure:"generated"`
+	WarnUnused  bool          `mapstructure:"warn-unused"`
+	Default     []string      `mapstructure:"default"`
+	Rules       []ExcludeRule `mapstructure:"rules"`
+	Paths       []string      `mapstructure:"paths"`
+	PathsExcept []string      `mapstructure:"paths-except"`
+}
+
+func (e *LinterExclusions) Validate() error {
```
bombsimon, line 25: Should we validate the values in `Default` as well? Looks like we blindly pass it to `getDefaultLintersExclusions` which assumes the values will exist in `defaultLintersExclusions`?

pkg/fsutils/basepath.go
```diff
@@ -0,0 +1,77 @@
...
+
+		return wd, nil
+
+	default:
+		return "", errors.New("unknown relative path mode")
+	}
+}
+
+func gitRoot(ctx context.Context) (string, error) {
+	cmd := exec.CommandContext(ctx, "git", "rev-parse", "--show-toplevel")
```
bombsimon, line 70: This is a new dependency right, is it reasonable to add `git` as a dependency? Probably worth mentio this in the future close to where the mode is described.

pkg/golinters/gocritic/testdata/gocritic.yml
```diff
@@ -17,4 +17,7 @@ linters-settings:
         #   The ruleguard files used in functional tests cannot be under the 'testdata' directory.
         #   This is because they import the 'github.com/quasilyte/go-ruleguard/dsl' package,
         #   which needs to be added to go.mod. The testdata directory is ignored by go mod.
-        rules: '${configDir}/ruleguard/preferWriteString.go,${configDir}/ruleguard/stringsSimplify.go'
+        rules: '${base-path}/ruleguard/preferWriteString.go,${base-path}/ruleguard/stringsSimplify.go'
```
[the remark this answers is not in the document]
bombsimon: I think it's more important to have this visible close to the description of `${base-path}` than in the test. The person who will maintain this test (or similar in new linters) should consult the docs for how paths work.

pkg/result/processors/exclusion_default.go
```diff
@@ -0,0 +1,136 @@
...
+			},
+		},
+	},
+}
+
+func getDefaultLintersExclusions(names []string) []config.ExcludeRule {
+	var rules []config.ExcludeRule
+
+	for _, name := range names {
+		rules = append(rules, defaultLintersExclusions[name]...)
```
bombsimon, line 132: Should we assert the map key exist before spreading?

pkg/result/processors/path_relativity.go
```diff
@@ -0,0 +1,60 @@
...
+)
+
+var _ Processor = (*PathRelativity)(nil)
+
+// PathRelativity computes [result.Issue.RelativePath] and [result.Issue.WorkingDirectoryRelativePath],
+// based on the base path.
+type PathRelativity struct {
+	log      logutils.Log
+	basePath string
+	wd       string
```
bombsimon, line 19: personal pref: I prefer avoiding abbreviations like this on struct fields since having the full name avoids any ambiguity and makes it clearer. Don't seem like we refer to it an awful lot to make it an issue in the code.

```suggestion
	workingDirectory       string
```

pkg/result/processors/path_relativity.go
```diff
@@ -0,0 +1,60 @@
...
+			return nil
+		}
+
+		return &newIssue
+	}), nil
+}
+
+func (*PathRelativity) Name() string {
+	return "path_relativity"
+}
```
bombsimon, line 58: nit: Other processors seem to put `Name()` after the constructor before `Process` (which I think makes sense).

bombsimon (approved):
LGTM, good stuff!

pkg/config/loader.go
```diff
@@ -68,6 +68,24 @@ func (l *Loader) Load(opts LoadOptions) error {
 
 	l.applyStringSliceHack()
 
+	if l.cfg.Linters.LinterExclusions.Generated == "" {
+		// `l.cfg.Issues.ExcludeGenerated` is always non-empty because of the flag default value.
+		l.cfg.Linters.LinterExclusions.Generated = cmp.Or(l.cfg.Issues.ExcludeGenerated, "strict")
```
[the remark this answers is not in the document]
ldez: It's in the next PR :wink:

pkg/config/linters_exclusions.go
```diff
@@ -0,0 +1,41 @@
...
+	DefaultExclusionCommonFalsePositives = "commonFalsePositives"
+	DefaultExclusionLegacy               = "legacy"
+)
+
+const excludeRuleMinConditionsCount = 2
+
+type LinterExclusions struct {
+	Generated   string        `mapstructure:"generated"`
+	WarnUnused  bool          `mapstructure:"warn-unused"`
+	Default     []string      `mapstructure:"default"`
```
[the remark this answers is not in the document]
ldez: `presets` or `builtin`? :thinking: 

I think `preset` is right :thinking:

pkg/config/loader.go
```diff
@@ -68,6 +68,24 @@ func (l *Loader) Load(opts LoadOptions) error {
...
+	}
+
+	if l.cfg.Issues.UseDefaultExcludes {
```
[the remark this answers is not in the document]
ldez: This is the current field name of the "old" option, I don't think we need to change the name as we will remove it.

pkg/fsutils/basepath.go
ldez: we already have `git` as "dependency": the diff processor is based on it.

ldez: Also, this is only called if the mode `gitroot` is used.

The modes are documented.

pkg/golinters/gocritic/testdata/gocritic.yml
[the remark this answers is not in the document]
ldez: I already removed the comment inside those test files.

ldez:
I'm so excited to merge this PR, I know this is mainly internal but I spend a lot of time on it, and the next PR (I will open it 1 minute after the merge of this one) will add the `fmt` command :tada:

ldez:
I will press the merge button :tada: 
I'm so excited

pkg/fsutils/basepath.go
bombsimon: Ah, I tried to look for any other places but couldn't find one, at least not one that shells out like this.

ldez: This is inside `revgrep`: https://github.com/golangci/revgrep/blob/main/patch.go

PR #5357, opened 2025-01-28
feat: new `fmt` command with dedicated formatter configuration

### LINKED ISSUE #5296: 🌟 Let's talk about "formatters"
> [!IMPORTANT]
> The proposal may evolve.

---

| tool        | format | imports order        | fix imports | rewrite-rules | simplify | extra |
|-------------|--------|----------------------|-------------|---------------|----------|-------|
| `gofmt`     | x      | x (alphabetically)   |             | x             | x        |       |
| `goimports` | x      | x (multiple imports) | x           |               |          |       |
| `gofumpt`   | x      | x (one import)       |             |               | x        | x     |
| `gci`       | x      | x (customizable)     |             |               |          |       |

### Types

The formatters don't require types, but when they are run with other linters that require types, the possibilities of formaters to apply format and fix the code are limited.

Today, for example, you can use the following command to apply the format without being constraint by the other linters:

```bash
golangci-lint run --enable-only="gofumpt,goimports" --fix
```

It works but this is limited.

## 💭 The Proposal

The idea is to create a dedicated section `formatters` like `linters` (internally they will be used as the other linters).

And adds a specific command `fmt` that runs only the formatters defined inside the configuration, like a variant of the `run` command but limited to formatters and with "autofix" always enabled, so no reports.

It allows us to clarify the difference between the formatters and the linters.

Also, we can set up a dedicated default for this section.
I suggest `goimports` as default because this is an official Go tool, more powerful than `gofmt`, and fits the needs of many projects.

```yml
formatters:
  enable:
    - gofumpt
    - gofmt
    - goimports
    - gci
  settings:
    gofumpt:
      opt1: true
    # ...
```

Enabling all the formatters make no sense as they provide the same base functionality (format) with some variants but their changes could be incompatible (mainly on imports) unless they are run in a specific order and without incompatible settings. So no need for `enable-all` and/or a `disable` section.

ldez:
> [!IMPORTANT]
> This PR deliberately omits documentation about the new options and section (but the JSONSchema is updated).
> This is because I don't want to deprecate the previous elements for now: I think it is better to handle that when all the v2 proposals are managed to avoid multiple migrations.

---

The command `golangci-lint fmt` always format the code.

The current default is same format used by `go fmt` (!= `gofmt`).

There is no support for stdin.

The formatters defined in the `formatters.enable` are added automatically to `linters.enable`.

If a formatter is enabled inside `formatters` and  `linters`, the configuration of `formatters.settings` will override `linters-settings` even if the configuration is omitted or empty inside `formatters.settings` section.
This means the formatter activation and configurations should be defined only on the `formatters` or `linters`+`linters-settings` section and not mixed.

The exclusions from `formatters.exclusions[].paths` are converted to `linters.exclusions[].rules` automatically.

There are only 2(3) flags:

```
Flags:
  -c, --config PATH      Read config from file path PATH
      --no-config        Don't read config file
  -E, --enable strings   Enable specific formatter
```

Note, the `-E, --enable` flags works like `--enable-only` for linters: it overrides the formatters defined inside the configuration file.
It is the opposite of the behavior the `-E, --enable` flags of the `run` command where this flag always adds linters to those defined inside the configuration file.

In v2, the configuration of formatters will not be allowed inside the `linters-settings` section.

```yml
formatters:
  enable:
    - gofumpt
    - gofmt
    - goimports
    - gci
  settings:
    gofumpt:
      # ...
    gofmt:
      # ...
    goimports:
      # ...
    gci:
      # ...
  exclusions:
    generated: strict
    paths:
      - foo
      - bar
      # ...
```

Fixes #5296


pkg/config/loader.go
```diff
@@ -495,6 +500,59 @@ func (l *Loader) handleEnableOnlyOption() error {
...
+	for _, path := range l.cfg.Formatters.Exclusions.Paths {
+		l.cfg.Linters.LinterExclusions.Rules = append(l.cfg.Linters.LinterExclusions.Rules, ExcludeRule{
+			BaseRule: BaseRule{
+				Linters: l.cfg.Formatters.Enable,
+				Path:    path,
+			},
+		})
+	}
+}
+func (l *Loader) handleFormatterDeprecations() {
```
alexandear, line 542: The `formatters` configuration is completely new. Do we really need to create deprecations for it? I know you are mimicking `linters`, and this will be removed in v2, but still, this complicates the logic.

cmd/golangci-lint/main.go
```diff
@@ -64,11 +64,10 @@ func createBuildInfo() commands.BuildInfo {
 		}
 	}
 
-	revision = cmp.Or(revision, "unknown")
-	modified = cmp.Or(modified, "?")
 	info.Date = cmp.Or(info.Date, "(unknown)")
 
-	info.Commit = fmt.Sprintf("(%s, modified: %s, mod sum: %q)", revision, modified, buildInfo.Main.Sum)
+	info.Commit = fmt.Sprintf("(%s, modified: %s, mod sum: %q)",
+		cmp.Or(revision, "unknown"), cmp.Or(modified, "?"), buildInfo.Main.Sum)
```
alexandear, line 70: Nit: It's hard to understand how this change is connected to the topic of the PR.

pkg/commands/fmt.go
```diff
@@ -0,0 +1,159 @@
...
+	if len(args) == 0 {
+		abs, err := filepath.Abs(".")
+		if err != nil {
+			return nil, err
+		}
+
+		return []string{abs}, nil
+	}
+
+	var expended []string
```
alexandear, line 148: Perhaps here is a typo:
```suggestion
	var expanded []string
```

pkg/commands/fmt.go
```diff
@@ -0,0 +1,159 @@
...
+		return fmt.Errorf("build walk options: %w", err)
+	}
+
+	c.runner = goformat.NewRunner(c.log, metaFormatter, matcher, opts)
+
+	return nil
+}
+
+func (c *fmtCommand) execute(_ *cobra.Command, args []string) error {
+	if !logutils.HaveDebugTag(logutils.DebugKeyLintersOutput) {
```
alexandear, line 102: Maybe we need to introduce a separate debug key for formatters.

pkg/goformat/runner.go
```diff
@@ -0,0 +1,166 @@
+package goformat
```
alexandear, line 1: I propose renaming the package to `fmt`. Functions from this package are used only in `pkg/commands/fmt.go`, so it will be okay. Additionally, this avoids a possible collision with the `goformatters` package.

pkg/goformat/runner.go
```diff
@@ -0,0 +1,166 @@
...
+	return filepath.Walk(root, func(path string, f fs.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+
+		if f.IsDir() && skipDir(f.Name()) {
+			return fs.SkipDir
+		}
+
+		// Ignore non-Go files.
```
alexandear, line 61: Looks like we can remove this comment, as the check below is clear enough.

pkg/goformat/runner.go
```diff
@@ -0,0 +1,166 @@
...
+			return RunnerOptions{}, fmt.Errorf("compile path pattern %q: %w", pattern, err)
+		}
+
+		opts.patterns = append(opts.patterns, exp)
+	}
+
+	return opts, nil
+}
+
+func (o RunnerOptions) MatchPatterns(path string) (bool, error) {
```
alexandear, line 128: Perhaps a better name is `MatchAnyPattern`

pkg/goformat/runner.go
```diff
@@ -0,0 +1,166 @@
...
+
+	case "third_party", "builtin": // For compatibility with `exclude-dirs-use-default`.
+		return true
+
+	default:
+		if strings.HasPrefix(name, ".") {
+			return true
+		}
+
+		return false
```
alexandear, line 160: nit: can be
```suggestion
		return strings.HasPrefix(name, ".")
```

pkg/goformat/runner.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+}
+
+func isGoFile(f fs.FileInfo) bool {
+	return !f.IsDir() && !strings.HasPrefix(f.Name(), ".") && strings.HasSuffix(f.Name(), ".go")
```
alexandear, line 165: Can we use `filepath.Ext(f.Name()) == ".go"` instead of `strings.HasSuffix(f.Name(), ".go")`?

alexandear (approved):

pkg/config/loader.go
ldez: I did it voluntarily to ease the cleaning of the v2 code.

pkg/config/formatters_settings.go
```diff
@@ -0,0 +1,52 @@
...
+	Gci: GciSettings{
+		Sections:      []string{"standard", "default"},
+		SkipGenerated: true,
+	},
+}
+
+type FormatterSettings struct {
+	Gci       GciSettings       `mapstructure:"gci"`
+	GoFmt     GoFmtSettings     `mapstructure:"gofmt"`
+	GoFumpt   GoFumptSettings   `mapstructure:"gofumpt"`
```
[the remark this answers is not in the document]
ldez: because you pronounce "go imports" and not "Goim port".

They are internal structures, not user-facing elements.

pkg/goformat/runner.go
ldez: https://github.com/golang/go/blob/e81f7155154c0f5d40363e84a8f24a5b559b5eed/src/cmd/fix/main.go#L264

pkg/goformat/runner.go
```diff
@@ -0,0 +1,161 @@
...
+	return false, nil
+}
+
+func skipDir(name string) bool {
+	switch name {
+	case "vendor", "testdata", "node_modules":
+		return true
+
+	case "third_party", "builtin": // For compatibility with `exclude-dirs-use-default`.
+		return true
```
bombsimon, line 152: Is this the same list as for the processors? Could/should that be re-used?

https://github.com/golangci/golangci-lint/blob/fc94060165905335e67333cd98b37ee5e054e359/pkg/result/processors/skip_dirs.go#L15-L22

bombsimon (approved):
LGTM!

I think making this still supported as a linter and available with the `--fix` flag is a great suggestion.

Have you done any testing on performance? I'd imagine `golangci-lint fmt` on big codebases would potentially be faster than `golangci-lint run --fix` or are they the same? When would you suggest one over the other? On save in your editor?

pkg/goformat/runner.go
ldez: the SkipDirs processors will be removed in v2, so I prefer not to share code.

ldez:
> Have you done any testing on performance? I'd imagine golangci-lint fmt on big codebases would potentially be faster than golangci-lint run --fix or are they the same?

Yes, I did it.

The `fmt` is significantly faster than `run`.
Because it's only focused on pure syntax analyzing, no type loading, no cache filling, etc.

> When would you suggest one over the other? On save in your editor?

The 2 commands have different goals, I cannot suggest one over the other.

But I suggest always running `fmt` first because it's quick and safe.

ldez:
I will not merge this PR for now: based on Go [release cycle](https://go.dev/wiki/Go-Release-Cycle), go1.24 will be released in 2 weeks.

I want to keep this feature for v2 and I want the last v1 release will support go1.24.

Antonboom:
Can we validate and warn:

- if gci and goimports enabled both
- if gofmt and gofumt enabled both
- if formatter key presented in linters-settings, but linter enabled in formatters (and visa versa)

?

Sorry, If its already presented, but I missed it 🙏

ldez:
> if gci and goimports enabled both

We don't need to do that because the way I implemented the run the conflicts are not possible.

> if gofmt and gofumt enabled both

The 2 linters should be able to run together because they don't have the same possibilities, and some are complementary.

> if formatter key presented in linters-settings, but linter enabled in formatters (and visa vers

In v2, the formatters will not be allowed inside linter settings and definitions.

pkg/result/processors/exclusion_generated_file_matcher_test.go
```diff
@@ -0,0 +1,154 @@
...
+		file, err := parser.ParseFile(token.NewFileSet(), tc.fpath, nil, parser.PackageClauseOnly|parser.ParseComments)
+		require.NoError(t, err)
+
+		doc := getComments(file)
+		assert.Equal(t, tc.doc, doc)
+	}
+}
+
+// Issue 954: Some lines can be very long, e.g. auto-generated
+// embedded resources. Reported on file of 86.2KB.
```
alexandear, line 146: nit:
```suggestion
// Issue #954: Some lines can be very long, e.g. auto-generated embedded resources.
// Reported on file of 86.2KB.
```

When GitHub sees `#954` (or `golangci/golangci-lint#954`), it renders it as a link to the issue `https://github.com/golangci/golangci-lint/issues/954`

image

This issue-link rendering from `#<num>` also works in popular IDEs.

pkg/result/processors/exclusion_generated_file_matcher.go
```diff
@@ -0,0 +1,107 @@
...
+	}
+
+	doc := getComments(file)
+
+	return p.isGeneratedFileLax(doc), nil
+}
+
+// isGeneratedFileLax reports whether the source file is generated code.
+// The function uses a bit laxer rules than isGeneratedFileStrict to match more generated code.
+// See https://github.com/golangci/golangci-lint/issues/48 and https://github.com/golangci/golangci-lint/issues/72.
```
alexandear, line 67: Can be simple:
```suggestion
// See #48 and #72.
```

ldez: The `#` links are not rendering in IDEs.

pkg/result/processors/exclusion_generated_file_matcher_test.go
alexandear: Oh, maybe it's only VS Code renders them.

Screenshot 2025-02-14 at 14 11 51

Can be resolved.

PR #5394, opened 2025-02-12
dev: fix JSONSchema publication

ldez:
The destination of the copy of the latest schema was wrong.


ldez:
I bypass the requirements to fix the problem quickly.

PR #5395, opened 2025-02-12
dev: fix JSONSchema publication

ldez:
Multiple reads of the same `io.Reader`.


ldez:
I bypass the requirements to fix the problem quickly.

alexandear (commented):
Maybe we can add tests that decrease chances of problems during a future release.

PR #5396, opened 2025-02-12
Specify `go` and `toolchain` versions separately

AlekSi:
That makes it easier to consume via 1.24's `go tool`.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/golangci/golangci-lint
 
-go 1.23.0
+go 1.23
+
+toolchain go1.23.6
 
 require (
 	4d63.com/gocheckcompilerdirectives v1.2.1


alexandear (commented):
Could you please provide a more detailed description of the problem? It is currently difficult to understand the issue that this PR is intended to address.

AlekSi:
Ok, actually, there is no _problem_ (or at least I can't reproduce it anymore after switching to 1.24).
But I think it is just weird to require the 1.23.0 _language version_ (that's what the `go` directive means now) and not just 1.23.
The somewhat related fix was done in #5112.

ldez:
The min Go version can be a release version or the family version, the Go documentation is not clear but this is perfectly right:
- https://go.dev/ref/mod#go-mod-file-go
- https://go.dev/doc/toolchain#version

So we don't need to have an explicit toolchain version.

Thank you anyway.

AlekSi:
I know that it _can_ be. But I think it _shouldn't_ be in that case: there is nothing about 1.24.0 specifically that golangci-lint requires. For example, you would want to use 1.24.1 once it is released. Then, you would update `go` directive to use 1.24.1 without adding `toolchain` directive. In turn, that would force all users of golangci-lint that specify it in the `tool` directive to update to 1.24.1 – and there is no reason for that if they are not affected by 1.24.0 bugs.

ldez:
In fact, we don't update the min Go version with a patched version for the same reasons we are using min go 1.23 instead of go1.24.

PR #5402, opened 2025-02-12
gci: fix standard packages list

ldez:
Allows to fix gci before the merge of the PR https://github.com/daixiang0/gci/pull/227


alexandear (approved):

pfi79:
@ldez Thank you for your change.

But it didn't work

pkg/goformatters/gci/internal/config/config.go
```diff
@@ -0,0 +1,106 @@
...
+	section.DotType:         4,
+	section.AliasType:       5,
+	section.LocalModuleType: 6,
+}
+
+type Config struct {
+	config.BoolConfig
+	Sections          section.SectionList
+	SectionSeparators section.SectionList
+}
```
pfi79, line 27: I think the problem here

ldez: this is a copy https://github.com/daixiang0/gci/blob/dca258a399bf32ab20c73ab4edb726cfff7224ef/pkg/config/config.go#L32-L36

pkg/goformatters/gci/internal/config/config.go
```diff
@@ -0,0 +1,106 @@
+package config
+
+import (
+	"sort"
+	"strings"
+
+	"gopkg.in/yaml.v3"
+
+	"github.com/daixiang0/gci/pkg/config"
+	"github.com/daixiang0/gci/pkg/section"
```
pfi79, line 10: I think the problem here.

"github.com/daixiang0/gci/pkg/section"
vs
"github.com/golangci/golangci-lint/pkg/goformatters/gci/internal/section"

pkg/goformatters/gci/internal/config/config.go
pfi79: I got it. You don't have `github.com/golangci/golangci-lint/pkg/goformatters/gci/internal/section` used anywhere

pkg/goformatters/gci/internal/config/config.go
ldez: I don't so because `Section` is an interface

pfi79: you created the package `github.com/golangci/golangci-lint/pkg/goformatters/gci/internal/section`
but the programme won't get into it.

I went through this in debugging

ldez: The problem is related to the `Parse`, I will fix that.

pfi79: Thank you for the quick fix

ldez: https://github.com/golangci/golangci-lint/pull/5407

Issue #5225, opened 2024-12-13
go1.24 support


ldez:
**EDIT: since v1.64 golangci-lint supports go1.24 if compiled with go1.24 :tada:**

#### Additional Notes

Some package systems (like **homebrew**, nix, etc.) are "slow" in propagating the new Go version.
After a new release of Go and golangci-lint, those systems can use an old Go version to compile golangci-lint resulting in a non-working binary (high memory usage, missing symbol, etc.).

In this situation, you should use the official binaries, or compile golangci-lint manually with the right Go version, or help these package systems to support this new Go version.

We are not responsible for package systems that don't use official binaries.

To validate the Go version used to compile: `golangci-lint version`

```console
$ golangci-lint version
golangci-lint has version 1.64.2 built with go1.24.0 from 1dae906b on 2025-02-11T21:44:55Z
```

---

#### Is golangci-lint support go1.24?

No, and yes.

Basically, golangci-lint supports Go versions that are lower or equal to the Go version used **to build it**.

In addition, of the Go version used to build it, some linters could need to be updated to support newer versions of Go, and internal pieces of golangci-lint could also need to be updated.

The current binaries from the releases page and the Docker images have been built with go1.23 and will be built with go1.23 as long as go1.24 is not GA.
So those binaries/images don't support go1.24 (new elements and internal Go changes).

https://golangci-lint.run/welcome/faq/#which-go-versions-are-supported

```mermaid
---
title: Is golangci-lint support go1.24?
---
flowchart
    A([compiled with go1.24?]) -- yes --> B([needs update of the internals of golangci-lint?])
    A -- no --> C((NO))
    B -- yes --> D([uses right internal pieces?])
    B -- no --> E((YES))
    D -- yes --> G((YES))
    D -- no --> F((NO))
```

You can follow the PR https://github.com/golangci/golangci-lint/pull/5224 to see the progress on this topic.

https://go.dev/wiki/Go-Release-Cycle

#### Additional Notes

Some package systems (like **homebrew**, nix, etc.) are "slow" in propagating the new Go version.
After a new release of Go and golangci-lint, those systems can use an old Go version to compile golangci-lint resulting in a non-working binary (high memory usage, missing symbol, etc.).

In this situation, you should use the official binaries, or compile golangci-lint manually with the right Go version, or help these package systems to support this new Go version.

We are not responsible for package systems that don't use official binaries.

To validate the Go version used to compile: `golangci-lint version`

```console
$ golangci-lint version
[...]

PR #5417, opened 2025-02-14
Add betteralign linter

dkorunic:
[betteralign](https://github.com/dkorunic/betteralign) is a tool to detect structs that would (theoretically) use less memory if their fields were sorted and optionally sort such fields. This is a fork of an official Go [fieldalignment](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/fieldalignment) tool, but there are some notable changes in regards how it automatically skips over various generated files, test files, structs marked with `betteralign:ignore` comments, etc.

The main feature is the ability to retain comment positions (field, doc, floating comments etc.) when rewriting files, although at least in Golangci-lint situation we can't have it enabled as it doesn't work through SuggestedFixes due to regular AST not retaining comment offsets/positions, so we dump decorated AST on our own, rewriting files directly.

Given how it primarily works through [decorated AST](https://github.com/dave/dst), it might not be the best fit for golangci-lint, but I have been asked a [few times](https://github.com/dkorunic/betteralign/issues/1) to send a PR...


ldez:

- [ ] The CLA must be signed

### Pull Request Description

### Linter

- [ ] It must use Go version <= 1.23

### Others Requirements

### Recommendations

---

ldez:
The main differences between betteralign and fieldalignment are related to CLI options, not to the core of the analysis.

Those options are already handled by golangci-lint (ignore generated, tests, etc.).

So this is a duplicate.

dkorunic:
While I don't mind having the PR closed (I really expected this, it's all fine and I understand), main difference is in *decorated AST* which means all comments locations/offsets are preserved when rewriting code. CLI is really irrelevant there, it's barebones singlechecker.

ldez:
The decorated AST is not compatible with the suggested fixes and so with the golangci-lint approach.

Golangci-lint should be compiled with go1.23 and go1.24 and not only go1.24.

ldez:
> In terms of why Go 1.24, go/analysis fails to lint Go 1.24-compiled code using Go <1.24 linter and throws error

You don't need to define the min go version to 1.24 to compile with go1.24: the min go version is related to the syntax used inside your code.

The toolchain version is used to control the go version used to compile.

dkorunic:
@ldez My apologies, but it doesn't seem we are talking about the same thing. I am referring to how linters (for instance golangci-lint in the following example) don't work with Go 1.24 programs that have both go and toolchain set to 1.24 (for instance to force Abseil-variant of the map):

```
$ golangci-lint run
Error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24)
Failed executing command with error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24)
```

Same applies for other linters.

ldez:
golangci-lint works with go1.24 when compiled with go1.24, but not when compiled with go1.23, this is expected because golangci-lint can be used with the tool pattern (even if we discourage this usage) and so we should allow golangci-lint to be compiled with go1.23.

- https://golangci-lint.run/welcome/faq/#which-go-versions-are-supported
- https://github.com/golangci/golangci-lint/issues/5225

The usage of the min Go version to force the compilation with go1.24 is an "abuse"/hack of the system.

All our binaries are compiled with go1.24 and work with go1.24.

PR #5426, opened 2025-02-17
Require type for oneOf mutual exclusion

Namnamseo:
Hello!

The following config:

```yaml
linters-settings:
  custom:
    foo:
      path: "bin/linter/myplugin.so"
```

is considered broken under current JSON schema for config:

```jsonc
{
  "custom": {
    "type": "object",
    "patternProperties": {
      "^.*$": {
        "type": "object",
        "additionalProperties": false,
        "properties": {
          "type": {
            "enum": [ "module", "goplugin" ],
            "default": "goplugin"
          },
          "path": {
            "type": "string",
            "examples": [ "/path/to/example.so" ]
          }, /* ... */
        },
        "oneOf": [
          {
            "properties": {
              "type": { "enum": [ "module" ] }
            }
          },
          {
            "required": [ "path" ]
          }
        ]
      }
    }
  }
}
```

for two reasons:
* In JSON Schema, `oneOf` requires *exactly one of each options*. When both sub-schemas match the object, the validation fails.
* In JSON Schema, all properties are deemed optional unless explicitly stated to be `"required"`.

The given YAML:
* matches the first sub-schema, which states: "If it has a `type` property, it should be one among `["module"]`.". Since it doesn't have that property, the statement is (vacuously) true.
* matches the second sub-schema, which states: "It should have a `"path"` property.". It does.

I think the `oneOf` clause was there to do a mutual exclusion between "module" and "goplugin" type. Since the default is considered the latter, the "module" case should explicitly require for the property to be present.
This PR adds the requirement, and adds some test cases for that.

--------------

Since I'm new to the code base, I don't know which files among `jsonschema/*.json` needs to change; please point me if I'm wrong.

### DIFF
--- a/jsonschema/golangci.jsonschema.json
+++ b/jsonschema/golangci.jsonschema.json
@@ -3846,7 +3846,8 @@
                 {
                   "properties": {
                     "type": {"enum": ["module"] }
-                  }
+                  },
+                  "required": ["type"]
                 },
                 {
                   "required": ["path"]
--- a/jsonschema/golangci.next.jsonschema.json
+++ b/jsonschema/golangci.next.jsonschema.json
@@ -3846,7 +3846,8 @@
                 {
                   "properties": {
                     "type": {"enum": ["module"] }
-                  }
+                  },
+                  "required": ["type"]
                 },
                 {
                   "required": ["path"]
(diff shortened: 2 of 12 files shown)


ldez (approved):
LGTM

PR #5463, opened 2025-02-22
govet: skip fieldalignment test on 32bit platforms

realsn0w:
the govet fieldalignment test currently fails on 32bit targets because it assumes `uintptr` always has a size of 8 bytes.
this pr (edited after some discussion) adds build tags to skip it on platforms where that is not the case.

### DIFF
--- a/pkg/golinters/govet/testdata/govet_fieldalignment.go
+++ b/pkg/golinters/govet/testdata/govet_fieldalignment.go
@@ -1,3 +1,4 @@
+//go:build !(386 || arm || mips || mipsle)
 //golangcitest:args -Egovet
 //golangcitest:config_path testdata/govet_fieldalignment.yml
 package testdata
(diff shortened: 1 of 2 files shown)


ldez:
Hello, in which context do you need to run our tests inside a 32-bit arch?

realsn0w:
ran into this while working on the golangci-lint void package.
ci there runs a check stage after build for non-cross targets.
[here](https://github.com/void-linux/void-packages/actions/runs/12097732789/job/33733423002?pr=53270#step:7:3045) is an example of a build failure caused by this.

also just in general, i feel like it would be nice if the tests work for all of the platforms you currently support / provide official binaries for (which currently includes armv6/7 and i386).

realsn0w:
alright, that sounds like a fair compromise.
thanks for the quick replies :slightly_smiling_face:

ldez (approved):
LGTM

PR #5497, opened 2025-03-03
docs: add GitLab documentation

nachtjasmin:
Since there's some kind of official documentation for this, I decided to link it.

edit: we could even link to the [official "CI/CD component"](https://gitlab.com/components/code-quality-oss/codequality-os-scanners-integration/-/blob/main/templates/golangci.yml) if wanted.

### DIFF
--- a/docs/src/docs/welcome/install.mdx
+++ b/docs/src/docs/welcome/install.mdx
@@ -25,6 +25,18 @@ Also, the action creates GitHub annotations for found issues (you don't need to 
 
 ![GitHub annotations of the action](./annotations.png)
 
+### GitLab CI
+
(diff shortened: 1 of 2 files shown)


ldez (changes requested):
Hello,

I think the content of "CI Integration" should be replaced by a link to https://golangci-lint.run/welcome/install/#ci-installation

And your modifications should be moved to [this section](https://golangci-lint.run/welcome/install/#ci-installation).

nachtjasmin:
Just added one additional note about the domain of CI components, just so that self-hosted users do not have to search for that in the GitLab documentation first.

ldez (approved):
LGTM

PR #5501, opened 2025-03-03
dev: fix schema not accepting valid timeout

SuperSandro2000:
Tested with https://regex101.com/r/7h45UC/1

While doing https://github.com/sapcc/go-makefile-maker/pull/248 we noticed that valid time.Durations are not accepted in the schema which got validated by the github action with verify=true (https://github.com/sapcc/go-makefile-maker/actions/runs/13635720861/job/38113890543?pr=248) but the command itself ran fine locally.

After this change everything works fine when running the following locally:

```
./golangci-lint config verify -c ~/go-makefile-maker/.golangci.yaml --schema jsonschema/golangci.jsonschema.json
```

### DIFF
--- a/jsonschema/golangci.v1.57.jsonschema.json
+++ b/jsonschema/golangci.v1.57.jsonschema.json
@@ -351,7 +351,7 @@
         "timeout": {
           "description": "Timeout for the analysis.",
           "type": "string",
-          "pattern": "^(\\d+m\\d+s|\\d+m|\\d+s|\\d)$",
+          "pattern": "^((\\d+h)?(\\d+m)?(\\d+(?:\\.\\d)?s)?|0)$",
           "default": "1m",
           "examples": ["30s", "5m", "5m30s"]
         },
--- a/jsonschema/golangci.v1.57.jsonschema.json
+++ b/jsonschema/golangci.v1.57.jsonschema.json
@@ -351,7 +351,7 @@
         "timeout": {
           "description": "Timeout for the analysis.",
           "type": "string",
-          "pattern": "^(\\d*m)?\\d*[sm]$",
+          "pattern": "^(\\d+m\\d+s|\\d+m|\\d+s|\\d)$",
           "default": "1m",
           "examples": ["30s", "5m", "5m30s"]
         },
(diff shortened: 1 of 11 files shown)


jsonschema/golangci.v1.57.jsonschema.json
```diff
@@ -351,9 +351,9 @@
         "timeout": {
           "description": "Timeout for the analysis.",
           "type": "string",
-          "pattern": "^\\d*[sm]$",
+          "pattern": "^(\\d*m)?\\d*[sm]$",
```
felix-kaestner, line 354:
```suggestion
          "pattern": "^(\\d*m)?\\d+[sm]$",
```
Shouldn't this at least require 1 digit? 🤔

ldez (changes requested):
I think this regexp works: `^(\d+[sm]?)+$`

ldez:
The suggested regexp doesn't work because it matches `3m0m` and doesn't match `0`.

My previous suggestion doesn't work to match `3m0m` or `3s0m`.

I think the right expression is `^(\d+m\d+s|\d+m|\d+s|\d)$`

ldez:
Inside `time.ParseDuration` we can found this regexp `[-+]?([0-9]*(\.[0-9]*)?[a-z]+)+`.
https://github.com/golang/go/blob/0312e31ed197b3bf1434e8dbb283f0d2374d7457/src/time/format.go#L1622

But this regexp is not a validation regexp because it accepts everything.

In the context of `run.timeout`:
- The values that use "ns", "us" (or "µs"), "ms" are off-topic.
- The value "0" is accepted.
- The negative values are off-topic.

So the best compromise is `^((\d+h)?(\d+m)?(\d+(?:.\d)?s)?|0)$`.

ldez:
The duration parsing is a bit surprising: https://go.dev/play/p/-NSON-ylfvm

So, at the end the regexp can be `^((\d+[hms])+|0)$`.

I don't know if I want to follow the "official" duration format (without negative numbers and too small units) or keep my [latest suggestion](https://github.com/golangci/golangci-lint/pull/5501#issuecomment-2695167707) with something more "logical". :thinking:

ldez:
:thinking: I think it's better to have something more strict and less human error prone.
Using `3m3m3m` as value to have `9m` for the timeout is "unexpected", I cannot see a use-case for that.

ldez (approved):
LGTM

SuperSandro2000:
Thanks! that was quick

PR #5506, opened 2025-03-05
feat: migration command

ldez:
I want to explain my journey on this topic because it was not an easy one.

My first thought was to try to keep comments by using several YAML parsing libs, but it was very complex, only for YAML (we should handle YAML, TOML, JSON), and at the end it was not working mainly due to the migration path of some elements.

So I gave up on migrating the comments.

I tried to use the v2 configuration structures directly, but it's not possible to serialize the configuration as expected (even with additional struct tags), mainly because of default non-zero values (e.g. boolean `true` by default, explicit `0` or `false`, ...). So it didn't work.

My next try was the working solution: using a clone of the v2 configuration that uses pointer for builtin types and struct tags.

I created a "cloner" which clones the v2 configuration structures (only) and modifies them by changing types and adds struct tags.
So the "cloned" configuration structures can be recreated if needed (`make clone_config`).

The solution is stable, maintainable, and tested, but I'm sad that I couldn't find a solution to keep the comments.

The migration file format is based on the extension of the configuration file.
The format can be overridden by using `--format` flag. (ex: `golangci-lint migrate --format yml`) 

Before the migration, the previous configuration file is copied and saved to a file named `<config_file_name>.bck.<config_file_extention>`. 

By default, before the migration process the configuration file is validated on the JSONSchema of the configuration v1.
If you want to skip this validation, you can use `--skip-validation`.  (ex: `golangci-lint migrate --skip-validation`)

<details><summary>Some notable elements</summary>

The migration command also enforced some new default values:
- `run.timeout`: the existing value is ignored, because, in v2, by default, there is no timeout.
- `issues.show-stats`: the existing value is ignored, in v2, by default, the stats are enabled.
- `run.concurrency`: if the existing value was `0`, as it's the new default, the key is removed.
- `run.relative-path-mode`: if the existing value was `cfg`, as it's the new default, the key is removed.

`issues.exclude-generated` has a new default value (v1 `lax`, v2 `strict`), then this field will be added during the migration to keep the previous behavior.

`issues.exclude-dirs-use-default` has been removed, so this is converted to `linters.exclusions.paths` and, if needed, `formatters.exclusions.paths`.

**Deprecated options from the v1 or unknown fields are not migrated.**

Other fields, which are explicitly defined in the configuration file, are migrated even if the value is the same as the default value.

A migration guide will be added to the documentation https://github.com/golangci/golangci-lint/pull/5439

</details>

I created more than 150 test configuration files, and added dedicated tests on complex migration paths.

The command documentation will be added to the migration guide.

```
Migrate configuration file from v1 to v2

Usage:
  golangci-lint migrate [flags]

Flags:
  -c, --config PATH       Read config from file path PATH
      --no-config         Don't read config file
      --format string     By default, the file format is based on the configuration file extension.
                          Overrides file format detection.
                          It can be 'yml', 'yaml', 'toml', 'json'.
      --skip-validation   Skip validating the configuration file against the JSON schema of the v1.

Global Flags:
```


pkg/commands/migrate.go
```diff
@@ -0,0 +1,239 @@
...
+	fs.SortFlags = false // sort them as they are defined here
+
+	setupConfigFileFlagSet(fs, &c.opts.LoaderOptions)
+
+	fs.StringVar(&c.opts.format, "format", "",
+		color.GreenString("By default, the file format is based on the configuration file extension.\n"+
+			"Overrides file format detection.\nIt can be 'yml', 'yaml', 'toml', 'json'."))
+
+	fs.BoolVar(&c.opts.skipValidation, "skip-validation", false,
+		color.GreenString("Skip validating the configuration file on the JSONSchema of the v1."))
```
bombsimon, line 78:
```suggestion
		color.GreenString("Skip validation of the configuration file against the JSONSchema for v1."))
```

bombsimon (commented):
Nice!! I've yet to review the individual migrations but everything else looks great! I don't think there's much to add for the rest of the code given all the tests but I saw you did a great job on documenting all permutations of enabled and disabled linters that I plan to give an extra read.

🔥

pkg/commands/internal/migrate/migrate.go
```diff
@@ -0,0 +1,19 @@
+package migrate
+
+import (
+	"github.com/golangci/golangci-lint/pkg/commands/internal/migrate/one"
+	"github.com/golangci/golangci-lint/pkg/commands/internal/migrate/ptr"
+	"github.com/golangci/golangci-lint/pkg/commands/internal/migrate/two"
```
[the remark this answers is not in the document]
ldez: I was traumatized by the use of k8s libs, with their package versioning.
Also, `v1`, `v2`, `vX` have a meaning for Go module, it can create confusion.
I would prefer to keep `one` and `two`, if you are not against.

pkg/commands/internal/migrate/one/base_rule.go
```diff
@@ -0,0 +1,9 @@
+package one
```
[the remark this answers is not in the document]
ldez: Thoses files are modified copy of v1.
I will add a comment.

pkg/commands/migrate.go
```diff
@@ -0,0 +1,239 @@
...
+	c := &migrateCommand{
+		viper:     viper.New(),
+		cfg:       one.NewConfig(),
+		buildInfo: info,
+		log:       log,
+	}
+
+	migrateCmd := &cobra.Command{
+		Use:               "migrate",
+		Short:             "Migrate configuration file from v1 to v2",
```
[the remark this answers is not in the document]
ldez: Pre-mature optimization, cleary.

pkg/commands/internal/migrate/one/config.go
```diff
@@ -0,0 +1,29 @@
+package one
+
+type Config struct {
+	cfgDir string // Path to the directory containing golangci-lint config file.
+
+	Version string `mapstructure:"version"` // From v2, to be able to detect already migrated config file.
```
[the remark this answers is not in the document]
ldez: It's impossible to be perfect here, but remember that the file is checked against the v1 JSONSchema.

This is just a shortcut.

pkg/commands/internal/migrate/cloner/cloner.go
```diff
@@ -0,0 +1,197 @@
...
+
+	return false
+}
+
+func processFile(file *ast.File) {
+	file.Name.Name = newPkgName
+
+	var newDecls []ast.Decl
+	for _, decl := range file.Decls {
+		switch d := decl.(type) {
```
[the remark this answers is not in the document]
ldez: The switch is used: `*ast.FuncDecl`, `*ast.GenDecl`, and default.

The 3 cases are different and important.

pkg/commands/migrate.go
```diff
@@ -0,0 +1,239 @@
...
+		c.log.Warnf("No config file detected")
+		os.Exit(exitcodes.NoConfigFileDetected)
+	}
+
+	err := c.backupConfigurationFile(srcPath)
+	if err != nil {
+		return err
+	}
+
+	c.cmd.Println("Migrating v1 configuration file:", srcPath)
```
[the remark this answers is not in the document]
ldez: It was intentional because I wanted the command to be "verbose" by default.
I will change the default log level (it will be more verbose than my expectation but it's not a problem).

pkg/commands/internal/migrate/cloner/cloner.go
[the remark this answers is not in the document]
ldez: I will remove the first switch: if we need it in the future, we can restore it.

But I will keep the 2nd because the 2nd is required (`continue` != nothing).

pkg/commands/migrate.go
```diff
@@ -0,0 +1,241 @@
...
+	err := c.backupConfigurationFile(srcPath)
+	if err != nil {
+		return err
+	}
+
+	c.log.Infof("Migrating v1 configuration file: %s", srcPath)
+
+	ext := filepath.Ext(srcPath)
+	if strings.TrimSpace(c.opts.format) != "" {
+		ext = "." + strings.TrimPrefix(c.opts.format, ".")
```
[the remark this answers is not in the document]
alexandear: Also, why do we need to trim prefix `.`? The format can be only: `yml`, `yaml`, `toml`, `json`. Let's add a validation and fail earlier before starting migration.

Because now, `migrate` changes files and then fails:

```sh
$ go run cmd/golangci-lint/main.go migrate --format="txt" --config pkg/commands/internal/migrate/testdata/issues_01_a.yml
INFO golangci-lint has version  built with go1.24.1 from (unknown, modified: ?, mod sum: "") on (unknown) 
INFO [config_reader] Used config file pkg/commands/internal/migrate/testdata/issues_01_a.yml 
INFO Validating v1 configuration file: pkg/commands/internal/migrate/testdata/issues_01_a.yml 
INFO Saving the v1 configuration to: pkg/commands/internal/migrate/testdata/issues_01_a.bck.yml 
INFO Migrating v1 configuration file: pkg/commands/internal/migrate/testdata/issues_01_a.yml 
Failed executing command with error: saving configuration file: unsupported file type: .txt
exit status 3

$ git status
On branch feat/migration-cmd
Your branch is up to date with 'ldez/feat/migration-cmd'.

  (use "git add/rm <file>..." to update what will be committed)
        deleted:    pkg/commands/internal/migrate/testdata/issues_01_a.yml

Untracked files:
  (use "git add <file>..." to include in what will be committed)
        pkg/commands/internal/migrate/testdata/issues_01_a.bck.yml pkg/commands/internal/migrate/testdata/issues_01_a.txt

no changes added to commit (use "git add" and/or "git commit -a")
```

pkg/commands/internal/migrate/migrate_formatters.go
```diff
@@ -0,0 +1,79 @@
...
+		ModulePath: old.ModulePath,
+		ExtraRules: old.ExtraRules,
+	}
+}
+
+func toGoImportsSettings(old versionone.GoImportsSettings) versiontwo.GoImportsSettings {
+	var localPrefixes []string
+
+	if ptr.Deref(old.LocalPrefixes) != "" {
+		localPrefixes = strings.Split(ptr.Deref(old.LocalPrefixes), ",")
```
alexandear, line 73:
```suggestion
	if prefixes := ptr.Deref(old.LocalPrefixes); prefixes != "" {
		localPrefixes = strings.Split(prefixes, ",")
```

pkg/commands/internal/migrate/migrate_linter_names.go
```diff
@@ -0,0 +1,930 @@
...
+			results = append(results, n)
+			continue
+		}
+
+		results = append(results, name)
+	}
+
+	slices.Sort(results)
+
+	return slices.Compact(results)
```
alexandear, line 929: `slices.Sort` + `slices.Compact` duplicates 4 times in the file. I propose creating a helper function and use it:

```go
func unique[S ~[]E, E cmp.Ordered](s S) S {
	return slices.Compact(slices.Sorted(slices.Values(s)))
}
```

pkg/commands/internal/migrate/ptr/ptr.go
```diff
@@ -0,0 +1,12 @@
...
+func Deref[T any](v *T) T {
+	if v == nil {
+		var zero T
+		return zero
+	}
+
+	return *v
+}
+
+func Pointer[T any](v T) *T { return &v }
```
alexandear, line 12: It's debatable, but do you consider `Of` as an alternative name? Like https://pkg.go.dev/k8s.io/utils/ptr#To

It's also commonly used: https://sourcegraph.com/search?q=context:global+ptr.Of+lang:go+&patternType=standard&sm=1

ldez: I prefer an explicit name.

bombsimon (approved):

bombsimon (approved):
Awesome, LGTM

pkg/commands/internal/migrate/migrate_run.go
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	var concurrency *int
+	if ptr.Deref(old.Run.Concurrency) != 0 {
+		// 0 is the new default
+		concurrency = old.Run.Concurrency
+	}
+
+	return versiontwo.Run{
+		Timeout:               0, // Enforce new default.
```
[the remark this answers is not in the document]
ldez: [As we discussed together](https://github.com/golangci/golangci-lint/pull/5470#discussion_r1968522617), most projects don't care about the timeout.

This option comes from a time when the `golang.org/x/tools/go/analysis` was not exiting.
It lost a lot of its meaning during the time.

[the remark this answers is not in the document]
ldez: I think that users are more limited by CPU or memory than time.

I'm not aware of real usages of timeout as an effective limitation for a project.
But I don't know all the use-cases, the option is still here, if they really want it, they can still set it.
I think it's better to remove it, from the migrated configuration, because this option was more pain than a positive thing by default.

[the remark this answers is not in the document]
bombsimon: To me the discussion was about sensible defaults and how reasonable different values were depending on common use cases. Since a lot of users uses a high timeout the 1m is not a sane default and we landed on not setting a default.

This to me doesn't mean that no one cares or that every value set is arbitrary and can be wiped. Not saying that's not the case, just that it's not how I interpreted the discussion.

If we plan to wipe the timeout I would consider adding a log row saying that the new config doesn't have any timeout set and that the user must set one if they want or need it.

And speaking of this and removing config, I think it's worth logging/printing in general that any not known value will be removed, such values might exist due to legacy config or for abuse such as [gclv](https://github.com/bombsimon/gclv) that I use.

pkg/commands/internal/migrate/migrate_linter_names_test.go
```diff
@@ -0,0 +1,473 @@
+package migrate
```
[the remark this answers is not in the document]
ldez: IMO, the implementation of v1 was overcomplicated despite its apparent "simplicity".
Even I, I learned some cases during the writing of the migration command.
I understand how it became like that, but recreating the system to obtain effective configuration was not easy.

pkg/commands/internal/migrate/migrate_run.go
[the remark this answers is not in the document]
ldez: > And speaking of this and removing config, I think it's worth logging/printing in general that any not known value will be removed, such values might exist due to legacy config or for abuse such as [gclv](https://github.com/bombsimon/gclv) that I use.

Technically, the custom field you are using for gclv or unknown values will not pass the JSON Schema validation.
So it's not possible to add a log.

And even without the JSON Schema validation, the configuration is loaded into a structure, so it's impossible to access unknown fields.


> This to me doesn't mean that no one cares or that every value set is arbitrary and can be wiped. Not saying that's not the case, just that it's not how I interpreted the discussion.

If we don't remove it, the users will continue to increase the value instead of removing it, it will just reproduce the default timeout problem, but the problem will be created by the users themself.

But if you think this is a critical point, I keep the value.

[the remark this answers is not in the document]
ldez: I added a log about `run.timeout`.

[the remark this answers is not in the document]
bombsimon: I think we can remove it but I also think we should notify the user about it.

> And even without the JSON Schema validation, the configuration is loaded into a structure, so it's impossible to access unknown fields.

Yep that I'm aware of, I'm mostly thinking it's enough to add a single print statement with this information and combine it with other changes we do. 

> Unknown settings removed. 'timeout' has been removed; default is now no timeout.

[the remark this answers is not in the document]
ldez: I think we can add a log with the link to the section about the command inside the migration guide.
Inside the section, we can add the same details I explained in the issue description.

But this could not be done inside this PR, because the migration guide is not ready for now.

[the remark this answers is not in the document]
alexandear: > I think we can add a log with the link to the section about the command inside the migration guide.

I will add.

pkg/commands/migrate.go
alexandear, line 1: I propose supporting the case when a user wants to migrate a configuration that has already been migrated.

Actual:
```
$ go run cmd/golangci-lint/main.go migrate --config pkg/commands/internal/migrate/testdata/issues_01_a.golden.yml
INFO golangci-lint has version  built with go1.24.1 from (unknown, modified: ?, mod sum: "") on (unknown) 
INFO [config_reader] Used config file pkg/commands/internal/migrate/testdata/issues_01_a.golden.yml 
INFO Validating v1 configuration file: pkg/commands/internal/migrate/testdata/issues_01_a.golden.yml 
jsonschema: "" does not validate with "/additionalProperties": additional properties 'version' not allowed
Failed executing command with error: the configuration contains invalid elements
exit status 3
```

Expected:
```
$ go run cmd/golangci-lint/main.go migrate --config pkg/commands/internal/migrate/testdata/issues_01_a.golden.yml
INFO golangci-lint has version  built with go1.24.1 from (unknown, modified: ?, mod sum: "") on (unknown) 
INFO [config_reader] Used config file pkg/commands/internal/migrate/testdata/issues_01_a.golden.yml 
INFO Configuration already migrated to v2. Nothing do.
```

What do you think?

pkg/commands/internal/migrate/testdata/linters_02.golden.yml
```diff
@@ -0,0 +1,110 @@
...
+    - recvcheck
+    - revive
+    - rowserrcheck
+    - sloglint
+    - spancheck
+    - sqlclosecheck
+    - staticcheck
+    - tagalign
+    - tagliatelle
+    - tenv
```
alexandear, line 94: Do we need to keep deprecated in v1 linters in the `linters.disable` section?

I am expecting that the `migrate` command produces a valid config, and we do not need to do a manual update afterward.

E.g., if I remove `mycustomlinter1, mycustomlinter2` from `linters_02.golden.yml`, run golangci-lint, and get an error:
```
$ go run cmd/golangci-lint/main.go run --config=pkg/commands/internal/migrate/testdata/linters_02.golden.yml
Error: unknown linters: 'tenv', run 'golangci-lint help linters' to see the list of supported linters
Failed executing command with error: unknown linters: 'tenv', run 'golangci-lint help linters' to see the list of supported linters
exit status 3
```

`tenv` is a valid linter for v1 (but deprecated), but is invalid for v2. I think we can add a warning to the user that `tenv` is removed or migrate `tenv` to `usetesting` with some settings.

pkg/commands/migrate.go
ldez: I prefer not to modify the v1 JSON Schema.

pkg/commands/internal/migrate/testdata/linters-settings_gocritic.yml
```diff
@@ -0,0 +1,269 @@
...
+      rangeExprCopy:
+        sizeThreshold: 516
+        skipTestFuncs: false
+      rangeValCopy:
+        sizeThreshold: 32
+        skipTestFuncs: false
+      ruleguard:
+        debug: 'emptyDecl'
+        failOn: dsl,import
+        rules: '${configDir}/ruleguard/rules-*.go,${configDir}/myrule1.go'
```
alexandear, line 259: Should `${configDir}` be replaced with `${base-path}`?

pkg/commands/internal/migrate/migrate_linters_exclusions.go
alexandear, line 1: This part of the v1 configuration:
```yaml
issues:
  exclude-rules:
    # Skip goimports check on generated files
    - path: \\.(generated\\.deepcopy|pb)\\.go$
      linters:
        - goimports
```

migrates to an invalid v2 configuration:
```yaml
linters:
  exclusions:
    rules:
      - path: \\.(generated\\.deepcopy|pb)\\.go$
```

I found this case here https://github.com/cilium/cilium/blob/dd2295f795974ee05dbb84b451a9a7635c5bc990/.golangci.yaml#L174-L175

```
$ golangci-lint-v2 migrate
INFO golangci-lint has version v1.64.6-0.20250307140015-345e67421361 built with go1.24.1 from (345e674213610af908bf0753a1a5b28a058c6769, modified: false, mod sum: "") on 2025-03-07T14:00:15Z 
INFO [config_reader] Config search paths: [./ /Users/alexandear/src/github.com/cilium/cilium /Users/alexandear/src/github.com/cilium /Users/alexandear/src/github.com /Users/alexandear/src /Users/alexandear /Users /] 
INFO [config_reader] Used config file .golangci.yaml 
INFO Validating v1 configuration file: /Users/alexandear/src/github.com/cilium/cilium/.golangci.yaml 
INFO Saving the v1 configuration to: /Users/alexandear/src/github.com/cilium/cilium/.golangci.bck.yaml 
WARN The configuration comments are not migrated. 
INFO Migrating v1 configuration file: /Users/alexandear/src/github.com/cilium/cilium/.golangci.yaml 
WARN The configuration `run.timeout` is ignored. By default, in v2, the timeout is disabled. 
INFO Migration done: /Users/alexandear/src/github.com/cilium/cilium/.golangci.yaml

$ golangci-lint-v2 run
Error: can't load config: error in exclude rule #4: at least 2 of (text, source, path[-except], linters) should be set
Failed executing command with error: can't load config: error in exclude rule #4: at least 2 of (text, source, path[-except], linters) should be set
...

Comment (ldez) [reply pkg/commands/internal/migrate/migrate_linters_exclusions.go:1]: I handled the case, but these kinds of exclusion rules don't have a lot of meaning: excludes only one linter on generated files, but lets the other linters be, which is not logical, especially excluding a formatter.

Either files are excluded for all linters (and formatters) or those files should be linted.

So this configuration is wrong by itself.

**Comment by SuperSandro2000** [comment]
Did you think about using the new omitzero tag from go 1.24?

We used that in https://github.com/majewsky/gg . See https://github.com/sapcc/keppel/pull/489 as an example where we used it.

**Comment by ldez** [comment]
> Did you think about using the new omitzero tag from go 1.24?

Yes, but:
1. We need to compile with go1.23
2. it doesn't solve default non-zero value (`true`, `8`, etc.)

File: pkg/commands/internal/migrate/cloner/cloner.go [hunk 16]
```diff
@@ -0,0 +1,198 @@
...
+		return true
+	}
+
+	if filepath.Base(path) == "base_loader.go" {
+		return true
+	}
+
+	if filepath.Base(path) == "loader.go" {
+		return true
+	}
```
alexandear, line 78: nit:
```suggestion
	switch filepath.Base(path) {
	case "base_loader.go", "loader.go":
		return true
	defautl:
		return true
	}
```

pkg/commands/internal/migrate/migrate_formatters.go
```diff
@@ -0,0 +1,107 @@
...
+	return results
+}
+
+func toGciSettings(old versionone.GciSettings) versiontwo.GciSettings {
+	return versiontwo.GciSettings{
+		Sections:         old.Sections,
+		NoInlineComments: old.NoInlineComments,
+		NoPrefixComments: old.NoPrefixComments,
+		CustomOrder:      old.CustomOrder,
+		NoLexOrder:       old.CustomOrder,
```
alexandear, line 70: Why not
```suggestion
		NoLexOrder:       old.NoLexOrder,
```
?

pkg/commands/internal/migrate/migrate_formatters.go
```diff
@@ -0,0 +1,107 @@
...
+		names := onlyFormatterNames(allNames)
+		if len(names) == 0 {
+			continue
+		}
+
+		if ptr.Deref(rule.Path) == "" {
+			continue
+		}
+
+		results = append(results, ptr.Deref(rule.Path))
```
alexandear, line 58: nit:
```suggestion
		rulePath := ptr.Deref(rule.Path)
		if rulePath == "" {
			continue
		}

		results = append(results, rulePath)
```

pkg/commands/internal/migrate/migrate_linter_names.go
```diff
@@ -0,0 +1,922 @@
...
+
+	return false
+}
+
+func ProcessEffectiveLinters(old versionone.Linters) (enable, disable []string) {
+	switch {
+	case ptr.Deref(old.DisableAll):
+		return disableAllFilter(old), nil
+	case ptr.Deref(old.EnableAll):
+		return nil, enableAllFilter(old)
```
alexandear, line 42: These lines confuse me and can confuse future readers. I propose to extend function names with `enable` and `disable` words for more clarity.

E.g.,
```suggestion
	case ptr.Deref(old.DisableAll):
		return enableWhenDisableAllFilter(old), nil
	case ptr.Deref(old.EnableAll):
		return nil, disableLintersWhenEnableAllFilter(old)
```

pkg/commands/internal/migrate/migrate_linter_names.go
```diff
@@ -0,0 +1,922 @@
...
+				allLinters(), onlyPresets(old), keepFast(old), // presets - [slow]
+			),
+			allEnabled(old, allLinters()), // + enable
+		),
+	)
+
+	return slices.Concat(names, unknownLinterNames(old.Enable, allLinters()))
+}
+
+// enableAllFilter generates the value of `disable` when `enable-all` is `true`.
```
alexandear, line 77: Thanks for the comment.

pkg/commands/internal/migrate/migrate_linter_names.go
```diff
@@ -0,0 +1,922 @@
...
+func convertAlternativeNames(names []string) []string {
+	altNames := map[string]string{
+		"gas":       "gosec",
+		"goerr113":  "err113",
+		"gomnd":     "mnd",
+		"logrlint":  "loggercheck",
+		"megacheck": "staticcheck",
+		"vet":       "govet",
+		"vetshadow": "govet",
+	}
```
alexandear, line 904: This map duplicates info from the `allLinters` function. Maybe better is to use the function `allLinters` instead of map?

pkg/commands/internal/migrate/migrate_linters_exclusions.go
```diff
@@ -0,0 +1,146 @@
...
+	if len(excludedTestLinters) > 0 {
+		results = append(results, versiontwo.ExcludeRule{
+			BaseRule: versiontwo.BaseRule{
+				Linters: excludedTestLinters,
+				Path:    ptr.Pointer(`(.+)_test\.go`),
+			},
+		})
+	}
+
+	return results
```
alexandear, line 135: We can remove `results` var:

```suggestion
	if len(excludedTestLinters) == 0 {
		return []versiontwo.ExcludeRule{}
	}

	return []versiontwo.ExcludeRule{
		{
			BaseRule: versiontwo.BaseRule{
				Linters: excludedTestLinters,
				Path:    ptr.Pointer(`(.+)_test\.go`),
			},
		},
	}
```

pkg/commands/internal/migrate/migrate_linters_settings.go
```diff
@@ -0,0 +1,1017 @@
...
+		AnalyzeTypes:         old.AnalyzeTypes,
+	}
+
+	for _, pattern := range old.Forbid {
+		if pattern.Pattern == nil {
+			buffer, err := pattern.MarshalString()
+			if err != nil {
+				// impossible case
+				panic(err)
+			}
```
alexandear, line 268: If this is impossible case, can we simply ignore an error?
```suggestion
			buffer, _ := pattern.MarshalString() // an error is impossible
```

pkg/commands/internal/migrate/migrate_linters_settings.go
```diff
@@ -0,0 +1,1017 @@
...
+		Simple:     old.Simple,
+		RangeLoops: old.RangeLoops,
+		ForLoops:   old.ForLoops,
+	}
+}
+
+func toPredeclaredSettings(old versionone.PredeclaredSettings) versiontwo.PredeclaredSettings {
+	var ignore []string
+	if ptr.Deref(old.Ignore) != "" {
+		ignore = strings.Split(ptr.Deref(old.Ignore), ",")
```
alexandear, line 694: nit:
```suggestion
	if oldIgnore := ptr.Deref(old.Ignore); oldIgnore != "" {
		ignore = strings.Split(oldIgnore, ",")
```

pkg/commands/internal/migrate/migrate_linters_settings.go
```diff
@@ -0,0 +1,1017 @@
...
+	}
+}
+
+func toStaticCheckSettings(old versionone.LintersSettings) versiontwo.StaticCheckSettings {
+	checks := slices.Concat(old.Staticcheck.Checks, old.Stylecheck.Checks, old.Gosimple.Checks)
+
+	slices.Sort(checks)
+
+	return versiontwo.StaticCheckSettings{
+		Checks:                  slices.Compact(checks),
```
alexandear, line 797: We can change:

```suggestion
		Checks:                  Unique(checks),
```

And remove the `slices.Sort(checks)` above.

pkg/commands/internal/migrate/migrate_linter_names.go
ldez: I don't agree

pkg/commands/internal/migrate/migrate_formatters.go
ldez: It's useless

pkg/commands/internal/migrate/migrate_linter_names.go
ldez: this is not a duplicate.

pkg/commands/internal/migrate/migrate_linters_settings.go
ldez: because ignoring an error is not a good idea.

pkg/commands/internal/migrate/migrate_linters_settings.go
ldez: ifshort is not my style, and in this context it's useless.

pkg/commands/internal/migrate/testdata/linters-settings_staticcheck.golden.yml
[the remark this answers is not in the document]
ldez: Consider this as a linter update.

pkg/commands/internal/migrate/migrate_formatters.go
alexandear, line 1: Found another case related to the migration of formatters in the project: https://github.com/mymmrac/telego

```sh
$ golangci-lint-v2 migrate
// success

$ golangci-lint-v2 run
// A lot of `gci` issues which are not expected
common_test.go:6:1: File is not properly formatted (gci)
        "github.com/stretchr/testify/require"
^
```

Migrated `formatters` settings:
```yaml
linters:
  default: all
  disable:
    # - ...
formatters:
  settings:
    gci:
      sections:
        - standard
        - default
        - prefix(github.com/mymmrac/telego)
  exclusions:
    generated: lax
```

Imports are correctly sorted:
```go
import (
	"testing"

	"github.com/stretchr/testify/require"
	"go.uber.org/mock/gomock"

	"github.com/mymmrac/telego/internal/json"
	ta "github.com/mymmrac/telego/telegoapi"
	mockapi "github.com/mymmrac/telego/telegoapi/mock"
)
```

So I'm expecting no `gci` issues.

Note, when I explicitly enable `gci` issues disappear:
```yaml
formatters:
  enable:
     - gci
  settings:
    gci:
      sections:
        - standard
        - default
        - prefix(github.com/mymmrac/telego)
  exclusions:
    generated: lax
```

pkg/commands/internal/migrate/migrate_linters.go
alexandear, line 1: I propose to remove the `typecheck` internal linter from the `linters.enable` list in v2.

Found this case in https://github.com/vektra/mockery/blob/325bff2a1e4f63d3d1ef37620b4127f5d462c03c/.golangci.yml#L14

v1:
```yaml
linters:
  disable-all: true
  enable:
    - typecheck
```

v2:
```yaml
version: "2"
linters:
  default: none
```

It's a rare case.

pkg/commands/internal/migrate/testdata/linters-settings_depguard.yml
alexandear, line 1: Is it possible to preserve case for a depguard rule name?

Currently, the following depguard rule `DenyGogoProtobuf` from https://github.com/istio/istio/blob/be46bb10b1750938b41cadd172abb2f187f9731e/common/config/.golangci.yml#L192

```yaml
linters-settings:
  depguard:
    rules:
      DenyGogoProtobuf:
        files:
          - $all
        deny:
          - pkg: github.com/gogo/protobuf
            desc: "gogo/protobuf is deprecated, use golang/protobuf"
```

migrates to `denygogoprotobuf`:

```yaml
linters:
  settings:
    depguard:
        denygogoprotobuf:
          files:
            - $all
          deny:
            - pkg: github.com/gogo/protobuf
              desc: gogo/protobuf is deprecated, use golang/protobuf
```

ldez: no

ldez: I will rewrite the configuration loading because, after some tests, there's a problem with some linter configuration (goheader, tagliatelle, etc.).

PR #5524, opened 2025-03-11
feat: check version of the configuration

ldez:
Checks the version of the configuration.


bombsimon (commented):
Why is this important? Do we plan to rely on the version somewhere?

ldez:
This is important because I'm 100% sure that people, that use `latest`, will try to run the v2 with v1 configuration.

bombsimon:
> This is important because I'm 100% sure that people, that use `latest`, will try to run the v2 with v1 configuration.

I think you're 100% right 😅

bombsimon (approved):

PR #5543, opened 2025-03-13
feat: error if typecheck is enabled or disabled

ldez:
Detects if a user tries to enable or disable `typecheck`.

`typecheck` is not a linter.

- https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors
- https://golangci-lint.run/welcome/faq/#why-is-it-not-possible-to-skipignore-typecheck-errors

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -60,6 +60,14 @@ func (l *Loader) Load(opts LoadOptions) error {
 		l.cfg.Linters.Exclusions.Generated = GeneratedModeStrict
 	}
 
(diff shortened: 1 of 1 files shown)


bombsimon (approved):
Now let's spin the lottery wheel and see if this result in more or less issues 😅

alexandear (approved):

ldez:
We will create a dedicated donation "box" like the box used for children when they say bad words.

Each user that opens an issue about typecheck will have to put a coin in it :smile_cat:

Issue #3280, opened 2022-10-06
revive: Invalid argument to the context-as-argument rule. Unrecognized key allowtypesbefore


pmalek:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).
- [X] Yes, I've tried with the standalone linter if available. (https://golangci-lint.run/usage/linters/)

### Description of the problem

It seems that golangci-lint (starting with v1.50?) passes the arguments to linters in lower case and not "as is".

This causes in our cases `panic` in `revive`

```
panic: Invalid argument to the context-as-argument rule. Unrecognized key allowtypesbefore

goroutine 3957 [running]:
github.com/mgechev/revive/rule.getAllowTypesFromArguments({0xc002f5eb30?, 0x0?, 0x0?})
	/home/runner/go/pkg/mod/github.com/mgechev/revive@v1.2.4/rule/context-as-argument.go:98 +0x379
github.com/mgechev/revive/rule.(*ContextAsArgumentRule).Apply(0x22a5ac0, 0xc00146e300, {0xc002f5eb30?, 0x13?, 0x90a6c696e203d00?})
	/home/runner/go/pkg/mod/github.com/mgechev/revive@v1.2.4/rule/context-as-argument.go:22 +0xa5
github.com/mgechev/revive/lint.(*File).lint(0xc00146e300, {0xc002f5ec50, 0x1, 0x1}, {0x0, 0x3fe999999999999a, {0x16cc8c0, 0x7}, 0x0, 0xc002f5ab70, ...}, ...)
	/home/runner/go/pkg/mod/github.com/mgechev/revive@v1.2.4/lint/file.go:105 +0x19b
github.com/mgechev/revive/lint.(*Package).lint.func1(0xc002b82480?)
	/home/runner/go/pkg/mod/github.com/mgechev/revive@v1.2.4/lint/package.go:185 +0x85
created by github.com/mgechev/revive/lint.(*Package).lint
	/home/runner/go/pkg/mod/github.com/mgechev/revive@v1.2.4/lint/package.go:184 +0xac
```

which can be tracked down to: https://github.com/mgechev/revive/blob/v1.2.4/rule/context-as-argument.go#L82-L110.

The same config with the same code base works fine with v1.49

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
[...]

PR #5548, opened 2025-03-14
📌 Pin `viper` module to `v1.12.0` in `go.mod`

busser:
I've been trying to get `golangci-lint` to work with `go tool`. The only
issue I've run into so far is related to dependency versions.

I know that `golangci-lint`'s documentation warns against compiling
`golangci-lint` ourselves, since the resulting binary is untested.
However, in practice it's very close to working.

I'm importing the `golangci-lint` module into a module that also imports
`viper`. However, while `golangci-lint` imports `v1.12.0`, my module
imports `v1.19.0` (the latest version). When I run `go tool golangci-lint`
the Go compiler compiles `golangci-lint` with `viper@v1.19.0`, which
breaks part of the `revive` linter (see issue #3280). It possibly breaks
parts of other linters, but my specific configuration did not reveal it.

I'm currently working around this by removing the configuration that
`golangci-lint` can't handle when using `viper@v1.19.0`. However I
believe we can get the Go compiler to use `viper@v1.12.0` for
`golangci-lint` and whatever other version for any module that would
import `golangci-lint`. The trick is to add a replacement to
`golangci-lint`'s `go.mod`:

```gomod
replace github.com/spf13/viper => github.com/spf13/viper v1.12.0
```

I've experimented with this fix by using Go workspaces and it seems to
work as desired. The parent module can still use `viper@1.19.0` and
the issue in the `revive` linter does not occur.

This change requires disabling the `gomoddirectives` linter in a couple
of `golangci-lint`'s tests. Those tests are completely unrelated to that
specific linter or to `go.mod`, so I figure it's fine.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ If you value it, consider supporting us, we appreciate it! ❤️
 [![Open Collective backers and sponsors](https://img.shields.io/badge/OpenCollective-Donate-blue?logo=opencollective&style=for-the-badge)](https://opencollective.com/golangci-lint)
 [![GitHub Sponsors](https://img.shields.io/badge/GitHub-Donate-blue?logo=github&style=for-the-badge)](https://github.com/sponsors/golangci)
 
+### v1.64.7
+
+1. Linters bug fixes
+   * `depguard`: from 2.2.0 to 2.2.1
+   * `dupl`: from 3e9179ac440a to f665c8d69b32
+   * `gosec`: from 2.22.1 to 2.22.2
+   * `staticcheck`: from 0.6.0 to 0.6.1
+2. Documentation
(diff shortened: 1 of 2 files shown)


ldez:
sorry I made a mistake, I wanted to clean your previous PR.

But `replace` directives are not transitive, so it doesn't "pin" a dependency.

ccoVeille:
Thanks for sharing this @ldez 

Could you consider to document it in the README and website?

I know you keep warning people about the risks. I saw you posting warnings again and again over the past months.
But now, `go tool` is here, more and more people will try using it anyway. 

What you documented is very interesting, I think it's a valuable solution

ldez:
https://github.com/golangci/golangci-lint/pull/5549

PR #5583, opened 2025-03-19
Update unparam to fix https://github.com/mvdan/unparam/issues/85

FGasper:
Dependabot seems to have missed this one, which was fixed weeks ago.


ldez:
The branch `master` is related to v1 but the v1 is now locked: https://github.com/golangci/golangci-lint/issues/5300#issuecomment-2718881988

FGasper:
@ldez Are you saying this crash can’t be fixed except in v2?

ldez:
Your question feels a bit aggressive.

Do you read the related issue?

FGasper:
@ldez I didn’t mean to be aggressive and am sorry to have come across that way.

I’m just wondering how this issue can be fixed. `main` doesn’t work for us because it currently rejects our v1 config files. The contributor docs still say to open PRs against `master`.

ldez:
@FGasper I think you can understand that I will not create a bug fix for v1 unless special circumstances because the maintenance of the v1 should stop at some point, even if there are bugs.

If I begin to accept v1 modifications, the maintenance will never stop.

So I need to define a "rule", and this "rule" will be valid only before the v2 release: I invite your company (MongoDB Inc.) to sponsor the project.

- https://donate.golangci.org/ 
- https://donate.ldez.dev/

Like that, I limit the possible v1 modifications.

The v2 release should happen very soon.

Issue #5597, opened 2025-03-24
[v2] Failed executing command with error: can't load config: gofmt is a formatter


yurishkuro:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

go install

### Description of the problem

Trying to upgrade to v2 and it fails on the existing config. I have these previously valid linters in the config:
```
    - gofmt
    - gofumpt
    - goimports
```

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version v2.0.0 built with go1.24.1 from (unknown, modified: ?, mod sum: "h1:RQWk8VCuMQv9bBDy3x23yds2yf9aRZ86C9MWGIdNRuU=") on (unknown)
```

</details>

### Configuration

<details>

```console
issues:
  # Excluding configuration per-path, per-linter, per-text and per-source
  exclude-rules:
    # Exclude some linters from running on tests files.
    - path: _test\.go
      linters: [gosec, bodyclose, noctx]
    - linters: [gocritic]
      text: "dupArg"
    - linters: [gocritic]
      text: "exitAfterDefer"
    - linters: [gocritic]
      text: "appendAssign"

  exclude-dirs-use-default: false
  exclude-files:
    - ".*.pb.go$"
    - "proto-gen/.*"
    - "model/v1/prototest/.*"

  max-issues-per-linter: 0
  max-same-issues: 0

linters:
  disable:
    - errcheck
  enable:
    # Plain ASCII identifiers.
    - asciicheck

    # Checks for dangerous unicode character sequences.
    - bidichk

    # Checks whether HTTP response body is closed successfully.
    # TODO enable this but maybe find a way to disable in tests.
    - bodyclose

    # Check whether the function uses a non-inherited context.
    - contextcheck

[...]

PR #5599, opened 2025-03-24
fix: validate version before configuration

ldez:
The version validation was done after the configuration validations.

Related to https://github.com/golangci/golangci-lint/issues/5597#issuecomment-2748533645

### DIFF
--- a/pkg/config/loader.go
+++ b/pkg/config/loader.go
@@ -60,6 +60,10 @@ func (l *Loader) Load(opts LoadOptions) error {
 		l.cfg.Linters.Exclusions.Generated = GeneratedModeStrict
 	}
 
+	if l.cfg.GetConfigDir() != "" && l.cfg.Version != "2" {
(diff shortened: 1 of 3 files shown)


pkg/config/loader.go
```diff
@@ -127,6 +132,14 @@ func (l *Loader) appendStringSlice(name string, current *[]string) {
 	}
 }
 
+func (l *Loader) checkConfigurationVersion() error {
+	if l.cfg.GetConfigDir() != "" && l.cfg.Version != "2" {
+		return fmt.Errorf("unsupported version of the configuration: %q (require configuration v2)", l.cfg.Version)
```
yurishkuro, line 137:
```suggestion
		return fmt.Errorf("unsupported version of the configuration: %q (require configuration v2). See https://golangci-lint.run/product/migration-guide for migration instructions.", l.cfg.Version)
```

bombsimon (approved):

PR #5626, opened 2025-03-25
docs: clean old configuration examples

ldez:


docs/src/docs/usage/false-positives.mdx
```diff
@@ -33,90 +34,83 @@ Exclude issue by text using command-line option `-e` or config option `issues.ex
...
 ```
 
 ### Exclude Issues by Path
 
-Exclude issues in path by `issues.exclude-dirs`, `issues.exclude-files` or `issues.exclude-rules` config options.
-
-Beware that the paths that get matched here are relative to the current working directory.
-When the configuration contains path patterns that check for specific directories,
-the `--path-prefix` parameter can be used to extend the paths before matching.
+Exclude issues in path by `linters.exclusions.paths`, or `issues.exclude-rules` config options.
```
alexandear, line 71:
```suggestion
Exclude issues in path by `linters.exclusions.paths` or `linters.exclusions.rules` config options.
```

docs/src/docs/usage/false-positives.mdx
alexandear, line 1: Let also remove this sentence at the beginning:
```
For example, we have a default enabled set of [exclude patterns](/usage/configuration#command-line-options).
```

And the "Exclusion Presets" section at the end of this page.

docs/src/docs/usage/false-positives.mdx
```diff
@@ -17,12 +17,13 @@ Otherwise, some linters have dedicated configuration to exclude or disable rules
...
+linters:
+  settings:
+    staticcheck:
+      checks:
+        - all
+        - '-SA1000' # disable the rule SA1000
+        - '-SA1004' # disable the rule SA1004
 ```
 
 ## Exclude or Skip
```
alexandear, line 29: `Skip` is no longer relevant:
```suggestion
## Exclude
```

docs/src/docs/usage/false-positives.mdx
```diff
@@ -33,90 +34,83 @@ Exclude issue by text using command-line option `-e` or config option `issues.ex
 It's helpful when you decided to ignore all issues of this type.
 Also, you can use `issues.exclude-rules` config option for per-path or per-linter configuration.
```
alexandear, line 35: Perhaps `issues.exclude-rules` should be replaced with `linters.exclusions.rules`.

docs/src/docs/usage/false-positives.mdx
```diff
@@ -33,90 +34,83 @@ Exclude issue by text using command-line option `-e` or config option `issues.ex
 It's helpful when you decided to ignore all issues of this type.
```
alexandear, line 34: The sentence above should be updated, since we don't have `-e` and `issues.exclude`:
```
Exclude issue by text using command-line option `-e` or config option `issues.exclude`.
```

docs/src/docs/usage/false-positives.mdx
```diff
@@ -33,90 +34,83 @@ Exclude issue by text using command-line option `-e` or config option `issues.ex
...
+  exclusions:
+    paths:
+      - path/to/a/file.go
 ```
 
-In the following example, all the reports related to the directories (`exclude-dirs`) are excluded:
+In the following example, all the reports related to the directories (`paths`) are excluded:
 
 ```yaml
 issues:
```
alexandear, line 110:
```suggestion
linters:
```

alexandear (changes requested):

docs/src/docs/plugins/go-plugins.mdx
alexandear, line 1: I guess, we should migrate to v2 https://github.com/golangci/example-plugin-module-linter/blob/main/.golangci.example.yml

docs/src/docs/usage/false-positives.mdx
ldez: I think you misplaced your comment, this is not related to a specific line.

docs/src/docs/plugins/module-plugins.mdx
alexandear, line 1: The lines https://github.com/golangci/golangci-lint/blob/487aee999af1d002447d35c643d8ec7266e2ca36/docs/src/docs/plugins/module-plugins.mdx#L11 and https://github.com/golangci/golangci-lint/blob/487aee999af1d002447d35c643d8ec7266e2ca36/docs/src/docs/plugins/module-plugins.mdx#L52

should be changed to
```
- Define the plugin inside the `linters.settings.custom` section with the type `module`.
```

docs/src/docs/usage/false-positives.mdx
ldez: The "Exclusion Presets" section should not be removed.

https://golangci-lint.run/usage/false-positives/#exclusion-presets

docs/src/docs/plugins/go-plugins.mdx
```diff
@@ -50,18 +50,19 @@ If you're looking for instructions on how to configure your own custom linter, t
 1. If the project you want to lint does not have one already, copy the [.golangci.yml](https://github.com/golangci/golangci-lint/blob/HEAD/.golangci.yml) to the root directory.
 2. Adjust the YAML to appropriate `linters-settings.custom` entries as so:
```
alexandear, line 51:
```suggestion
2. Adjust the YAML to appropriate `linters.settings.custom` entries as so:
```

docs/src/docs/plugins/go-plugins.mdx
```diff
@@ -50,18 +50,19 @@ If you're looking for instructions on how to configure your own custom linter, t
...
   ```yaml title=.golangci.yml
```
alexandear, line 52: Do we need to add `version: "2"` here and below?

docs/src/docs/plugins/go-plugins.mdx
ldez: I will do that after this PR.

ldez: Done

alexandear, line 1: One last thing.

The line should be updated
https://github.com/golangci/golangci-lint/blob/80c45e7a901a5387a42cca47afc34641ad031d85/docs/src/docs/plugins/go-plugins.mdx#L74

- `.golang.yml` -> `.golangci.yml`
- `linters.disable-all: true` -> `linters.default: none`

alexandear (approved):

docs/src/docs/usage/false-positives.mdx
alexandear: > I think you misplaced your comment, this is not related to a specific line.

GitHub doesn’t allow me to add a comment to a place not modified by this PR. That’s why I added it to the whole file.

PR #5650, opened 2025-03-28
feat: add config path placeholder

### LINKED ISSUE #5646: Separate relative path mode for config vs output
### Welcome

Official GitHub Action

Specifying a relative path mode of cfg is needed to reliably detect additional config files such as for ruleguard.

Unfortunately this then results in linter outputs using paths relative to the cfg directory too so they start with eg `../../../../../../` breaking CI workflows from referencing the right file

A way to specify ruleguard configuration relative to the config directory but path outputs relative to either the working directory, gitroot, or go mod root would fix the issue.

This seemed to work fine under golangci-lint 1.x. Specifying gitroot as the relative-path-mode breaks loading ruleguard rules but output paths are fine, specifying cfg allows ruleguard rules to load fine but output paths are wrong.

### Additional context

ruleguard config:

```
        ruleguard:
          failOn: all
          rules: ${base-path}/ruleguard/*.go
```

### Supporter

ldez:
In the rare cases, the linter configuration should be relative to the configuration path, even when using other relative path modes other than `cfg`.

Those cases are relative to paths that reference "external files".

Fixes #5646

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -335,6 +335,7 @@ linters:
           # List of file globs that will match this list of settings to compare against.
           # By default, if a path is relative, it is relative to the directory where the golangci-lint command is executed.
           # The placeholder '${base-path}' is substituted with a path relative to the mode defined with `run.relative-path-mode`.
+          # The placeholder '${config-path}' is substituted with a path relative to the configuration file.
           # Default: $all
           files:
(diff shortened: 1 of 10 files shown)


alexandear (commented):
Could you add a few tests?

bombsimon (approved):

PR #5651, opened 2025-03-28
feat: add an option to display absolute paths

### LINKED ISSUE #5649: Add an option to show the full/absolute path of files
### Welcome

Official binary

In my workflow, I run golangci-lint on several projects at the same time (from a script).
When an issue is reported, the file path is relative to the directory of each projets.
It's not very convenient, because I don't know from which project the error is coming from.
My terminal can linkify the files paths, but it doesn't work if the path is relative an match several projects.

It would be nice to have a flag/option similar to the `-fullpath` of the Go command that show the full/absolute path of files.

I could use the `--path-prefix` flag and provide the full/absolute path of each repository.
But it's not convenient.

### Additional context

_No response_

### Supporter

ldez:
Adds an option to display absolute paths instead of relative paths.

For now, there is only one option, but if there are other concrete use cases, we can extend the behavior.

Fixes #5649


alexandear (approved):

.golangci.next.reference.yml
```diff
@@ -4155,9 +4155,16 @@ output:
       path: ./path/to/output.json
 
   # Add a prefix to the output file references.
+  # This option is ignored when using `output.path-prefix: abs` mode.
```
pierrre, line 4158:
```suggestion
  # This option is ignored when using `output.path-mode: abs` mode.
```

?

pkg/commands/flagsets.go
```diff
@@ -62,6 +62,8 @@ func setupRunFlagSet(v *viper.Viper, fs *pflag.FlagSet) {
 func setupOutputFlagSet(v *viper.Viper, fs *pflag.FlagSet) {
 	internal.AddFlagAndBind(v, fs, fs.String, "path-prefix", "output.path-prefix", "",
 		color.GreenString("Path prefix to add to output"))
+	internal.AddFlagAndBind(v, fs, fs.String, "path-mode", "output.path-mode", "",
+		color.GreenString("Path mode to use"))
```
alexandear, line 66: I propose extending descriptions for `path-prefix` and `path-mode` with possible values.

It's for consistency with the rest of descriptions.
image

ldez: There are no "possible values" `path-prefix`

PR #5652, opened 2025-03-28
feat: colored diff for fmt command

ldez:
<details><summary>--diff</summary>

Screenshot_20250328_210820

</details>

<details><summary>--diff-colored</summary>

Screenshot_20250328_210752

</details>

### DIFF
--- a/pkg/commands/fmt.go
+++ b/pkg/commands/fmt.go
@@ -20,8 +20,9 @@ import (
 type fmtOptions struct {
 	config.LoaderOptions
 
-	diff  bool // Flag only.
-	stdin bool // Flag only.
+	diff        bool // Flag only.
+	diffColored bool // Flag only.
+	stdin       bool // Flag only.
 }
 
 type fmtCommand struct {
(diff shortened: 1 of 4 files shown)


pkg/commands/fmt.go
```diff
@@ -70,6 +71,7 @@ func newFmtCommand(logger logutils.Log, info BuildInfo) *fmtCommand {
 	setupFormattersFlagSet(c.viper, fs)
 
 	fs.BoolVarP(&c.opts.diff, "diff", "d", false, color.GreenString("Display diffs instead of rewriting files"))
+	fs.BoolVar(&c.opts.diffColored, "diff-colored", false, color.GreenString("Display diffs instead of rewriting files (with colors)"))
```
ccoVeille, line 74: What about having just --color ?

Using --diff --color would be more classic

ldez: I think about it when I created the flag, but I prefer one flag because it's easier to use.

ldez: Also because `--color` can only be used with `--diff` :wink:

alexandear (approved):

bombsimon (approved):

PR #5692, opened 2025-04-04
docs: add `redundant-test-main-exit` to `revive` rules

G-Rath:
This rule has been available for a while but apparently never got added to the schema - I built the list by copying the table of contents from [the docs](https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md)


ldez (approved):
LGTM, thank you.

PR #5697, opened 2025-04-05
docs: explicitly describe that the `migrate` command automatically migrate `linters.presets`

kyu08:
Thanks for maintaining golangci-lint always!
golangci-lint helps me and my team prevent bugs by making static analysis easy to execute. 👏

When I read the migration guide to `v2`, I wondered, "How should I migrate preset linters to v2 config file? Should I indentify what linters belong to the preset, and write out each linter name to `v2` config file one by one?".
After executing the `migrate` command, I realized the `migrate` command did it instead of me. 😂

I believe it would be easier to understand if this were described in migration guide.
So I suggest adding description that shows the `migrate` command automatically migrates `linters.presets` to `linters.enable`.

<details>

image
image

</details>


ldez (approved):
LGTM

PR #5707, opened 2025-04-09
Bump goconst to v1.8.0

kalexmills:
I realized too late that I can't expect this PR to be merged unless I am the author of the linter. Alas, the work is done.

In any case, we made some changes upstream to release v1.8.0 and `goconst` now uses the typechecker as a result. I'll open an issue downstream to alert the linter author.

goconst now finds duplicate constants and can evaluate constant expressions in some cases.

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -632,6 +632,12 @@ linters:
       # Exclude strings matching the given regular expression.
       # Default: ""
       ignore-strings: 'foo.+'
+      # Look for duplicate consts with matching values.
+      # Default: false
+      find-duplicates: true
+      # Evaluate constant expressions when searching for constants.
+      # Default: false
+      eval-const-expressions: true
 
     gocritic:
       # Disable all checks.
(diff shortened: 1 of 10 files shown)


ldez:
The rules are the rules, sorry.

ldez:
It's useless to open an issue on the goconst repo, as the PR template says: the dependabot handles the updates.

kalexmills:
@ldez thanks for the quick response. We added features in v1.8.0 which will require integration support to use, that's mainly what this PR handles.

In any case, I've opened an issue upstream and pointed maintainers to this PR.

ldez:
I will explain the reason for this rule: it takes more time to review a PR than do the modifications.
Also, the Dependabot uses a specific commit format and PR description with useful information for the update.

So, even for authors of linters, in most cases, I prefer to handle that with Dependabot.

ldez:
In all cases, the update to 1.8.0 will be rejected because of the usage of `types.NewChecker`: a linter should not do that because it will lead to extremely poor performance.
The information about types is inside the `pass`.

PR #5718, opened 2025-04-11
feat: add arangolint linter

Crocmagnon:
https://github.com/Crocmagnon/arangolint

Opinionated linter for [ArangoDB go driver v2](https://github.com/arangodb/go-driver).

Only one feature for now: Enforce explicit `AllowImplicit` in transactions (see repo readme)


ldez:
At the moment, I'm wondering about the scope of this linter, because it's extremely niche.

Crocmagnon:
Based on the number of stars on the github driver it may seem so but it looks like there are some large companies using it: https://arangodb.com/solutions/solutions-customers/
I don't know if they're using the go driver though.

ldez (approved):
We have linters for SQL, Ginkgo, etc., and Arango seems popular, so I think it is acceptable to add it.

LGTM

PR #5735, opened 2025-04-15
docs: fix default value of linters.exclusions.generated

chris-reeves:
The default value for `linters.exclusions.generated` as documented in the reference config file (and
thus https://golangci-lint.run/usage/configuration/#linters-configuration) conflicts with the
documentation in https://golangci-lint.run/product/migration-guide/#command-migrate. After a bit of
digging I confirmed that the it's the reference config that's incorrect.

This PR fixes the reference config file to reflect that correct default value.


ldez (approved):
Good catch, tank you :+1:

PR #5741, opened 2025-04-16
fix: toStaticCheckSettings

sivchari:
Currently, toStaticCheckSettings only treat order to make unique slice. So if we receive such as []string{"all", "-S1000"} from staticcheck, the output results in []string{"-S1000", "all"}. The rule that has minus symbol means the rule will be disabled from previous one. It means that current order must be failed if developer doesn't fix it, so I fixed it.


ldez (approved):
LGTM

PR #5745, opened 2025-04-17
feat: updating funcorder to v0.3.0, support for alphabetical.

manuelarte:
Updating funcorder to `v0.3.0` to add alphabetical setting.


ldez:
Closed in favor of #5743

manuelarte:
do I still need to create a PR to add the setting?

ldez:
No, the updates of linters are handled inside the dependabot PR, it's easier for multiple reasons.

It takes more time to review a PR than to make the modifications.
Also, the Dependabot uses a specific commit format and PR description with useful information for the update and the history.

So, even for authors of linters, in most cases, I prefer to handle that with Dependabot.

PR #5793, opened 2025-05-11
feat: sign release artifacts with cosign

scop:
Closes https://github.com/golangci/golangci-lint/issues/2462

Sample results in my fork (do not mind the changelog, scroll down to assets): https://github.com/scop/golangci-lint/releases/tag/v0.0.0

PR #5794, opened 2025-05-11
feat: verify installer downloads with cosign

scop:
Depends on https://github.com/golangci/golangci-lint/pull/5793

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -42,6 +42,7 @@ execute() {
   log_debug "downloading files into ${tmpdir}"
   http_download "${tmpdir}/${TARBALL}" "${TARBALL_URL}"
   http_download "${tmpdir}/${CHECKSUM}" "${CHECKSUM_URL}"
+  cosign_verify "${tmpdir}/${CHECKSUM}"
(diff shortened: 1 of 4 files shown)


install.sh
```diff
@@ -390,6 +391,20 @@ End of functions from https://github.com/client9/shlib
 ------------------------------------------------------------------------
 EOF
 
+cosign_verify() {
+  if ! is_command cosign; then
+    log_info "cosign is not available, skipping verify"
+    return 0
+  fi
+  checksums=$1
+  http_download "${tmpdir}/${CHECKSUM_COSIGN_BUNDLE}" "${CHECKSUM_COSIGN_BUNDLE_URL}"
```
scop, line 400: This will fail for releases that don't have signed checksums. I suppose we should address it somehow, but I'm not sure what would be the best way. Some thoughts:
- Hardcode some version comparison (chore in shell script code), don't try it for versions known to not have it?
- Skip verify if downloading the cosign bundle fails due to HTTP 404? Not sure if there's a good way to implement that for wget.

scop: An example of this problem can be seen in the CI failure of this PR.

ldez:
please use your fork.

scop:
> please use your fork.

#5806

PR #5795, opened 2025-05-13
docs: update new linter template to specify that fix should also be s…

manuelarte:
…upported and covered in tests.

Stupid question:

Does it make sense to update the New Linter template specifying that the fix scenario should also be covered if applicable? Or is it kind of obvious that that's the case.


ldez:
This is "obvious", and it's more a part of the review itself.

PR #5796, opened 2025-05-13
dev: simplify linter constructors

ldez:
90% of linters use only one analyzer, so a `Linter` constructor that takes only one analyzer can be used.

Also, the configuration map for those linters always uses the name of the analyzer as the first and only key, then, a method `WithConfig`, that handles this key, can be used.

This simplifies and removes a lot of code.

### DIFF
--- a/pkg/goanalysis/linter.go
+++ b/pkg/goanalysis/linter.go
@@ -13,11 +13,6 @@ import (
 	"github.com/golangci/golangci-lint/v2/pkg/result"
 )
 
-const (
-	TheOnlyAnalyzerName = "the_only_name"
-	TheOnlyanalyzerDoc  = "the_only_doc"
-)
-
 type LoadMode int
 
 func (loadMode LoadMode) String() string {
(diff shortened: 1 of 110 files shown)


pkg/goanalysis/linter.go
```diff
@@ -71,6 +70,24 @@ func (lnt *Linter) LoadMode() LoadMode {
...
+func (lnt *Linter) WithDesc(desc string) *Linter {
+	lnt.desc = desc
+
+	return lnt
+}
+
+func (lnt *Linter) WithConfig(cfg map[string]any) *Linter {
+	if len(cfg) == 0 {
+		return lnt
+	}
```
bombsimon, line 82: When do we faulty pass an empty config? Maybe it's not used but I'd expect this to just clear the config if any was set rather than a noop.

bombsimon (approved):
LGTM!

pkg/goanalysis/linter.go
ldez: Technically, the configuration is only set once, I don't see a use case where the configuration needs to be reset.

The current change split the map of map to 2 maps, and so before that, the 1st map was empty when the second map was empty.

So I check if the 2nd map is empty to avoid creating the 1st map if not needed.

It's not simple to explain.

PR #5799, opened 2025-05-14
dev: remove AUR publishing

ldez:
The AURs have been "moved" to the official 'extra' repository, based on a unilateral decision without discussion.

https://gitlab.archlinux.org/archlinux/packaging/packages/golangci-lint

This is annoying because they removed `golangci-lint` AUR but also `golangci-lint-bin` AUR.
The official 'extra' repository always compiles golangci-lint instead of using our official binaries, so this is a problem because we lost control of the Go version used to build.
This is a major problem.

Also, `date`, `version`, and `commit` fields are not set.

I regret that there was no discussion before doing this.
I also regret that it is not possible to create an account on the Arch GitLab, and so not possible to discuss.

This is the third time I've seen my AUR "migrated" to the official Arch repositories; this is always the same bad feeling.


bombsimon (approved):

carlsmedstad:
Hey, sorry about the lack of notification - I usually do not write a comment about potential moves from the AUR if the move is uncontroversial which I thought was the case in this case, but I will make sure to do so in the future.

> The official 'extra' repository always compiles golangci-lint instead of using our official binaries, so this is a problem because we lost control of the Go version used to build.
> This is a major problem.

Just so I understand correctly - `golangci-lint` supports the Go versions up until and including the one it was compiled with?  If so we can ensure that the package is rebuilt when new Go versions are made available in the repos.

>  I also regret that it is not possible to create an account on the Arch GitLab, and so not possible to discuss.

Please write an email to accountsupport@archlinux.org to get an account - We really appreciate input (issues, MRs, etc.), and that's the best place for packages in the official repositories.

ldez:
> golangci-lint supports the Go versions up until and including the one it was compiled with? 

The answer is more complex than yes or no.

```mermaid
---
title: Is golangci-lint support go1.24?
---
flowchart
    A([compiled with go1.24?]) -- yes --> B([needs update of the internals of golangci-lint?])
    A -- no --> C((NO))
    B -- yes --> D([uses right internal pieces?])
    B -- no --> E((YES))
    D -- yes --> G((YES))
    D -- no --> F((NO))
```

ldez:
@carlsmedstad can you restore `golangci-lint-bin`?

You can keep `golangci-lint` (with some updates if possible), but the `-bin` variant is important for us.

carlsmedstad:
@ldez Absolutely, I've republished the `-bin` package and made you maintainer: https://aur.archlinux.org/packages/golangci-lint-bin

We do unfortunately not have a way to restore comments and votes though, so this is the best I can do :/

Feel free to ping me if you're having issues creating an account for the Arch Linux GitLab btw so I can push the process along if needed.

PR #5806, opened 2025-05-15
feat: verify installer downloads with cosign

scop:
Depends on https://github.com/golangci/golangci-lint/pull/5793

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -42,6 +42,7 @@ execute() {
   log_debug "downloading files into ${tmpdir}"
   http_download "${tmpdir}/${TARBALL}" "${TARBALL_URL}"
   http_download "${tmpdir}/${CHECKSUM}" "${CHECKSUM_URL}"
+  cosign_verify "${tmpdir}/${CHECKSUM}"
   hash_sha256_verify "${tmpdir}/${TARBALL}" "${tmpdir}/${CHECKSUM}"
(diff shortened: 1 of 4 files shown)


install.sh
```diff
@@ -390,6 +391,20 @@ End of functions from https://github.com/client9/shlib
 ------------------------------------------------------------------------
 EOF
 
+cosign_verify() {
+  if ! is_command cosign; then
+    log_info "cosign is not available, skipping verify"
+    return 0
+  fi
+  checksums=$1
+  http_download "${tmpdir}/${CHECKSUM_COSIGN_BUNDLE}" "${CHECKSUM_COSIGN_BUNDLE_URL}"
```
scop, line 400: This will fail for releases that don't have hose signed checksum .cosign.bundles. I suppose we should address it somehow, but I'm not sure what would be the best way. Some thoughts:
* Hardcode some version comparison, don't try it for versions known to not have it? This is somewhat a chore to do in the shell script.
* Skip verify if downloading the cosign bundle fails due to HTTP 404? Not sure if there's a good way to implement that for wget.

scop:
Closing as I just closed the parent PR. Will likely have a look at making another on top of GH artifact attestations when we have them, https://github.com/golangci/golangci-lint/pull/6257

PR #5824, opened 2025-05-22
updated go version

nirrozenbaum:
This PR updates the go version used in the go.mod file.
current version causes lint not to work when running it for projects that use go version 1.24.
When trying to run lint against a 1.24 project the following error is returned:
```
Error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24.0)
Failed executing command with error: can't load config: the Go language version (go1.23) used to build golangci-lint is lower than the targeted Go version (1.24.0)
```

this minor change is fixing this issue.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/golangci/golangci-lint/v2
 
-go 1.23.0
+go 1.24.0
 
 require (
 	4d63.com/gocheckcompilerdirectives v1.3.0


nirrozenbaum:
@bombsimon thanks for your quick reply.

I went over the attached issues quickly. haven't found what I'm looking for.
is there an official tag/version I can use to download golangci-lint using `go install` that is built using version 1.24?

bombsimon:
> is there an official tag/version I can use to download golangci-lint using go install that is built using version 1.24?

When you run `go install` _you_ are building it, so you need to ensure that _your_ pipeline is using Go 1.24. All tags after 1.64 can be bult with Go 1.24. What version do you see when you run `go version`?

I suspect you have strong reasons not to download the pre-built binaries? If not, there are plenty ways:

- [In GitHub CI, use the GitHub action](https://golangci-lint.run/welcome/install/#github-actions)
- [Download pre-built binaries with curl](https://golangci-lint.run/welcome/install/#other-ci)
- With package manager such as `brew`, `apt`, `apk`, `Chocolatey`, `Scoop` and many more, [see here](https://golangci-lint.run/welcome/install/#binaries)

We really try to make it so you never have to build it yourself.

nirrozenbaum:
no strong reason to build myself. it works.
thanks

PR #5826, opened 2025-05-23
Add noinlineerr linter

AlwxSin:
Adding [noinlineerr](https://github.com/AlwxSin/noinlineerr) linter.

Purpose - allow only one style of error handling.

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -91,6 +91,7 @@ linters:
     - nilnil
     - nlreturn
     - noctx
+    - noinlineerr
     - nolintlint
     - nonamedreturns
     - nosprintfhostport
@@ -200,6 +201,7 @@ linters:
     - nilnil
     - nlreturn
     - noctx
+    - noinlineerr
     - nolintlint
     - nonamedreturns
     - nosprintfhostport
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	4d63.com/gochecknoglobals v0.2.2
 	github.com/4meepo/tagalign v1.4.2
 	github.com/Abirdcfly/dupword v0.1.4
+	github.com/AlwxSin/noinlineerr v1.0.1
 	github.com/Antonboom/errname v1.1.0
 	github.com/Antonboom/nilnil v1.1.0
 	github.com/Antonboom/testifylint v1.6.1
--- a/go.sum
+++ b/go.sum
@@ -39,6 +39,8 @@ github.com/4meepo/tagalign v1.4.2 h1:0hcLHPGMjDyM1gHG58cS73aQF8J4TdVR96TZViorO9E
 github.com/4meepo/tagalign v1.4.2/go.mod h1:+p4aMyFM+ra7nb41CnFG6aSDXqRxU/w1VQqScKqDARI=
 github.com/Abirdcfly/dupword v0.1.4 h1:jxqkdgO0L7FSwq4fMu3d+fAA1E3bBQAtl4utA3tcQZM=
 github.com/Abirdcfly/dupword v0.1.4/go.mod h1:s+BFMuL/I4YSiFv29snqyjwzDp4b65W2Kvy+PKzZ6cw=
+github.com/AlwxSin/noinlineerr v1.0.1 h1:xiNjgRwY1ExRJ8xy6NVfpxU5KyAssTH0h5cuoZipAd4=
+github.com/AlwxSin/noinlineerr v1.0.1/go.mod h1:FjvQoRSDujHgtQilUOsiFEp0Rh1J8qAVw53fxqSuI58=
 github.com/Antonboom/errname v1.1.0 h1:A+ucvdpMwlo/myWrkHEUEBWc/xuXdud23S8tmTb/oAE=
 github.com/Antonboom/errname v1.1.0/go.mod h1:O1NMrzgUcVBGIfi3xlVuvX8Q/VP/73sseCaAppfjqZw=
 github.com/Antonboom/nilnil v1.1.0 h1:jGxJxjgYS3VUUtOTNk8Z1icwT5ESpLH/426fjmQG+ng=
--- a/jsonschema/golangci.next.jsonschema.json
+++ b/jsonschema/golangci.next.jsonschema.json
@@ -791,6 +791,7 @@
             "nilnil",
             "nlreturn",
             "noctx",
+            "noinlineerr",
             "nolintlint",
             "nonamedreturns",
             "nosprintfhostport",
--- a/pkg/golinters/noinlineerr/noinlineerr.go
+++ b/pkg/golinters/noinlineerr/noinlineerr.go
@@ -1,0 +1,18 @@
+package noinlineerr
+
+import (
+	"github.com/AlwxSin/noinlineerr"
+	"golang.org/x/tools/go/analysis"
+
+	"github.com/golangci/golangci-lint/v2/pkg/goanalysis"
+)
+
+func New() *goanalysis.Linter {
+	analyzer := noinlineerr.NewAnalyzer()
+	return goanalysis.NewLinter(
+		analyzer.Name,
+		analyzer.Doc,
+		[]*analysis.Analyzer{analyzer},
+		nil,
+	).WithLoadMode(goanalysis.LoadModeTypesInfo)
+}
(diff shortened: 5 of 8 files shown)


ldez:

- [ ] The CLA must be signed

### Pull Request Description

### Linter

- [ ] It must use Go version <= 1.23.0

### Others Requirements

### Recommendations

---

AlwxSin:
@ldez thank you for checklist.

Fixed:
* It must use Go version <= 1.23.0
* They must have at least one std lib import.
* The version in WithSince(...) must be the next minor version (v2.X.0) of golangci-lint.
* added `.gitignore`

Quick question:
> The files (tests and linter) inside golangci-lint must have the same name as the linter.

I thought they already do – did I miss something? Could you clarify what exactly should be renamed or matched?

ldez:
I detected at least one false-positive:
```go

func _() {
	if ok, _ := strconv.ParseBool("1"); ok {
		fmt.Println("ok")
	}
}
```

The linter is very limited: it only handles explicit `error` type references, so false negatives can happen.

AlwxSin:
@ldez Thanks for the feedback!

Do you have any specific ideas in mind?

Some thoughts I had:

Adding an option to choose the preferred style (though I’m not sure if it makes sense to only prefer the inline variant).

Adding autofix support — though that’s a bit more complex for me at the moment and would need some digging in.

Happy to hear any other suggestions you might have!

ldez:
The proposed options are not what I expected.

- `allow-inline-in-tests` will not be used by golangci-lint because this must be handled inside the golangci-lint exclusions and not a the linter level.
- `allow-multi-assign` I don't see why this should be a specific topic.


I'm talking about a broader perspective of inlined `if`, not to the restricted scope of errors.

AlwxSin:
About the broader perspective — I get your point, but I’d prefer to keep the scope limited to `err := ...` cases for now.
This is where most readability issues tend to show up in practice, and it matches the linter's intent (and name — `noinlineerr`).

I’ve often seen inline error handling being overused or misused, whereas inline Init with other values is way less common —
so I focused on the pattern that was actually causing pain in the codebases I work with.

That said, building a separate linter to cover all inline `Init` statements (`if x := ...; cond {}`) could definitely be interesting —
but that feels like a broader concern and a different project altogether.

ldez:
ok, so we agree that if a new linter is added that handles the whole topic of inlined `if`, your linter will be deprecated inside golangci-lint and removed.

If we agree on that, we can accept this linter.

ldez (approved):
LGTM

G-Rath:
@ldez fyi it looks like this linter is not listed in https://golangci-lint.run/usage/linters/

ldez:
@G-Rath refresh the page: https://golangci-lint.run/usage/linters/#noinlineerr

ruanxin:
Hi @AlwxSin:

I just noticed your contribution while our team was trying to adopt the latest golangci-lint. First, thanks for your work on this, and I understand the intent behind it, but I’d like to offer a different perspective.

One important advantage of inline error handling is that **it limits the scope of the error variable to exactly where it’s relevant**. When an error is only used for a single operation, there’s no need to keep it alive in a wider scope. This helps avoid accidental reuse or confusion with unrelated errors, and keeps reasoning about the code more localized.

When inline handling is disallowed, you’re forced to declare a top-level err and reuse it for multiple unrelated calls, which can subtly reduce readability. For example:

```
err := doSomething()
if err != nil {
    return err
}

err = doAnotherThing()
if err != nil {
    return err
}

err = doLastThing()
if err != nil {
    return err
}
```

Here, a single `err` variable is reused across unrelated operations, and readers must mentally track its value over a longer stretch of code.

With inline checks, each error is self-contained and unambiguous:

This approach makes it immediately clear where each error came from, and the variable’s scope exactly matches its relevance.

It’s also worth noting that major style guides explicitly recommend inline error handling in certain cases:

[Google Code Style – Concision](https://google.github.io/styleguide/go/guide#concision)
image


[Uber Go Style Guide – Reduce Scope of Variables](https://github.com/uber-go/guide/blob/master/style.md#reduce-scope-of-variables)
image

For these reasons, I believe your `noinlinerr` is basically against those recommendations. It removes a useful language feature that often improves scope control, readability, and maintainability.  

cc @ldez, enabling this linter by default really goes against widely known, established style recommendations. Could we at least mark it as not recommended by default?

bombsimon:
The linter is not enabled by default, no new linters are. If you don't agree with the style (I don't), don't enable the linter.

ldez:
This linter, like other linters, is related to style.
The style is personal, so I don't want to provide any recommendations about using or not using a linter.

ruanxin:
Hi @bombsimon @ldez , thanks for the quick feedback.

First, I have to apologise, after checking our golangci.yaml, I realised the issue happened because we have `default: all` enabled and then selectively disable specific linters. That means when we upgraded, `noinlineerr` became active automatically.

But this actually highlights a point: our team (and I suspect many others) **treat golangci-lint as a trusted guideline for recommended code style.** When new linters are added, the implicit expectation is that they are **widely recognized, well-accepted, and non-controversial**. In this case, as I mentioned earlier, many in the Go community consider banning inline error handling a code smell rather than a style preference, as I mentioned in the [Uber’s guideline](https://github.com/uber-go/guide/blob/master/style.md#reduce-scope-of-variables), which explicitly encourages the opposite.

By including a specific linter, golangci-lint still implicitly signals that this style is worth enforcing, since golangci-lint is trusted by a huge portion of the Go community, that signal matters I believe.

However, if the position of golangci-lint is that style linters are **purely neutral and not meant to guide or educate on best practices**, then I understand, but that’s a different philosophy than what I think many users expect. Perhaps we’ve simply placed too much expectation on golangci-lint as a de facto style authority, but that expectation is real.

Follow-up question: if someone later proposed a linter that enforces the use of inline error handling, which directly conflicts with this `noinlineerr`,  as it's exactly the opposite style, but is backed by reputable guidelines. Would you accept such linter? If the answer is yes, I probably will contribute to that linter 😄

AlwxSin:
@ruanxin I completely understand that handling errors inline is largely a matter of personal preference. Personally, I find it harder to read code where errors are inlined, because it shifts the focus from the function call itself to the error handling logic. That’s the main reason why I created this linter.

I also intentionally decided not to enable this linter by default, knowing that in existing codebases this could cause significant issues. In my view, this is a highly opinionated linter, so the decision to enable it should be left entirely to the user.

P.S.I respect the guidelines from Uber and Google, though I see them as recommendations, not mandatory rules, and I prefer a different approach in this case.

ldez:
golangci-lint is "neutral" but tries not to recommend bad practices.

I would emphasize that the user guides you are quoting don't enforce the usage of error inlining.
They use terms like "Where possible" because **always inlining errors is a bad practice**.

I have a lot of examples where error inlining created unwanted complexity by hiding factorization or creating shadow declarations.

Always not inlining is not a bad practice; this is just a point of view.

So, I don't think we will accept a linter that forces the usage of error inlining, but you can create a plugin if you want.

https://golangci-lint.run/docs/plugins/module-plugins/

AlwxSin:
> ok, so we agree that if a new linter is added that handles the whole topic of inlined `if`, your linter will be deprecated inside golangci-lint and removed.
> 
> If we agree on that, we can accept this linter.

@ruanxin [ldez’s comment](https://github.com/golangci/golangci-lint/pull/5826#issuecomment-2917089058) actually touches on exactly this point — if there is a linter that comprehensively covers all cases of inline if usage and allows enforcing or forbidding any variation, then my noinlineerr linter would become unnecessary.

bombsimon:
> But this actually highlights a point: our team (and I suspect many others) **treat golangci-lint as a trusted guideline for recommended code style.** When new linters are added, the implicit expectation is that they are **widely recognized, well-accepted, and non-controversial**.

If so, `golangci-linter` would only accept linters that would make sense if enabling all linters would be a recommended way. In fact, there are other considerations of not doing that, e.g. #4661 which is about duplicated/overlapping linters. As mentioned, `golangci-lint` is supposed to be neutral, maybe this is something that could be documented better.

I think the discussion so far has been around how to make sharable presets/sharable standards such as the `eslint-config-airbinb` has been for JS/TS which would promote styles you like such as Uber or Google style guide.

Some reference to previous issues/discussions where recommended/sharable configuration has been discussed:

- #1141
- #4393 (see the answer here)
- #3954
- #2137
- #2618
- #4661

PR #5847, opened 2025-06-01
perf: nolint simplifications and improvements

scop:
See individual comments for details.

I have not run actual benchmarks on these exact changes, but from experience I expect them to show some performance and memory use improvements, in addition to the code being simpler.

### DIFF
--- a/pkg/golinters/nolintlint/internal/nolintlint.go
+++ b/pkg/golinters/nolintlint/internal/nolintlint.go
@@ -26,7 +26,7 @@ type Needs uint
 
 const commentMark = "//"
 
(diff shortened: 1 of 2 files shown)


ldez:
nolint code is considered frozen.

scop:
That sounds odd to me. Anyway, just to make sure it did not go unnoticed, this PR contains changes to both nolint code, as well as nolintlint code. Are _both_ frozen, or is the nolintlint change acceptable if I removed the nolint commit? Reopening for comments.

ldez:
They are both frozen because it's the same topic.

PR #5854, opened 2025-06-02
dev: use slice of pointers

ldez:
`result.Issue` are large structures, and we have to create a copy to iterate over them, so it's better to use slices of pointers instead of slices of structures.

https://github.com/google/go-github/issues/180


bombsimon (approved):
LGTM! Did you benchmark this? Manually or with some bench tests?

Mostly just curious how big the difference is, would be interesting to know.

ldez:
The advantage is more related to usage than performance because we mainly used indexes.

bombsimon:
> The advantage is more related to usage than performance because we mainly used indexes.

Yeah makes sense, saw the linked issue and changes, was mostly curious if it had any implications at all. The changes looks good and simplifies code though so LGTM

PR #5882, opened 2025-06-16
Add workflowcheck linter

Kamefrede:
## Description

This PR adds the `workflowcheck` linter which is a static analysis tool for Temporal.io workflows.

https://pkg.go.dev/go.temporal.io/sdk/contrib/tools/workflowcheck
https://github.com/temporalio/sdk-go


ldez:
From the [readme](https://github.com/temporalio/sdk-go/blob/master/contrib/tools/workflowcheck/README.md):
> This will not catch all cases of non-determinism such as global var mutation. This is just a helper and developers should still scrutinize Workflow code for other non-determinisms.


So, from the POV of golangci-lint, this is a detector, not a linter.

I recommend using it as a plugin: https://golangci-lint.run/plugins/module-plugins/

Kamefrede:
Fair enough, thank you for taking the time and considering it though! :)

love1120520:
________________________________

PR #5884, opened 2025-06-18
fix: typecheck memory leak

### LINKED ISSUE #5883: memory usage very high and not exit when code has some compile error
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).

go install

We found the lint stage take many memory and got killed in the CI recently.

Image

### Version of golangci-lint

<details>

```console
golangci-lint has version  built with go1.24.0 from (unknown, modified: ?, mod sum: "") on (unknown)
```

</details>

### Configuration

<details>

```console
../golangci-lint/golangci-lint-debug run ./...
```

</details>

### Go environment

<details>

```console
go version go1.24.0 darwin/amd64
AR='ar'
CC='clang'
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_ENABLED='1'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
CXX='clang++'
GCCGO='gccgo'
GO111MODULE=''
GOAMD64='v1'
GOARCH='amd64'
GOAUTH='netrc'
GOBIN=''
GOCACHE='/Users/xxx/Library/Caches/go-build'
GOCACHEPROG=''
GODEBUG=''
GOENV='/Users/xxx/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFIPS140='off'
GOFLAGS=''
GOGCCFLAGS='-fPIC -arch x86_64 -m64 -pthread -fno-caret-diagnostics -Qunused-arguments -fmessage-length=0 -ffile-prefix-map=/var/folders/bh/g2h9d2vx35b2b4k8pn4k2rt80000gn/T/go-build664656937=/tmp/go-build -gno-record-gcc-switches -fno-common'
GOHOSTARCH='amd64'
GOHOSTOS='darwin'
GOINSECURE=''
GOMOD='/Users/xxx/work/xxx/yyy/kubernetes/go.mod'
GOMODCACHE='/Users/xxx/go/pkg/mod'
GONOPROXY='git.in.xxx.com'
GONOSUMDB='git.in.xxx.com'
GOOS='darwin'
GOPATH='/Users/xxx/go'
GOPRIVATE='git.in.xxx.com'
GOPROXY='https://goproxy.in.xxx.com,direct'
GOROOT='/usr/local/go'
GOSUMDB='sum.golang.org'
GOTELEMETRY='on'
GOTELEMETRYDIR='/Users/xxx/Library/Application Support/go/telemetry'
GOTMPDIR=''
GOTOOLCHAIN='auto'
GOTOOLDIR='/usr/local/go/pkg/tool/darwin_amd64'
GOVCS=''
GOVERSION='go1.24.0'
GOWORK='/Users/xxx/work/xxx/yyy/kubernetes/go.work'
PKG_CONFIG='pkg-config'
```

</details>

### Verbose output of running

<details>

```console
```

</details>

I have reproduced the bug in  https://github.com/alingse/golangci-lint/tree/dump-heap  and  https://github.com/alingse/kubernetes/tree/add-small-crack , but I don't know is this only way to trigger the bug.

only a small compile error in "_test.go"  cause the golangci-lint take 25GB memory and keep growing

<details>

Image

Image

Image

Image

</details>

I research the code and pprof, found this code will create a tree of `joinError`, and cause memory use when `fmt.Errorf`

the check has alread done, but the string format process never exit (maybe I wait long time it can return)

```go
	var depErrors error
	for _, dep := range act.Deps {
		if dep.Err != nil {
			depErrors = errors.Join(depErrors, errors.Unwrap(dep.Err))
		}
	}
	if depErrors != nil {
		act.Err = fmt.Errorf("failed prerequisites: %w", depErrors)
		return
	}
```

the `errror.Join` not implement the `func Unwrap() error`, it only implement the `func Unwrap() []error`

> // In particular Unwrap does not unwrap errors returned by [Join].

and `fmt.Errorf("failed prerequisites: %w", depErrors)` will format every error `Error()` string in the tree.

I try change the code like this ( https://github.com/golangci/golangci-lint/commit/b7bc902fcfb8c0f3d451adf5f931abdd50ea1b99 ) , it exit and report compile error now.

```go
	var depErrors []error
	for _, dep := range act.Deps {
		if dep.Err != nil {
			depErrors = append(depErrors, errors.Unwrap(dep.Err))
		}
	}
	if depErrors != nil {
		act.Err = errors.Join(depErrors...)
		return
	}
```

```console
test/integration/cronjob/cronjob_test.go:19:1: expected 'package', found 'import' (typecheck)
import (
^
test/integration/cronjob/cronjob_test.go:20:2: expected ';', found "context" (typecheck)
	"context"
	^
2 issues:
* typecheck: 2
```

### Validation

- [x] Yes, I've included all information above (version, config, etc.).

### Supporter

ldez:
Fixes #5883

### DIFF
--- a/pkg/goanalysis/runner_checker.go
+++ b/pkg/goanalysis/runner_checker.go
@@ -90,12 +90,12 @@ func (act *action) analyze() {
 	// Report an error if any dependency failures.
 	var depErrors error
 	for _, dep := range act.Deps {
-		if dep.Err != nil {
+		if dep.Err != nil && errors.Unwrap(dep.Err) != nil {
 			depErrors = errors.Join(depErrors, errors.Unwrap(dep.Err))
 		}
 	}
 	if depErrors != nil {
-		act.Err = fmt.Errorf("failed prerequisites: %w", depErrors)
+		act.Err = errors.Join(depErrors, errors.New("failed prerequisites"))
 		return
 	}
 


pkg/goanalysis/runner_checker.go
```diff
@@ -90,12 +90,12 @@ func (act *action) analyze() {
 	// Report an error if any dependency failures.
 	var depErrors error
 	for _, dep := range act.Deps {
-		if dep.Err != nil {
+		if dep.Err != nil && errors.Unwrap(dep.Err) != nil {
 			depErrors = errors.Join(depErrors, errors.Unwrap(dep.Err))
```
ccoVeille, line 94: dep.Err could be unwrapped only once

pkg/goanalysis/runner_checker.go
```diff
@@ -90,12 +90,12 @@ func (act *action) analyze() {
...
 		}
 	}
 	if depErrors != nil {
-		act.Err = fmt.Errorf("failed prerequisites: %w", depErrors)
+		act.Err = errors.Join(depErrors, errors.New("failed prerequisites"))
```
ccoVeille, line 98: In this PR, it's clear why you are doing the change, but someone later might bump on this line and update it back to use fmt.Errorf as it sounds logic and natural, while this code sounds odd.

I'm unsure there is a way to test a possible regression, nor it's worth trying to detect the regression.

But adding a comment could be interesting.

pkg/goanalysis/runner_checker.go
ldez: nope: a value can be read multiple times.

alingse: // Unwrap only calls a method of the form "Unwrap() error".
// In particular Unwrap does not unwrap errors returned by [Join].
func Unwrap(err error) error {
	u, ok := err.(interface {
		Unwrap() error
	})
	if !ok {
		return nil
	}
	return u.Unwrap()
}

if dep.Err is return by Join, it is always nil 

because

func (e *joinError) Unwrap() []error {
	return e.errs
}

bombsimon (approved):

pkg/goanalysis/runner_checker.go
ccoVeille: my point was to somehow raise the fact unwrapping as a cost, and I felt like this could be better

```suggestion
		if dep.Err != nil {
			unwrappedErr := errors.Unwrap(dep.Err)
			if unwrappedErr != nil {
				depErrors = errors.Join(depErrors, unwrappedErr)
			}
```

But you might consider this as a micro optimization

bombsimon: Since we only add the unwrapped version, we don't even need the extra check. It's safe to pass `nil` to `errors.Unwrap` and both `nil` and non wrapped errors will return  `nil` and not be joined.

```go
if unwrappedErr := errors.Unwrap(dep.Err); unwrappedErr != nil {
    depErrors = errors.Join(depErrors, unwrappedErr)
}
```

alingse:
@ldez I still have a question,  see my review upper.

`errors.Unwrap` don't support `errors.Join`, so the action dep.Err will be nil ? and finally action Err will be nil too?

Or something I miss understanding?

ldez:
There is no review above from you.

Neither `dep.Err` nor `act.Err` will be nil: the unwrapped `dep.Err` errors are added to `depErrors`.

alingse:
`errors.Unwrap(dep.Err)` will got `nil` now.

if `dep.Err` is return by `errors.Join`, it is always `nil`

```go
// Unwrap only calls a method of the form "Unwrap() error".
// In particular Unwrap does not unwrap errors returned by [Join].
func Unwrap(err error) error {
	u, ok := err.(interface {
		Unwrap() error
	})
	if !ok {
		return nil
	}
	return u.Unwrap()
}
```

`errors.Unwrap` don't support `errors.Join` error, and will be nil. so the `errors.Unwrap(dep.Err) != nil` will always be false 


```go
func (e *joinError) Unwrap() []error {
	return e.errs
}
```
the code is ok before because `act.Err` is wraped by `fmt.Errorf`, the `UnWrap` can got real err.

if `act.Deps` is N size , the loop will create a N element `*joinError` link
I would like the `var depErrs []error` solution, simple and just append. finally got a `joinError` with N size `[]error`

I check it again see https://go.dev/play/p/PBgeIKmCrJt



github loss my review comment. 
<details>

<details>

ldez:
Your review is "pending" because you didn't submit it.

You misinterpreted the code, you must read it again:
1. The `errors.Unwrap` was already here.
2. `errors.Join` is on `depErrors` and not `dep.Err`
3. `errors.Unwrap(dep.Err)` is not nil

FYI, the 2 following snippets are the same identical thing:

```go
	var depErrors []error
	for _, dep := range act.Deps {
		if dep.Err != nil {
			depErrors = append(depErrors,  errors.Unwrap(dep.Err))
		}
	}
	if depErrors != nil {
		act.Err = errors.Join(depErrors...)
		return
	}
```

```go
	var depErrors error
	for _, dep := range act.Deps {
		if dep.Err != nil {
			depErrors = errors.Join(depErrors, errors.Unwrap(dep.Err))
		}
	}
	if depErrors != nil {
		act.Err = depErrors
		return
	}
```

pkg/goanalysis/runner_checker.go
ccoVeille: thanks for confirming @bombsimon

And good point the unwrap(nil) == nil

Let's see my where things are going with the follow up comments @alingse posted

ldez: The if `errors.Unwrap` is useless because `errors.Join` ignore `nil`, I will remove it.

alingse:
I explain my comment, I don't mean this once code logic. 

I means the `dep` is also a `action`. so it's error is return by another `errors.Join` logic.

steps

1. `action` has `Deps` and collect it's errors, and return a  final `joinError` err
2. `dep` is also a `action`, so `dep.Err` is might be `originErr` err or another `joinError`err 
3. `error.Unwrap` do not support `errors.Join`, it return nil on  `joinError`err 
5.  current `action` visit every `deps` error, and call `Unwrap` on it, and finally got nil
6. act.Err is nil, when some deps is err

I commit some error.log code in https://github.com/alingse/golangci-lint/tree/show-err-and-errlog

you can run it on some broken repo. (I create one  https://github.com/alingse/demos )

you will got logs like this.

```console
act Analyzer.Name spancheck
act depErrors <nil>
act dep.Err &errors.joinError{errs:[]error{(*errors.joinError)(0xc0000118a8), (*errors.errorString)(0xc000f6e360)}}
act errors.Unwrap(dep.Err) <nil>
act Package.Name dep2
act Analyzer.Name lostcancel
act depErrors &errors.joinError{errs:[]error{(*pkgerrors.IllTypedError)(0xc001b28038)}}
act dep.Err &errors.joinError{errs:[]error{(*errors.joinError)(0xc0000118a8), (*errors.errorString)(0xc000f6e360)}}
act errors.Unwrap(dep.Err) <nil>
act Package.Name dep2
act Analyzer.Name fact_purity
act depErrors <nil>
act someDepErr true
```

ldez:
Theoretically, yes, but in reality, `nil` on `act.err` doesn't change the behavior.

ldez:
I will revert this PR in favor of a better approach https://github.com/golangci/golangci-lint/pull/5885

PR #5894, opened 2025-06-25
feat: Clear analysis cache when running custom command

k-omotani:
## Overview
This PR adds automatic cache clearing when running the custom command to prevent outdated cache results from affecting the behavior.

## Changes
- Added cache directory removal in the runE method of custom.go
- Added appropriate log messages for successful/failed cache clearing

## Reason
When using the custom command, the linter itself is being modified with custom plugins or configurations. In this scenario, previously cached analysis results become invalid and can lead to incorrect linting results or unexpected behavior. 

By ensuring a fresh start with a clean cache directory for each custom build, we guarantee that the analysis results accurately reflect the current state of the modified linter, providing more consistent and reliable results.

### DIFF
--- a/pkg/commands/custom.go
+++ b/pkg/commands/custom.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/spf13/cobra"
 
+	"github.com/golangci/golangci-lint/v2/internal/cache"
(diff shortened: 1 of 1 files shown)


ldez:
The `custom` command doesn't run the analysis; this is a build command.

If you have a real problem, open an issue.

PR #5896, opened 2025-06-25
fix: unique version per custom build

ldez:
The version of the binary is used as a salt for the analysis cache.

This PR creates a unique version per custom build to invalidate the cache when modifying a custom linter without modifying the golangci-lint configuration.

The new version format is: `2.1.6-custom-gcl-JVlKbdkNNiWQLK19xGSNKdgLE7pWskPwSGwnFZtCnfc`.

I think behind the AI slop PR #5894 there is a real issue.


ldez:
In fact, this is only a "problem" when the command is run with a local path because the `go.mod` is not modified.

But my modification based on time will always invalidate the cache when using CI, and this should not happen.

So I need to change the approach.

ldez:
I found a solution by using the sumdb hash functions used by Go to compute the hash of local modules.

bombsimon (approved):

PR #5897, opened 2025-06-26
docs: point towards goimports when autofixing perfsprint

scop:
I recently found out that perfsprint autofixes can break code because it does not fix imports, and that not fixing them is actually a request from here: https://github.com/catenacyber/perfsprint/issues/19

I think it would be good to mention this in docs. Not sure what's the best way, but here's one.

### DIFF
--- a/assets/linters-info.json
+++ b/assets/linters-info.json
@@ -744,7 +744,7 @@
   },
   {
     "name": "perfsprint",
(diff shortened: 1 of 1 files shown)


ldez:
This file is generated, and regenerated only during the release.

There are several linters that suggest fixes that require extra actions, I'm not sure adding this inside the linter description is the right way.

PR #5901, opened 2025-06-28
docs:  add jsonschema step inside the new linters doc

uudashr:
Add guideline to modify `jsonschema/golangci.next.jsonschema.json` for new linter


ldez:
I consider updating the JSON Schema as optional because it requires different knowledge.

This is why it's not inside the doc and only an option inside the checklist.

uudashr:
Been thinking so because is not in the guidline. But when I run the test against the project, it failed.

PR #5902, opened 2025-06-28
dev: add suggested migration for linter configuration

ldez:
The goal is to display a suggested migration configuration when running golangci-lint.

~~The code related to the command `migrate` is ambiguous due to the difference between the default values of v4 and v5.~~
Edit: I removed the part related to the `migrate` command because I'm not fully happy with the approach.


ccoVeille:
Commit messages and PR could mention it's about wsl.

bombsimon (approved):
🚀

PR #5920, opened 2025-07-07
fix: skip type check in ExclusionPaths Processor

chenyahui:
In the scenario of exclusion.paths, we should skip all checks, including typecheck. For example, some legacy code.


ldez:
Hello, this is expected to not filter `typecheck` errors.

typecheck is not a linter, and errors are not reports.

- https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors
- https://golangci-lint.run/welcome/faq/#why-is-it-not-possible-to-skipignore-typecheck-errors

PR #5937, opened 2025-07-13
Add clearslice linter

zcross:
Adding [clearslice](https://github.com/zcross/clearslice) linter.

Motivation: Naive detection of slice reuse patterns that might lead to longer object liveness, potentially contributing to unnecessary longer or indefinite object retention. For a better-written (and visualized) explanation, see [Robust generic functions on slices](https://go.dev/blog/generic-slice-functions) from The Go Blog

Example:
```
type someHotPathProcessor struct {
  itemsToProcess []string
}

func processTick() {
  doWork(itemsToProcess)
  itemsToProcess = itemsToProcess[:0]
}
```

Recommended replacement:
```
type someHotPathProcessor struct {
  itemsToProcess []string
}

func processTick() {
  doWork(itemsToProcess)
  itemsToProcess = slices.Delete(itemsToProcess, 0, len(itemsToProcess))
}
```

This is my first golangci-lint linter contribution, so here are my attempts to minimize annoyance for users (ei.e,. false positives)
* Recognize when `x = x[:0]` pattern is immediately preceded by a `clear(x)` and do not fail lint
* Recognize when slice elements are strictly primitive types (or structs of strictly primitive fields, recursively)
* PR this in disabled-by-default state, allowing people to opt-in if interested
* Provide working fix option that uses simple same-line one-liner replacement
    *  (`x = x[:0]` -> `x = slices.Delete(x, 0, len(x))`

And some potential discussion points:
* `slices.Delete(x, 0, N)` is indeed O(N) due to the iteration+zeroing work done, just like `clear()`. 
    * This fix can provide peace of mind but one can imagine consciously dropping `//nolint` ignores in code that is explicitly aware of the risk and/or is actually structurally safe (beyond what the naive linter can recognize).
* Given the linter's fix suggestion only holds for Go ≥ 1.22, is it reasonable to associate that minimum version with this linter rather than attempting to provide multiline fix suggestions (e.g., preceding `clear()` when on an older Go)
* I've not written many analyzers outside of very simple ones. Would it be feasible to do escape analysis or something like it to determine when a slice allocated in a given scope does not escape that scope, even if said slice is used repeatedly in said scope with the length-only resizing pattern?

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -137,6 +137,7 @@ linters:
     - bidichk
     - bodyclose
     - canonicalheader
+    - clearslice
     - containedctx
     - contextcheck
     - copyloopvar
--- a/pkg/golinters/clearslice/testdata/fix/out/simple.go
+++ b/pkg/golinters/clearslice/testdata/fix/out/simple.go
@@ -1,0 +1,182 @@
+//golangcitest:args -Eclearslice
+//golangcitest:expected_exitcode 0
+package simple
+
+import (
+	"fmt"
+	"slices"
+)
+
+type ReferenceAliasTypeA string
+type ReferenceAliasTypeB *int
+type PrimitiveAliasType float64
+
+type FlatStruct struct {
+	Field1 int
+	Field2 float64
+	Field3 bool
+}
+
+type NestedFlatStruct struct {
+	ThingOne FlatStruct
+	ThingTwo FlatStruct
+}
+
+type FlatLookingStructWithSlice struct {
+	Flats []FlatStruct
+}
+
+type ReferenceHoldingStruct struct {
+	Field1 *FlatStruct
+}
+
+type ReferenceHoldingStructUnexported struct {
+	field1 *FlatStruct
+}
+
+type NestedRefStruct struct {
+	Refs []ReferenceHoldingStruct
+}
+
+type NestedRefStructUnexported struct {
+	refs []ReferenceHoldingStruct
+}
+
+type SomeThingWithFlatSliceMember struct {
+	Flats []FlatStruct
+}
+
+func _() {
+	// Safe: slice of primitive types (int)
+	s := []int{1, 2, 3}
+	s = s[:0]
+	fmt.Printf("%+v\n", s)
+}
+
+func _() {
+	// Safe: slice of primitive types (float)
+	s := []float64{1.1, 2.2, 3.3}
+	s = s[:0]
+	fmt.Printf("%+v\n", s)
+}
+
+func _() {
+	// Safe: slice of struct free of reference types
+	s := []FlatStruct{{1, 1.1, true}, {2, 2.2, false}}
+	s = s[:0]
+	fmt.Printf("%+v\n", s)
+}
+
+func _() {
+	// Safe: slice of struct free of reference types
+	s := []NestedFlatStruct{
+		{ThingOne: FlatStruct{1, 1.1, true}, ThingTwo: FlatStruct{2, 2.2, false}},
+	}
+	s = s[:0]
+	fmt.Printf("%+v\n", s)
+}
+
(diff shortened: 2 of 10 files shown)


ldez:

- [x] The CLA must be signed

### Pull Request Description

### Linter

- [x] It must have a valid license (AGPL is not allowed), and the file must contain the required information by the license, ex: author, year, etc.
- [x] It must use Go version <= 1.23.0
- [x] It must use [`go/analysis`](https://golangci-lint.run/contributing/new-linters/).

### Others Requirements

- [x] The `.golangci.yml` of golangci-lint itself must not be edited and the linter must not be added to this file.
- [x] The linters must be sorted in the alphabetical order (case-insensitive) in the `lintersdb/builder_linter.go` and `.golangci.next.reference.yml`.

### Recommendations

- [x] The file `jsonschema/golangci.next.jsonschema.json` should be updated.
- [x] A tag should never be recreated.
- [x] use `main` as the default branch name.

---

.golangci.next.reference.yml
```diff
@@ -137,6 +137,7 @@ linters:
     - bidichk
     - bodyclose
     - canonicalheader
+    - clearslice
```
zcross, line 140: Disabled by default to minimize adoption annoyance (and anticipating some iteration/improvement over time)

ldez:
Hello,

pkg/golinters/clearslice/testdata/fix/out/simple.go
```diff
@@ -0,0 +1,182 @@
...
+	s = s[:0]
+	fmt.Printf("%+v\n", s)
+}
+
+func _() {
+	// Unsafe: slice of struct that has a slice field (even if slice elements do not contain references)
+	s := []FlatLookingStructWithSlice{
+		{Flats: []FlatStruct{{1, 1.1, true}, {2, 2.2, false}}},
+	}
+	s = slices.Delete(s, 0, len(s))
```
zcross, line 84: BTW, the linter fix suggestion uses `len(s)` instead of `cap(s)` along the lines of reasoning that, given this is the sole driver of slice "shrinking," it inductively ends up being equivalent to always clearing up to capacity without the extra time cost (on wasted iterations to zero items past `len`)

.golangci.next.reference.yml
ldez: The "default" management is more complex than the `standard` set.

A lot of users use the set `all` and so your linter will be enabled "by default".

So this is not a way to handle false positives.

And this is the documentation and not real configuration, so you must add your analyzer in the two sections.

zcross:
> Inside golangci-lint we differentiate 2 types of analyzers:
> 
>     * the linters (find problems with nearly 0 false positives)
> 
>     * the detectors (detect potential problems, and so a majority of false positives)
> 
> 
> The detectors are not allowed because of the number of false positives.
> 
> My first impression is that your analyzer is a detector and not a linter.

Ah, ok, TIL and thanks for the quick feedback! I'll close this if getting to nearly 0 false positives is a requirement, because it seems very unlikely (without a much "smarter" analysis that is probably beyond what I'm planning on maintaining). 

Longer form reasoning: while replacing the specific pattern with the recommended fix should always be _logically_ correct, it may be tricky to generalize without introducing _performance_ regressions in code paths that are structurally set up to regularly and completely "clobber" a slice's contents with new items. There's an easy to imagine long tail of cases where it's totally reasonable to only alter `len` in O(1) time, and it's probably going to be hard to avoid missing such cases in a simple analyzer. My mental model had been that this linter could drive down the overall usage of the "risky" pattern, with folks hypothetically using `//nolint:clearslice` to signal explicit intent on resizing operations and the linter providing some forcing function to reason about this more explicitly. Based on the detector/linter distinction, it doesn't sound like that mental model is actually in line with the idea of linters for sure.

PR #5941, opened 2025-07-16
Add safebigint linter

winder:
[safebigint](https://github.com/winder/safebigint) checks for potential misuse of the `math/big` package, in particular the `Uint64()` and `Int64()` functions which silently truncate numbers which don't fit in the returned type.

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -105,6 +105,7 @@ linters:
     - recvcheck
     - revive
     - rowserrcheck
+    - safebigint
     - sloglint
     - spancheck
     - sqlclosecheck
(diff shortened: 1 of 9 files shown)


ldez:
Hello,

If I'm wrong, please add a comment to explain why your analyzer is not a detector.

winder:
Thank you @ldez for the quick reply. My primary use case is to avoid using `Uint64()` and `Int64()` entirely. In that sense there would be 0 false positives. I could imagine other users might have a different policy for the `math/big` package and they shouldn't use it.

The other check in `safebigint` would have many false positives. Do you think this package would be suitable for golangci-lint if I were to limit it to the case without false positives?

I'll look into creating a plugin.

ldez:
:thinking: your context seems very precise and personal, so it feels like a niche style.

Maybe you can use [`forbidigo`](https://golangci-lint.run/usage/linters/#forbidigo) to detect `Uint64()` and `Int64()`.

Another possibility is to use `ruleguard` rules for [`gocritic`](https://golangci-lint.run/usage/linters/#gocritic)

PR #5962, opened 2025-07-28
fix: improve HTTP download error handling for `wget` and `curl`

tjasko:
### Summary

This PR improves the error handling and consistency between `http_download_curl` and `http_download_wget` in `install.sh`.

### Changes

* Updated `http_download_curl` to use `log_err` for non-200 HTTP status codes instead of `log_debug`.
* Enhanced `http_download_wget` to:
	 * Capture and evaluate the HTTP status code using `wget --server-response`.
		* Return an exit status of `1` for non-200 HTTP responses, matching the behavior of `http_download_curl`.
		* Log errors via `log_err` for better visibility.
	* Simplified `http_download` to directly call `http_download_wget` after its availability check.

### Why

Previously, `http_download_wget` did not validate HTTP response codes, which led to misleading behavior. For example, when GitHub returned an HTTP 503, the installer incorrectly reported that the requested version did not exist:

```sh
$ ./install.sh v1.64.8
golangci/golangci-lint info checking GitHub for tag 'v1.64.8'
golangci/golangci-lint crit unable to find 'v1.64.8' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details
```

In this case, the tag `v1.64.8` did exist, but the underlying issue was a GitHub service error returning a `503`. By properly checking and logging HTTP response codes, this update makes such scenarios clearer to users and avoids misleading "version not found" messages.


install.sh
```diff
@@ -311,10 +311,15 @@ http_download_wget() {
...
   if [ -z "$header" ]; then
-    wget -q -O "$local_file" "$source_url"
+    code=$(wget --server-response --quiet -O "$local_file" "$source_url" 2>&1 | awk '/^  HTTP/{print $2}' | tail -n1)
   else
-    wget -q --header "$header" -O "$local_file" "$source_url"
+    code=$(wget --server-response --quiet --header "$header" -O "$local_file" "$source_url" 2>&1 | awk '/^  HTTP/{print $2}' | tail -n1)
   fi
+  if [ "$code" != "200" ]; then
+    log_err "http_download_wget received HTTP status $code"
+    return 1
```
ldez, line 328: this doesn't work in all cases, and it can hide other errors.

```console
$ wget --server-response --quiet -O foo http://example.g 2>&1 | awk '/^  HTTP/{print $2}' | tail -n1
$ echo $?
0
$ wget --server-response --quiet -O foo http://example.g
$ echo $?
4
$ wget --server-response -O foo http://example.g
--2025-07-28 23:19:59--  http://example.g/
Resolving example.g (example.g)... failed: Name or service not known.
wget: unable to resolve host address 'example.g'
```

tjasko: Ah, good catch. Will look into it soon enough. Thanks for the quick review!

tjasko: Fixed in a8f43c243e5e8169d08f4c145ac74317c765e7f2

ldez (approved):
LGTM

Issue #4683, opened 2024-04-28
docs: support sorting / filtering linters table


derekperkins:
### Your feature request related to a problem? Please describe

I regularly visit the documentation site to see if there have been any new linters added that I want to introduce to our codebase. The problem I run into is that the list is sorted alphabetically, so it's difficult to see just changes since the last version of golangci-lint.
https://golangci-lint.run/usage/linters/#disabled-by-default

### Describe the solution you'd like

Add filtering / sorting to the table

### Describe alternatives you've considered

I have copy / pasted the table into Excel

ccoVeille:
I also faced problems you describe.

I used to look for Ctrl+F on the linters page with version to see what was added.

Lately, I found the changelog page, and read it. It's its purpose after all.

https://golangci-lint.run/product/changelog/#changelog

I'm unsure there is a need to sort the linters on the page, but why not.

But there is a need to understand what was added from the linters page. I would say adding a visible message + link to the changelog page would be a good thing.

ldez:
Hello,

this is in our plan (filtering).

PR #5965, opened 2025-08-01
docs: new website

ldez:
I tried a lot of Hugo themes, but most of them don't fit our needs.

I found [hextra](https://github.com/imfing/hextra): the features and the style are "extra" :smile_cat:

There are some constraints for the search feature, so I split some pages.

I also improved the page contents and adapted the scripts.

I think I will replace the scripts in the future with Hugo partials or shortcodes.

Note: URLs are different from the previous site.

<details>
<summary>URLs</summary>

TLDR: adds the path prefix `/docs`

| OLD                                                  | NEW                                                          |
|------------------------------------------------------|--------------------------------------------------------------|
| https://golangci-lint.run/welcome/install/           | https://golangci-lint.run/docs/welcome/install/              |
| https://golangci-lint.run/welcome/quick-start/       | https://golangci-lint.run/docs/welcome/quick-start/          |
| https://golangci-lint.run/welcome/integrations/      | https://golangci-lint.run/docs/welcome/integrations/         |
| https://golangci-lint.run/welcome/faq/               | https://golangci-lint.run/docs/welcome/faq/                  |
| https://golangci-lint.run/usage/configuration/       | https://golangci-lint.run/docs/configuration/                |
| https://golangci-lint.run/usage/linters/             | https://golangci-lint.run/docs/linters/                      |
| https://golangci-lint.run/usage/linters/#xxx         | https://golangci-lint.run/docs/linters/configuration/#xxx    |
| https://golangci-lint.run/usage/false-positives/     | https://golangci-lint.run/docs/linters/false-positives/      |
| https://golangci-lint.run/usage/formatters/          | https://golangci-lint.run/docs/formatters/                   |
| https://golangci-lint.run/usage/formatters/#xxx      | https://golangci-lint.run/docs/formatters/configuration/#xxx |
| https://golangci-lint.run/product/thanks/            | https://golangci-lint.run/docs/product/thanks/               |
| https://golangci-lint.run/product/changelog/         | https://golangci-lint.run/docs/product/changelog/            |
| https://golangci-lint.run/product/migration-guide/   | https://golangci-lint.run/docs/product/migration-guide/      |
| https://golangci-lint.run/product/roadmap/           | https://golangci-lint.run/docs/product/roadmap/              |
| https://golangci-lint.run/contributing/quick-start/  | https://golangci-lint.run/docs/contributing/                 |
| https://golangci-lint.run/contributing/workflow/     | https://golangci-lint.run/docs/contributing/workflow/        |
[...]

PR #5986, opened 2025-08-09
dev: new way to fill the documentation

ldez:
The previous method for filling data inside the documentation was cumbersome.
It was modifying the Markdown files directly by replacing placeholders with the actual data.

Also, the script to update the documentation was not able to run when the placeholders were already filled.
You were forced to reset the changes made by the script before modifying the files.
But when you modified the same files as the script and reverted the changes, all the changes were reverted.

So it was a pain.

The placeholders are removed from the files; instead, I replaced placeholders with shortcodes and partials that use the data from the dedicated files.

Now, modifying and running locally the documentation is easier.

There is more information about the new way to fill out the documentation on the dedicated page inside the website.

It was not easy to create this new way to fill in the documentation.
The Hugo documentation is complex.
I found bugs inside the theme, and I had to fix them.

There are also a few improvements:
- Better linter's authors section on the "thanks" page
- Basic filtering inside the "linters" and "formatters" pages
- Less aggressive image size for GitHub Actions illustrations inside the "installation" page
- Add the description of the commands inside the "Command line" page

Fixes https://github.com/golangci/golangci-lint/issues/4683

Follows #5965


docs/content/docs/contributing/website.md
```diff
@@ -30,14 +49,24 @@ GitHub deploys the website to production after merging anything to a `main` bran
...
+# (in the root of the repository)
 make docs_serve
-# OR (if inside the docs/ folder)
+```
+
+or
+
+```bash
+# (in the root of the repository)
+make make website_expand_templates
```
bombsimon:
```suggestion
make website_expand_templates
```

docs/layouts/_partials/clickable-badge.html
```diff
@@ -0,0 +1,39 @@
+{{- /* Modified version of https://github.com/imfing/hextra/blob/v0.9.7/layouts/partials/shortcodes/badge.html */ -}}
```
bombsimon, line 1: Can we add a not somewhere either in the docs of templating or here in `_partials` why we have a hard copy of these and what the changes are? 

Didn't find any issues or PRs upstream so I guess the changes are not something that is expected to be upstreamed?

docs/hugo.yaml
```diff
@@ -66,7 +66,7 @@ menu:
         icon: github
   sidebar:
     - identifier: donate
-      name: "Donate"
+      name: "Supporting Us"
```
bombsimon: I think the imperative form would work better here since it's just linking to where you can support and feels more natural and common to me


```suggestion
      name: "Support Us"
```

docs/layouts/_partials/clickable-badge.html
ldez: It's more of a hack than something "upstreamable".
If I want to upstream it, I need to rethink the approach.

bombsimon (approved):

alexandear (approved):

PR #5989, opened 2025-08-11
dev: improve search index

ldez:
Parse all headings to retrieve the precise section, not the parent, when heading > h2.

I submitted the PR to the upstream with this approach https://github.com/imfing/hextra/pull/740


ldez:
The theme is using zeroVer like Hugo, the current main branch contains breaking changes (i.e. Tailwind 4), also this project uses milestones as release scope; the v0.10.0 is not ready.

PR #6000, opened 2025-08-13
update upstream tags for golang dockerfile

taigrr:
I noticed the CI for goreleaser failed.

I think golang uses a slightly different architecture tag for arm64 now, calling it arm64v8.

Hopefully this is the fix!

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -295,7 +295,7 @@ dockers:
       - '--label=org.opencontainers.image.created={{.Date}}'
       - '--label=org.opencontainers.image.revision={{.FullCommit}}'
       - '--label=org.opencontainers.image.version={{.Version}}'
-      - '--platform=linux/arm64'
+      - '--platform=linux/arm64v8'
 
   - use: buildx
     goos: linux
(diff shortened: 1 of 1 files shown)


ldez:
https://github.com/golangci/golangci-lint/pull/5872#issuecomment-3185863078

taigrr:
It's specifically the arm64 image that's missing, it's been removed from the list of supported architectures and replaced with arm64v8, which is available now, hence my PR

Unless I am misreading the situation! Maybe they always delete the docker repo before changing tags?

https://hub.docker.com/r/arm64v8/golang/

vs 

https://hub.docker.com/r/arm64/golang/

ldez:
I don't want to create a bug fix release just because people don't want to wait a few times for the images to be available.

taigrr:
Not trying to be pushy. Maybe I just misunderstood, I thought this was a tag renaming not images not available yet, hence the links above. I'm happy to wait. I probably misread the context. Thanks!

Edit: I scrolled up and read the rest of the conversation and now completely understand, my PR is not necessary or even correct. Absolutely just need to wait. Thanks again, sorry to bother!

Issue #6005, opened 2025-08-14
golangci-lint run ./... only checks formatting, it doesn’t actually format the code.


Zacama:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/docs/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/docs/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

Official binary

### Description of the problem

<img width="1109" height="337" alt="Image" src="https://github.com/user-attachments/assets/c8168924-2141-4eaf-a9b3-9f8cfbdfff3a" />
The documentation says that `golangci-lint run ./...` will automatically run `golangci-lint fmt`, but it seems that `golangci-lint run ./...` only checks formatting issues according to the defined formatters and doesn’t actually format the files. In contrast, `golangci-lint fmt` directly formats the files.

### Version of golangci-lint

<details>

```console
 ~  golangci-lint --version                                                                                                                                                                                   ✔
golangci-lint has version 2.4.0 built with go1.25.0 from 43d03392 on 2025-08-13T23:36:29Z
```

</details>

### Configuration

<details>

```console
# paste configuration file or CLI flags here
```

</details>

### Go environment

<details>

```console
$ go version && go env
# paste output here
```

</details>

### Verbose output of running

<details>

```console
$ golangci-lint cache clean
$ golangci-lint run -v
# paste output here
```

</details>

### A minimal reproducible example or link to a public repository

<details>

```go
// add your code here
```

</details>

### Validation

- [x] Yes, I've included all information above (version, config, etc.).

### Supporter

- [ ] I am a sponsor/backer through [GitHub](https://github.com/sponsors/golangci/) or [OpenCollective](https://opencollective.com/golangci-lint)

ldez:
Hello,

The documentation doesn't say:
> `golangci-lint run ./...` will automatically run `golangci-lint fmt`

[...]

PR #6006, opened 2025-08-14
docs: improve wording about formatter as linter

ldez:
Improve the wording because it seems confusing.

Related to #6005

### DIFF
--- a/docs/content/docs/configuration/cli.md
+++ b/docs/content/docs/configuration/cli.md
@@ -14,9 +14,7 @@ weight: 1
 {{< /cards >}}
 
 > [!NOTE]
-> The formatters explicitly defined in the configuration section [`formatters`](/docs/configuration/file/#formatters-configuration) are automatically run by the command [`golangci-lint run`](/docs/configuration/cli/#run).
-> 
-> You don't need to run [`golangci-lint fmt`](/docs/configuration/cli/#fmt) before [`golangci-lint run`](/docs/configuration/cli/#run)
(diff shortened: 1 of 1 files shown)


docs/content/docs/configuration/cli.md
```diff
@@ -14,9 +14,7 @@ weight: 1
 {{< /cards >}}
 
 > [!NOTE]
-> The formatters explicitly defined in the configuration section [`formatters`](/docs/configuration/file/#formatters-configuration) are automatically run by the command [`golangci-lint run`](/docs/configuration/cli/#run).
-> 
-> You don't need to run [`golangci-lint fmt`](/docs/configuration/cli/#fmt) before [`golangci-lint run`](/docs/configuration/cli/#run)
+> The formatters explicitly defined in the configuration section [`formatters`](/docs/configuration/file/#formatters-configuration) are automatically run (as linter) by the command [`golangci-lint run`](/docs/configuration/cli/#run).
```
bombsimon: Can we mention something around the difference between checking if the code is formatted vs. actually formatting it? Thinking similar to `ruff check` or `prettier --check`?

And also show the difference between _only_ formatting the code with formatters vs. with `--fix` that also applies fixes from linters.

Maybe not perfect but you get the gist:

```suggestion
> The [`golangci-lint run`](/docs/configuration/cli/#run) command executes all enabled linters and the formatters defined in [`formatters`](/docs/configuration/file/#formatters-configuration) to check whether code is correctly formatted. It does not modify the code.  
> To reformat code, use `golangci-lint fmt`. To apply both formatting and linter fixes, use `golangci-lint run --fix`.
```

PR #6019, opened 2025-08-19
result/processors: make exclusion_paths apply to typecheck issues and…

135yshr:
… ensure robust path matching

- Use filterIssuesUnsafe in exclusion_paths to honor explicit path exclusions for typecheck reports
- Fallback to absolute/WD-relative paths in path_relativity when Rel() fails or yields empty, so exclusions can match reliably
- Update docs to clarify difference between linters.exclusions.paths and formatters.exclusions.paths

### DIFF
--- a/docs/content/docs/linters/false-positives.md
+++ b/docs/content/docs/linters/false-positives.md
@@ -113,6 +113,28 @@ linters:
       - path/to/a/dir/
 ```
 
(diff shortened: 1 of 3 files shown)


ldez:
The PR has been closed, but I want to avoid any other PRs with the same content.

1. The filtering of typecheck errors cannot work: https://golangci-lint.run/docs/welcome/faq/#why-is-it-not-possible-to-skipignore-typecheck-errors
2. The documentation about exclusions is confusing.
3. The handling of the path is not required or suitable; there are no real use cases.
4. You should never create a PR with multiple topics.
5. Please check issues and documentation before working on a PR.

PR #6026, opened 2025-08-22
Refactor to use reflect.TypeFor

cuiweixie:


ldez:
Thank you, but this is not needed.

PR #6030, opened 2025-08-26
Update ineffassign and add new check-escaping-errors config option

nvx:
This exposes the `check-escaping-errors` flag in the `ineffassign` linter.

Note the most recent tagged version of this linter is quite old with a few commits since then without a tagged release, so I don't think dependabot would pick up non-tagged versions.

If this is a showstopper I can raise an issue in the ineffassign repo asking for a new tag to be created but the other changes will still be needed to expose the new config flag.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -56,7 +56,7 @@ require (
 	github.com/golangci/revgrep v0.8.0
 	github.com/golangci/swaggoswag v0.0.0-20250504205917-77f2aca3143e
 	github.com/golangci/unconvert v0.0.0-20250410112200-a129a6e6413e
-	github.com/gordonklaus/ineffassign v0.1.0
+	github.com/gordonklaus/ineffassign v0.1.1-0.20250824073611-44548fda05c1
 	github.com/gostaticanalysis/forcetypeassert v0.2.0
 	github.com/gostaticanalysis/nilerr v0.1.1
 	github.com/hashicorp/go-version v1.7.0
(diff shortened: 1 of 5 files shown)


ldez:
Hello,

https://github.com/golangci/golangci-lint/blob/0515ab20087fa7a35df1ead71d3aed4cfdedcbcc/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L3-L8

we will not update to a pseudo version because it will prevent the dependabot from handling this linter in the future.

Also, the update of implementation inside golangci-lint is not right.

nvx:
> we will not update to a pseudo version because it will prevent the dependabot from handling this linter in the future.

Noted, I will open an issue in the ineffassign repo to get a tag added. I did see that part of the PR template but note it does not mention anything about pseudo versions and there are a number of linters in the go.mod file pinned to pseudo versions already so it would be worth updating that to be more clear.

> Also, the update of implementation inside golangci-lint is not right.

Can you elaborate on this? After getting a tag created on ineffassign repo config is still needed to expose the new flag but this comment does not provide any indication of what is needed on that front.

ldez:
I think the sentence is already clear enough:

> No pull requests to update a linter will be accepted unless you are the author of the linter AND specific changes are required.

nvx:
A tag has been made now so I assume dependabot will pick this up next week when it runs.

> Also, the update of implementation inside golangci-lint is not right.

I'm happy to resubmit this PR without the dependency change, but I'm not sure what this is referring to specifically. I was able to set the config option and the new behaviour worked fine.

If changes are needed to how the config is exposed I'm happy to make them, I copied how another linter that uses flags was done but if I've missed something I'm happy to clean that up.

ldez:
This is not how the golangci-lint process works: the PR of the dependabot is used to update the implementation of the linter, so no need for a PR after the dependabot PR.

But if you want to do something, I recommend adding test inside the [ineffassign repository](https://github.com/gordonklaus/ineffassign) about the new option.
IMHO, this option doesn't work.

---

>  there are a number of linters in the go.mod file pinned to pseudo versions already
> ...
> (despite this being the case for a number of other linters today) 

This is factually false.

The only linters that used a pseudo-version are:
- forks inside the golangci organization (for various reasons): `dupl`, `gofmt`, `golines`, `swaggoswag`, `unconvert`
- linters that historically don't have tags: `bodyclose`, `go-err113`, `unparam`

8 linters among hundreds of linters, and not at random.

Issue #828, opened 2019-10-20
Add support for multiple subdirs each with their own go.mod


fkautz:
Thank you for creating the issue!

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```bash
$ golangci-lint --version
golangci-lint has version 1.20.1 built from 849044b on 2019-10-15T19:11:27Z
```

</details>

<details><summary>Config file</summary>

```bash
$ cat .golangci.yml
---
run:
  concurrency: 6
  deadline: 15m
  issues-exit-code: 1
  tests: true
  skip-dirs:
    - build
    - conf
    - controlplane/scripts
    - forwarder/vppagent/build
    - forwarder/vppagent/conf
    - forwarder/scripts
    - deployments
    - docks
    - scripts
linters-settings:
  errcheck:
    check-type-assertions: false
    check-blank: false
  govet:
    check-shadowing: true
    settings:
      printf:
        funcs:
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Infof
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Warnf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Errorf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
  golint:
    min-confidence: 0.8
  goimports:
    local-prefixes: github.com/networkservicemesh/networkservicemesh
  gocyclo:
    min-complexity: 20
  maligned:
    suggest-new: true
  dupl:
    threshold: 100
  goconst:
    min-len: 3
    min-occurrences: 3
  depguard:
    list-type: blacklist
    include-go-root: false
    packages:
      - github.com/davecgh/go-spew/spew
  misspell:
    locale: US
  unused:
    check-exported: true
  unparam:
    check-exported: true
  nakedret:
    max-func-lines: 30
  prealloc:
    simple: true
    range-loops: true
    for-loops: false
  gocritic:
    enabled-checks:
      - rangeValCopy
      - boolExprSimplify
      - badCond
      - methodExprCall
      - paramTypeCombine
      - ptrToRefParam
      - rangeExprCopy
      - captLocal
      - caseOrder
      - defaultCaseOrder
      - dupBranchBody
      - dupSubExpr
[...]

PR #6057, opened 2025-09-05
support running against other directories

ecordell:
I found it frustrating that you couldn't run:
```sh
golangci-lint run --fix ../mydir
```

This PR adds support for this, by checking if `../mydir` is part of a go module, finding the associated golangci config and running the configured linters.

For completeness, this PR also validates that all passed paths are part of the same module, and provides a nice error if not:

Example running against multiple modules:
```sh
$ golangci-lint run module1/subdir module2/subdir
ERRO Running error: context loading failed: failed to load packages: multiple Go modules detected: [/private/tmp/multi-subdir-test/module1 /private/tmp/multi-subdir-test/module2]

Multi-module analysis is not supported. Each module should be analyzed separately:
  golangci-lint run /private/tmp/multi-subdir-test/module1
  golangci-lint run /private/tmp/multi-subdir-test/module2
```

Example running against multiple packages in the same module: 
```sh
$ golangci-lint run ../module/bar ../module/baz
../../../private/tmp/sibling-test/module/bar/main.go:1: : # testmodule/bar
bar/main.go:5:3: "os" imported and not used (typecheck)
package bar
../../../private/tmp/sibling-test/module/baz/main.go:1: : # testmodule/baz
baz/main.go:5:3: "os" imported and not used (typecheck)
package baz
2 issues:
* typecheck: 2
```

This doesn't directly address https://github.com/golangci/golangci-lint/issues/828 but does make solving it a bit simpler (no need to change directories).

### DIFF
--- a/pkg/config/base_loader.go
+++ b/pkg/config/base_loader.go
@@ -130,8 +130,9 @@ func (l *BaseLoader) getConfigSearchPaths() []string {
 	}
 
 	// find all dirs from it up to the root
-	searchPaths := []string{"./"}
+	searchPaths := []string{}
 
+	// Add the target directory and its parents first (highest priority)
 	for {
 		searchPaths = append(searchPaths, currentDir)
 
@@ -141,6 +142,15 @@ func (l *BaseLoader) getConfigSearchPaths() []string {
 		}
 
 		currentDir = parent
+	}
+
+	// Add current working directory if it's not already included and we haven't found a config yet
+	cwd, err := os.Getwd()
+	if err == nil {
+		absCwd, err := filepath.Abs(cwd)
+		if err == nil && !slices.Contains(searchPaths, absCwd) {
+			searchPaths = append(searchPaths, "./")
+		}
 	}
 
 	// find home directory for global config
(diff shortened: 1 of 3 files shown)


ldez:
I don't understand your problem.

Why do you want to run golangci-lint on a parent or sibling directories of the location you are running it?

ldez:
1. I don't understand the need
2. It can create bugs/regressions
3. It will have a notable impact on performance
4. It breaks our tests.

I will decline this PR I recommend opening an issue to explain your problem, and we will discuss it inside this issue.

mauriciabad:
@ldez 

First reason and VERY important: Because if we have a monorepo, the VS Code extension crashes.

Second reason: the [documentation](https://golangci-lint.run/docs/welcome/quick-start/) says: 
   > You can choose which directories or files to analyze: `golangci-lint run dir1 dir2/...`

Which is incorrect. That command only works if the working directory is a go module. 

As a new user, I expected `golangci-lint` to run wherever I want in my code. It should check if the file/path I pass as argument is inside any go module recursively and use those configurations automatically. Like ESLint does.

Issue #5437, opened 2025-02-18
Intermittent errors when re-running linters (eventually succeeding, seems to be cache-related)


vegas503:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

Docker

### Description of the problem

When running inside Docker, errors that are reported are changing among runs, eventually reducing to zero.

Initial errors are somewhat absurd: first it reports names that are clearly known as unknown, then it throws some errors about "no export data", then it finally succeeds. Re-running in new container from scratch reproduces the problem over and over.

Sometimes (very rarely) the check succeeds at first try.

Example:
```console
$ docker run --rm -it -v $PWD:/code --entrypoint sh golangci/golangci-lint:v1.64.5-alpine
$ cd /code/master/
$ golangci-lint run
internal/infrastructure/database/database.go:85:14: a.Transaction undefined (type *Adapter has no field or method Transaction) (typecheck)
        if err := a.Transaction(func(tx *gorm.DB) error {
                    ^
pkg/auth/jwt2.go:83:29: claims.Audience undefined (type *Claims has no field or method Audience) (typecheck)
        for _, aud := range claims.Audience {
                                   ^
pkg/auth/jwt2.go:87:30: claims.Subject undefined (type *Claims has no field or method Subject) (typecheck)
        return j.GenerateJWT(claims.Subject, claims.Name, audiences, ttl)
                                    ^
pkg/auth/jwt2.go:131:40: claims.Subject undefined (type *Claims has no field or method Subject) (typecheck)
        subjectUUID, err := uuid.Parse(claims.Subject)
                                              ^
$ golangci-lint cache status
Dir: /root/.cache/golangci-lint
Size: 453.3KiB
$ golangci-lint run
[...]

Issue #5428, opened 2025-02-17
failed to load package spdy after upgrading to go 1.24


ocofaigh:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

After upgrading to go 1.24, we are seeing the following error when running version 1.64.5 of golangci-lint:

```
ERRO Running error: can't run linter goanalysis_metalinter
buildir: failed to load package spdy: could not load export data: no export data for "k8s.io/client-go/transport/spdy" 
```

### Version of golangci-lint

<details>

```console
% golangci-lint --version
golangci-lint has version 1.64.5 built with go1.24.0 from 0a603e49 on 2025-02-13T21:19:55Z
```

</details>

### Configuration

<details>

```console
GL_DEBUG="loader,env,gocritic" golangci-lint run --fix --timeout=15m -v
```

</details>

### Go environment

<details>

```console
% go version && go env
go version go1.24.0 darwin/arm64
AR='ar'
CC='cc'
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_ENABLED='1'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
CXX='c++'
GCCGO='gccgo'
GO111MODULE=''
GOARCH='arm64'
GOARM64='v8.0'
GOAUTH='netrc'
GOBIN=''
GOCACHE='/Users/conall/Library/Caches/go-build'
GOCACHEPROG=''
GODEBUG=''
GOENV='/Users/conall/Library/Application Support/go/env'
GOEXE=''
GOEXPERIMENT=''
GOFIPS140='off'
GOFLAGS=''
[...]

Issue #5184, opened 2024-12-01
Running error: can't run linter goanalysis_metalinter


QZAiXH:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [X] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

```
golangci-lint run --fix
WARN [runner] Can't run linter goanalysis_metalinter: inspect: failed to load package app: could not load export data: no export data for "fyne.io/fyne/v2/app"
ERRO Running error: can't run linter goanalysis_metalinter
inspect: failed to load package app: could not load export data: no export data for "fyne.io/fyne/v2/app"
```
**I skipped the directory where that file is located and I still get the same error.**

### Version of golangci-lint

```
golangci-lint --version
golangci-lint has version v1.62.2 built with go1.23.2 from (unknown, modified: ?, mod sum: "h1:b8K5K9PN+rZN1+mKLtsZHz2XXS9aYKzQ9i25x3Qnxxw=") on (unknown)
```

### Configuration

<details>

```yml
run:
  timeout: 5m

linters:
  disable-all: true
  enable:
    - gofumpt
    - govet
    - misspell
    - revive
    - lll

linters-settings:
  gofumpt:
    extra-rules: true

  lll:
    line-length: 160
    tab-width: 1

  govet:
    # Matches what we use in corp as of 2023-12-07
    enable:
      - asmdecl
      - assign
      - atomic
      - bools
      - buildtag
      - cgocall
      - copylocks
      - deepequalerrors
      - errorsas
      - framepointer
      - httpresponse
      - ifaceassert
      - loopclosure
      - lostcancel
      - nilfunc
      - nilness
      - printf
      - reflectvaluecompare
      - shift
      - sigchanyzer
      - sortslice
      - stdmethods
      - stringintconv
      - structtag
      - testinggoroutine
      - tests
      - unmarshal
[...]

Issue #5051, opened 2024-10-01
Can't run linter goanalysis_metalinter: inspect: failed to load package jsonpatch: could not load export data: no export data for \"github.com/evanphx/json-patch/v5\"


lucastheisen:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [X] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [X] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [X] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### Description of the problem

When running as part of a [gitlab action](https://github.com/pastdev/clconf/actions/runs/11132412282/job/30936179345?pr=52), the step fails with:

```console
  Running [/home/runner/golangci-lint-1.61.0-linux-amd64/golangci-lint run] in [/home/runner/work/clconf/clconf] ...
  Received 232188 of 232188 (100.0%), 0.2 MBs/sec
  level=warning msg="[runner] Can't run linter goanalysis_metalinter: inspect: failed to load package jsonpatch: could not load export data: no export data for \"github.com/evanphx/json-patch/v5\""
  level=error msg="Running error: can't run linter goanalysis_metalinter\ninspect: failed to load package jsonpatch: could not load export data: no export data for \"github.com/evanphx/json-patch/v5\""
  
  Error: golangci-lint exit with code 3
  Ran golangci-lint in 6975ms
```

When running local, i cannot reproduce this error:

```console
ltheisen@ltop ~/egit/pastdev-clconf
$ ~/tmp/glcil/golangci-lint-1.61.0-linux-amd64/golangci-lint run --max-same-issues 100000 --max-issues-per-linter 1000000

ltheisen@ltop ~/egit/pastdev-clconf
$ echo $?
0
```

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
# Paste output here
```

I cannot run this command directly as it is part of the gitlab action, but the configuration explicitly says:

```yaml
  lint:
    name: lint
    runs-on: ubuntu-latest
    steps:
[...]

PR #6058, opened 2025-09-05
pkg/goanalysis: expose underlying errors

### LINKED ISSUE #6056: "no export data": clearer error message
### Welcome

go install

"no export data" frequently obscures a more tractable error message - for example a type check failure or other build failure. Many issues reported on Github have this error message attached.

If it is possible for goanalysis_metalinter to expose the underlying error, that would be very helpful. Failing that, if the metalinter could detect if the error message contains "no export data", and give a hint - "are you sure the project can be built with no errors?" - that would also help. 

We could just keep presenting unclear error messages to end users 

### Additional context

_No response_

### Supporter

kevinburkesegment:
If compilation or building of a Go package fails, export data will not be available. Previously, we would mask the underlying failure and display only a message that "no export data" was available. This has led to lots of confusion among golangci-lint users (and users at my company, which is how I discovered this problem).

#3996, #4630, #4912, #4964, #5037, #5051, #5184, #5428, #5437

In addition to numerous discussions. The only common thread between the responses to these discussions, is that the "no export data" error message is obscuring whatever the underlying problem is, and making it more difficult to troubleshoot.

#2489, #2191, #2751, #3363, #4829, ~~#4880~~

This fix simply walks the import tree and appends any found error messages to the "no export data" error message. This will make it much clearer for end users to discover the problem that is causing the linter to fail.

Example output:

    ERRO Running error: can't run linter goanalysis_metalinter
    inspect: failed to load package errs: could not load export data: no export data for "github.com/segmentio/ctlstore/pkg/errs" because of error in imported package(s): error in package "github.com/segmentio/stats/v5": # github.com/segmentio/stats/v5
    vendor/github.com/segmentio/stats/v5/tag.go:38:35: in call to slices.IsSortedFunc, type func(a Tag, b Tag) int of tagCompare does not match inferred type func(a Tag, b Tag) bool for func(a E, b E) bool

Fixes #6056

### DIFF
--- a/pkg/goanalysis/runner_loadingpackage.go
+++ b/pkg/goanalysis/runner_loadingpackage.go
@@ -239,34 +239,11 @@ func (lp *loadingPackage) loadFromExportData() error {
 			return fmt.Errorf("dependency %q hasn't been loaded yet", path)
 		}
 	}
-	if pkg.ExportFile == "" {
-		seen := make(map[string]struct{})
 
-		count := 0
-		var sb strings.Builder
-		packages.Visit([]*packages.Package{pkg}, func(p *packages.Package) bool {
-			if _, ok := seen[p.ID]; ok {
-				return false
-			}
-			if count >= 3 {
-				return false
-			}
-			seen[p.ID] = struct{}{}
-			// Optionally skip test variants:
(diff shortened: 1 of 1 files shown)


ldez:
Closed in favor of #6061

PR #6059, opened 2025-09-05
pkg/goanalysis: add better error message for 'no export data' failure

kevinburkesegment:
If compilation or building of a Go package fails, export data will not be available. Previously, we would mask the underlying failure and display only a message that "no export data" was available. This has led to lots of confusion among golangci-lint users (and users at my company, which is how I discovered this problem).

golangci#3996
golangci#4630
golangci#4912
golangci#4964
golangci#5037
golangci#5051
golangci#5184
golangci#5428
golangci#5437

In addition to numerous discussions. The only common thread between the responses to these discussions, is that the "no export data" error message is obscuring whatever the underlying problem is, and making it more difficult to troubleshoot.

golangci#2489
golangci#2191
golangci#2751
golangci#3363
golangci#4829
golangci#4880

This fix indicates to users what they should try next if they receive this error message from golangci-lint. Hopefully, this will help users solve their own problem, and reduce the amount of issues and discussions related to this issue that people post on the discussion boards.

### DIFF
--- a/pkg/goanalysis/runner_loadingpackage.go
+++ b/pkg/goanalysis/runner_loadingpackage.go
@@ -240,7 +240,8 @@ func (lp *loadingPackage) loadFromExportData() error {
 		}
 	}
 	if pkg.ExportFile == "" {
-		return fmt.Errorf("no export data for %q", pkg.ID)
+		// https://github.com/golangci/golangci-lint/issues/6056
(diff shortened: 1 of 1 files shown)


ldez:
By creating the PR after I closed #6058, what was your expectation?
This kind of behavior is unacceptable.

PR #6062, opened 2025-09-07
feat: add godoclint linter

babakks:
This PR adds `godoclint` to available linters ([`godoc-lint/godoc-lint`](https://github.com/godoc-lint/godoc-lint)).

Godoc-Lint is a little opinionated linter for Go documentation practice.

<details>

### About the linter

*Godoc-Lint* is a linter for godocs, applying the rules set by the Go team in [*Go Doc Comments*](https://go.dev/doc/comment) blog post. I made Godoc-Lint as a small/encapsulated linter to enforce best practices around godocs. There might be other linters that support some of the Godoc-Lint rules, but I couldn't find one single linter encompassing them all. The linter is rather new, but is already used in a few projects, including [`googleapis/librarian`](https://github.com/googleapis/librarian/blob/f890a370773f89754683f69d8e44c901fcb94a18/all_test.go#L182), [`googleapis/google-cloud-rust`](https://github.com/googleapis/google-cloud-rust/blob/77183b333026ff0bb5f4cf24bcf5dc9f3d00b9fa/tools/golints/all_test.go#L68), and a few others, where it's being used outside of the golang-lint suite. That's why I decided it makes sense for the linter to be integrated with golangci-lint as the current industry standard. More about the linter and its rules can be found in the repo's [`README.md`](https://github.com/godoc-lint/godoc-lint/blob/main/README.md) file.

### Integration

I made sure the linter satisfies all technical criteria on the checklist (e.g. no panics, CI tests, etc). The current latest version, [`v0.5.0`](https://github.com/godoc-lint/godoc-lint/releases/tag/v0.5.0) has been released for this purpose. Note that `os.Exit` has been used a couple of times in the `cmd/main` package of the linter. That package is the CLI binary entrypoint and is not reachable through the integration with golangci-lint.

Regarding the tests, there are exhaustive tests in the linter's repo, which seemed too much to be duplicated in here in golangci-lint. Instead I added a minimal set of test cases to assert certain aspects of the integration (e.g. default config behaviour, or configuration mapping). However, I couldn't find a decent, already implemented way of package-wide testing. There's this specific rule, named `single-pkg-doc`, that needs to check all files in a Go package to make sure only one of the `package` statements has godoc, if any. That specific rule is not covered by the tests in this PR.

Also, as pointed out in the checklist, the test files include an stdlib import. I'm not sure if it's the same reason, but during early days of developing the linter I noticed `go/analysis` iterates over all imported packages, and therefore I had to make the linter automatically exclude those packages to ensure it stays within the codebase.

Regarding the default configuration, the linter has a sensible set of defaults that makes it usable out of the box while bringing value.

</details>


ldez:

- [x] The CLA must be signed

### Pull Request Description

### Linter

- [x] It must have a valid license (AGPL is not allowed), and the file must contain the required information by the license, ex: author, year, etc.
- [x] It must use [`go/analysis`](https://golangci-lint.run/contributing/new-linters/).
- [x] It must not have false positives/negatives (the team will help to verify that).

### Others Requirements

- [x] The `.golangci.yml` of golangci-lint itself must not be edited and the linter must not be added to this file.
- [x] The linters must be sorted in the alphabetical order (case-insensitive) in the `lintersdb/builder_linter.go` and `.golangci.next.reference.yml`.

### Recommendations

- [x] The file `jsonschema/golangci.next.jsonschema.json` should be updated.
- [x] The linter repository should have a readme and linting.
- [x] A tag should never be recreated.
- [x] use `main` as the default branch name.

---

ldez:
The linter description claims to follow Go recommendations from [this page](https://go.dev/doc/comment), but several rules are not those recommendations (ex, line length).

The linter applies the rules to non-exposed elements, but those elements are not a part of the godoc render (ex, a non-exported constant doesn't need to have a doc beginning with its own name), and so they don't need to follow the same rules as exposed elements.
Also, this is, in some ways, opposed to the recommendations themselves:
> Every exported (capitalized) name should have a doc comment.

This can be seen as false positives, IMO, this will be the main criterion for users to not use this linter.

Even the examples of "good" from the linter README don't pass the rules of the linter:
> // Check [docs] here.                    // (Good)
> //
> // [docs]: https://foo.com/docs
> const foo = 0

babakks:
Thanks for taking the time reviewing the linter, @ldez! :pray:

Let me first provide a categorised table of the rules so that I can refer to them later:

| Kind | Rules | Notes |
|-|-|-|
| Basic (default) | `pkg-doc`, `single-pkg-doc`, `start-with-name` | recommended by [*Go Doc Comments*][godoc], and low-effort |
| Strict | `require-doc`, `require-pkg-doc` | recommended by [*Go Doc Comments*][godoc], and high-effort |
| Extra | `max-len`, `no-unused-link` | extra but compatible with [*Go Doc Comments*][godoc] |

[godoc]: https://go.dev/doc/comment

By default, the linter only applies the *Basic* rules. They're low-effort because they don't require user to add more godocs. Rather, they just check if the existing godocs match the recommended formatting.

The *Strict* rules are costly to enable. Although they're recommended by [*Go Doc Comments*][godoc], but the linter do not apply them by default, because it will be too much, and in some cases unnecessary. So, that's up to the user to enable them based on their situation.

And finally the *Extra* rules which are not part of the blog post, are additional rules to help improve godocs. They're not breaking/opposing the recommended rules. These are also not enabled by default. 

Now, back to your points:

> The linter description claims to follow Go recommendations from [this page](https://go.dev/doc/comment), but several rules are not those recommendations (ex, line length).

True. `max-len` and `no-unused-link` are *Extra* rules, to help further improve godocs. Note that they're opt-in and not enabled by default.

> The linter applies the rules to non-exposed elements, but those elements are not a part of the godoc render (ex, a non-exported constant doesn't need to have a doc beginning with its own name), and so they don't need to follow the same rules as exposed elements. Also, this is, in some ways, opposed to the recommendations themselves:
> 
> > Every exported (capitalized) name should have a doc comment.
> 
> This can be seen as false positives, IMO, this will be the main criterion for users to not use this linter.

Good point, and that was one of my initial challenges. Note that godocs are not only used with the renderer or on `pkg.go.dev`. The Go language server (gopls) uses them to improve the IDE experience. So, godocs on unexported symbols are still visible to developers and it makes sense to apply the *Basic* rules to them. That's one of the reasons why I called the linter *a little opinionated* in the README.

That said, I think I'm open to exclude unexported symbols from the default behaviour, of course, in favour of a configuration parameter to enable the rule for them. Please let me know your thoughts.

> Even the examples of "good" from the linter README don't pass the rules of the linter:
> 
> > // Check [docs] here.                    // (Good)
> > //
> > // [docs]: https://foo.com/docs
> > const foo = 0

Makes sense, and thanks for the feedback. I'll update those examples to avoid confusion for the reader.

ldez:
> That said, I think I'm open to exclude unexported symbols from the default behaviour, of course, in favour of a configuration parameter to enable the rule for them. Please let me know your thoughts.

I think it should be an option of `start-with-name` (and false by default)

ldez:
Note: golangci-lint will not follow your configuration naming style with `/`.
The options will be on structures.

babakks:
> I think it should be an option of `start-with-name` (and false by default)

Makes sense. I'll push the changes for this and README good/bad examples, and will then rebase the PR.

> Note: golangci-lint will not follow your configuration naming style with `/`. The options will be on structures.

Just to be clear, you saying this format is not acceptable, right?

```yaml
linters:
  settings:
    godoclint:
      options:
        max-len/length: 127
```

And if yes, how about using dashes like this?

```yaml
linters:
  settings:
    godoclint:
      options:
        max-len-length: 127
```

Or even dots?

```yaml
linters:
  settings:
    godoclint:
      options:
        max-len.length: 127
```

Any particular recommendation?

ldez:
```yml
linters:
  settings:
    godoclint:
      options:
        max-len:
          length: 127
```

babakks:
> ```yaml
> linters:
>   settings:
>     godoclint:
>       options:
>         max-len:
>           length: 127
> ```

I'll push the changes for this as well.

babakks:
I made the following changes and published v0.6.1 of the linter:
- The `start-with-name` rule skips unexported symbol by default.
  - A new option added, named `start-with-name/include-unexported` (default `false`), to override the behaviour.
- The examples in the linter README are improved so that "good" cases are actually okay with the linter.
- The same categorised table I mentioned in https://github.com/golangci/golangci-lint/pull/6062#issuecomment-3263739443 is now added to the linter's README to proivde a quick overview of different types of rules.
- A few typos are fixed.

And here are the changes I (force) pushed to this PR branch:
- The fine tuning options are now nested structs (rather than being forward-slash separated).
  - The next version JSON schema and config reference are both updated accordingly.
- The test files are renamed with a `godoclint_` prefix.

pkg/golinters/godoclint/godoclint.go
```diff
@@ -9,26 +9,41 @@ import (
...
+		// - Options.RequireDocIncludeTests
+		// - Options.StartWithNameIncludeTests
+		// - Options.NoUnusedLinkIncludeTests
+
+		pcfg = glconfig.PlainConfig{
+			Enable:  settings.Enable,
+			Disable: settings.Disable,
+			Options: &glconfig.PlainRuleOptions{
+				MaxLenLength:                   settings.Options.MaxLen.Length,
+				MaxLenIncludeTests:             pointer(true),
```
babakks, line 45: I think `IncludeTests` options should be set to `false`, as this is the default behaviour of the linter. Correct me if I'm missing sth.

ldez: The values don't depend on the default values of the linter, but rather on the possibility to manage them globally.

babakks: But it's now set to `true` and that tells the linter to always include test files, which is a rare case in practice. So, if we're going to drop the option from config, it's better to populate it with `false`. I mean, otherwise, it'd be very annoying for the user, right?

ldez: I think it's better for consistency between godocs to enable this option by default.

The test files can be excluded with the global configuration:
```yml
linters:
  exclusions:
    rules:
      - path: (.+)_test\.go
        text: godoc exceeds max length
        linters:
          - godoclint
```

ldez: But we could also remove `max-len` because it's already handled by a dedicated linter (`lll`) and a formatter (`golines`).
I don't see cases where users want to use this kind of limit only on godocs.
IMO, this rule is a bit off-topic.

babakks: I don't know about `lll`, but godoclint is careful when limiting a line length because, for example, lines in a code block must not be altered at all.

ldez: We can keep the option: it's not enabled by default, and it's a bit more smart than `lll`.
I bet on the fact that users will enable everything and complain about about conflicts, but it's my problem to educate users not to enable everything.

ldez:
As an improvement of the linter, I think some messages can be more precise.

ex:
```go
// Foo is a thing.
type Bar struct {}
```

Current message:
```
foo.go:14:2: godoc should start with symbol name (pattern "((A|a|An|an|THE|The|the) )?%") (godoclint)
// Foo is a thing.
^
```

Maybe something like that:
```
foo.go:14:2: godoc should start with symbol name ('Bar') (godoclint)
// Foo is a thing.
^
```

babakks:
Good point, and I like the nicer message. I'll make a release for it and rebase this PR. Thanks for the feedback. 🙏

ldez:
Please don't rebase.

> The reviews should be addressed as commits (no squash).
> https://github.com/golangci/golangci-lint/pull/6062#issuecomment-3263349680

ldez:
From my experience, when users don't understand a report:
1. In most cases, they disabled the linter or don't enable it.
2. A few of them will try to understand, and will understand, but because they are working with a team, and teams don't like confusion, they will disable the linter.
3. A few of the few will understand and keep the linter.

Report messages are as critical as good defaults for the adoption of a linter.

I recommend reviewing the report messages to ensure they are easily understandable by Go newcomers and non-native English speakers.
I think this can be frustrating for native English speakers, but this is important.

example:
```
godoc exceeds max length
```

This is confusing because the problem is related to a line inside the godoc and not the godoc by itself.

babakks:
Valuable insight, @ldez! With that I'll review the report messages and will come back with a new release. I was leaning toward short messages, but the cost of losing adoption due to this is a real thing that I would never have thought. Thanks again! ✌️

Antonboom:
Finally someone did it!

Related to

- https://github.com/golang/go/issues/57963
- https://github.com/golang/go/issues/63929
- https://github.com/golang/go/issues/73056
- https://github.com/go-critic/go-critic/issues/1479

ldez:
It doesn't verify links, but it verifies that links are used.
And only element with the form `[link]: URL`

babakks:
@ldez is right. But checking for valid links is on my list of todos.

babakks:
I just pushed the changes to issue messages. You can see https://github.com/godoc-lint/godoc-lint/pull/33 for an overvall look.

ldez:
I don't have the answers, but I wonder if a godoc starting with `// Deprecated:` should be skipped.

https://go.dev/doc/comment#deprecations
> The paragraph does not have to be the last paragraph in the doc comment.


```go
// My constants.
const (
	A = "A"
	B = "B"
	C = "C"

	// Deprecated: use A instead.
	D = "D"
)
```



Same thing on notes (at least on non-exported symbols):
https://go.dev/doc/comment#notes

```go
// TODO(user1): refactor to use standard library context
// BUG(user2): not cleaned up
var ctx context.Context
```

This is not blocking from the POV of golangci-lint, this is only some questions.

babakks:
The `// Deprecated:` thing is already on my radar (recently noticed it as a valid use case when I saw https://github.com/kubernetes/kubernetes/issues/133566). Notes can also be handled, I think. The good thing about the notes on unexported symbols is that now they don't need to start with the symbol name by default.

Anyway, definitely both use cases are on my list, and yeah, they're not seem like blockers.

ldez:
Maybe you can create one or several issues to expose what is on your list.
It may encourage contributions or at least avoid having issues on already known topics.

babakks:
Good point. I'll create one issue for each for now to cover the ideas so far.

ldez:
The new message of `start-with-name` when using a explicit `pattern` is unexpected.

```go
// Foo is a thing.
type Bar struct {}
```

```yml
        start-with-name:
          pattern: "GODOC %"
```


```
main.go:8:1: godoc should start with symbol name ("Bar") (godoclint)
// Foo is a thing.
^
```


I think the message is right in 99% of cases, but the problem is maybe the option `pattern`.

ldez:
I found the previous problem, because I dug a bit more about `// Deprecated:` comment, and it seems to be a common pattern.

I wanted to quickly hack around the topic by modifying the pattern `(%|Deprecated:)`, and the reports were confusing.
And the pattern doesn't seem to work. :thinking:

babakks:
I think I already have a fix for the `// Deprecated:` case. Just need to add a couple of tests to maje sure ot's the right change.

Regarding the pattern thing, I think I can modify the report message to be different based on the configured pattern.

I'll push a commit when I'm done with the fix. Probably tomorrow.

ldez:
> Regarding the pattern thing, I think I can modify the report message to be different based on the configured pattern.

I think the `pattern` option is not useful; maybe it can be removed.
Do you have some real cases where this option is useful?

babakks:
A pattern is needed, because as of the godoc reference godocs should start with a complete senetence, which is okay to start with an article, for example. But I'm not sure there's a point in making the pattern configurable. 🤔

babakks:
Just released changes, using v0.8.0 of the linter. The `default` option is added to config, and pattern options are dropped from `start-with-name` and `pkg-doc`, so the config is much simpler now. Also, deprecations are better handled now (more on that in https://github.com/godoc-lint/godoc-lint/issues/36).

babakks:
I'm going to work on https://github.com/godoc-lint/godoc-lint/issues/41 to detect wrong usages of the `Deprecated:` marker. Should I just come back and update this PR's branch?

ldez:
For now, you can update the PR.
I will consider this as the last modification of the PR.

babakks:
Done, and thanks! :pray:

ldez:
maybe `deprecated` should be a part of `basic`? :thinking:

babakks:
I'm torn about that, actually. You think so? I'd like it.

ldez:
The pattern seems strict enough not to have false positives.
So I'm also torn about that, but I think it can be good.

babakks:
Yeah, that's my thought, too. Let's do it. I'll update here shortly.

ldez:
This linter has identical rules to other linters, but:
1. Those linters are meta-linters.
2. This linter is focused on one topic in the right way.
3. This linter has more rules than the others.

In this specific context, the rule about "no duplicate rules" doesn't apply, and then the linter can be accepted.

This is a specific case.
It's neither a rule nor a dogma.
It's only based on the context of this specific situation.

babakks:
By the way, I have already added validation to the `PlainConfig` type ([here](https://github.com/godoc-lint/godoc-lint/blob/d57d2744a4b2177bf956f788a65ca4f63ae89254/pkg/config/plain.go#L76-L105)). I doesn't check for conditions like *"if `none`, then no `disable`"*, but it mostly matches with what you did in https://github.com/golangci/golangci-lint/pull/6062/commits/3922fe58e7a36ce03363ce5854b5198f8d3d2de6, except for it also returns a single error with all issues.

Let me know if we can reuse it, or update it to accommodate with golangci-lint requirements.

ldez (approved):
LGTM

PR #6070, opened 2025-09-10
feat: add gocheckerrbeforeuse linter

T-Sh:
Adding [gocheckerrbeforeuse](https://github.com/T-Sh/go-check-err-before-use) linter.

The linter checks whether the error is handled before using the other returned values.

For example, the linter should catch the following situation, when `print` is called before err handling:

```go
func wrongCustom2Func() {
	res, err := returns2Values()

	print(res)

	if err != nil {
	}
}
```

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -52,6 +52,7 @@ linters:
     - funlen
     - ginkgolinter
     - gocheckcompilerdirectives
+    - gocheckerrbeforeuse
     - gochecknoglobals
     - gochecknoinits
     - gochecksumtype
@@ -163,6 +164,7 @@ linters:
     - funlen
     - ginkgolinter
     - gocheckcompilerdirectives
+    - gocheckerrbeforeuse
     - gochecknoglobals
     - gochecknoinits
     - gochecksumtype
@@ -656,6 +658,12 @@ linters:
       # Force adding assertion descriptions to gomega matchers.
       # Default: false
       force-assertion-description: true
+
+    gocheckerrbeforeuse:
+      # Maximum allowable distance between receiving an error and handling it.
+      # Value must always be greater than or equal to 1.
+      # Default: 1
+      max-allowed-distance: 1
 
     gochecksumtype:
       # Presence of `default` case in switch statements satisfies exhaustiveness, if all members are not listed.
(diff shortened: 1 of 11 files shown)


ldez:

- [x] The CLA must be signed

### Pull Request Description

### Base Requirements

### Linter

    - [x] The default values must be defined in a comment.
    - [x] The option must have a short description.

### Other Requirements

- [x] The `.golangci.yml` of golangci-lint itself must not be edited, and the linter must not be added to this file.
- [x] The linters must be sorted in alphabetical order (case-insensitive) in the `lintersdb/builder_linter.go` and `.golangci.next.reference.yml`.

### Recommendations

- [x] The linter repository should have a readme and linting.
- [x] Use `main` as the default branch name.

### Workflow

---

ldez:
> The reviews should be addressed as commits (no squash).

https://github.com/golangci/golangci-lint/pull/6070#issuecomment-3274446040

T-Sh:
@ldez sorry for force push, i didnt expect you'd be so fast with review)

Fixes:
- updates for .golangci.next.reference.yml, add linter and settings
- extended tests, add test with configuration
- linter replacement in builder_linter for sorting

ldez:
1. The detection is based on [names](https://github.com/T-Sh/go-check-err-before-use/blob/a2edf705cf74d6f7a1273895145c9a3f686ca262/pkg/analyzer/checks.go#L14): this is too naive, and it will lead to false negatives.
2. The rule itself is too strict: there are valid use cases where errors are not checked right after the instantiation. This will lead to false positives.
3. The `distance` option, as it's a "global" option, is "all or nothing". This will result in either false positives or false negatives, depending on the value of the option.
4. The real problem is not the distance between the error and the check of the error, but the usage of the result(s) of a function that also emits the error before checking the error.

My conclusion, even if I understand the need behind this linter, is not to accept this linter because the problems are related to the approach of the linter.
Some points can be improved, but the linter cannot be improved easily to handle all of the previous points.

Thank you anyway for suggesting the addition of your linter.

---

Some extra notes:

The linter implementation is overusing inlined `if` and nested `if`.
In 90% of the cases, this can be simplified by using "quick return" or `switch`.
We recommend reading this article: https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88

The licence file has been copied from another linter, and so [the file is not right](https://github.com/T-Sh/go-check-err-before-use/blob/a2edf705cf74d6f7a1273895145c9a3f686ca262/LICENSE#L3).

The linter name is explicit, maybe too explicit: it's almost a complete sentence.
- The prefix `go` doesn't seem useful.
- Module names that contain characters other than lower-case ASCII letters should be avoided.
- The name needs to be rethought.

Issue #4458, opened 2024-03-05
Incorrect information about gocritic disabled/enabled-checks being mutually exclusive?


scop:
### Welcome

- [X] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [X] Yes, I've searched similar issues on GitHub and didn't find any.
- [X] Yes, I've read the typecheck section of the FAQ (https://golangci-lint.run/usage/faq/#why-do-you-have-typecheck-errors).
- [X] Yes, I've tried with the standalone linter if available (e.g., gocritic, go vet, etc.). (https://golangci-lint.run/usage/linters/)

### Description of the problem

https://golangci-lint.run/usage/linters/#gocritic states that disabled-checks can not be combined with enabled-checks, and vice versa.

However, some tests of mine indicate that combining works just fine when combining also with enabled-tags. I don't know if there are more configurations in which they both work combined. Would be nice to clarify this in the docs.

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version 1.56.2 built with go1.22.0 from 58a724a0 on 2024-02-15T18:01:51Z
```

</details>

### Configuration

<details>

```console
linters:
  disable-all: true
  enable:
    - gocritic
linters-settings:
  gocritic:
    disabled-checks:
      - appendCombine # enabled through performance tag
    enabled-checks:
      - importShadow  # disabled by default
    enabled-tags:
      - performance
```

</details>

### Go environment

<details>

```console
$ go version && go env
# paste output here
```

</details>

### Verbose output of running

<details>

```console
$ golangci-lint cache clean
$ golangci-lint run -v
INFO [config_reader] Config search paths: [./ /home/scop/Documents/code /home/scop/Documents /home/scop /home /] 
INFO [config_reader] Used config file .golangci.yml 
INFO [lintersdb] Active 1 linters: [gocritic]     
INFO [loader] Go packages loading at mode 575 (compiled_files|imports|name|types_sizes|deps|exports_file|files) took 71.266967ms 
INFO [runner/filename_unadjuster] Pre-built 0 adjustments in 158.85µs 
INFO [linters_context/goanalysis] analyzers took 384.533975ms with top 10 stages: gocritic: 384.529978ms, typecheck: 3.997µs 
[...]

PR #6090, opened 2025-09-19
docs: gocritic enabled-checks and disabled-checks can be combined

skitt:
The documentation currently indicates that enable-checks and disabled-checks can't be combined for gocritic. That is however not accurate: as per InferEnabledChecks in the gocritic wrapper,

* if enable-all is set, all available checks are added to the set of enabled checks
* otherwise, unless disable-all is set, the default checks are added to the set of enabled checks
* then the checks corresponding to each enabled tag are added
* then all enabled checks are added (with a warning if a check is enabled multiple times)
* then the checks corresponding to each disabled tag are removed
* finally, all disabled checks are removed (with a warning if a check wasn't enabled)

Using both enabled-checks and disabled-checks is useful when keeping the default set of checks as reference.

This adjusts the documentation to match the implementation.

related to #4458

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -717,8 +717,8 @@ linters:
       # Disable all checks.
       # Default: false
       disable-all: true
-      # Which checks should be enabled in addition to default checks; can't be combined with 'disabled-checks'.
-      # By default, list of stable checks is used (https://go-critic.com/overview#checks-overview):
+      # Which checks should be enabled in addition to default checks (or at all, if disable-all is true).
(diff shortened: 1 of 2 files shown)


ldez:
Hello @skitt, could you sign the CLA?

skitt:
> Hello @skitt, could you sign the CLA?

Sorry, I missed that part of the contribution workflow — I’ll have to get approval from our legal department.

ldez:
@skitt any news?

skitt:
Nothing yet :frowning_face:

PR #6098, opened 2025-09-26
golangci-lint custom follows plugin's go.mod replace directives

sublee:
When developing a golangci-lint plugin, it’s common to first implement the analyzer using the go/analysis interface and then wrap it with the golangci-lint plugin interface in a separate module. This separation is useful because the go/analysis interface is a generic abstraction that can be reused outside golangci-lint.

During development, a plugin may depend on an unreleased analyzer implementation and declare a replace directive in its `go.mod`. However, the current `golangci-lint custom` command ignores these directives. As a result, the custom build only succeeds if the plugin and the analyzer are bundled within the same `go.mod`.

This PR copies `replace` directives from the plugin's `go.mod` into the temporary `go.mod` used by `golangci-lint custom`. This ensures that the custom build correctly recognizes both the plugin and any modules it references via `replace`. As a result, plugins and analyzers can now be managed as separate modules.

### DIFF
--- a/pkg/commands/internal/builder.go
+++ b/pkg/commands/internal/builder.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/base64"
+	"encoding/json"
 	"fmt"
 	"io"
 	"os"
@@ -115,6 +116,11 @@ func (b Builder) addToGoMod(ctx context.Context) error {
 				return err
 			}
 
+			err = b.mergeReplaceDirectives(ctx, plugin.Path)
+			if err != nil {
+				return err
+			}
+
 			continue
 		}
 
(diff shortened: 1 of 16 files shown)


ldez:
You don't need that.

You only need to use `path` with the "Automatic way": https://golangci-lint.run/docs/plugins/module-plugins/#the-automatic-way

Or to follow the "manual way": https://golangci-lint.run/docs/plugins/module-plugins/#the-manual-way

sublee:
Thanks for the review, @ldez!

You're right that a `replace` in the plugin module itself is fine. The issue is that `golangci-lint custom` builds a temporary repo whose `go.mod` does not inherit the plugin's `replace`, so it tries to fetch the analyzer from VCS.

For example.
1. `.custom-gcl.yml` points to `./myplugin` by path option
2. `./myplugin/go.mod` has `replace example.com/myanalyzer => ../myanalyzer` (unpublished)
3. Running `golangci-lint custom` fails downloading `example.com/myanalyzer`

I've thought the root cause is the temporary `go.mod` generated by `custom` lacks the plugin's `replace`. This PR copies the plugin's `replace` into that temporary `go.mod`, so local analyzer dependency can be resolved.

If there's an existing "as-is" automatic way that makes custom adopt the plugin's `replace` without this change, could you point me to it?

ldez:
A plugin must contain the Analyzer so I don't understand what your plugin is.

During development, you don't need `replace` directive for your plugin; you only need to reference your plugin path inside the `path` field.

sublee:
You are write. It could be solved with both `import` and `path` options:

```yaml
plugins:
  - module: example.com/myanalyzer
    import: example.com/myanalyzer/myplugin
    path: .
```

When the repository looks like:

```
myanalyzer/
  .custom-gcl.yml
  go.mod -- does not depend on golangci-lint
  ...
  myplugin/
    go.mod -- depends on golangci-lint
    ...
```

Thanks for your advice!

ldez:
A plugin should not depend on golangci-lint, but on the [`plugin-module-register`](https://github.com/golangci/plugin-module-register)

See the example: https://github.com/golangci/example-plugin-module-linter

PR #6108, opened 2025-10-03
feat: add recovercheck linter

cksidharthan:
This PR adds support for [recovercheck](https://github.com/cksidharthan/recovercheck) linter.

A Go static analysis tool that finds goroutines created without panic recovery logic.

Below is a quick example of what this linter catches.
```
// ❌ Bad - will crash the goroutine
go func() {
    panic("oops")
}()

// ✅ Good - panic is recovered
go func() {
    defer func() {
        if r := recover(); r != nil {
            log.Printf("recovered: %v", r)
        }
    }()
    panic("oops")
}()
```


ldez:
Hello,

I recommend creating a plugin: https://golangci-lint.run/docs/plugins/module-plugins/

Issue #5873, opened 2025-06-11
go1.25 support


ldez:
**EDIT: since v2.4.0 golangci-lint supports go1.25 if compiled with go1.25 :tada:**

**Our official binaries are built with go1.25**

## Additional Notes

Some package systems (like **homebrew**, nix, etc.) are "slow" in propagating the new Go version.
After a new release of Go and golangci-lint, those systems can use an old Go version to compile golangci-lint resulting in a non-working binary (high memory usage, missing symbol, etc.).

In this situation, you should use the official binaries, or compile golangci-lint manually with the right Go version, or help these package systems to support this new Go version.

We are not responsible for package systems that don't use official binaries.

To validate the Go version used to compile: `golangci-lint version`

<details>

#### Is golangci-lint support go1.25?

No, and yes.

Basically, golangci-lint supports Go versions that are lower or equal to the Go version used **to build it**.

In addition to the Go version used to build it, some linters could need to be updated to support newer versions of Go, and internal pieces of golangci-lint could also need to be updated.

The current binaries from the releases page and the Docker images have been built with go1.24 and will be built with go1.24 as long as go1.25 is not GA.
So those binaries/images don't support go1.25 (new elements and internal Go changes).

https://golangci-lint.run/welcome/faq/#which-go-versions-are-supported

```mermaid
---
title: Is golangci-lint support go1.25?
---
flowchart
    A([compiled with go1.25?]) -- yes --> B([needs update of the internals of golangci-lint?])
    A -- no --> C((NO))
    B -- yes --> D([uses right internal pieces?])
    B -- no --> E((YES))
    D -- yes --> G((YES))
    D -- no --> F((NO))
```

You can follow the PR https://github.com/golangci/golangci-lint/pull/5872 to see the progress on this topic.

https://go.dev/wiki/Go-Release-Cycle

#### Additional Notes

Some package systems (like **homebrew**, nix, etc.) are "slow" in propagating the new Go version.
After a new release of Go and golangci-lint, those systems can use an old Go version to compile golangci-lint resulting in a non-working binary (high memory usage, missing symbol, etc.).

In this situation, you should use the official binaries, or compile golangci-lint manually with the right Go version, or help these package systems to support this new Go version.

We are not responsible for package systems that don't use official binaries.

To validate the Go version used to compile: `golangci-lint version`

```console
$ golangci-lint version     
[...]

PR #6114, opened 2025-10-06
-

pypaut:
Projects using Go 1.25+ are not compatible with the current `golangci-lint`. This PR aims at using a later version of Go to fix this dependency.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/golangci/golangci-lint/v2
 
-go 1.24.0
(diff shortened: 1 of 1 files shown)


ldez:
The minimum Go version is the **minimum language version** used to write golangci-lint: golangci-lint must support go1.24 and go1.25.

golangci-lint already supports go1.25 since v2.4.0: https://github.com/golangci/golangci-lint/issues/5873

This PR is useless.

PR #6122, opened 2025-10-11
Add unnecessaryerror linter

sollniss:
Add support for the [unnecessaryerror](https://github.com/sollniss/unnecessaryerror) linter.

unnecessaryerror suggests replacing errors that are only ever nil-checked, with bools.
Errors returned from exported functions, or errors that are used in some way are not reported.

```go
func foo() error { // error is only ever nil-checked; consider returning a bool instead
	return errors.New("error")
}

func Bar() {
	err := foo()
	if err != nil {
		doA()
	}
	doB()
}
```

### DIFF
--- a/jsonschema/golangci.next.jsonschema.json
+++ b/jsonschema/golangci.next.jsonschema.json
@@ -869,7 +869,7 @@
             "thelper",
             "tparallel",
             "unconvert",
-						"unnecessaryerror",
+            "unnecessaryerror",
             "unparam",
             "unused",
             "usestdlibvars",


ldez:

- [x] The CLA must be signed

### Pull Request Description

### Base Requirements

### Linter

  - [ ] AGPL is not allowed
- [ ] It must not contain:
  - [ ] `init()`
  - [ ] `panic()`

### Other Requirements

### Recommendations

### Workflow

---

ldez:
Hello,

I don't understand the reasoning behind this linter: why returning a bool is better than returning an error?

Also, the rule of the linter is not consistent because it's only on explicit unexported functions.
Methods, exported functions, or functions from types are not reported.

This leads to many false negatives.

I will not accept this linter, thank you anyway.

I recommend creating a plugin: https://golangci-lint.run/docs/plugins/module-plugins/ this linter.

sollniss:
>I don't understand the reasoning behind this linter: why returning a bool is better than returning an error?

It's mainly a micro optimization for hot paths that helps reduce allocations.

>Also, the rule of the linter is not consistent because it's only on explicit unexported functions.
>Methods, exported functions, or functions from types are not reported.

Errors returned from exported functions/methods could be relied upon by clients checking their type. This linter only reports errors that could be replaced without side-effects. I'm not aware of any false negatives that break this rule and would appreciate some examples.

>I will not accept this linter, thank you anyway.

Fair enough, maybe a bit too specialized for this project.

PR #6125, opened 2025-10-12
chore: freebsd/arm64 builds

gogolok:
This unblocks goreleaser from creating a freebsd/arm64 binary. I'm not sure why it has been blocked since golangci-lint is working fine on FreeBSD arm64/AArch64 when built manually.

Also, while here, adjust the `install.sh` script to support freebsd/arm64.

There is already a FreeBSD port of golangci-lint available at [devel/golangci-lint](https://www.freshports.org/devel/golangci-lint/) and the builds for arm64 seem to look fine.

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -36,8 +36,6 @@ builds:
     ignore:
       - goos: darwin
         goarch: 386
-      - goos: freebsd
-        goarch: arm64
 
 archives:
   - formats: [ 'tar.gz' ]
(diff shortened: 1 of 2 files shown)


ldez:
Could you add a description? Maybe with some references.

You also need to sign the CLA

ldez:
I noticed that you are working for StackIT.

I sent an email to your company on 2025-05-14 (I used the email from [GitHub organization profile](https://github.com/stackitcloud)), but I didn't receive any response.

Do you have a contact to share? You can use the email from [our GitHub profile](https://github.com/golangci)

ldez:
The exclusion of `freebsd/arm64` was not documented.

There is only [a commit](https://github.com/golangci/golangci-lint/commit/839ce6906b1b629bb1b31ab54f8c3fad0b0b8018), no related PR.

But this commit happens during go1.13, and `freebsd/arm64` was added in go1.14.

> Go now supports the 64-bit ARM architecture on FreeBSD 12.0 or later (the freebsd/arm64 port).
> https://tip.golang.org/doc/go1.14#freebsd

Issue #5747, opened 2025-04-18
revive: overriding one linter rule disables other rules


kkettinger:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/usage/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

Official binary

### Description of the problem

Using this minimal `.golangci.yml` yields around 12 errors in my project (`exported`, `package-comments`):
```
version: "2"
linters:
  default: none
  enable:
    - revive
```

```sh
12 issues:
* revive: 12
```

Now i add linter settings to ignore `package-comments` errors for revive:
```
version: "2"
linters:
  default: none
  enable:
    - revive
  settings:
    revive:
      rules:
        - name: package-comments
          disabled: true
```

Running golangci-lint again, suddenly the `exported` errors are gone:
```
0 issues.
```

### Version of golangci-lint

<details>

```console
$ golangci-lint --version
golangci-lint has version 2.1.2 built with go1.24.2 from 00a561d9 on 2025-04-15T13:26:47Z
```

</details>

### Configuration

<details>

```console
version: "2"
linters:
  default: none
  enable:
    - revive
  settings:
    revive:
      rules:
        - name: package-comments
          disabled: true
```

</details>

### Go environment

<details>

```console
$ go version && go env
go version go1.24.1 linux/amd64
AR='ar'
CC='gcc'
CGO_CFLAGS='-O2 -g'
CGO_CPPFLAGS=''
CGO_CXXFLAGS='-O2 -g'
CGO_ENABLED='1'
CGO_FFLAGS='-O2 -g'
CGO_LDFLAGS='-O2 -g'
CXX='g++'
GCCGO='gccgo'
GO111MODULE='on'
GOAMD64='v1'
GOARCH='amd64'
GOAUTH='netrc'
GOBIN=''
GOCACHE='/home/kk/.cache/go-build'
[...]

Issue #1745, opened 2021-02-18
Revive linter rules configuration


vtvz:
Some revive rules are not configurable due to invalid type. You can reproduce the error using these rule config one by one:

```yaml
- name: cognitive-complexity
  arguments: [ 7 ]
- name: line-length-limit
  arguments: [ 110 ]
- name: function-result-limit
  arguments: [ 3 ]
- name: argument-limit
  arguments: [ 4 ]
- name: cyclomatic
  arguments: [ 10 ]
- name: max-public-structs
  arguments: [ 3 ]
```

These broken rules are commented in the full config listed below

- [x] Yes, I'm using a binary release within 2 latest major releases. Only such installations are supported.
- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] Yes, I've included all information below (version, config, etc).

Please include the following information:

<details><summary>Version of golangci-lint</summary>

```console
$ golangci-lint --version
golangci-lint has version 1.37.0 built from 34e5fc63 on 2021-02-17T09:32:37Z
```

</details>

<details><summary>Config file</summary>

```yaml
# cat .golangci.yml
linters-settings:
  dupl:
    threshold: 100
  funlen:
    lines: 100
    statements: 50
  gci:
    local-prefixes: github.com/golangci/golangci-lint
  goconst:
    min-len: 2
    min-occurrences: 2
  gocritic:
    enabled-tags:
      - diagnostic
      - experimental
      - opinionated
      - performance
      - style
    disabled-checks:
      - dupImport # https://github.com/go-critic/go-critic/issues/845
      - ifElseChain
      - octalLiteral
      - whyNoLint
      - wrapperFunc
  gocyclo:
    min-complexity: 15
  goimports:
    local-prefixes: app
  golint:
    min-confidence: 0
  gomnd:
    settings:
      mnd:
        # don't include the "operation" and "assign"
        checks: [ argument,case,condition,return ]
  govet:
    check-shadowing: true
    settings:
      printf:
        funcs:
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Infof
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Warnf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Errorf
          - (github.com/golangci/golangci-lint/pkg/logutils.Log).Fatalf
  lll:
    line-length: 140
  maligned:
    suggest-new: true
  misspell:
    locale: US
  nolintlint:
[...]

PR #6146, opened 2025-10-23
add enable-default-rules: setting for revive

cce:
This adds an `enable-default-rules:` setting for revive. If set, the entries in the `rules:` list will be able to enable/disable individual rules assuming the default set has already been enabled. This makes it easier for users to exclude specific rules beyond the default configuration, and not have to check during upgrades if new rules have appeared in the default set.

Related: #1745, #5747

### DIFF
--- a/pkg/config/linters_settings.go
+++ b/pkg/config/linters_settings.go
@@ -834,15 +834,16 @@ type RecvcheckSettings struct {
 }
 
 type ReviveSettings struct {
-	Go             string            `mapstructure:"-"`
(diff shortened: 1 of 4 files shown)


ldez:
This is a choice to follow the same way as revive to handle the configuration.

cce:
I hope you will reconsider this patch some day. It would be much easier for users to configure the revive linter if this option were available. Again, thank you for your work on this project.

cce:
I will submit a patch upstream, thank you for the feedback

PR #6147, opened 2025-10-26
feat: add perfcheck linter and wire docs/schema/migrations

m-v-kalashnikov:
The PR adds support for [perfcheck](https://github.com/m-v-kalashnikov/perfcheck) linter

perfcheck is a linter that guides on writing code in performance-by-default manner.


ldez:
Most of the rules already exist in other linters; some are not related to performance.

IMHO, this is an "early optimization" analyzer, and will lead to wrong recommendations / false-positives.

So, this is a detector, not linter.

Also, the PR contains many unwanted changes.

Smells like AI slop.

PR #6151, opened 2025-10-26
feat: add varasconst linter

MeenaAlfons:
This PR adds `varasconst` to available linters.

The Go linter [MeenaAlfons/varasconst](https://github.com/MeenaAlfons/varasconst) allows developer to mark global vars as constant using `// const`.

<details>

### General

I followed all the steps in this [docs](https://golangci-lint.run/docs/contributing/new-linters/) and from what I have seen in other PRs for adding a new linter.

The linter doesn't take any configurations (at least for now).

### Integration

Regarding testing, I copied most of the use cases from the test cases in the original repo. One case that cant be tested using golangci-lint test suite is changing a `var` marked as const that is imported from another package. I couldn't find a way to make the golangci-lint test suite run on multiple packages under `testdata`. See [tests for this case](https://github.com/MeenaAlfons/varasconst/blob/main/testdata/src/p2/p2.go) in the original repo.

</details>

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -125,6 +125,7 @@ linters:
     - unused
     - usestdlibvars
     - usetesting
+    - varasconst
     - varnamelen
     - wastedassign
     - whitespace
@@ -240,6 +241,7 @@ linters:
     - unused
     - usestdlibvars
     - usetesting
+    - varasconst
     - varnamelen
     - wastedassign
     - whitespace
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/golangci/golangci-lint/v2
 
-go 1.24.0
+go 1.24.5
 
 require (
 	4d63.com/gocheckcompilerdirectives v1.3.0
@@ -15,6 +15,7 @@ require (
 	github.com/Antonboom/testifylint v1.6.4
 	github.com/BurntSushi/toml v1.5.0
 	github.com/Djarvur/go-err113 v0.1.1
+	github.com/MeenaAlfons/varasconst v0.1.2
 	github.com/MirrexOne/unqueryvet v1.2.1
 	github.com/OpenPeeDeeP/depguard/v2 v2.2.1
 	github.com/alecthomas/chroma/v2 v2.20.0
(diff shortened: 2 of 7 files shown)


ldez:

- [x] The CLA must be signed

### Pull Request Description

### Base Requirements

### Linter

  - [ ] AGPL is not allowed
- [ ] It must not contain:
  - [ ] `init()`
  - [ ] `panic()`

### Other Requirements

### Recommendations

### Workflow

---

ldez:
hello,

The variable names in documentation and tests are the opposite of the Go convention: `_` should not be used.

The examples are not right: a string can be a constant, so using `var` in this context makes no sense.

This analyzer will lead to false positives because golangci-lint works per package, and not on the whole code.

Using global variables is bad by design.
This linter will create an illusion because the global variables can be modified externally, and it will lead to bad practices.

You cannot fight against the language.

MeenaAlfons:
That's fair. 

Let me share a use case of a `var` that can't be const and is intended not to be changed later.

```go
const (
	tableUsers = "users"

	columnFirstname = "firstname"
	columnLastname  = "lastname"
)

var columns = []string{
	columnFirstname,
	columnLastname,
}

// Adopt a rule to define queries outside func() to keep query construction away from user input.
var updateFirstnameQuery = fmt.Sprintf("UPDATE %s SET %s = ?", tableUsers, columnFirstname)

func updateFirstname(firstname string) {
	db.Query(ctx, updateFirstNameQuery, firstname)
}

var getUserQuery = fmt.Sprintf("SELECT %s from %s", strings.Join(columns, ","), tableUsers)

func getUsers() {
	db.Query(ctx, getUserQuery)
}
```

Real world mistake:
```go
func updateFirstname(firstname string) {
	db.Query(ctx, fmt.Sprintf("UPDATE %s SET %s = %s", tableUsers, columnFirstname, firstname))
}
```

So maybe there is a time and place where global `var` could be useful and is intended to be constant.

Happy to hear your further thoughts. In general I wouldn't mind stopping this effort. Just sharing the purpose behind it in case it has a chance.

P.S. the first two comments are fixable. Hence I'm not discussing them.

PR #6152, opened 2025-10-27
build(deps): bump mvdan.cc/unparam to HEAD

mjgarton:
https://github.com/mvdan/unparam/compare/0df0534333a4...5beb8c8f8f15

---

**EDIT**
Note from golangci-lint maintainer:

**For future contributors: please don't open a PR to update a linter in these cases.**
**You should wait or open an issue.**


ldez:
This is a special dependency because there is no version, so dependabot is not able to update it automatically.

I regularly check those non-versioned dependencies, and I handle them.

**For future contributors: please don't open a PR to update a linter in these cases.**
**You should wait or open an issue.**

Exceptionally, I will reopen and merge this PR.

PR #6156, opened 2025-10-28
dev: prealloc intermediate slice in max same issues

ferhatelmas:
Prevents unnecessary memory allocations.


ldez:
This is micro-optimization, there is no real impact.

PR #6165, opened 2025-10-31
feat!: warning when finding unknown linters

luoling8192:
Allow the program to continue running when unknown linters are found, and print a warning, so that it can be compatible with new versions of configuration files on old versions of golangci-lint. 

Before this, if a new version-specific linter was set, the old version would report an error and exit directly. This PR improves compatibility.

```
> ./golangci-lint run ./...
WARN [lintersdb] unknown linters: 'xxx', run 'golangci-lint help linters' to see the list of supported linters 
0 issues.
```

### DIFF
--- a/pkg/lint/lintersdb/validator.go
+++ b/pkg/lint/lintersdb/validator.go
@@ -1,7 +1,6 @@
 package lintersdb
 
 import (
-	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


luoling8192:
> I'm not sure if we need forward compatibility in that way. Could you create a discussion or a feature request first, collect feedback, and then implement it?

But I think for a problem that does not affect the program's operation, it should not report an error and directly interrupt the execution.

PR #6175, opened 2025-11-06
docs: improve fix flag description

ldez:
Related to #6174

### DIFF
--- a/.golangci.next.reference.yml
+++ b/.golangci.next.reference.yml
@@ -4495,7 +4495,7 @@ issues:
   # Default: false
   whole-files: true
 
-  # Fix found issues (if it's supported by the linter).
+  # Apply the fixes detected by the linters and formatters (if it's supported by the linter).
   # Default: false
   fix: true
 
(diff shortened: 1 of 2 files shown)


pkg/commands/flagsets.go
```diff
@@ -138,5 +138,5 @@ func setupIssuesFlagSet(v *viper.Viper, fs *pflag.FlagSet) {
 	internal.AddFlagAndBind(v, fs, fs.Bool, "whole-files", "issues.whole-files", false,
 		color.GreenString("Show issues in any part of update files (requires new-from-rev or new-from-patch)"))
 	internal.AddFlagAndBind(v, fs, fs.Bool, "fix", "issues.fix", false,
-		color.GreenString("Fix found issues (if it's supported by the linter)"))
+		color.GreenString("Apply the fixes detected by the linters and formatters (if it's supported by the linter)"))
```
MadLittleMods, line 141: There are a few more places in the codebase where the `Fix found issues (if it's supported by the linter)` language is used that could be updated.

ldez: The other files are generated.

alexandear (approved):

PR #6180, opened 2025-11-08
fix: follows symbolic links with formatter exclusion paths

chandler-solo:
`ln -s worktree wt && cd wt && golangci-lint fmt` now works

### DIFF
--- a/pkg/goformat/runner.go
+++ b/pkg/goformat/runner.go
@@ -223,8 +223,13 @@ func NewRunnerOptions(cfg *config.Config, diff, diffColored, stdin bool) (Runner
 		return RunnerOptions{}, fmt.Errorf("get base path: %w", err)
 	}
 
+	evaluatedBasePath, err := fsutils.EvalSymlinks(basePath)
+	if err != nil {
+		return RunnerOptions{}, fmt.Errorf("evaluate base path: %w", err)
+	}
+
 	opts := RunnerOptions{
-		basePath:            basePath,
+		basePath:            evaluatedBasePath,
 		generated:           cfg.Formatters.Exclusions.Generated,
 		diff:                diff || diffColored,
 		colors:              diffColored,
@@ -251,7 +256,6 @@ func (o RunnerOptions) MatchAnyPattern(path string) (bool, error) {
 		return false, nil
 	}
 
-	// The basePath is already resolved via `fsutils.Getwd()` in `NewRunnerOptions`.
 	evaluatedPath, err := fsutils.EvalSymlinks(path)
 	if err != nil {
 		return false, err
--- a/pkg/goformat/runner_test.go
+++ b/pkg/goformat/runner_test.go
@@ -3,14 +3,12 @@ package goformat
 import (
 	"os"
 	"path/filepath"
-	"runtime"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 
 	"github.com/golangci/golangci-lint/v2/pkg/config"
-	"github.com/golangci/golangci-lint/v2/pkg/fsutils"
 )
 
 func TestRunnerOptions_MatchAnyPattern(t *testing.T) {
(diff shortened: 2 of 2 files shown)


ldez:
The problem supposedly fixed by this PR "issue with the 'fmt' command and symbolic links" is not fixed.

At first, I was thinking the PR title was wrong because the only change was in the exclusions system.
So I made some changes to fit that assumption.

But in fact, this PR fixes nothing, because the problem with symlinks is not here.

This is why it's better to open an issue before opening a PR, and when opening a PR, a real description is important.

I close this PR, and I will try to create the real fix.

chandler-solo:
> ln -s worktree wt && cd wt && golangci-lint fmt

This is how to reproduce, on MacOS 15.7.1, with golangci-lint 2.6.1. You can work around the issue with `golangci-lint fmt $(pwd -P)`

This has nothing to do with formatter exclusion paths. My test case fails on `main` and passes with my changes.

ldez:
Your [commit](https://github.com/golangci/golangci-lint/pull/6180/commits/4ec7afc1fb23977e6c7e66e80afb253d1365ecf4) is **only** modifying `RunnerOptions.MatchAnyPattern` method.

1. This method is only related to formatter exclusion paths (`patterns` -> `cfg.Formatters.Exclusions.Paths`).
If there is no pattern (exclusion paths), this method is skipped.

https://github.com/golangci/golangci-lint/blob/7bcb236705fdd83923d5dfe2487d9244cc6585c2/pkg/goformat/runner.go#L236-L244

https://github.com/golangci/golangci-lint/blob/7bcb236705fdd83923d5dfe2487d9244cc6585c2/pkg/goformat/runner.go#L250-L252

2. Your comment `The basePath is already resolved via fsutils.Getwd() in NewRunnerOptions.` is wrong: `fsutils.Getwd()` is only called when `wd` mode is used. This mode is only one of the four modes (and not the default).

https://github.com/golangci/golangci-lint/blob/7bcb236705fdd83923d5dfe2487d9244cc6585c2/pkg/fsutils/basepath.go#L56-L57

3. All the tests, inside your [commit](https://github.com/golangci/golangci-lint/pull/6180/commits/4ec7afc1fb23977e6c7e66e80afb253d1365ecf4), are only testing formatter exclusion paths, and in the wrong way by playing with unexposed fields.

<details>
<summary>A concrete reproducer</summary>

```bash
#!/bin/sh -e

ROOT=$(mktemp -d)
mkdir -p $ROOT/project/

## Create module
cat > $ROOT/project/go.mod <<EOF
module github.com/golangci/sandbox

go1.24.0
EOF

# Create golangci-lint configuration file.
cat > $ROOT/project/.golangci.yaml <<EOF
version: "2"

formatters:
  enable:
    - gofmt
EOF

## Create main.go file.
cat > $ROOT/project/main.go <<EOF
package main

import "fmt"

func main() {
fmt.Println("hello world") // Formatting problem.
}
EOF

## Create symlink to the project directory.
ln -s $ROOT/project/ $ROOT/workdir


## Run golangci-lint inside the real project directory.
cd $ROOT/project

echo
echo "--------------------------"
echo "Run on the real project directory ($(pwd)):"
echo

golangci-lint fmt --diff || true

## Run golangci-lint inside the symlinked project directory.
cd $ROOT/workdir

echo
echo "--------------------------"
echo "Run on the symlinked project directory ($(pwd)):"
echo

golangci-lint fmt --diff || true

rm -rf $ROOT
```


```console
$ ./setup.sh

--------------------------
Run on the real project directory (/tmp/tmp.5C7AJyRYRC/project):

diff /tmp/tmp.5C7AJyRYRC/project/main.go.orig /tmp/tmp.5C7AJyRYRC/project/main.go
--- /tmp/tmp.5C7AJyRYRC/project/main.go.orig
+++ /tmp/tmp.5C7AJyRYRC/project/main.go
@@ -3,5 +3,5 @@
 import "fmt"
 
 func main() {
-fmt.Println("hello world") // Formatting problem.
+       fmt.Println("hello world") // Formatting problem.
 }

--------------------------
Run on the symlinked project directory (/tmp/tmp.5C7AJyRYRC/workdir):

$
```

</details>

I will create a PR with the real fix.

chandler-solo:
That is the perfect reproducer. I'll open an issue next time with a real reproducer.

PR #6198, opened 2025-11-14
dev: isolating GOCACHE for integration tests

abecli:
Every integration test now runs golangci-lint with its own temporary GOCACHE, so Go drops build artifacts into a folder the test owns instead of `~/Library/Caches/go-build`. That stops the “operation not permitted” errors we were seeing in sandboxes/CI when tests tried to write into the user cache, and it doesn’t affect normal CLI usage, only the test harness changed.

### DIFF
--- a/test/testshared/integration/run.go
+++ b/test/testshared/integration/run.go
@@ -64,6 +64,8 @@ func testOneSource(t *testing.T, log *logutils.StderrLog, binPath, sourcePath st
 		"--max-issues-per-linter=100",
 	}
 
+	cacheDir := t.TempDir()
+
 	for _, addArg := range []string{"", "-Etypecheck"} {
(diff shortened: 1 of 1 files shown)


ldez:
In which context do you need to run our tests?

abecli:
I’m running the integration suites inside a sandbox environment that can’t write to the user Go build cache `~/Library/Caches/go-build`. Without redirecting GOCACHE, the tests fail with “operation not permitted” when Go tries to populate that directory. The change forces each test run to use a temp cache the sandbox can write to, so the tests complete successfully in that case.

ldez:
Maybe I was not clear: why do you need to run our integration tests?

ldez:
Please open an issue and describe the problem in detail.

I will close this PR because the context is unclear, and I think the change is incorrect.

PR #6212, opened 2025-11-26
feat: add iotyper linter

fujiwaraizuho:
Adding [iotyper](https://github.com/CyberAgent/iotyper-lint) linter.

This linter detects iota usage in const declarations without explicit type specification. 
Explicit type annotation improves code clarity and type safety.

### Example
This linter flags the following code:

```go
// ❌ Bad: iota without type specification (will be flagged)
const (
    StatusPending = iota
    StatusActive
    StatusClosed
)
```

```go
// ✅ Good: explicit type specified
const (
    StatusPending int = iota
    StatusActive
    StatusClosed
)
```


ldez:

- [x] The CLA must be signed

### Pull Request Description

### Base Requirements

### Linter

- [x] It must have a valid license:
  - [x] AGPL is not allowed
  - [x] The file must start with `LICENSE` (capitalized)
  - [x] The file must contain the required information by the license, ex: author, year, etc.
- [ ] It must not contain:
  - [ ] `init()`
  - [ ] `panic()`

### Other Requirements

### Recommendations

### Workflow

---

ldez:
I have several remarks.

`iotyper` is a bad name because it seems related to `io` and not `iota`.

It's better if the repository and module names are the same as the base package. (avoid non-letter characters)

I don't see the point with having an builtin standard type for `iota`.
First because a `iota` is an untyped integer, so not exactly a `int`.
Using `int` can be a problem on 32-bit arch if the value is higher than `math.MaxInt32`.

This doesn't feel like a good practice: I can understand the usage of explicit custom type when defining enum, in some cases.
But your analyzer will lead to the usage of builtin type (`int`, etc.) and IMO this is false-positives.

```go
// ❌ Bad
const (
    StatusPending int = iota
    StatusActive
    StatusClosed
)
```

```go
// ✅ Good but only if this is useful to have a type.
type Status uint64

const (
	StatusPending Status = iota
	StatusActive
	StatusClosed
)
```

Based on my previous remarks, as a standalone linter, I think this is not a good idea, but I also think this can be an option (not enabled by default) of [`iotamixing`](https://github.com/AdminBenni/iota-mixing).

fujiwaraizuho:
@ldez 
Thanks a lot for your feedback.
First, I will update the name of iotyper, as well as the repository and module names.

Sorry for the confusing example — to clarify:
iotyper is not intended to enforce using a standard type (like `int`) for `iota`.
Its purpose is to enforce that `iota` always has an explicit type, not to encourage untyped or standard integer usage.

So the intention is not to allow something like:

```go
const (
    Foo int = iota
)
```

but rather to ensure that when `iota` is used for enum-like values,
a custom, explicitly defined type is used, such as:

```go
type Status uint64

const (
    StatusPending Status = iota
    StatusActive
    StatusClosed
)
```

ldez:
> Its purpose is to enforce that iota always has an explicit type, not to encourage untyped or standard integer usage.

I already understood the objective, but your linter will result into encouraging builtin types usage.
The users will see the reports, add a random type or ignore the reports.
This will be interpreted as the behavior of an analyzer with false-positives.

> but rather to ensure that when iota is used for enum-like values

This is not what your linter is doing: it reports all missing explicit types.
And all your examples and your tests are illustrating the problem.

fujiwaraizuho:
@ldez 
Thanks for the clarification — that makes a lot of sense.

You're absolutely right that, as currently designed, the linter ends up encouraging the use of standard types, because users may simply add `int` (or another arbitrary type) to silence the warning.
That is definitely not the intended behavior, and I agree this would effectively look like false positives from the analyzer.

And yes, you are also correct that the current implementation reports all missing explicit types, not just enum-like usages. The examples and tests indeed reflect this unintended behavior.

Given your feedback, I'm thinking of revising the rule so that it only applies when `iota` is used in a context where an enum-like custom type is expected — and not enforcing explicit types globally.

Before I move in that direction, I want to confirm:
Would a rule that only enforces explicit types for enum-style constants (with custom types) be more appropriate in your view?

Thanks again for pointing out the issue — it’s been very helpful.

ldez:

IMO, your analyzer is currently a detector: it detects missing explicitly typed `iota`.
But having an explicit typed `iota` is not a requirement, and doesn't really improve the readability, and can lead to usage of builtin types.
So, this can be interpreted as false-positives.

I think we agree on that.

Maybe, if your analyzer was forcing the usage of dedicated types, so being more opinionated, it could be see as a linter, but even in this case I'm not sure, I need to think about.

ldez:
If your analyzer is able to find the usage of `iota` inside a `const` group and reports missing type and usage of builtin types, and so have no false-positives, we may accept it.

I use "may" because I will have to evaluate the linter to be sure everything works as expected.

I will close this PR, feel free to open a new PR with the new analyzer when it will be changed and renamed.

Issue #6232, opened 2025-12-03
modernize and revive linters in the latest 2.7.0


alex-aizman:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/docs/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/docs/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

I don't know

### Description of the problem

it is a minor problem that boils down to false positives and wrong recommendations

reproducible if you just

```sh
git clone https://github.com/NVIDIA/aistore
make lint
```

The result:

```sh
$ golangci-lint version
golangci-lint has version 2.7.0 built with go1.25.4 from e6ebea01 on 2025-12-03T19:50:52Z
$ 
$ make lint
Running lint...
ais/backend/azure.go:194:11: stringscut: strings.Index can be simplified using strings.Cut (modernize)
                if i := strings.Index(line, azErrCode); i > 0 {
                        ^
ais/dpq.go:244:10: stringscut: strings.IndexByte can be simplified using strings.Cut (modernize)
        if i := strings.IndexByte(s, '='); i > 0 {
                ^
cmn/bck.go:120:9: stringscut: strings.IndexByte can be simplified using strings.Cut (modernize)
        idx := strings.IndexByte(s, apc.NsNamePrefix)
               ^
cmn/certloader/certloader.go:314:7: stringscut: strings.Index can be simplified using strings.Cut (modernize)
        i := strings.Index(s, " +")
             ^
cmn/cos/template.go:405:12: stringscut: strings.IndexByte can be simplified using strings.Cut (modernize)
                right := strings.IndexByte(template, '@')
                         ^
cmn/k8s/k8s.go:113:9: stringscut: strings.IndexByte can be simplified using strings.Cut (modernize)
        idx := strings.IndexByte(msg, ',')
               ^
[...]

PR #6233, opened 2025-12-04
modernize: disable stringscut analyzer

ldez:
`stringscut` analyzer reports false positives.
And this is a more important problem because it applies suggested fixes that could creates bugs.

Related to https://github.com/golangci/golangci-lint/issues/6232#issuecomment-3609413970

### DIFF
--- a/pkg/golinters/modernize/modernize.go
+++ b/pkg/golinters/modernize/modernize.go
@@ -14,9 +14,9 @@ func New(settings *config.ModernizeSettings) *goanalysis.Linter {
 	var analyzers []*analysis.Analyzer
 
 	if settings == nil {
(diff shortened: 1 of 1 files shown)


pkg/golinters/modernize/modernize.go
scop, line 1: Looks like `.golangci.*reference.yml` updates would be in order

leonklingele:
Has this been reported to the Go team already? Might be worth to let them know and fix the analyzer. If not already reported, I'll do that tomorrow.

Thanks for the quick fix! 😃

ferhatelmas:
> Related to https://github.com/golangci/golangci-lint/issues/6232#issuecomment-3609413970

PR #6234, opened 2025-12-04
fix(fixer): dedupe error message on conflicts from the same linter

scop:
For example, modernize's omitzero produces this. Better to say

> conflicting edits from modernize on ...

...than:

> conflicting edits from modernize and modernize on ...


ldez:
I prefer to keep the existing message.

PR #6236, opened 2025-12-05
Add propro linter

digitalstraw:
[`propro`](https://github.com/digitalstraw/propro) stands for **Protected Properties**. It prevents direct writes to public/exported properties of structs that represent domain entities.

## Why propro?

ORMs like GORM or tools like JSON encoding require access to exported struct fields.  
However, when a domain entity is updated, validation logic may need to run and domain events may need to be raised.  
Bypassing this logic through direct writes to entity properties can compromise domain integrity.

To enforce domain state mutation through methods, developers often keep entity properties private and create separate “model” 
structs for GORM or JSON with public properties. This approach has several drawbacks: it introduces significant boilerplate, 
requires maintaining two parallel structures, and adds extra unit tests for mapping logic in both directions.

`propro` solves this at the **linter level** by preventing direct writes to exported properties of structs marked as protected by the user.  
This allows entities to keep their properties exported and still be used directly with GORM and JSON.


ldez:

- [x] The CLA must be signed

### Pull Request Description

### Base Requirements

### Linter

  - [ ] AGPL is not allowed
- [ ] It must not contain:
  - [ ] `init()`
  - [ ] `panic()`

### Other Requirements

### Recommendations

### Workflow

---

ldez:
Hello,

this analyzer will not be accepted, because this is not a real protection (you cannot fight against the language), and because it will not work outside of a package (because goalngci-lint works only per package).

digitalstraw:
Hello and thanks for your review and feedback, even if it is negative. 

I explored your argument "_that it will not work outside of a package_" and must say I don't understand what do you mean.
I have it tested and it works well across the codebase in a real project with entities defined and used across multiple packages. So that for instance,
```go
func SomeFunc19() {
	e := &outside.PackageEntity{}
	e.ProtectedField = "value" // want "assignment to exported field OutsidePackageEntity.ProtectedField is forbidden outside its methods"
}
```
reports well.

Regarding "it is not a real protection (you cannot fight against the language)":
Could you please be more specific here? I don't understand what is the difference in "protection" provided by let's say `cyclop` or `embeddedstructfieldcheck` from that provided by my linter. And how these already merged linters "do not fight against the language" when one for instance reports wrong order of fields in the struct and the other reports that user has written too deep "ifs"...  I'm convinced that my linter plays the same game, it's the same kind of "fight".

This linter will help teams which implement domain driven design within hexagonal architecture and CQRS, which rely on domain integrity provided by domain events and Outbox pattern. Domain integrity is crucial, and if let's say newly hired developer unintentionally writes to exported property, it skips important routines that are needed to keep integrity of the domain.

So you decided that domain integrity is not worth protecting at the linter level, but on the other hand, we're happy to have here the protection against wrong ordrer of fields in the struct in the `embeddedstructfieldcheck`. 

Anyway, as I said, please don't take this as an offensive reaction, I'm fine with your decision whatever it will be. I just wanted to share what I found useful in my team.

Thank you and greetings to France.

ldez:
> please don't take this as an offensive reaction

No problem, my comment was a bit too short to be understanable.

---

First, sorry, my sentence "it will not work outside of a package" is offtopic, it was related to previous analyzers with the same kind of goal but on other elements.

Regarding the "protection": your analyzer will be used to "protect" the fields but the fields will not really be "protected" against direct writes.

The accessors on structures are here to protect the data, at least, this is how, in OOP with a layering situation, we (as dev) express the idea.
You are using this vocabulary: "Protected Properties"
The concept behind that is "immutability".


`embeddedstructfieldcheck` is an analyzer related to code style: functions/methods order.
`cyclop` is also related to style: you can defined your targeted complexity.

But those analyzers don't claim to protect against anything related to the language.

Maybe I missevaluated your analyzer, but I am always a little reluctant with analyzers that claim to offer "protection" that the language does not offer.

digitalstraw:
Thanks for your comment, understood, all is fine. I understand, that it must be pretty time consuming volunteering work. Thank you for doing it.

I will just be a little bit more specific and please accept my apologies if I'm explaining topics that you are absolutely familiar with. I just want to paint the bigger picture. 

I'm not sure if I understood you correctly, but immutability is not the case here, because entities are mutable by definition and multiple fields and multiple entities may change during a single operation. 

Mutation of the domain state is not as simple as writting a new value into a property (I know, you did not suggest anything like that).
- First, we have to check, if the value is already there, and run a domain validation of the new value (if it is alligned with business rules).
- Then, original value is saved to a temp variable and new value is written into the property. 
- Next comes raising the domain event which contains both old value and new value. 
- Then, the entity is sent to the repository for persistance. 
- Repository creates a transacation which saves the entity data along with outbox data. Outbox data contain all domain events that arised during the operation.
- The domain state was mutated only when this transaction is commited successfully.
- Then, the outbox data is read from the db (in another process) and are published to the message broker so that subscribers will be notified.
- If anybody directly writes the value to the entity property outside of the method, all this is skipped and domain integrity is lost because subscribers will not be notified.

Now where's the root of the problem: 

1/ if you want to use Object-Relational Mapping tool like GORM:
- It needs to work with struct that has exported fields which must be tagged for GORM. 
- If we would accept the fact that we will rely on the LANGUAGE level of field protection (fields are not exported), we must map language-protected data in the entity to some intermediate GORM-compatibile struct. 
- This introduces boilerplate - mappings in both directions: Entity -> GORM struct and GORM struct -> entity after it's loaded from the database. 
- It's not comfortable to keep and maintain two structs and multiple unit tests for a single entity. 
- We have multiple aggregates in the system that contain two or more entities each. Large systems have 100s of entities....
- The codebase would then explode with additional mappers and unit tests.

2/ optimistic concurrency 
- You don't want to overwrite data that was saved from the same entity by a concurrent user. 
- That's why you need to introduce some kind of a `Version` property into each entity which prevents this. 
- Simply, all database updates are then performed with `WHERE version=xxxxx` which - if new data was written by a concurrent user before my update, the update will not happen.
- If anybody by mistake writes directly a value into the `Version` property, it might break the optimistic concurrency solution.

So, by rejecting this pull request you enforce teams to rely on language features and write boilerplate. I don't feel it would be a good decision, but I fully respect it is in your hands.

Thank you for your time and attention.

ldez:
I will keep the PR close for now, but I will take time sometime to reevaluate your analyzer.
In all cases, I will comment here even if I stay on my previous decision.

digitalstraw:
FYI various fixes were made in the linter repo and docs. Latest version is 2.0.1.
Branch of the closed PR was updated.

leonklingele:
Hey @ldez, did you get a chance to check the updated version already?

digitalstraw:
Hello @ldez, is there any chance to know your final verdict about this linter? Thanks a lot.

PR #6239, opened 2025-12-06
docs: add thanks to the creator of golangci-lint

jirfag:
Add thanks to the creator of golangci-lint


jirfag:
@ldez please take a look is text ok for you, I've added explicit thanks to you (and me) in docs.

PR #6273, opened 2025-12-19
Support flag to clear outputs from config

timofurrer:
This change set adds a new command line flag called `--clear-config-outputs`. When setting this flag all configured outputs from the config file will be cleared and only the ones from the command line respected or if none, fallback to the default happens.

This is useful for applications like the golangci-lint LSP that invoke golangci-lint and want to ensure that only the JSON output is enabled without respecting the users config.


timofurrer:
@alexandear hey 👋Happy to see you at the "other side of the table" this time 🎉 

Would you mind having a look at this contribution? 🙏

ldez:
1. The name of the flag is unclear and confusing.
2. I'm not sure about the proposed approach.
3. The tests are not portable and written in a "Describe" style.
4. `docs/data/cli_help.json` should not be changed.

I would have preferred an issue, focused on the explanation of the problem, before opening a PR.

I need to think about this more.

For now, please don't change the PR content to avoid wasting time on this.

timofurrer:
@ldez does it make sense open that issue then ... before I waste more time on this ?

timofurrer:
Anyways, here we go: https://github.com/golangci/golangci-lint/issues/6274

PR #6275, opened 2025-12-21
docs: add mise alternative installation method

jylenhof:


ldez:
Hello,

I don't think we need to explain to `mise` users how to use `mise` to install something.

ldez:
:thinking: maybe we can adding but not in this way.

Issue #6259, opened 2025-12-11
install.sh fails to find valid release in CI (GitHub API rate limiting)


kaovilai:
### Welcome

- [x] Yes, I've searched similar issues on GitHub and didn't find any.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

install.sh script

### Description of the problem

The install.sh script fails to find valid releases when run in CI environments (e.g., GitHub Actions during Docker builds), reporting "unable to find 'v2.5.0'" even though the release exists with all necessary assets.

**Reproduction:**
```bash
# In CI environment (GitHub Actions, Docker build)
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b /tmp/test v2.5.0
```

**Error output:**
```
golangci/golangci-lint info checking GitHub for tag 'v2.5.0'
golangci/golangci-lint crit unable to find 'v2.5.0' - use 'latest' or see https://github.com/golangci/golangci-lint/releases for details
```

**However:**
- The release exists: https://github.com/golangci/golangci-lint/releases/tag/v2.5.0
- Has all necessary binaries (linux-amd64, darwin-arm64, etc.)
- Local installation works fine
- Installation in Docker locally works fine

### Root Cause

The install script's `github_release()` function makes unauthenticated GitHub API calls:
```bash
giturl="https://github.com/${owner_repo}/releases/${version}"
json=$(http_copy "$giturl" "Accept:application/json")
```

In CI environments with many builds, these unauthenticated API calls hit GitHub's rate limits, causing the script to fail even for valid releases.

### Proposed Solution

1. **Support GITHUB_TOKEN environment variable** for authenticated API requests
2. **Add retry logic** with exponential backoff for rate limit errors
3. **Document the go install alternative** as the recommended CI approach:
   ```bash
   go install github.com/golangci/golangci-lint/cmd/golangci-lint@v2.5.0
   ```

The `go install` method is more reliable in CI since it doesn't hit GitHub API rate limits and is the official Go way to install tools.

### Version of golangci-lint

v2.5.0 (and likely affects other versions)

### Go environment

```
Docker: golang:1.25-bookworm
GOOS=linux
GOARCH=amd64
```

### Additional context

This appears to be a CI-specific issue related to GitHub API rate limiting. The install script works fine in environments with fewer API requests.

> [!Note]
[...]

PR #6280, opened 2025-12-23
chore: expose install.sh on the main domain

ldez:
The current `install.sh` is exposed through https://raw.githubusercontent.com/golangci/golangci-lint/HEAD/install.sh

But GitHub has changed its rate limit policy on `raw.githubusercontent.com`.

> These changes will apply to operations like cloning repositories over HTTPS, anonymously interacting with our REST APIs, and downloading files from **raw.githubusercontent.com**.
> https://github.blog/changelog/2025-05-08-updated-rate-limits-for-unauthenticated-requests/

I think this is can only impact rare cases.

The previous URL still works.

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -74,4 +74,4 @@ jobs:
         os: [ubuntu-latest, macos-latest, windows-latest]
     runs-on: ${{ matrix.os }}
     steps:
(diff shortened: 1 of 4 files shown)


ccoVeille:
Is it a try to mitigate #6259 ?

- #6259

PR #6335, opened 2026-01-23
dev: fix modernize lint issues

alexandear:
This PR fixes `modernize` lint issues as it suggests a bit "cleaner" code.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -41,7 +41,6 @@ linters:
     - ineffassign
     - lll
     - misspell
-    - modernize
     - nakedret
     - noctx
     - nolintlint


ldez:
You know that we do not add new linters to the repository configuration, I have already told you this several times.

And the file has a clear documentation about this:
https://github.com/golangci/golangci-lint/blob/a0a46d1c5f4333f10c702ac035534f4c81655316/.golangci.yml#L3-L8

alexandear:
> You know that we do not add new linters to the repository configuration

Yeah, I know about this. But I thought that since the `modernize` was enabled in v2.6 and we are currently on v2.8, we already catch all the issues and can enable it.

But I'm fine with only the fixes without touching .golangci.yml.

ldez:
You misinterpreted the word "new": this is not related to a fresh new linter, but to the fact to enable a linter, so add a "new" linter to the list.

Once again the documentation is clear about this:
https://github.com/golangci/golangci-lint/blob/a0a46d1c5f4333f10c702ac035534f4c81655316/.golangci.yml#L3-L8

And once again, I have already told you this several times.

PR #6369, opened 2026-02-14
chore: improve the support of large numbers of packages (FilenameUnadjuster)

ran-arigur:
Currently (before this change), when linting a large number of packages, NewFilenameUnadjuster is liable to exhaust the thread limit (which defaults to 10,000). That's because it creates a separate goroutine for each package, and each goroutine does file I/O so it needs its own thread (at least part of the time).

After this change, NewFilenameUnadjuster will limit itself to at most 2000 goroutines: if there are more than 2000 packages, then it will divide them into between 1000 and 2000 chunks, with one goroutine processing each chunk.

Some notes:

- I don't know exactly how many goroutines is reasonable, but 1000&ndash;2000 seems safe in both directions: small enough to definitely avoid the issue, large enough that the chunks will be singletons for the vast majority of codebases (and still pretty small even for unusually large codebases).
- The exact number of goroutines can be anywhere between 1000 and 2000, because this just sets the chunk size to ⌊len(pkgs)/2000⌋+1 and lets the number of goroutines come out as a consequence of that. (If that's a concern, let me know and I can rework it to precisely hit a specific number of goroutines. I just didn't want to overengineer when I don't even have a specific target in mind.)
- While I was here, I simplified slightly by using some newer features of Go: [loop variable scope was fixed in Go 1.22](https://go.dev/blog/loopvar-preview) (so we don't have to avoid capturing the variable in a goroutine), and [sync.WaitGroup added a 'Go' method in Go 1.25](https://pkg.go.dev/sync#WaitGroup.Go) (so we don't need to explicitly 'Add' and mark 'Done').

### DIFF
--- a/pkg/result/processors/filename_unadjuster.go
+++ b/pkg/result/processors/filename_unadjuster.go
@@ -3,6 +3,7 @@ package processors
 import (
 	"go/parser"
 	"go/token"
+	"slices"
 	"strings"
 	"sync"
 	"time"
@@ -41,14 +42,14 @@ func NewFilenameUnadjuster(pkgs []*packages.Package, log logutils.Log) *Filename
 	startedAt := time.Now()
 
 	var wg sync.WaitGroup
-	wg.Add(len(pkgs))
 
-	for _, pkg := range pkgs {
-		go func(pkg *packages.Package) {
-			// It's important to call func here to run GC
-			processUnadjusterPkg(&m, pkg, log)
-			wg.Done()
-		}(pkg)
+	for chunk := range slices.Chunk(pkgs, len(pkgs)/2000+1) {
+		wg.Go(func() {
+			for _, pkg := range chunk {
+				// It's important to call func here to run GC
+				processUnadjusterPkg(&m, pkg, log)
+			}
+		})
 	}
 
 	wg.Wait()


ldez:
Your analysis is theoretical, do you have some concrete proofs of the change?

ran-arigur:
> Your analysis is theoretical, do you have some concrete proofs of the change?

Good question!

On a codebase of about 14,000 packages, we occasionally get fatal errors of this form:

```
runtime: program exceeds 1000-thread limit
fatal error: thread exhaustion
```

and I'm confident that this part of the codebase is the issue, because the error is accompanied by a goroutine dump, and more than 99% of the goroutines start at pkg/result/processors/filename_unadjuster.go line 47.

So the issue itself is concrete.

---

The fix, admittedly, is theoretical. Verifying the fix isn't 100% straightforward, because the error happens only occasionally; but I'm pretty sure I can manage to do so, if you're concerned? (Please let me know.)

ldez:
OK, thank you.

K8s is about 1400 packages, and it's a large project.
How is it possible to have 14 000 packages???

ldez:
Can you run with the verbose flag (`golangci-lint run -v`) with and without your change on your project?

You will see this line:
```
INFO [runner/filename_unadjuster] Pre-built ...
```

This line contains the duration of this specific step.

ran-arigur:
Thanks for the guidance!

---

> K8s is about 1400 packages, and it's a large project.
> How is it possible to have 14 000 packages???

The repo in question is Samsara Inc.'s main internal repo, which has most of our microservices, dev tooling, CI/CD tooling, customer-facing APIs, DB accessors, etc., etc., etc.

(At some companies this would be done as many separate repos, or as many independent modules within a flexibly-structured repo; but in our case it's all in one top-level Go module, with a very small number of nested modules for special cases, e.g. one nested module for our custom golangci-lint linters that we build with `golangci-lint custom`.)

---

> Can you run with the verbose flag (`golangci-lint run -v`) with and without your change on your project?
> 
> You will see this line:
> 
> ```
> INFO [runner/filename_unadjuster] Pre-built ...
> ```
> 
> This line contains the duration of this specific step.

_Without_ the change from this PR:

```
level=info msg="[runner/filename_unadjuster] Pre-built 0 adjustments in 3.685458125s"
```

_With_ the change from this PR:

```
level=info msg="[runner/filename_unadjuster] Pre-built 0 adjustments in 3.68088695s"
```

So the latency seems to be the same either way; but I'm not sure if the "0 adjustments" part means that it didn't really test what you wanted it to?

Please let me know if there's anything else you want me to try. 🙂

---

Thanks again!

ldez:
Given that your PR appears to be company-related, I would like to ask if you could help us by asking your company to sponsor our project.

My only source of income is from donations.
This will be appreciated, thank you :heart:

- https://donate.golangci.org
- https://donate.ldez.dev

ran-arigur:
I'll ask my manager about that. ❤️

ldez (approved):
LGTM

I did benchmarks on "smaller" projects and I didn't detect performance regressions.

If we detect some problems we will adjust the limit in the future.

PR #6381, opened 2026-02-18
dev: remove redundant check

alexandear:
`loopclosure` is already disabled when Go < 1.22, according to [the line](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.42.0:go/analysis/passes/loopclosure/loopclosure.go;l=58) and [tests](https://cs.opensource.google/go/x/tools/+/refs/tags/v0.42.0:go/analysis/passes/loopclosure/testdata/src/versions/go22.txtar). No need to explicitly disable it.

Or maybe we should remove `loopclosure` altogether, as the TODO suggests?


ldez:
I prefer to keep it for now.

alexandear:
I added tests to prove that this check is unnecessary.

PR #6407, opened 2026-03-04
errcheck: ignore Rollback errors in std-error-handling preset

danrneal:
## Description

This PR updates the `std-error-handling` exclusion preset to ignore errors from `Rollback` calls.

Currently, `golangci-lint` with `errcheck` reports an error when `defer tx.Rollback()` is not checked. Since it's a common pattern to ignore the error returned by `Rollback` when deferring it (as the transaction will either be committed or rollback will just do nothing if already committed/rolled back), this update prevents false positives.

### Changes made:
* Appended `|.*Rollback` to the regex pattern in the `std-error-handling` preset defined in `pkg/result/processors/exclusion_presets.go`.
* Updated the corresponding documentation reference in `docs/data/exclusion_presets.json`.

Tests were executed locally to ensure this change causes no regressions.

### DIFF
--- a/docs/data/exclusion_presets.json
+++ b/docs/data/exclusion_presets.json
@@ -82,7 +82,7 @@
       "linters": [
         "errcheck"
       ],
-      "text": "(?i)Error return value of .((os\\.)?std(out|err)\\..*|.*Close|.*Flush|os\\.Remove(All)?|.*print(f|ln)?|os\\.(Un)?Setenv). is not checked"
+      "text": "(?i)Error return value of .((os\\.)?std(out|err)\\..*|.*Close|.*Flush|os\\.Remove(All)?|.*print(f|ln)?|os\\.(Un)?Setenv|.*Rollback). is not checked"
     }
   ]
 }
--- a/pkg/result/processors/exclusion_presets.go
+++ b/pkg/result/processors/exclusion_presets.go
@@ -55,7 +55,7 @@ var LinterExclusionPresets = map[string][]config.ExcludeRule{
 			// Almost all programs ignore errors on these functions and in most cases it's ok.
 			BaseRule: config.BaseRule{
 				Text: "(?i)Error return value of .((os\\.)?std(out|err)\\..*|.*Close" +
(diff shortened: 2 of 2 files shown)


ldez:
Technically, this is a breaking change because it will create false-negatives.

ldez:
I confirm this is breaking: if a user has defined `nolint` directives and error will happen.
And from what I can see on GitHub the error is handled in a lot of cases.

https://github.com/search?q=+language%3AGo+%22tx.Rollback%22+NOT+%22defer+tx.Rollback%22&type=code

Also `Rollback`, as a function/method name, is not specific to SQL, so it will produce more false negatives.

So thank you, but we will decline the changes.

You can add this exclusions inside your configuration:

```yml
linters:
  exclusions:
    rules:
      - linters:
          - errcheck
        text: ".*Rollback is not checked"
```

danrneal:
> Technically, this is a breaking change because it will create false-negatives.

That's a fair point. However, `defer tx.Rollback()` is an extremely common idiom in Go, and ignoring its error is standard practice, much like `defer f.Close()`. Since `.*Close` and `.*Flush` are already included in this preset for the same reason, I thought `.*Rollback` would be a fitting addition.

I could scope the regex to be more specific, like `(\*sql\.Tx)\.Rollback` or `.*[tT]x\.Rollback`? Alternatively, I understand if this needs to wait for a major version since it alters existing behavior.

ldez:
As I explained [here](https://github.com/golangci/golangci-lint/pull/6407#issuecomment-3994855810) this will produce unwanted errors in existing code bases.

So I will not accept this change or a more specific change.

Feel free to open an issue, and we will evaluate it later when a next major will be created.

danrneal:
> As I explained [here](https://github.com/golangci/golangci-lint/pull/6407#issuecomment-3994855810) this will produce unwanted errors in existing code bases.
> 
> So I will not accept this change or a more specific change.
> 
> Feel free to open an issue, and we will evaluate it later when a next major will be created.

Apologies, you were too fast!  I posted my comment before I realized you had already answered my question. :) Gotta love active maintainers.

I'll go ahead an open an issue! Thanks for the detailed and prompt response.

PR #6429, opened 2026-03-11
internal/cache: stabilize package cache keys across worktrees

sheki:


ldez:
Humanity's razor:

What an AI has created can be dismissed without explanation.

PR #6431, opened 2026-03-11
internal/cache: stabilize package cache keys across worktrees

### LINKED ISSUE #6430: Golangci-lint is slow on a git worktree checkout as it does not share cache with parent repo
### Welcome

go install

`golangci-lint`'s package cache appears to be path-sensitive.

I'm seeing cache misses across identical checkouts/worktrees when the repo lives at a different absolute path, even with the same contents, config, Go version, and shared `GOLANGCI_LINT_CACHE`.

This matters more now because ephemeral git worktrees are increasingly common in agent-driven workflows.

I would like different git workspaces to use the same cache

None

### Additional context

This seems to be because:
- `internal/cache/cache.go` includes both file path and file hash in the package key:
```go
fmt.Fprintf(key, "file %s %x\n", f, h)
```
- `go/packages` provides absolute `CompiledGoFiles` paths.

So identical repos at different absolute paths get different cache keys.

Minimal repro:
1. Warm cache in worktree A.
2. Create worktree B for the same commit at a different absolute path.
3. Run the same `golangci-lint run ...` command.
4. Package analysis runs again instead of reusing cache.

### Supporter

sheki:
Closes #6430


leonklingele:
@sheki why not reopen https://github.com/golangci/golangci-lint/pull/6429?

@ldez this seems reasonable. Why do you think this is AI generated? And even if it is, what's the problem with such a small diff?

sheki:
Ping.

ldez:
This PR has absolutely no impact other than slowing down golangci-lint and create bugs: the implementation is just wrong because the diagnosis is wrong.
I will not spent time to explain to an AI what is wrong.

PR #6442, opened 2026-03-19
feat: support custom config file path for custom command

delamainer:
## Summary

- add `--config` / `-c` support to `golangci-lint custom`
- allow `custom` to load a config file from an explicit path instead of always auto-discovering `.custom-gcl.yml`
- keep existing behavior unchanged when `--config` is not provided


ldez:
I don't see the point of doing that, there is no explanation of the need, and no related to issue.

PR #6444, opened 2026-03-19
feat(custom): propagate replace directives from local plugin go.mod files

pkarakal:
## Problem

When a plugin is referenced by local path in `.custom-gcl.yml`, the `golangci-lint custom` build fails if the plugin's `go.mod` contains any replace directives. This affects three common real-world scenarios:

**1. Local/unpublished dependencies**
A plugin that wraps an internal shared library that has never been published to a module proxy:

```
    replace example.com/mylib => ../mylib
```

`go mod tidy` tries to fetch `example.com/mylib` from VCS and fails with "repository not found".

**2. Forked dependencies**
A plugin depending on a patched fork of an upstream library:

```
    replace github.com/upstream/lib => github.com/myfork/lib v1.2.3
```

`go mod tidy` attempts to resolve the original module path and may fail if the version doesn't exist upstream, or silently use the wrong code.

**3. Pre-release / in-development dependencies**
A plugin being developed alongside a library that hasn't cut a release yet, using a local path to track HEAD:

```
    replace example.com/lib => /home/user/projects/lib
```

`go mod tidy` fails to resolve the module from the proxy because the required version has not been published.

The root cause is a Go modules constraint: `replace` directives are only honoured in the *main module*. When the builder clones `golangci-lint` into a temporary directory and runs `go mod tidy` there, `golangci-lint` itself becomes the main module. Any `replace` directives declared inside the plugin's own go.mod are silently ignored.

This effectively forces plugin authors to either duplicate source code inside the plugin module, or publish every dependency before  `golangci-lint custom` will work, neither of which may be acceptable for private, forked, or actively-developed modules.

## Fix

After `addReplaceDirective` registers the plugin's own path replace in the `golangci-lint` `go.mod`, it now calls the new `propagatePluginReplaces` function. This function parses the plugin's `go.mod` using `golang.org/x/mod/modfile` (already a transitive dependency) and forwards every replace directive it finds into the `golangci-lint` `go.mod` via `go mod edit -replace`, before `go mod tidy` runs.

Two cases are handled:
*  Local directory replaces (empty New.Version): the path is resolved to an absolute path relative to the plugin directory, since the relative path would be meaningless inside the cloned golangci-lint tree in /tmp.
* Versioned replaces (non-empty New.Version): forwarded verbatim, e.g. `github.com/upstream/lib=github.com/myfork/lib@v1.2.3`.

The fix is intentionally one level deep and propagates **only** the replace directives from the plugin's own `go.mod`, not recursively from transitive dependencies. Deeper dependency graphs are better served by publishing
modules.

## Testing

Four table-driven unit tests are added to builder_test.go:
- Local path replace is resolved to an absolute path relative to the plugin directory
- Versioned replace (fork) is forwarded unchanged
- Plugin with no replace directives is a no-op
- Plugin with multiple replace directives has all of them propagated

### DIFF
--- a/pkg/commands/internal/builder.go
+++ b/pkg/commands/internal/builder.go
@@ -14,6 +14,7 @@ import (
 	"time"
 	"unicode"
 
+	"golang.org/x/mod/modfile"
 	"golang.org/x/mod/sumdb/dirhash"
 
 	"github.com/golangci/golangci-lint/v2/pkg/logutils"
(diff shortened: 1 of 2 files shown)


ldez:
Duplicate of #6098

pkarakal:
Sorry for wasting your time, dumb mistake from my side. I only checked the open issues but necer checked in the PRs.

PR #6445, opened 2026-03-19
feat: decrease cache entropy

### LINKED ISSUE #6430: Golangci-lint is slow on a git worktree checkout as it does not share cache with parent repo

ldez:
A project inside different locations will be able to reuse the cache.

As the entropy is decreased, parallel runs for the same project, in different locations, will never work because it will create unsafe concurrent access to the files used by the cache.

Fixes #6430

Notes:
- The cache usage can be see by using `GL_DEBUG="goanalysis/issues/cache"`
- No need to use a worktree to check the PR, you only need to copy a project in 2 different places.

### DIFF
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -713,7 +713,8 @@ func computeGoModSalt() (string, error) {
 		return "", fmt.Errorf("failed to read go.mod: %w", err)
 	}
 
-	sum, err := dirhash.Hash1([]string{goModPath}, func(string) (io.ReadCloser, error) {
+	// NOTE: the filename `goModPath` is not used here to have the same hash for the same content.
+	sum, err := dirhash.Hash1([]string{"go.mod"}, func(string) (io.ReadCloser, error) {
 		return io.NopCloser(bytes.NewReader(data)), nil
 	})
 	if err != nil {
--- a/pkg/commands/run.go
+++ b/pkg/commands/run.go
@@ -713,7 +713,7 @@ func computeGoModSalt() (string, error) {
 		return "", fmt.Errorf("failed to read go.mod: %w", err)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/commands/run.go
```diff
@@ -713,7 +713,8 @@ func computeGoModSalt() (string, error) {
 		return "", fmt.Errorf("failed to read go.mod: %w", err)
 	}
 
-	sum, err := dirhash.Hash1([]string{goModPath}, func(string) (io.ReadCloser, error) {
+	// NOTE: the filename `goModPath` is not used here to have the same hash for the same content.
```
ccoVeille: What about this


```suggestion
	// NOTE: the filename `goModPath` is not used here to have the same hash for go.mod with the same content in different folders.
```

alexandear (approved):

bombsimon (approved):

silverwind:
Looking forward to a release with this, this should make working in git worktrees much less painful than it currently is.

PR #6468, opened 2026-03-30
perf: parallelize cache hash computation

jerenkrantz:
## Summary
- Replace `hashFileCache` global mutex with `sync.Map` + `singleflight.Group`
  for lock-free cache hits and deduplicated concurrent misses
- Hash all files within a package concurrently in `computePkgHash`
- Add `PrecomputePackageHashes` that walks the full dependency graph in
  topological order, computing hashes leaf-to-root across all available cores

## Motivation
Cache key computation (`loadIssuesFromCache`) can take 600ms+ even for a
single-package run because it recursively hashes every file in every
transitive dependency, sequentially, on one goroutine. The global mutex
in `FileHash` further serializes concurrent callers.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/cache/cache.go
+++ b/internal/cache/cache.go
@@ -12,6 +12,7 @@ import (
 	"slices"
 	"strings"
 	"sync"
+	"sync/atomic"
 
 	"golang.org/x/tools/go/packages"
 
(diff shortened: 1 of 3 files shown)


jerenkrantz:
I opened #6469 - our lint on a large go codebase runs take a long time to compute the dependency hashes.

PR #6472, opened 2026-03-30
dev: improve migration suggestion for gomodguardv2

ldez:
When there is no explicit configuration, the current suggestion is:
```yml
linters:
  enable:
    - gomodguard_v2
  settings:
    gomodguard_v2:
      allowed: []
      blocked: []
      local-replace-directives: false 
```

with this PR, the suggestion is:
```yml
linters:
  enable:
    - gomodguard_v2
```

Same thing when the other fields are not explicitly set.

### DIFF
--- a/pkg/lint/linter/config.go
+++ b/pkg/lint/linter/config.go
@@ -189,10 +189,10 @@ func Replacement[T any](replacement string, mgr func(T) any, data T) func(*Depre
 				"enable": []string{d.Replacement},
 			}
 
-			settings := mgr(data)
+			replacementSettings := mgr(data)
 
-			if settings != nil {
-				linters["settings"] = map[string]any{d.Replacement: settings}
+			if replacementSettings != nil {
+				linters["settings"] = map[string]any{d.Replacement: replacementSettings}
 			}
 
 			suggestion := map[string]any{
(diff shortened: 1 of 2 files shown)


pkg/lint/linter/config.go
```diff
@@ -185,15 +185,18 @@ func Replacement[T any](replacement string, mgr func(T) any, data T) func(*Depre
...
 
+			linters := map[string]any{
+				"enable": []string{d.Replacement},
+			}
+
+			settings := mgr(data)
+
+			if settings != nil {
+				linters["settings"] = map[string]any{d.Replacement: settings}
+			}
```
bombsimon: Should this be `replacement`? It's only for the `d.Replacement` settings key and not `linter["settings"]`.


```suggestion
			replacement := mgr(data)

			if settings != nil {
				linters["settings"] = map[string]any{d.Replacement: replacement}
			}
```

bombsimon (approved):

pkg/lint/linter/config.go
ldez: `d.Replacement` is the linter name, and `settings` is the settings of this linter.

bombsimon: Yeah I still think it's less clear with just `settings` since the code is mutating the whole `liter["settings"]` but this is just a personal pref and hence why I approved it.

Maybe `replacementSettings` if you really want to emphasize it's settings for the `d.Replacement` but feel free to keep `settings` if you don't agree!

PR #6477, opened 2026-04-02
paralleltest: expose checkcleanup option

xcapaldi:
Add support for the `checkcleanup` option in [paralleltest](https://github.com/kunwardeep/paralleltest).

`make build`, `./golangci-lint run -v` and `make test` all pass.


ldez (approved):
LGTM

Issue #6513, opened 2026-04-20
golangci-lint custom corrupts git index when run inside a pre-commit hook (inherited GIT_INDEX_FILE)


Maxattax97:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/docs/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/docs/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

Other

### Description of the problem

When `golangci-lint custom` is invoked inside a git pre-commit hook, the `git clone` it runs internally (in `pkg/commands/internal/builder.go`) **overwrites the repository's git index**, silently destroying all staged changes.

  ### Root cause

  During `git commit`, git sets `GIT_INDEX_FILE` (and `GIT_DIR`, `GIT_WORK_TREE`, etc.) as environment variables for hook subprocesses. The `clone()` method in `builder.go` uses `exec.CommandContext` without setting `cmd.Env`, so the child `git clone` process inherits these variables. The clone's internal checkout then
   writes its index to `GIT_INDEX_FILE` — which points to the **original repository's index file** — overwriting it with golangci-lint's file tree.

  After this:
  - All staged changes are permanently lost
  - `git fsck` reports dozens of "missing blob" errors (the index references objects from the golangci-lint repo, not the user's repo)
  - `git ls-files` shows golangci-lint's files (`.github/peril/`, `cmd/golangci-lint/main.go`, etc.) instead of the project's files

  The corruption is completely silent — `golangci-lint custom` exits successfully.

  ### Why it's intermittent

  `golangci-lint custom` only runs `git clone` when rebuilding the binary. If the binary already exists and matches the expected version, the clone is skipped. This means **the first commit in a new worktree** (where the custom binary doesn't exist yet) triggers the bug. Version bumps in `.custom-gcl.yml` also trigger
  it.

[...]

PR #6514, opened 2026-04-20
fix: strip GIT env vars from subprocesses in golangci-lint custom (#6513)

Maxattax97:
When invoked inside a git pre-commit hook, git sets `GIT_INDEX_FILE`, `GIT_DIR`, and `GIT_WORK_TREE` as environment variables. The `git clone` in `Builder.clone()` inherits these, causing the clone's checkout to write its index to `GIT_INDEX_FILE` -- silently overwriting the original repository's git index with golangci-lint's file tree. This destroys all staged changes and corrupts the repo (`git fsck` reports missing blobs, `git diff` fails).

Filter `GIT_*` variables from the environment for all subprocesses in `builder.go` (clone, go get, go mod tidy, go build), preserving transport/auth vars like `GIT_SSH_COMMAND` and `GIT_EXEC_PATH`.

### DIFF
--- a/pkg/commands/internal/builder.go
+++ b/pkg/commands/internal/builder.go
@@ -29,6 +29,42 @@ type Builder struct {
 	repo string
 }
 
+// cleanGitEnv returns a copy of the current environment with GIT_* variables
+// removed that could interfere with subprocess git/go operations.
+//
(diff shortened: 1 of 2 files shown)


ldez:
you win https://github.com/golangci/golangci-lint/issues/6513#issuecomment-4282009016

PR #6515, opened 2026-04-20
fix(custom): filter env vars when cloning the repository

### LINKED ISSUE #6513: golangci-lint custom corrupts git index when run inside a pre-commit hook (inherited GIT_INDEX_FILE)

ldez:
Fixes #6513

### DIFF
--- a/pkg/commands/internal/builder.go
+++ b/pkg/commands/internal/builder.go
@@ -306,8 +306,6 @@ func filterGitEnviron(envs []string) []string {
 			"GIT_ALLOW_PROTOCOL",
 			"GIT_ASKPASS":
 			filtered = append(filtered, env)
-		default:
-			continue
 		}
 	}
 
--- a/pkg/commands/internal/builder.go
+++ b/pkg/commands/internal/builder.go
@@ -96,6 +96,7 @@ func (b Builder) clone(ctx context.Context) error {
 		"https://github.com/golangci/golangci-lint.git",
 	)
 	cmd.Dir = b.root
+	cmd.Env = filterGitEnviron(os.Environ())
 
 	output, err := cmd.CombinedOutput()
 	if err != nil {
@@ -279,4 +280,36 @@ func sanitizeVersion(v string) string {
 	}
 
 	return strings.Join(strings.FieldsFunc(v, fn), "")
+}
+
+// Inspired by https://github.com/pre-commit/pre-commit/blob/f5678bf4ac35cffc0ff7174ad85f7fdc2a5c977e/pre_commit/git.py#L27
+func filterGitEnviron(envs []string) []string {
+	var filtered []string
+
+	for _, env := range envs {
+		if !strings.HasPrefix(env, "GIT_") {
+			filtered = append(filtered, env)
+			continue
+		}
+
+		if strings.HasPrefix(env, "GIT_CONFIG_KEY_") || strings.HasPrefix(env, "GIT_CONFIG_VALUE_") {
+			filtered = append(filtered, env)
+			continue
+		}
+
+		key, _, _ := strings.Cut(env, "=")
+
+		switch key {
+		case "GIT_EXEC_PATH", "GIT_SSH", "GIT_SSH_COMMAND", "GIT_SSL_CAINFO",
+			"GIT_SSL_NO_VERIFY", "GIT_CONFIG_COUNT",
+			"GIT_HTTP_PROXY_AUTHMETHOD",
+			"GIT_ALLOW_PROTOCOL",
+			"GIT_ASKPASS":
+			filtered = append(filtered, env)
+		default:
+			continue
+		}
+	}
+
+	return filtered
 }
--- a/pkg/commands/internal/builder_test.go
+++ b/pkg/commands/internal/builder_test.go
@@ -1,6 +1,7 @@
 package internal
 
 import (
+	"slices"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


pkg/commands/internal/builder_test.go
```diff
@@ -54,3 +55,52 @@ func Test_sanitizeVersion(t *testing.T) {
 		})
 	}
 }
+
+func Test_filterEnviron(t *testing.T) {
```
Maxattax97: I like your tests better than mine, good job!

pkg/commands/internal/builder.go
```diff
@@ -96,6 +96,7 @@ func (b Builder) clone(ctx context.Context) error {
 		"https://github.com/golangci/golangci-lint.git",
 	)
 	cmd.Dir = b.root
+	cmd.Env = filterGitEnviron(os.Environ())
```
Maxattax97, line 99: It's a keen observation that technically it is only strictly necessary to filter the env for this clone path. Good thinking.

Maxattax97 (commented):
✅  Appreciate the quick turnaround 🙏

pkg/commands/internal/builder_test.go
```diff
@@ -54,3 +55,52 @@ func Test_sanitizeVersion(t *testing.T) {
...
+		"GIT_CONFIG_VALUE_1=bot@example.com",
+		"GIT_EXEC_PATH=/usr/lib/git-core",
+		"GIT_HTTP_PROXY_AUTHMETHOD=basic",
+		"GIT_SSH=/usr/bin/ssh",
+		"GIT_SSH_COMMAND=ssh -i ~/.ssh/id_ed25519 -o StrictHostKeyChecking=no",
+		"GIT_SSL_CAINFO=/etc/ssl/certs/ca-certificates.crt",
+		"GIT_SSL_NO_VERIFY=true",
+	}
+
+	assert.Equal(t, expected, envs)
```
alexandear, line 102: We can change to [ElementsMatch](https://pkg.go.dev/github.com/stretchr/testify/assert#ElementsMatch):
```suggestion
	assert.ElementsMatch(t, expected, envs)
```

and remove `slices.Sort`.

pkg/commands/internal/builder.go
```diff
@@ -280,3 +281,35 @@ func sanitizeVersion(v string) string {
...
+
+		switch key {
+		case "GIT_EXEC_PATH", "GIT_SSH", "GIT_SSH_COMMAND", "GIT_SSL_CAINFO",
+			"GIT_SSL_NO_VERIFY", "GIT_CONFIG_COUNT",
+			"GIT_HTTP_PROXY_AUTHMETHOD",
+			"GIT_ALLOW_PROTOCOL",
+			"GIT_ASKPASS":
+			filtered = append(filtered, env)
+		default:
+			continue
```
alexandear: nit: redundant `continue`

alexandear (approved):

PR #6525, opened 2026-04-26
dev: switch to crazy-max/ghaction-github-pages

ldez:
Switch from the unmaintained https://github.com/peaceiris/actions-gh-pages to https://github.com/crazy-max/ghaction-github-pages


bombsimon (approved):
Always surprised at how often you find abandoned/unmaintained repos without pinned issues or notes in the docs but you clearly se the owner is still active. 

Didn't vet `crazy-max/ghactions-github-pages` other than a quick look but seems fine!

PR #6527, opened 2026-04-28
dev: improve Docker login during release

ldez:
Use the official Docker login action.
https://github.com/docker/login-action

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -75,7 +75,10 @@ jobs:
         uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
 
       - name: Login do docker.io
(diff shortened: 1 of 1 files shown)


bombsimon (approved):

.github/workflows/release.yml
```diff
@@ -75,7 +75,10 @@ jobs:
         uses: docker/setup-buildx-action@4d04d5d9486b7bd6fa91e7baf45bbb4f8b9deedd # v4.0.0
 
       - name: Login do docker.io
-        run: docker login -u golangci -p ${{ secrets.GOLANGCI_LINT_DOCKER_TOKEN }}
+        uses: docker/login-action@v4
```
alexandear, line 78: Please use the commit hash instead of v4

alexandear: Why was this resolved without an explanation?

ldez: Because I'm why on the topic but in another PR.

PR #6538, opened 2026-05-01
docs: update GitHub Action assets

ldez:
The post-release workflow has not been triggered, and I don't why for now.
So, I updated the file manually.


ldez:
The CI fails because https://github.com/golangci/golangci-lint/discussions/6537

I'm on the topic.

I will merge this PR anyway because the script doesn't impact the gha asserts and the JSONSchema.

Issue #6555, opened 2026-05-04
v2.12.1 linux/amd64: SHA256 checksum mismatch causes install.sh to fail


souravgupta-msft:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/docs/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/docs/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

Official binary

## Description
Running the official `install.sh` script to install the latest version fails with a checksum verification error on `v2.12.1` for `linux/amd64`.
The SHA256 hash of the downloaded tarball (4a1fed9...) does not match the hash listed in the release's .sha256 checksum file (82c7932...). This suggests a mismatch between the binary artifact and its checksum file published in the v2.12.1 release on GitHub.

## Steps to replicate
```
curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh \ | sh -s -- -b $(go env GOPATH)/bin
```

## Error output
```
golangci/golangci-lint info checking GitHub for latest tag
golangci/golangci-lint info found version: 2.12.1 for v2.12.1/linux/amd64
golangci/golangci-lint err hash_sha256_verify checksum for '/tmp/tmp.n8PBmscCrk/golangci-lint-2.12.1-linux-amd64.tar.gz' did not verify 82c7932d9aa10c34e22ef56961ac5aad71d4a14806a1a31b06931add2688b368
4a1fed90469250d59ab1d623521c3584027ed9cf94ea7aba3c2477164746e2cf vs 82c7932d9aa10c34e22ef56961ac5aad71d4a14806a1a31b06931add2688b368
```

## Environment
- OS: Linux (amd64)
- Install method: official install.sh from master branch
- Version resolved: v2.12.1

### Version of golangci-lint

v2.12.1

### Configuration

<details>
NA
</details>

### Go environment

<details>

```console
go version go1.26.2 linux/amd64
```

[...]

Issue #6540, opened 2026-05-01
err hash_sha256_verify checksum for


adamdecaf:
### Welcome

- [x] Yes, I'm using a binary release within 2 latest [releases](https://github.com/golangci/golangci-lint/releases). Only such installations are supported.
- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] Yes, I've read the `typecheck` section of the [FAQ](https://golangci-lint.run/docs/welcome/faq/#why-do-you-have-typecheck-errors).
- [x] Yes, I've tried with the standalone [linter](https://golangci-lint.run/docs/linters/) if available (e.g., gocritic, go vet, etc.).
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

Official binary

### Description of the problem

```
golangci/golangci-lint info checking GitHub for tag 'latest'
golangci/golangci-lint info found version: 2.12.0 for v2.12.0/linux/amd64
golangci/golangci-lint err hash_sha256_verify checksum for '/tmp/tmp.RqO2TT0noV/golangci-lint-2.12.0-linux-amd64.tar.gz' did not verify 6b89d77b6396f81decee882f20473486bda5ef28b160f9a13b08f24848d9003c
7976b1209947f3d94a851145bfb67ea4779bd431b0999a3cd8f38f4544afd0cf vs 6b89d77b6396f81decee882f20473486bda5ef28b160f9a13b08f24848d9003c
```

### Version of golangci-lint

2.12.0

### Configuration

https://github.com/moov-io/infra/blob/master/go/lint-project.sh#L352 

### Go environment

Actions run: https://github.com/moov-io/ach/actions/runs/25219675443/job/73948742491

<details>

```console
go version go1.26.2 linux/amd64
```

</details>

### Verbose output of running

Fresh Github Runner, no cache. 

### A minimal reproducible example or link to a public repository

https://github.com/moov-io/infra/blob/master/go/lint-project.sh#L352
https://github.com/moov-io/ach/actions/runs/25219675443/job/73948742491#logs

### Validation

- [x] Yes, I've included all the information above (example, version, config, etc.).

### Supporter

- [x] I am a [sponsor/backer](https://donate.golangci.org) of this project.

ldez:
Already fixed: #6539

ldez:
[...]

PR #6539, opened 2026-05-01
fix: install.sh script

ldez:
---

> [!IMPORTANT] 
> If you are using the install script from the `master` branch:
> `https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh`
> This branch is not used anymore, we are using `main`.
>
> We recommend using `https://golangci-lint.run/install.sh` as URL to install golangci-lint.
> 
> https://golangci-lint.run/docs/welcome/install/local/#binaries

---

The SBOM files has been added to the checksum file.

```
6b89d77b6396f81decee882f20473486bda5ef28b160f9a13b08f24848d9003c  golangci-lint-2.12.0-linux-amd64.tar.gz
7976b1209947f3d94a851145bfb67ea4779bd431b0999a3cd8f38f4544afd0cf  golangci-lint-2.12.0-linux-amd64.tar.gz.sbom.json
```

The script uses grep to find the right line, but there are two lines that match.

```console
$ grep "golangci-lint-2.12.0-linux-amd64.tar.gz" "golangci-lint-2.12.0-checksums.txt" 2>/dev/null | tr '\t' ' ' | cut -d ' ' -f 1 
6b89d77b6396f81decee882f20473486bda5ef28b160f9a13b08f24848d9003c
7976b1209947f3d94a851145bfb67ea4779bd431b0999a3cd8f38f4544afd0cf
```

~~So, I added a `head -n 1`, this is only a quick fix.~~
I edited the grep matcher to use `$`.

I need to think if this is a proper solution, but for now, I need to unlock the situation.

The job `check-local-install-script` of the workflow `pr-checks.yml` validates that my fix is right.

Fixes https://github.com/golangci/golangci-lint/discussions/6537

PR #6556, opened 2026-05-04
fix: improve checksum verification logic in hash_sha256_verify function

alexlovelltroy:
When installing via `install.sh` on linux-amd64, I noticed a checksum error.  The script was finding the checksum for the sbom rather than the tarball.  This small patch seems to work for me by falling back to shell syntax instead of using grep.  I've tested it on macos and ubuntu, so I believe it covers the various shell/coreutils behaviors properly.

## Bug Listing
```
$ curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh
golangci/golangci-lint info checking GitHub for latest tag
golangci/golangci-lint info found version: 2.12.1 for v2.12.1/linux/amd64
golangci/golangci-lint err hash_sha256_verify checksum for '/tmp/tmp.QAhs33Nb6n/golangci-lint-2.12.1-linux-amd64.tar.gz' did not verify 82c7932d9aa10c34e22ef56961ac5aad71d4a14806a1a31b06931add2688b368
4a1fed90469250d59ab1d623521c3584027ed9cf94ea7aba3c2477164746e2cf vs 82c7932d9aa10c34e22ef56961ac5aad71d4a14806a1a31b06931add2688b368
```

```
$ sha256sum /tmp/tmp.QAhs33Nb6n/golangci-lint-2.12.1-linux-amd64.tar.gz
82c7932d9aa10c34e22ef56961ac5aad71d4a14806a1a31b06931add2688b368  /tmp/tmp.QAhs33Nb6n/golangci-lint-2.12.1-linux-amd64.tar.gz

$ grep linux-amd64.tar.gz /tmp/tmp.QAhs33Nb6n/golangci-lint-2.12.1-checksums.txt
82c7932d9aa10c34e22ef56961ac5aad71d4a14806a1a31b06931add2688b368  golangci-lint-2.12.1-linux-amd64.tar.gz
4a1fed90469250d59ab1d623521c3584027ed9cf94ea7aba3c2477164746e2cf  golangci-lint-2.12.1-linux-amd64.tar.gz.sbom.json
```

## Listing after patch
```
$ cat install.sh | sh
golangci/golangci-lint info checking GitHub for latest tag
golangci/golangci-lint info found version: 2.12.1 for v2.12.1/linux/amd64
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -387,7 +387,17 @@ hash_sha256_verify() {
     return 1
   fi
   BASENAME=${TARGET##*/}
-  want=$(grep "${BASENAME}$" "${checksums}" 2>/dev/null | tr '\t' ' ' | cut -d ' ' -f 1)
+  want=""
+  # Parse checksum files by fields to avoid substring/regex matches.
+  while read -r sum file _; do
+    [ -z "$sum" ] && continue
+    file=${file#\*}
+    file=$(echo "$file" | tr -d '\r')
+    if [ "$file" = "$BASENAME" ]; then
+      want="$sum"
+      break
+    fi
+  done < "$checksums"
   if [ -z "$want" ]; then
     log_err "hash_sha256_verify unable to find checksum for '${TARGET}' in '${checksums}'"
     return 1


bombsimon:
The `raw.githubusercontent.com` URL is not recommended and the `master` branch is stale/not used. We recommend using `https://golangci-lint.run/install.sh` which is mentioned [in the docs](https://golangci-lint.run/docs/welcome/install/local/). 

This also means that this change won't be merged to `master` since it's unused.

Relates to #6555, #6540

EDIT: Re-opening because even though the fix won't land on `master` and fix your URL, it is related to the current way we parse this even on the recommended URL.

ldez:
The problem is already fixed by #6539

PR #6562, opened 2026-05-04
fix(gomodguard_v2): squash embedded base so blocked rules decode Module field

ryancurrah:
`GoModGuardv2Blocked` embeds `GoModGuardv2Base` without a `mapstructure:",squash"` tag. mapstructure does not promote embedded fields by default, so the user's flat YAML (`module:`, `match-type:`, `version:` keys directly under each `blocked` entry) was decoding into `Module=""`, `MatchType=""`, `Version=""` — only `Reason` and `Recommendations` survived.

Every blocked rule then arrived at gomodguard with an empty `Module`, collided on the same exact-match key, and never matched any required module from `go.mod`. Imports of supposedly blocked modules silently fell through to the allowed-list check, producing the misleading "is blocked because the module is not in the allowed modules list" message instead of the intended reason and recommendation.

Add `mapstructure:",squash"` to the embedded base, matching the pattern already used for `Tagliatelle*` settings in the same file. Update the gomodguard_v2 testdata `// want` expectations, which previously asserted the buggy "not in the allowed modules list" message and now correctly assert the blocked-rule message including the reason and recommendation.

### DIFF
--- a/pkg/config/linters_settings.go
+++ b/pkg/config/linters_settings.go
@@ -624,7 +624,7 @@ type GoModGuardv2Base struct {
 }
 
 type GoModGuardv2Blocked struct {
-	GoModGuardv2Base
+	GoModGuardv2Base `mapstructure:",squash"`
 
 	Recommendations []string `mapstructure:"recommendations"`
 	Reason          string   `mapstructure:"reason"`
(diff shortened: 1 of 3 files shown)


ldez:
already fixed :wink: -> #6561

PR #6568, opened 2026-05-06
install.sh: avoid grep matching .tar.gz.sbom.json as well.

garloff:
There already was a "$" supposed to prevent that, but it needs to be escaped as it's in double-quotes.

This hits on alpine with busybox shell, sone others possibly as well. The result is a want with two sha256sums, which of course does not match. Fix is to properly escape the "\$".

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -387,7 +387,7 @@ hash_sha256_verify() {
     return 1
   fi
   BASENAME=${TARGET##*/}
-  want=$(grep "${BASENAME}$" "${checksums}" 2>/dev/null | tr '\t' ' ' | cut -d ' ' -f 1)
+  want=$(grep " ${BASENAME}\$" "${checksums}" 2>/dev/null | tr '\t' ' ' | cut -d ' ' -f 1)
   if [ -z "$want" ]; then
     log_err "hash_sha256_verify unable to find checksum for '${TARGET}' in '${checksums}'"
     return 1


ldez:
This has been fixed 5 days ago: #6539

This is fun because the PR description is based on wrong assumptions.

This PR illustrates the problem when a context is not described: we don't know which URL is used.

My guess this is the URL based on `master` and so not the right one.
In the end, it came trying to fix a problem based on the fix of the problem.

To prove what I'm saying:

```console
$ docker run --rm -it golang:1.25-alpine sh
/go # wget -O- -nv https://golangci-lint.run/install.sh | sh -s v2.12.2
Connecting to golangci-lint.run (185.199.108.153:443)
writing to stdout
-                    100% |*************************************************************************************************************************************************************************************| 12091  0:00:00 ETA
written to stdout
golangci/golangci-lint info checking GitHub for tag 'v2.12.2'
golangci/golangci-lint info found version: 2.12.2 for v2.12.2/linux/amd64
/go # 
```

garloff:
Happy to sign the CLA.
We're lucky that shells don't eat the "$" before the closing ".
Normally, we'd need to escape it with "\\$", which is what my patch does.
So that change would make it more robust, but it's not required.
But you are completely right that the Dockerfile downloaded from the master branch and thus even lacked the $.
Thanks.

ldez:
For you, to understand my context: we already have 3 duplicate issues and 2 duplicate PRs on the same topic, despite the fact that I documented the topic inside the changelog, the release description, my PR to fix the problem, and all the duplicated elements.

At some point, this is painful for me to put effort to document the topic but be forced to repeat the same things during one week.

Not against you personally.

garloff:
@ldez - I can feel your pain. Been in that role in other projects. Sorry if I added to it a bit, even by delivering a valid (but not needed) improvement.

Issue #6498, opened 2026-04-14
Add retry to the curl http download


l-urbini:
### Welcome

- [x] Yes, I've searched similar [issues on GitHub](https://github.com/golangci/golangci-lint/issues) and didn't find any.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/golangci/golangci-lint?tab=coc-ov-file)

### How did you install golangci-lint?

go install

### Your feature request related to a problem? Please describe

we had several pipeline failures, occurring due to github network glitched causing 504s and breaking here https://github.com/golangci/golangci-lint/blob/479e5609d9ea67247c8ec2e1ca1f3aab46a392b4/install.sh#L291 

### Describe the solution you'd like

would be nice to have some retry logic to avoid this breaking things 

### Describe alternatives you've considered

n/a

### Supporter

- [ ] I am a [sponsor/backer](https://donate.golangci.org) of this project.

PR #6612, opened 2026-06-02
-

toller892:
## Problem

The install script's HTTP download functions have no retry logic. When GitHub returns transient errors (e.g. 504 Gateway Timeout), the installation fails immediately without any retry.

## Solution

Added exponential backoff retry logic to the `http_download` function that wraps both curl and wget backends:

- Retries up to **3 times** by default on failure
- Exponential backoff: 1s → 2s → 4s between attempts
- Configurable via `HTTP_RETRY_COUNT` environment variable
- Logs retry attempts at info level for debugging

This addresses the issue reported in #6498 where CI pipelines fail due to intermittent GitHub network issues.

## Testing

- Verified script syntax with `bash -n install.sh`
- Successfully tested download and installation: `bash install.sh -d -b /tmp/test`
- Confirmed retry behavior triggers on failure

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -332,14 +332,30 @@ http_download_wget() {
 }
 http_download() {
   log_debug "http_download $2"
-  if is_command curl; then
-    http_download_curl "$@"
-    return
-  elif is_command wget; then
-    http_download_wget "$@"
-    return
-  fi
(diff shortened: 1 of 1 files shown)


ldez:
Screenshot 2026-06-02 at 06-01-22

PR #6673, opened 2026-07-15
feat: Add enable setting to modernize linter

taran-p:
Adds an `enable` setting to the `modernize` linter to allow keeping a set list of enabled analyzers.


ldez:
Please open an issue to explain your need.

PR #6703, opened 2026-08-06
fix: resolve symlinks on both sides when computing relative paths

### LINKED ISSUE #6702: Wrong relative paths in output when the working directory is reached through a symlink

When the working directory is reached through a symlink, the paths in the output leave the base path and come back into it, instead of being relative to it.

Setup (Linux, `/linkroot` is a symlink to `/realroot`, module in `/realroot/try-go`, `main.go` contains an empty `for {}` so `staticcheck` reports `SA5002`):

```console
$ cd /realroot/try-go && golangci-lint run .
main.go:4:2: SA5002: this loop will spin, using 100% CPU (staticcheck)

$ cd /linkroot/try-go && golangci-lint run .
../../linkroot/try-go/main.go:4:2: SA5002: this loop will spin, using 100% CPU (staticcheck)
```

Same file, same directory, only the spelling of the working directory differs.

It depends on `run.relative-path-mode`, and the split is consistent:

| mode | base path | result |
|---|---|---|
| `wd` | resolved by `fsutils.Getwd` | ❌ `../../linkroot/try-go/main.go` |
| `gitroot` | resolved by `git rev-parse --show-toplevel` | ❌ `../../linkroot/try-go/main.go` |
| `gomod` | as returned by `go env GOMOD` | ✅ `main.go` |
| `cfg` | configuration directory | ✅ `main.go` |

The two broken modes are exactly the two whose base path is symlink-resolved, while the file names reported by the linters keep the spelling used by the Go toolchain. `filepath.Rel` in [`pkg/result/processors/path_relativity.go`](https://github.com/golangci/golangci-lint/blob/main/pkg/result/processors/path_relativity.go#L44) then compares two different spellings of the same directory.

For the text output this is cosmetic, but for `json`, `checkstyle` and `sarif` the path no longer resolves against the repository root, so CI annotations do not attach to the file.

This came up while reproducing #4099, which turned out to be `cmd/go` behaviour rather than a golangci-lint bug.

### Version of golangci-lint

Built from `2cce7356513cf1394bad5158205fe62a12eb0b41` (main, 2026-08-04).

### Configuration file

```yaml
version: "2"
run:
  relative-path-mode: "wd"
linters:
  default: none
  enable:
    - staticcheck
```

### Go environment

`go version go1.26.0 linux/amd64` (official `golang:1.26` image).

Eljees:
Fixes #6702

`PathRelativity` compared a symlink-resolved base path (`fsutils.Getwd`, `git rev-parse --show-toplevel`) with file names kept in the spelling used by the Go toolchain, so `filepath.Rel` returned paths leaving the base path and coming back into it (`../../link/pkg/main.go` instead of `pkg/main.go`) in the `wd` and `gitroot` modes.

Both sides are now resolved, which is what `fsutils.ShortestRelPath` already does. Paths that cannot be resolved are left unchanged, so file names that do not exist on disk keep working.

### DIFF
--- a/pkg/result/processors/path_relativity.go
+++ b/pkg/result/processors/path_relativity.go
@@ -27,7 +27,7 @@ func NewPathRelativity(log logutils.Log, basePath string) (*PathRelativity, erro
 
 	return &PathRelativity{
 		log:              log.Child(logutils.DebugKeyPathRelativity),
-		basePath:         basePath,
+		basePath:         evalSymlinks(basePath),
 		workingDirectory: wd,
 	}, nil
 }
@@ -40,14 +40,16 @@ func (p *PathRelativity) Process(issues []*result.Issue) ([]*result.Issue, error
 	return transformIssues(issues, func(issue *result.Issue) *result.Issue {
 		newIssue := *issue
 
+		filePath := evalSymlinks(issue.FilePath())
+
 		var err error
-		newIssue.RelativePath, err = filepath.Rel(p.basePath, issue.FilePath())
+		newIssue.RelativePath, err = filepath.Rel(p.basePath, filePath)
 		if err != nil {
 			p.log.Warnf("Getting relative path (basepath): %v", err)
 			return nil
 		}
 
-		newIssue.WorkingDirectoryRelativePath, err = filepath.Rel(p.workingDirectory, issue.FilePath())
+		newIssue.WorkingDirectoryRelativePath, err = filepath.Rel(p.workingDirectory, filePath)
 		if err != nil {
 			p.log.Warnf("Getting relative path (wd): %v", err)
 			return nil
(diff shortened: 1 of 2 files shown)


Eljees:
Sorry for the noise — the linked issue was closed automatically because I opened it through the API instead of the issue form. That one is on me, and I'll refile it properly.

Before I do: was the change itself declined, or was it just following an invalid issue? I'd rather not resend something you don't want.

For context, the behaviour I was trying to fix is that the `wd` and `gitroot` relative-path modes report paths like `../../link/pkg/main.go` when the working directory is reached through a symlink, because their base path is symlink-resolved while the file names coming from the linters are not. The `gomod` and `cfg` modes are unaffected. If that is working as intended, I'm happy to drop it.

